PR #171, opened 2013-11-05
Add a "mock" region for which the endpoint URL is not rewritten

mguillaume:
For offline testing (like, when you're on a train), it's nice to be able to use a fake S3 server (eg fakes3).
However, the fix_s3_host function in handlers.py rewrites them to things like http://test.s3.amazonaws.com

This commit adds a fake "mock" region in the RESTRICTED_REGIONS so that custom endpoint_url used with this fake region are not rewritten to valid S3 endpoints. (this may not be the best way to do it, but it has the advantage to be super simple and have little chance to break anything).

### DIFF
--- a/botocore/data/aws/_regions.json
+++ b/botocore/data/aws/_regions.json
@@ -28,5 +28,8 @@
     },
     "fips-us-gov-west-1": {
         "description": "AWS GovCloud (FIPS 140-2) S3 Only"
+    },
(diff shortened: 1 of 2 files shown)


jamesls:
Closing out old pull request.  The internals have changed in botocore have changed quite a bit such that this is possible without having to modify any botocore code.

PR #205, opened 2014-01-10
Add instance metadata retry config values to session

jamesls:
This adds two new (config file only) options:
- metadata_service_timeout: The number of seconds until we should time out a request to the metadata service.
- metadata_service_num_attempts:  The total number of attempts to make to the metadata service.

These options match what's in boto.

This PR also includes an enhancement to the session class for per-session config variable overrides.  This is useful for testing, but I think it's also useful as a general feature (different session instances can explicitly set config values).  In order for this to make the most sense, I also moved the module vars for the mapping of session variables into the session class itself.

This is the other half of https://github.com/boto/botocore/pull/202

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -215,7 +215,9 @@ class Session(object):
         self._provider = None
         self._profile = profile
 
-    def get_config_variable(self, logical_name, methods=('instance', 'env', 'config')):
+    def get_config_variable(self, logical_name,
+                            methods=('instance', 'env', 'config'),
+                            default=None):
         """
         Retrieve the value associated with the specified logical_name
         from the environment or the config file.  Values found in the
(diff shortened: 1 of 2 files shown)


rcorre:
I thought maybe I [had to use Config](https://stackoverflow.com/questions/34003350/change-the-number-of-request-retries-in-boto3) but the [config reference](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html) does not include these variables.

rcorre:
nvm, found it at https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html. You have to set them as environment variables:

- AWS_METADATA_SERVICE_TIMEOUT
- AWS_METADATA_SERVICE_NUM_ATTEMPTS

PR #393, opened 2014-11-30
Add service description to services lacking docs

pplu:
I've lifted the documentation from the official AWS documentation pages for each service. Except Route53Domains, for which I've made a small stub.


jamesls:
Hi, just want to follow up on this pull request. Looks like this has evolved quite a bit from the develop branch.  Was there anything you'd like us to merge into the develop branch?  I notice a few of these changes are for the JSON model files, which are generated from an upstream source so we aren't able to accept pull requests.  We'd need to work to get those models updated internally.

PR #413, opened 2014-12-12
Added fix and passing integration tests for kinesis put_records method.

TelegramSam:
I've added failing unit tests for the new Kinesis put_records method. The call returns an error similar to this:

    ClientError: An error occurred (SerializationException) when calling the PutRecords operation: 'foobar' can not be converted to a Blob: Base64 encoded length is expected a multiple of 4 bytes but found: 6

It is possible that I have the syntax wrong, as I was unable to find an example.


jamesls:
Merged, thanks for the PR!  I made a small change and added a unit test but otherwise this looked good.

PR #414, opened 2014-12-16
Refactor to decouple auth/endpoints and unify events.

danielgtaylor:
Refactor to decouple auth/endpoints and unify events.

This change decouples authentication from the `Endpoint` class and
introduces a new `RequestSigner` class to handle the signing of
requests. It fires the following events:

* `choose-signer.service.operation`: Allow overriding signature version
* `before-sign.service.operation`: Allow mutating request before signing

This also fixes the EC2 encrypted snapshot copying customization to work
with clients, as well as implements the changes needed for boto/boto3#39
(signature version overrides) and boto/boto3#30 (anonymous operations)
for the Botocore client interface.

The previous `before-auth` event is now removed, and all handlers that
previously used an endpoint's `.auth` property are now switched to
use the request signer. This may require changes upstream in the AWS CLI
as it is a potentially breaking change.

Additionally, the event emitter now supports an `emit_until_response`
method that will **stop** processing handlers once the first non-`None`
response has been received.

Tests have been updated/moved/removed accordingly.

cc @jamesls, @kyleknap


botocore/client.py
```diff
@@ -317,4 +349,5 @@ def __init__(self, events):
         self.events = events
 
     def __copy__(self):
-        return ClientMeta(**copy.deepcopy(self.__dict__))
```
kyleknap: What was the purpose in changing this line from deepcopy to copy?

botocore/endpoint.py
```diff
@@ -152,16 +130,13 @@ def _create_request_object(self, request_dict):
                              headers=headers)
         return request
 
-    def prepare_request(self, request, signer):
-        if signer is not None:
-            with self._lock:
```
kyleknap: Will the lock be needed in the RequestSigner class? I can see the case where a client is shared across multiple threads.

botocore/client.py
```diff
@@ -170,33 +172,49 @@ def _load_service_model(self, service_name):
...
+        endpoint = endpoint_creator.create_endpoint(
             service_model, region_name, is_secure=is_secure,
             endpoint_url=endpoint_url, verify=verify,
-            credentials=creds,
             response_parser_factory=self._response_parser_factory)
         response_parser = botocore.parsers.create_parser(protocol)
+
+        # Get endpoint heuristic overrides before creating the
+        # request signer.
+        resolver = endpoint_creator.resolver
```
kyleknap: Don't we already have the endpoint resolver through ``self._endpoint_resolver``? That way you do not need to make it a public attribute for ``EndpointCreator``.

botocore/endpoint.py
```diff
@@ -152,16 +130,13 @@ def _create_request_object(self, request_dict):
...
-                # Parts of the auth signing code aren't thread safe (things
-                # that manipulate .auth_path), so we're using a lock here to
-                # prevent race conditions.
-                event_name = 'before-auth.%s' % self._endpoint_prefix
-                self._event_emitter.emit(
-                    event_name, endpoint=self, request=request, auth=signer)
-                signer.add_auth(request=request)
+    def prepare_request(self, request, operation_model=None):
+        if operation_model:
+            event_name = 'request-created.{endpoint_prefix}.{op_name}'.format(
```
kyleknap: It is kind of weird that the event is called ``requested-created`` and it is located in the ``prepare_request`` method. Maybe make it ``request-prepared``? If you really need the ``request-created`` event maybe put it in the ``create_request`` method for the sake of making the code easier to understand.

botocore/handlers.py
```diff
@@ -275,8 +276,16 @@ def signature_overrides(service_data, service_name, session, **kwargs):
...
+        return signature_version_override
+
+
+def disable_signing(**kwargs):
+    """
+    This handler disables request signing by setting the signer
+    name to an empty string, similar to how the signature
+    overrides above work.
+    """
+    return ''
```
kyleknap: I think I made a change such that disable signing is by specifying ``None``. Default signing is then specified by ``NOT_SET`` which is a global object specified in ``endpoint.py``

botocore/handlers.py
```diff
@@ -265,9 +266,9 @@ def _register_for_operations(config, session, service_name):
                          handler, unique_id=unique_id)
 
 
-def signature_overrides(service_data, service_name, session, **kwargs):
-    scoped_config = session.get_scoped_config()
-    service_config = scoped_config.get(service_name)
+def signature_overrides(event_name, config, signature_version, **kwargs):
+    service_name = event_name.split('.')[1]
```
kyleknap: I am not sure if you can rely on parsing ``event_name``.  The service name used in this event comes from the model, and before it came from whatever you specified from ``get_service()`` call from the session. So, for like ``elb`` the event emitted will most likely be ``elasticloadbalancing`` because that is how it is defined in the model. I have not confirmed that is the case. I would need to check it.

botocore/operation.py
```diff
@@ -77,6 +80,30 @@ def call(self, endpoint, **kwargs):
...
         request_dict = self.build_parameters(**kwargs)
+
+        service_name = self.service.service_name
+        service_model = self.session.get_service_model(service_name)
+        credentials = self.session.get_credentials()
+        scoped_config = self.session.get_scoped_config()
+        event_emitter = self.session.get_component('event_emitter')
+        resolver = self.session.get_component('endpoint_resolver')
+        scheme = endpoint.host.split(':')[0]
+        if endpoint.region_name is None:
```
kyleknap: This code path seems pretty similar to what it was in clients. If you can consolidate some of the code into a utils function that would be nice. But if it is not easy, do not bother since we will be deprecating operations in the near future.

botocore/operation.py
```diff
@@ -86,8 +113,25 @@ def call(self, endpoint, **kwargs):
...
-                          operation=self)
+                          operation=self,
+                          request_signer=signer)
+
+        # This is kind of ugly, but it lets us hook into the request
+        # created event on a per-operation basis without mucking with
+        # the Endpoint/EndpointCreator objects.
+        real_emit = event_emitter.emit
+        def fake_emit(name, **kwargs):
+            if name.startswith('request-created'):
```
kyleknap: I am still confused why we are unable to hook into ``request-created`` from operation objects. Could you elaborate? I guess I am not seeing why you cannot just register to the ``request-created`` event. Is it because operation objects share event_emitters across sessions when clients have copies of the event emitter.

tests/integration/test_s3.py
```diff
@@ -483,7 +483,12 @@ def test_can_retrieve_presigned_object(self):
             region_name='us-east-1', service_name='s3', expires=60)
         op = self.service.get_operation('GetObject')
         params = op.build_parameters(bucket=self.bucket_name, key=key_name)
-        request = self.endpoint.create_request(params, signer)
+        def sign(request):
+            signer.add_auth(request)
+        print(self.endpoint)
```
kyleknap: remove print statements

botocore/operation.py
```diff
@@ -77,6 +80,30 @@ def call(self, endpoint, **kwargs):
                           model=self.model,
                           params=kwargs)
         request_dict = self.build_parameters(**kwargs)
+
+        service_name = self.service.service_name
```
jamesls: This is really confusing and hard to follow.  Can you explain why this code is necessary?  Also, given you're already passing in an endpoint, it has potential to be wrong because this is relying on the fact that it can construct the internals of an Endpoint the exact same as the passed in `endpoint` object, which may not be the case.

botocore/operation.py
```diff
@@ -86,8 +113,25 @@ def call(self, endpoint, **kwargs):
...
                           params=request_dict,
-                          operation=self)
+                          operation=self,
+                          request_signer=signer)
+
+        # This is kind of ugly, but it lets us hook into the request
+        # created event on a per-operation basis without mucking with
+        # the Endpoint/EndpointCreator objects.
+        real_emit = event_emitter.emit
+        def fake_emit(name, **kwargs):
```
jamesls: Not a fan of this.  This has a number of problems, and is also not thread safe.  Also, as a best practice, we should not be monkey patching code like this.  Let's come up with a cleaner way to do this.

tests/integration/test_s3.py
```diff
@@ -739,7 +744,8 @@ def test_verify_can_switch_sigv4(self):
         service = self.session.get_service('s3')
         endpoint = service.get_endpoint('us-east-1')
         # The set_config_variable should ensure that we use sigv4 for s3.
-        self.assertIsInstance(endpoint.auth, botocore.auth.S3SigV4Auth)
+        # TODO: How do we check this? It's not exposed anywhere anymore...
+        #self.assertIsInstance(endpoint.auth, botocore.auth.S3SigV4Auth)
```
jamesls: Is this still an issue?  Are you just saying that this is technically supported, we just have no way to actually verify this anymore?  If so we should probably get a test in place.  We want to ensure we don't regress on this.

kyleknap: You could try making a call with the S3Sigv4 signer mocked out and determine if the signer's ``add_auth`` method was called.

kyleknap: Wait, this is a integration test. Scratch my suggestion to mock stuff out. Just make a call using KMS. Sigv4 is required for that and you will get errors if you do not use it.

botocore/operation.py
```diff
@@ -86,8 +113,25 @@ def call(self, endpoint, **kwargs):
         self.session.emit(event, endpoint=endpoint,
                           model=self.model,
                           params=request_dict,
-                          operation=self)
+                          operation=self,
+                          request_signer=signer)
```
jamesls: Why is the request_signer added to before-call kwargs?  Seems unnecessary to me.

tests/unit/test_session.py
```diff
@@ -367,6 +367,15 @@ def test_credential_provider_not_called_when_creds_provided(self):
                          "explicit credentials were provided to the "
                          "create_client call.")
 
+class TestPerformOperation(BaseSessionTest):
+    def test_s3(self):
```
kyleknap: Is there anything asserted here?

kyleknap:
Looks good overall. My comments came from questions in how the code works in certain places and edge cases that I was not sure if were caught or not.

botocore/client.py
```diff
@@ -170,33 +172,49 @@ def _load_service_model(self, service_name):
...
+        endpoint_config = resolver.construct_endpoint(
+                service_model.endpoint_prefix,
+                region_name, scheme=scheme)
+        region_name = endpoint_config.get('properties', {}).get(
+            'credentialScope', {}).get('region', region_name)
+        signature_version = service_model.signature_version
+        if 'signatureVersion' in endpoint_config.get('properties', {}):
+            signature_version = endpoint_config['properties']\
+                                               ['signatureVersion']
+        signer = RequestSigner(service_model.service_name, region_name,
```
jamesls: After getting a better understanding of how this is used, it seems like scoped_config is only needed in the request signer to emit an event.  And in practice, this is only used so we can change the signature version based on the configuration values.

What if we moved this logic here?  This should be a fixed thing per session and does not need to be recalculated each time we sign a request.

jamesls:
Overall I think there were only a few questions/concerns I had:

1. The issue with recreating some of the endpoint dependencies in the operation.call method
2. The RequestSigner needing the scoped_config when this can be handled outside of that object (I'm ok if the signature_override from config is no longer a handler)
3.  The request signer being sent in an event which appears to be only to support the copy snapshot handler.  I'd rather not complicate that event just to support a single specific handler.  Perhaps there's another way of writing the handler that keeps everything simple.  Maybe there's another event besides `before-call` we can use.
4. The monkey patch.  I'd really recommend exploring alternatives.  Thinking out loud: 1) Just share handlers for signing (ensure it's thread safe) 2) Make the RequestSigner not tied to a specific service 3) Sign the request directly in operation (maybe don't use make_request) 4) Clone the event handlers when we create a service object, etc.  I haven't really thought through each of those options, but I think we should avoid monkey patching if we can avoid it.

Overall though, I think this is definitely heading in the right direction.

botocore/client.py
danielgtaylor: This was something that @jamesls was also going to do. The event emitter can't be deep copied (things like regexs and partials raise an error when you try), so we have to use a simple copy instead. Unfortunately it means we'll need to maintain this method - anytime you add something to `Meta` you need to make sure it gets copied properly in this method.

botocore/client.py
danielgtaylor: You're right, we do. Nice catch :+1:

botocore/client.py
danielgtaylor: I agree that it could be done here instead of as a customization, but I'm concerned that there might be other use cases for having access to the config in a customization. I'll think about it a bit, and if I don't come up with anything I'll move the logic here. We can always add the config to the events later if we need it, I suppose.

botocore/endpoint.py
danielgtaylor: I believe this was only needed for the `fix_s3_host` customization. It was updated to be thread-safe, and now we should no longer need a lock. Maybe @jamesls can elaborate?

jamesls: The develop branch should be thread safe, as auth_path is now set on the request instead of the auth class.  I'd still suggest double checking that this is the case though.

botocore/endpoint.py
danielgtaylor: Spoke in person, and I'm going to refactor this to create a request object for each retry. Then the event will be moved into `create_request` and everything will make sense, plus we have less of a chance of stale values causing issues later on.

botocore/handlers.py
danielgtaylor: Agreed. I've moved this code around a bit due to some other comments, but I also switched it to using `service_model.endpoint_prefix` instead.

botocore/handlers.py
danielgtaylor: It's still checking against `None` here:

https://github.com/danielgtaylor/botocore/blob/auth-refactor/botocore/signers.py#L79-L84

Can you elaborate a bit on what you changed?

tests/unit/test_session.py
danielgtaylor: Just that no exceptions are raised, i.e. is it possible to make a call with an operation. Apparently we didn't have any such test in place. Crazy!

I could write a bunch of tests for operation objects, but with their eminent removal it seems like a waste of time.

tests/integration/test_s3.py
danielgtaylor: This is now tested elsewhere by mocking out the RequestSigner and sending in different config files. I'm removing this test, unless you specifically want to read a file from disk and mock it out here instead. Thoughts?

tests/integration/test_s3.py
danielgtaylor: Done.

botocore/handlers.py
kyleknap: Yeah here is the PR: https://github.com/boto/botocore/pull/433

So in the CLI to turn off signing, you change the ``Service`` object's ``signature_version`` attribute to ``None``. So when an endpoint is created using the ``Service`` object it will pass through ``None`` to ``EndpointCreator``'s ``create_endpoint`` call to signify not to look for a signature version in the endpoint resolver or model:
https://github.com/boto/botocore/pull/433/files#diff-054181c44857a41f4e27bab9c8a6ba6eR333

But, I suppose both ``None`` and ``''`` will work for specifying no signing. Now that I think about it, the second part of my comment is a little off topic. What I really was getting at is it would be nice that we are consistent on what value disables signing ``None`` or ``''``.

tests/unit/test_session.py
kyleknap: Really? Interesting. In that case, I would just add an assertion of the response at the end to ensure the mocked response is properly returned by the ``call()``. I guess we were just relying on integration test for that.

botocore/operation.py
danielgtaylor: The endpoint which is passed in contains no auth information. The endpoint heuristic resolver is used to load an endpoint config based on the passed in endpoint, which in turn may override auth information before we create the request signer. In this case the information in `endpoint_config` cannot be retrieved from `endpoint`, because it doesn't exist there anymore. Clients have the same logic in them.

The code is necessary because we need to know when to override the auth based on things like the hostname (e.g. switch to sigv4 for eu-central-1). The endpoint no longer has any associated auth information, so instead we do this after an endpoint has been created but before the request signer is created. The endpoint is still used to make the HTTP(S) request, but the request signer, which is created after checking for these overrides, is the one that signs the request before going out over the wire.

botocore/operation.py
danielgtaylor: It's a copy of what's in clients, more of less. I'm on the fence about whether to try and make a utility method, because the client's version of this code is pretty clean right now.

botocore/client.py
danielgtaylor: Okay I can't think of a specific reason to keep this as a custom handler, so I've moved the code.

tests/integration/test_s3.py
kyleknap: What is even the purpose of this test? It is called ``test_can_retrieve_presigned_object`` but it never actually tries to retrieve the object using the presigned url. It just checks that sigv4 was used in the signing.

If its only purpose is to check if sigv4 is used, +1 on removing it. If we actually want to grab the object using the presigned url, we should keep the test and do exactly that.

botocore/operation.py
kyleknap: That's fine in that case.

tests/unit/test_session.py
danielgtaylor: Updated the test to include that assertion.

botocore/operation.py
danielgtaylor: I'm updating this now to register for the specific request-created.service.operation method, sign in a threadsafe fasion, do the request(s), then unregister. I think it's a nicer solution than the monkey patch. Thanks!

botocore/operation.py
danielgtaylor: I've redone this portion of the code. Please have another look as it now uses the public register/unregister functionality and should be thread-safe.

kyleknap:
Looks good now. :ship: Just as a reminder could you run the CLI's unit and integration tests before this gets merged so we can ensure functionality of the CLI.

botocore/operation.py
```diff
@@ -86,8 +127,28 @@ def call(self, endpoint, **kwargs):
...
+
+        # Here we register to the specific request-created event
+        # for this operation. Since it's possible to run the same
+        # operation in multiple threads, we used a lock to prevent
+        # issues. It's possible a request will be signed more than
+        # once. Once the request has been made, we unregister the
+        # handler.
+        def request_created(request, **kwargs):
+            with threading.Lock():
```
jamesls: Are we absolutely sure there isn't a problem with mutating the request as it's in flight through the requests/httplib code?  Seems like this is very sensitive to race conditions.  Also we read through the request body as we sign the code.  I assume this could cause issues if we're also trying to send the request at the same time.

This also means that the number of times this request is proportional to the concurrency level which I'm not really a big fan of.  I assume for things like sigv4 payload checksums this can get expensive.

botocore/client.py
```diff
@@ -170,33 +176,62 @@ def _load_service_model(self, service_name):
...
+                region_name, scheme=scheme)
+        region_name = endpoint_config.get('properties', {}).get(
+            'credentialScope', {}).get('region', region_name)
+        signature_version = service_model.signature_version
+        if 'signatureVersion' in endpoint_config.get('properties', {}):
+            signature_version = endpoint_config['properties']\
+                                               ['signatureVersion']
+
+        # Signature overrides from a configuration file
+        if scoped_config is not None:
```
jamesls: Can we break this up into multiple methods?  Not a huge fan of long methods.  As a heuristic, every block of code here that starts with a comment explaining what it does is a good candidate for a method.

jamesls:
Overall it looks good.  I'd suggest cutting down on method length in both the operation.py and client.py modules.  But my main concern is the operation.call().  I'm still concerned that this code has race conditions and performance implications.  Can we come up with an alternate way that does not have these problems?

jamesls:
:shipit:  Looks good.  Thanks for driving this to completion.

PR #432, opened 2015-01-07
Initial commit of glacier support in botocore

jamesls:
In addition to including the JSON model for the service,
there were also three customizations needed:

* Add the x-amz-glacier-version header to all glacier requests
* Add a default value of '-' for the accountId param if it's not
  provided
* Add (and compute) the x-amz-content-sha256 and the
  x-amz-sha256-tree-hash header (also the "checksum" param)
  automatically for the user.

I verified I got the same tree hash values from boto2 and I
also added some integration tests to double check this.  The
tree hash is somewhat similar to boto's version, with some
slight performance improvements (about ~33% faster tree hashes).

We'll need to make a corresponding change to the the CLI to
make the --account-id param optional since botocore will
fill that value in.

cc @kyleknap

### DIFF
--- a/botocore/compat.py
+++ b/botocore/compat.py
@@ -36,6 +36,7 @@ if six.PY3:
     from io import IOBase as _IOBase
     from base64 import encodebytes
     from email.utils import formatdate
+    from itertools import zip_longest
     file_type = _IOBase
     zip = zip
 
@@ -72,6 +73,7 @@ else:
     from email.Utils import formatdate
     file_type = file
     from itertools import izip as zip
+    from itertools import izip_longest as zip_longest
     from httplib import HTTPResponse
     from base64 import encodestring as encodebytes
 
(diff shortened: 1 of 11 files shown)


kyleknap:
This looks good. :ship: It be cool if we automatically calculated the sha256 tree-hash for CompleteMultipartUpload. But we would need to do something like we do for the s3 commands for the CLI, where we automatically handle multipart uploads.

I am also not sure how weird it is that you can specify the checksum in the input for UploadArchive and UploadMulitpartPart. Do you happen to know if the checksum specified by the user is used or if the checksum automatically calculated is used when the user specifies a checksum?

jamesls:
Yeah we should add some conditional logic there for specifying the checksum.  We only need to calculate the checksums if the user did not provide one for us.  That way if the user has precomputed this value we can avoid calculating it twice.  I'll update this.

jamesls:
@kyleknap Added unit tests for that case.  The code already handles this case, it just didn't have a unit test for it.

kyleknap:
Ok cool. I did not originally see that in the code. Thanks for the tests. :ship:

PR #433, opened 2015-01-12
Add option to specify endpoint's signature version

kyleknap:
Add option to pass ``signature_version`` to EndpointCreator's ``create_endpoint`` method.

When ``get_endpoint`` is called from Service objects, it will pass in its signature version to ``create_endpoint``. That is required for the CLI's ``--no-sign-request`` flag to work because the handler changes the service object's signature version.

I will be sending a PR out real soon for the CLI, that will add a test to catch that --no-sign-request is adhered to.

cc @jamesls @danielgtaylor

### DIFF
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -35,6 +35,7 @@ from botocore import parsers
 
 logger = logging.getLogger(__name__)
 DEFAULT_TIMEOUT = 60
+NOT_SET = object()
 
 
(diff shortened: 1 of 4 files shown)


kyleknap:
@jamesls Test now pass. I added a flag to check if the service object is using a custom signature version.

jamesls:
:shipit:

PR #434, opened 2015-01-14
Fix canonical query string sorting for Signature Version 4

jamesls:
We were previously sorting by first creating a list of strings
of `'%s=%s' % (key, value)`.  But really we should be:

1. Sorting by key names
2. In the case of repeated keys, sort then by values.

Fixes aws/aws-cli#976

Added unit tests.  Also verified the previously failing CLI command now works:

```
$ aws cloudsearchdomain search --endpoint-url myendpoint --search-query 'George Lucas' --query-options '{"defaultOperator": "and", "fields":["directors^10"]}' --query hits.hit[].fields.[title[0],rating[0]] --output table
---------------------------------------------------------
|                        Search                         |
+------------------------------------------------+------+
|  Star Wars                                     |  8.7 |
|  Star Wars: Episode III - Revenge of the Sith  |  7.7 |
|  Star Wars: Episode I - The Phantom Menace     |  6.5 |
|  Star Wars: Episode II - Attack of the Clones  |  6.7 |
|  American Graffiti                             |  7.5 |
|  THX 1138                                      |  6.7 |
+------------------------------------------------+------+
```

cc @kyleknap @danielgtaylor

### DIFF
--- a/tests/unit/auth/test_signers.py
+++ b/tests/unit/auth/test_signers.py
@@ -290,6 +290,31 @@ class TestS3SigV4Auth(unittest.TestCase):
         self.assertEqual('marker=%C3%A4%C3%B6%C3%BC-01.txt&prefix=', cqs)
 
 
+class TestSigV4(unittest.TestCase):
+    def setUp(self):
+        self.credentials = botocore.credentials.Credentials(
+            access_key='foo', secret_key='bar')
+
+    def test_canonical_query_string(self):
+        request = AWSRequest()
+        request.url = (
+            'https://search-testdomain1-j67dwxlet67gf7ghwfmik2c67i.us-west-2.'
+            'cloudsearch.amazonaws.com/'
+            '2013-01-01/search?format=sdk&pretty=true&q=George%20Lucas&'
+            'q.options=%7B%22defaultOperator%22%3A%20%22and%22%2C%20%22'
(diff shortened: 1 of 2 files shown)


tests/unit/auth/test_signers.py
```diff
@@ -290,6 +290,31 @@ def test_query_string_params_in_urls(self):
...
+    def setUp(self):
+        self.credentials = botocore.credentials.Credentials(
+            access_key='foo', secret_key='bar')
+
+    def test_canonical_query_string(self):
+        request = AWSRequest()
+        request.url = (
+            'https://search-testdomain1-j67dwxlet67gf7ghwfmik2c67i.us-west-2.'
+            'cloudsearch.amazonaws.com/'
+            '2013-01-01/search?format=sdk&pretty=true&q=George%20Lucas&'
```
kyleknap: Is this query params already sorted here? It may be beneficial to rearrange some of them to ensure they get sorted.

kyleknap:
Looks good. :ship: I just had a comment about making the query string unsorted to begin with in the test in order to better avoid any regressions.

tests/unit/auth/test_signers.py
jamesls: Good points I'll update.

jamesls:
Test updated.

kyleknap:
Thanks! :ship:

PR #442, opened 2015-01-26
Add compat for legacy operation methods on waiters

jamesls:
While pulling in https://github.com/boto/botocore/pull/437,
I noticed that we aren't handling the 'error' matcher properly
because of a CLI customization that will raise an error on
non 200 responses.  This handles that case so, for example, the
command
`aws rds wait db-instance-deleted --db-instance-identifier foo`
will work as expected.  Previously, the error was being propogated
and the CLI command would exit with a non-zero RC and a client error
message.

cc @kyleknap @danielgtaylor

### DIFF
--- a/botocore/waiter.py
+++ b/botocore/waiter.py
@@ -96,9 +96,27 @@ class LegacyOperationMethod(object):
         self._endpoint = endpoint
 
     def __call__(self, **kwargs):
-        http, parsed = self._operation_object.call(
-            self._endpoint, **kwargs)
+        try:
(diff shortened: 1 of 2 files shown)


kyleknap:
This looks good. :ship: As a note if you want to add an integration test for this in the CLI, ``bucket-not-exists`` and ``object-not-exists`` commands would be easy testing options.

jamesls:
Sounds good.  I'll send a PR for a CLI integration test.

PR #443, opened 2015-01-26
Add waiters for elasticache

Jc2k:
This adds 4 new waiters i've needed recently. I've used the same delays/attempts as RDS, which works for me quite well so far.

### DIFF
--- a/botocore/data/aws/elasticache/2014-09-30.waiters.json
+++ b/botocore/data/aws/elasticache/2014-09-30.waiters.json
@@ -44,8 +44,8 @@
       "maxAttempts": 60,
       "acceptors": [
         {
-          "expected": 404,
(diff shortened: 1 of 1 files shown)


jamesls:
Thanks for adding these.  The one thing I'd suggest changing is instead of using the 404 status code checker we use the error code returned by the service.  There was a bug in botocore that prevented us from using this before, but this issue is now fixed.  For example here's this change for RDS (https://github.com/boto/botocore/blob/develop/botocore/data/aws/rds/2014-10-31.waiters.json#L64-L68).

Jc2k:
I saw you released the fix for that, and so quickly too! Cheers! I've made the change you requested.

jamesls:
Merged, thanks!

PR #444, opened 2015-01-27
Use GMT for timestamp format

jamesls:
This changes our date serialization to
use an IMF-fixdate timestamp in headers for the rest serializers.

RFC 7231 describes the preferred date-time format in HTTP messages as an
IMF-fixdate timestamp. This timestamp format is a subset of the format
described in RFC 5322 except the format always used a fixed timezone offset of
"GMT". S3 also uses this timestamp format in their documentation and in the
response headers they return.

cc @kyleknap @danielgtaylor

### DIFF
--- a/botocore/serialize.py
+++ b/botocore/serialize.py
@@ -137,7 +137,7 @@ class Serializer(object):
         return int(calendar.timegm(value.timetuple()))
 
     def _timestamp_rfc822(self, value):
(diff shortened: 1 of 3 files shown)


danielgtaylor:
LGTM :+1:

kyleknap:
:ship:

PR #445, opened 2015-01-28
Account for squid proxy adding unexpected headers

jamesls:
Most servers (including S3) will just return the CLRF after the 100
continue response.  However, some servers (I've specifically seen this
for squid when used as a straight HTTP proxy) will also inject a
Connection: keep-alive header.  To account for this we'll read until we
read '\r\n', and ignore any headers that come immediately after the 100
continue response.

Fixes aws/aws-cli#1116

I've also added an additional commit, https://github.com/jamesls/botocore/commit/7ca5d35c1b3af2872d16800eb7a3fceefcf1e863, that cleans up the pep8 formatting.

cc @kyleknap @danielgtaylor

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -154,6 +154,18 @@ class AWSHTTPConnection(HTTPConnection):
             # we must run the risk of Nagle.
             self.send(message_body)
 
+    def _consume_headers(self, fp):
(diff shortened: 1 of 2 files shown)


danielgtaylor:
LGTM :ship:-it!

kyleknap:
Same :ship:

PR #446, opened 2015-01-30
Handle capitalized keys in error responses

jamesls:
For rest-json, we should handle the case where
the keys can come back as either code/Code and
message/Message.

cc @kyleknap @danielgtaylor


kyleknap:
LGTM :ship:

PR #447, opened 2015-02-02
fix #420 - use path from client URL

### LINKED ISSUE #420: Not compatible with endpoints that have a url_path (regression from boto)

Andrey-mp:
Because default url is '/' then urlparse library uses it for creating request.

### DIFF
--- a/botocore/data/aws/elasticache/2014-09-30.waiters.json
+++ b/botocore/data/aws/elasticache/2014-09-30.waiters.json
@@ -1,0 +1,139 @@
+{
+  "version": 2,
+  "waiters": {
+    "CacheClusterAvailable": {
+      "delay": 30,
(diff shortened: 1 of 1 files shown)


jamesls:
Not opposed to supporting this.  However, it looks like this is causing some failing tests which will need to be addressed before this is merged.  Also, as per our [contributing guidelines](https://github.com/boto/botocore/blob/develop/CONTRIBUTING.rst), bug fixes need to have tests added both to ensure the bug is fixed and to prevent regressions in the future.  Once these two issues have been fixed I'd be happy to merge this.

Andrey-mp:
Ok, I wrote two tests and made some cleanup in the test's code.
But I don't understand idea why this default URL is needed...
I didn't find a way how I can rewrite it in my client code.

PR #448, opened 2015-02-03
Implement opt-in unsigned requests for clients.

danielgtaylor:
This change makes it possible to override the signature version at client
creation time using a parameter. It also introduces the concept of an
advanced configuration object for clients:

```python
import botocore

config = Config(signature_version=botocore.UNSIGNED)
s3 = session.create_client('s3', config=config)
```

This does **not** move any existing parameters into the advanced
configuration object, and thus is not yet a breaking change. It
does include the following:

* Adds `botocore.UNSIGNED` as a sentinel to disable signing
* Updates `botocore.handlers.disable_signing` to use `botocore.UNSIGNED`
* Adds a `botocore.client.Config` object to store signature version
* Modifies client creation to use this new config
* Updated / new unit tests

cc @jamesls @kyleknap


tests/unit/test_session.py
```diff
@@ -367,6 +367,15 @@ def test_credential_provider_not_called_when_creds_provided(self):
                          "explicit credentials were provided to the "
                          "create_client call.")
 
+    @mock.patch('botocore.client.ClientCreator')
+    def test_config_passed_to_client_creator(self, client_creator):
+        config = mock.Mock()
```
jamesls: We've discussed in the past not mocking value objects.  Let's just use the actual Config object here.

botocore/__init__.py
```diff
@@ -51,6 +51,9 @@ def emit(self, record):
 
 BOTOCORE_ROOT = os.path.dirname(os.path.abspath(__file__))
 
+# Used to specify anonymous (unsigned) request signature
+UNSIGNED = ''
```
jamesls: We've been using an sentinel value of `object()` for these values.  It allows us to use identity checks instead of equality checks to avoid certain types of errors.  I'd prefer if we kept to that convention unless there's a strong reason not to.

tests/unit/test_client.py
```diff
@@ -111,6 +112,16 @@ def test_client_signature_override(self, request_signer):
         request_signer.assert_called_with(
             mock.ANY, mock.ANY, mock.ANY, 'foo', mock.ANY, mock.ANY)
 
+    @mock.patch('botocore.client.RequestSigner')
```
jamesls: This might not be possible, but can we add a test that's more "direct"?  I see that this test is essentially checking that we forward the value of config.signature_version to the request signer, but it'd be great if we had something that verified, if you set the signature version to UNSIGNED, then we do not sign the request.  I get there's another level of indirection here with the RequestSigner and the auth classes, but it seems possible to regress on this feature without any unit tests failing because we're never actually testing anything that explicitly sets the value to `botocore.UNSIGNED`.

botocore/__init__.py
danielgtaylor: That is a backward-incompatible change for the signature version events fired from the `RequestSigner`, which is why I kept the empty string.

tests/unit/test_client.py
danielgtaylor: I'll look into adding a more comprehensive test for this.

botocore/__init__.py
jamesls: Is the auth override stuff even in a public release yet?  If the only reason for this is that it's not backwards compatible, and not that there's an issue with using `object()` sentinel's, I'd vote for changing this to be consistent.  Pretty sure the impact here is minimal, if any.

tests/unit/test_client.py
danielgtaylor: So there is another test that covers some of this:

https://github.com/boto/botocore/blob/develop/tests/unit/test_signers.py#L117-L127

I've also added a new test to ensure the request is attempted without any auth headers/query args.

danielgtaylor:
@jamesls @kyleknap please have another look :+1:

botocore/signers.py
```diff
@@ -86,7 +87,7 @@ def sign(self, operation_name, request):
             signature_version=signature_version, request_signer=self)
 
         # Sign the request if the signature version isn't None or blank
-        if signature_version:
+        if signature_version != botocore.UNSIGNED:
```
kyleknap: Will this affect the CLI? We set the endpoint's signature version to ``None``. You can check real quick by running this test: https://github.com/aws/aws-cli/blob/develop/tests/integration/customizations/s3/test_plugin.py#L1594

kyleknap:
This looks good to me. :ship: If the integration test that I pointed out does fail, a subsequent pull request will have to be made to change ``None`` to ``UNSIGNED`` when using the ``--no-sign-request`` option.

danielgtaylor:
@kyleknap good call. I've created an associated pull request with the CLI for this. After it's been reviewed I'll merge in both.

botocore/signers.py
jamesls: Given it's now a sentinel, we should be using identity checks instead of equality check (`signature_version is not botocore.UNSIGNED`)

tests/unit/test_signers.py
```diff
@@ -118,7 +119,8 @@ def test_disable_signing(self):
         # Returning a blank string from choose-signer disabled signing!
```
jamesls: This comment is no longer accurate.

jamesls:
Looks good, just a couple small things.  :shipit:

PR #449, opened 2015-02-03
Update docs for IAM

kyleknap:
This updates the documentation for IAM. Most of the changes look like the removal of doc strings on the top level request and response shapes, but we do not use those so that should be fine.

cc @jamesls @danielgtaylor


jamesls:
:shipit:

Issue #266, opened 2014-03-28
botocore explodes when installed alongside pyOpenSSL, ndg-httpsclient and pyasn1 


bhuber:
I'm not an expert with boto or botocore, so apologies in advance if this is a duplicate or whatever, but I can't find anything related in the issues on GH.  To reproduce, create a new virtualenv and do the following:

```sh
pip install pyOpenSSL==0.14 ndg-httpsclient==0.3.2 pyasn1==0.1.7 botocore==0.38
python
>>> import botocore.vendored.requests
```

Output:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/bhuber/envs/sandbox/local/lib/python2.7/site-packages/botocore/vendored/requests/__init__.py", line 53, in <module>
    from .packages.urllib3.contrib import pyopenssl
  File "/home/bhuber/envs/sandbox/local/lib/python2.7/site-packages/botocore/vendored/requests/packages/urllib3/contrib/pyopenssl.py", line 55, in <module>
    orig_connectionpool_ssl_wrap_socket = connectionpool.ssl_wrap_socket
AttributeError: 'module' object has no attribute 'ssl_wrap_socket'
>>> 
```

This is a direct manifestation of a bug closed in requests==2.1.0: https://github.com/kennethreitz/requests/issues/1732 .  Updating the requests lib would probably fix it.

As a workaround, ```pip install pyopenssl==0.12``` currently fixes it for me.

guss77:
I have the same problem with installing awscli on Ubuntu 15.04. I worked around by doing:

    sudo apt-get purge python-ndg-httpsclient
    sudo pip install ndg-httpsclient==0.3.1

Rowanto:
Thanks @guss77 . I am using Debian testing, and those commands also fixed my eb cli. :+1:

jdanbrown:
:boom: Thanks @jamesls!
- Works for me with `awscli==1.7.7` + `botocore==0.88.0` + `requests==2.5.1`
- Works with either `ndg-httpsclient==0.3.3` or `0.3.2`, but not with `0.3.1` like in my comment above

```sh
$ python --version
Python 2.7.7

$ pip freeze | ack -i '^(awscli|botocore|pyopenssl|pyasn1|ndg-httpsclient|requests)\='
awscli==1.7.7
botocore==0.88.0
ndg-httpsclient==0.3.3
pyOpenSSL==0.14
pyasn1==0.1.7
requests==2.5.1

[...]

PR #451, opened 2015-02-05
Upgrade Requests dependency to 2.5.1

jamesls:
This is the latest request release.  It was relatively painless, with the exception of one change to how redirects are handled.

>  In newer versions of requests, the Authorization
header is removed if a redirect occurs where the new
host does not equal the original host.  In the case
of S3 redirects, we know that this is ok so we need
to add back the old behavior.  We do this by subclassing
the session and overriding its rebuild_auth() class to
be a no-op.  That is, it just keeps the auth from the
original request.

Only new change I added is https://github.com/jamesls/botocore/commit/0c9790d521e69be581057ccc40f804733ec58daa

Verified clean runs of the botocore/aws-cli unit and integration tests.

cc @kyleknap @danielgtaylor


jamesls:
Should also fix https://github.com/boto/botocore/issues/266

kyleknap:
LGTM :ship:

PR #453, opened 2015-02-09
Fix failing integ tests due to requests/urllib3 upgrade

jamesls:
We need to ignore warnings about missing SAN's in SSL certs
in py26.  This is due to a python bug and we have to validate
using the CN.

These actually were caught by AWS CLI integ tests.  Given they're
warnings, they don't actually cause anything in botocore to hard
fail, whereas the CLI is very sensitive to what shows up in stdout
and stderr which explains why the CLI tests are more likely to
catch this.

I've also updated our integ smoke tests in botocore.  We now will
also verify that the smoke tests generate no warnings.  This will
catch any regressions against this issue.

cc @kyleknap @danielgtaylor

### DIFF
--- a/botocore/compat.py
+++ b/botocore/compat.py
@@ -119,10 +119,31 @@ if sys.version_info[:2] == (2, 6):
     # will raise a plain old SyntaxError instead of
     # a real exception, so we need to abstract this change.
     XMLParseError = SyntaxError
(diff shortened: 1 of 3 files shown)


kyleknap:
Looks good. :ship:

PR #455, opened 2015-02-11
Add retry events to the client interface.

danielgtaylor:
This change loads the retry configuration and adds retry events and
handlers to the client interface. When calling `session.create_client(...)`
and the model is loaded the events are set up on the client's own
event emitter.

Various tests have been added to ensure the events and handlers are set
up properly. The handlers themselves are tested elsewhere.

cc @jamesls @kyleknap


botocore/client.py
```diff
@@ -14,19 +14,16 @@
...
-from botocore import xform_name
+from botocore.model import ServiceModel
 from botocore.paginate import Paginator
-from botocore.utils import CachedProperty
-import botocore.validate
-import botocore.serialize
-from botocore import credentials
 from botocore.signers import RequestSigner
-from botocore.endpoint import EndpointCreator
+from botocore.utils import CachedProperty
```
danielgtaylor: This is mostly just a reorganization into alphabetical order and proper grouping. The real change is the addition of `from botocore import ..., retryhandler, translate, ...` which is used by the code below.

kyleknap:
Code looks good to me. It looks like it was just copying some code over from the handler.

But, could you add an integration test though to make sure the the "end-to-end" functionality works?

What I would do is make a test like this but using clients:
https://github.com/boto/botocore/blob/9d4ae4b57aa3c90d3d4d9e503ee6581d5d8accff/tests/integration/test_smoke.py#L141

Note that in this test, for some reason, client's are instantiated there but service and operation objects are used to run the tests.

botocore/client.py
```diff
@@ -174,8 +171,46 @@ def waiter_names(self):
     def _load_service_model(self, service_name):
         json_model = self._loader.load_service_model('aws/%s' % service_name)
         service_model = ServiceModel(json_model, service_name=service_name)
+        self._register_retries(service_model)
         return service_model
 
+    def _register_retries(self, service_model):
```
jamesls: can we break this up into multiple methods.  I've been trying to keep the client method pretty clean and avoid long methods where possible.  I'd suggest instead of a comment block, extract a method instead.

botocore/client.py
```diff
@@ -174,8 +171,46 @@ def waiter_names(self):
...
+            retry_config, endpoint_prefix)
+        unique_id = 'retry-config-%s' % endpoint_prefix
+        self._event_emitter.register('needs-retry.%s' % endpoint_prefix,
+                                     handler, unique_id=unique_id)
+
+        # There's certainly a tradeoff for registering the retry config
+        # for the operations when the service is created.  In practice,
+        # there aren't a whole lot of per operation retry configs so
+        # this is ok for now.
+        for key in retry_config:
```
jamesls: On second thought, do we even have any per operation configs?  I've vote for just removing this functionality altogether if we're not using it.

botocore/client.py
```diff
@@ -174,8 +171,46 @@ def waiter_names(self):
...
+        if not original_config:
+            return
+
+        retry_config = translate.build_retry_config(
+            endpoint_prefix, original_config.get('retry', {}),
+            original_config.get('definitions', {}))
+
+        logger.debug("Registering retry handlers for service: %s",
+                     service_model.service_name)
+        handler = retryhandler.create_retry_handler(
```
jamesls: This is going to require a patch to test.  The client code has been pretty consistent with requiring dependencies being passed in.  It makes customization cleaner and tests easier to read.  I'd like to stick to that convention.  I realize some of the recent signature changes don't follow this, but I'd like to not veer off too far in that direction.

jamesls:
I'd also agree with Kyle.  We need integration tests that verify retries are working in the client.  This was missed for a while largely in part because we didn't have any integration tests verifying that clients had retries.  I'd like to ensure we don't regress on this in the future.

botocore/client.py
jamesls: Note opposed to this change, but do note that alphabetical order is not mentioned anywhere in pep8 imports (https://www.python.org/dev/peps/pep-0008/#imports).

botocore/exceptions.py
```diff
@@ -286,6 +286,10 @@ class ClientError(Exception):
         'operation: {error_message}')
 
     def __init__(self, error_response, operation_name):
+        if 'Errors' in error_response and isinstance(error_response['Errors'],
```
jamesls: Pretty sure this is a bug in our error response parser.  We shouldn't be working around it in the ClientError `__init__`, we should just fix it at the source to prevent an invalid `error_response` from being created in the first place.

kyleknap: +1

kyleknap:
Thanks for adding the integration tests. They look good. I would also like to get to the direct cause of that exception parsing bug tracked down, fixed, and tested, before the PR gets merged.

botocore/parsers.py
```diff
@@ -386,10 +386,13 @@ def _do_error_parse(self, response, shape):
...
-        # Once we've converted xml->dict, we need to make one more
-        # adjustment to be consistent with ResponseMetadata
-        # for non-error responses:
-        # {"RequestId": "id"} -> {"ResponseMetadata": {"RequestId": "id"}}
+        # Once we've converted xml->dict, we need to make one or two
+        # more adjustments to extract nested errors and to be consistent
+        # with ResponseMetadata for non-error responses:
+        # 1. {"Errors": {"Error": {...}}} -> {"Error": {...}}
+        # 2. {"RequestId": "id"} -> {"ResponseMetadata": {"RequestId": "id"}}
+        if 'Errors' in parsed and isinstance(parsed['Errors'], dict):
```
jamesls: I'm not able to find a unit test for this change.  If there's not one we should get one in place for this.

Also, why the isinstance check for dict?  Are there cases where there's an Errors key and it's not a dict?

danielgtaylor: @jamesls I've added another test in `test_parsers.py` that covers the SDB case. Since we do not control the response I was checking for a dict to ensure that we never call `parsed.update(...)` with a non-dictionary type. This prevents awkward and potentially confusing type or value errors if a service ever returns something weird.

kyleknap:
Looks good to me. :ship:

botocore/parsers.py
jamesls: But we know what the service response is.  Unless we have a concrete example of a service doing this, I'm against hypothetical these cases.  There's _many_ places in the parsing code where we make strong assumptions about how an AWS service responds.

PR #456, opened 2015-02-11
Let exceptions propogate from make_request

jamesls:
This prevents a value of None from ever being returned
from make_request.  Previously, exceptions were caught and
after the retry logic would just return a value of None.

Now, if things like `ConnectionError`s, `socket.error`s, etc.
are raised, then this will propogate.  This also fixed the
client interface that expected that a tuple was always returned
from `make_request`.  This is now the case.

Noticed this while working on another bug fix.  Now instead
of:

```
  File "t.py", line 10, in <module>
    ContentDisposition="attachment; filename=5小時接力起跑.jpg;")
  File "botocore/botocore/client.py", line 292, in _api_call
TypeError: 'NoneType' object is not iterable
```

you'll get:

```
  File "t.py", line 10, in <module>
    ContentDisposition="attachment; filename=5小時接力起跑.jpg;")
  File "botocore/botocore/client.py", line 292, in _api_call
  File "botocore/botocore/endpoint.py", line 109, in make_request
  File "botocore/botocore/endpoint.py", line 161, in _send_request
    raise exception
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe5 in position 43: ordinal not in range(128)
```

Though keep in mind the above traceback is actually also a bug and will be fixed
shortly.

cc @danielgtaylor @kyleknap

### DIFF
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -142,10 +142,10 @@ class Endpoint(object):
     def _send_request(self, request_dict, operation_model):
         attempts = 1
         request = self.create_request(request_dict, operation_model)
-        response, exception = self._get_response(
+        success_response, exception = self._get_response(
             request, operation_model, attempts)
         while self._needs_retry(attempts, operation_model,
-                                response, exception):
+                                success_response, exception):
             attempts += 1
             # If there is a stream associated with the request, we need
             # to reset it before attempting to send the request again.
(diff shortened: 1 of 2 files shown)


jamesls:
Also fixes https://github.com/boto/boto3/issues/63

kyleknap:
Looks good. :ship:

PR #457, opened 2015-02-12
Correctly handle non-ascii header values

jamesls:
Fixes aws/aws-cli#1124

cc @kyleknap @danielgtaylor

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -123,7 +123,9 @@ class AWSHTTPConnection(HTTPConnection):
 
     def _send_output(self, message_body=None):
         self._buffer.extend((b"", b""))
-        msg = b"\r\n".join(self._buffer)
+        msg = b"\r\n".join(
+            b.encode('utf-8') if isinstance(b, six.text_type) else b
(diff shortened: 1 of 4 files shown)


botocore/awsrequest.py
```diff
@@ -123,7 +123,9 @@ def _send_request(self, method, url, body, headers):
 
     def _send_output(self, message_body=None):
         self._buffer.extend((b"", b""))
-        msg = b"\r\n".join(self._buffer)
+        msg = b"\r\n".join(
+            b.encode('utf-8') if isinstance(b, six.text_type) else b
```
kyleknap: I'd prefer if it was not all in-line as it took me a little longer to understand. Code looks fine otherwise.

kyleknap:
Looks good. :ship: Had small comment.

botocore/awsrequest.py
jamesls: Ok, extracted that out into it's own method.

kyleknap: Thanks. Looks good.

PR #460, opened 2015-02-13
fix #420. Make available to use client host URL.

### LINKED ISSUE #420: Not compatible with endpoints that have a url_path (regression from boto)

Andrey-mp:
urlparse uses '/'(as taken from default request) instead of client URL.
I suggest to set default URL as empty to allow using path from client host URL
like 'http://host/path'

I couldn't merge this patch with HEAD in https://github.com/boto/botocore/pull/447
so I've created a new one...

### DIFF
--- a/tests/unit/protocols/input/ec2.json
+++ b/tests/unit/protocols/input/ec2.json
@@ -34,7 +34,7 @@
           "Bar": "val2"
         },
         "serialized": {
-          "uri": "/",
+          "uri": "",
           "body": "Action=OperationName&Version=2014-01-01&Foo=val1&Bar=val2"
         }
       }
@@ -82,7 +82,7 @@
           "Yuck": "val3"
         },
         "serialized": {
-          "uri": "/",
+          "uri": "",
           "body": "Action=OperationName&Version=2014-01-01&Foo=val1&BarLocationName=val2&yuckQueryName=val3"
         }
       }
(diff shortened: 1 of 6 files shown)


tests/unit/protocols/input/ec2.json
```diff
@@ -34,7 +34,7 @@
           "Bar": "val2"
         },
         "serialized": {
-          "uri": "/",
+          "uri": "",
```
jamesls: These tests are suppose to be describing what the serialized request should be, including the URI.  So the url_path should be non empty.  Another thing to keep in mind is that these tests are shared across various SDKs to ensure consistency (e.g for ruby: https://github.com/aws/aws-sdk-ruby/tree/master/aws-sdk-core/spec/fixtures/protocols/input).

jamesls:
I think this would be a helpful feature.  We may need to make some adjustments for how this is implemented so we can avoid changing the protocol compliance tests, but I think the test_endpoints.py are a great starting point.  I'll take a look at this more today.  Thanks for the PR.

tests/unit/protocols/input/ec2.json
Andrey-mp: but if this url will be non-empty then these tests will fail...
because test_protocols uses default requst object from library. and they fail without this change because I changed default path_url there from '/' to ''.

PR #461, opened 2015-02-13
Session.available_profiles is broken when  ~/.aws/credentials exists.

che2cbs:
Updated session.full_conifg() to ignore the key '_path' returned from config.raw_config_parse() when processing the credentials file.  If left in, it can break session.available_profiles(). I also added a test to validate session.available_profiles() in the presence of a ~/.aws/config and ~/.aws/credentials file.


jamesls:
Looks like the travis CI build is failing.  I don't think we want to remove the `_path` from the actual parsed dict.  I'd rather just fix the code to ignore this key rather than removing it.  This is a pretty small change, so I don't mind working on this.

jamesls:
Actually, I checked in the botocore/AWS CLI/boto3 codebase, and `_path` is not used anywhere.  I would actually be in favor of simplifying and just removing these values from not only the parsed credentials file, but the config file as well.  You can always get these locations via `get_config_variable` if you do need them.

@kyleknap @danielgtaylor thoughts?

jamesls:
Haven't heard anything, so let's just go ahead and remove this key entirely, even in the `config_file` case.  Superceded by https://github.com/boto/botocore/pull/477

PR #463, opened 2015-02-14
Fix error with parsing iam policies response

jamesls:
We were making an assumption that the member names were always
in the response, but in the case of ListPolicyVersions the
'Document' key is _not_ in the response.

Before you'd just get:

```
$ aws iam list-policy-versions --policy-arn 'arn:aws:iam::asdf:policy/AdministratorAccess-asdf'

u'Document'
```
cc @kyleknap @danielgtaylor


kyleknap:
Looks good. :ship:

danielgtaylor:
LGTM :ship:-it!

PR #464, opened 2015-02-18
Prevent override of region when using endpoint_url

kyleknap:
This ensures that user's region is not overwritten by the credentialScope if the user supplies an endpoint url. This is needed to be able to make signed calls to STS regionalized endpoints, without getting auth errors due to incorrect region . So for example, in the CLI, to make a GetSessionToken call to us-west-2 you would do:
```
$ aws sts get-session-token --region us-west-2 --endpoint-url https://sts.us-west-2.amazonaws.com
```

Before, this would cause a signing error if signed with sigv4 because the credentialScope was set to ``us-east-1`` and that overrode the region that you provided.

cc @jamesls @danielgtaylor

### DIFF
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -294,27 +294,28 @@ class EndpointCreator(object):
                 raise
         # We only support the credentialScope.region in the properties
         # bag right now, so if it's available, it will override the
-        # provided region name if an endpoint url was not manually set.
-        # If a endpoint url was specified, the user must specify a region.
-        region_name_override = None
-        if endpoint_url is None:
-            region_name_override = endpoint['properties'].get(
-                'credentialScope', {}).get('region')
+        # provided region name.
+        region_name_override = endpoint['properties'].get(
+            'credentialScope', {}).get('region')
         if signature_version is NOT_SET:
             signature_version = service_model.signature_version
             if 'signatureVersion' in endpoint['properties']:
                 signature_version = endpoint['properties']['signatureVersion']
-        if region_name_override is not None:
-            # Letting the heuristics rule override the region_name
-            # allows for having a default region of something like us-west-2
-            # for IAM, but we still will know to use us-east-1 for sigv4.
-            region_name = region_name_override
         if endpoint_url is not None:
+            # If an endpoint_url is provided, do not use region name override if a region
+            # was provided by the user.
+            if region_name is not None:
+                region_name_override = None
             # If the user provides an endpoint url, we'll use that
             # instead of what the heuristics rule gives us.
             final_endpoint_url = endpoint_url
         else:
             final_endpoint_url = endpoint['uri']
+        if region_name_override is not None:
+            # Letting the heuristics rule override the region_name
+            # allows for having a default region of something like us-west-2
+            # for IAM, but we still will know to use us-east-1 for sigv4.
+            region_name = region_name_override
         return self._get_endpoint(service_model, region_name,
                                   signature_version, final_endpoint_url,
                                   verify, response_parser_factory)
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -290,11 +290,6 @@ class EndpointCreator(object):
                 endpoint = {'uri': endpoint_url, 'properties': {}}
             else:
                 raise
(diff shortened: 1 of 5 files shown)


kyleknap:
Coveralls is wrong. Coverage should have went up.

danielgtaylor:
LGTM :ship:-it!

botocore/endpoint.py
```diff
@@ -294,9 +294,12 @@ def create_endpoint(self, service_model, region_name=None, is_secure=True,
                 raise
         # We only support the credentialScope.region in the properties
         # bag right now, so if it's available, it will override the
-        # provided region name.
-        region_name_override = endpoint['properties'].get(
-            'credentialScope', {}).get('region')
+        # provided region name if an endpoint url was not manually set.
+        # If a endpoint url was specified, the user must specify a region.
```
jamesls: Region is only required for sigv4, so you could previously specify an endpoint-url without a region provided you weren't using sigv4.  For example, this is a breaking change for commands such as this:

```
$ aws configure list | grep region
    region                <not set>             None    None
$ aws s3api list-buckets --endpoint-url https://s3.amazonaws.com/
```

kyleknap: Good point so how about instead?
``if endpoint_url is None and region_name is None:``

jamesls: I don't think that will fix the issue.  In the example above endpoint_url is not None so we don't pull in the region override from the heuristics file.

kyleknap:
@jamesls 
I think I got the s3 case covered now. The credentialScope region is only used now if it is specified in the endpoint JSON, an endpoint url was not specified, and then, if a region was not specified by the user.

jamesls:
Is there another change that's needed before the calls to us-west-2 work?  I'm seeing a signing error with the latest changes:

```
$ aws sts get-session-token --region us-west-2 --endpoint-url https://sts.us-west-2.amazonaws.com

A client error (SignatureDoesNotMatch) occurred when calling the GetSessionToken operation: Credential should be scoped to a valid region, not 'us-east-1'.
```

kyleknap:
To make it clear what I did in the latest commits, here is a summary:

1) Separate all of the region determining logic into a helper method ``_determine_region_name()`` in the endpoints class. The goal is that this method can be directly ported to clients when we deprecate operation objects and ``endpoint.region_name``.  Note that now both the clients and operations rely on this helper method for region name logic. This consolidation will ensure that clients and operation objects will use the same codepath when choosing a region name.

2) Removed some of the signature version arguments from endpoint methods. We were not using them at all.

3) Added integration test to test that sts regionalized calls can be made.

botocore/endpoint.py
```diff
@@ -224,17 +224,15 @@ def _get_proxies(url):
...
     event_emitter = session.get_component('event_emitter')
     user_agent = session.user_agent()
     return get_endpoint_complex(service_name, endpoint_prefix,
-                                signature_version,
                                 region_name, endpoint_url, verify, user_agent,
                                 event_emitter)
 
 
-def get_endpoint_complex(service_name, endpoint_prefix, signature_version,
+def get_endpoint_complex(service_name, endpoint_prefix,
```
danielgtaylor: This is a backward-incompatible change of a public function. It's possible we might break someone using Botocore.

danielgtaylor:
I think this looks good. We just need to decide if we really want to make that breaking change. :ship:-it.

tests/integration/test_sts.py
```diff
@@ -0,0 +1,60 @@
...
+from botocore.exceptions import ClientError
+
+class TestSTS(unittest.TestCase):
+    def setUp(self):
+        self.session = botocore.session.get_session()
+
+    def test_regionalized_endpoints(self):
+
+        sts = self.session.create_client('sts', region_name='ap-southeast-1')
+        response = sts.get_session_token()
```
jamesls: You can't call sts when you're using session credentials (say on a CI box using an IAM role), so we should put a check where we only run these tests if we have long term credentials.

kyleknap:
@jamesls 
I made the if statement in the setUp that if a session token is being used then skip the test.

jamesls:
:shipit:

Issue #465, opened 2015-02-18
Client event emitters are not entirely separate


kyleknap:
Take this sample code:
```
import botocore.session


def print_something(**kwargs):
    print('foo')


def main():
    session = botocore.session.get_session()

    client1 = session.create_client('s3')
    client2 = session.create_client('s3')
    client1.meta.events.register('after-call.s3.ListBuckets', print_something, 'unique_id')
    client2.meta.events.register('after-call.s3.ListBuckets', print_something, 'unique_id')

    print('Calling client 1:')
    client1.list_buckets()

    print('Calling client 2:')
    client2.list_buckets()


if __name__ == '__main__':
    main()
```

The expected output would be:
```
Calling client 1:
foo
Calling client 2:
foo
```

However the actual output is:
```
Calling client 1:
foo
Calling client 2:
```

My guess is that ``unique_id`` from ``client1`` is getting picked up in ``client2`` when it should not be because clients are suppose to have separate event emitters.

jamesls:
Yep, pretty sure your assessment is right: https://github.com/boto/botocore/blob/develop/botocore/hooks.py#L336

We only make a copy of _handlers, but the `_unique_id_cache` is in fact, not a cache and also needs to be explicitly copied.  We should probably rename it something else to make this clear.

PR #467, opened 2015-02-19
Copy unique_id handlers for events

kyleknap:
When event emitters were being copied they did not copy the dictionary of unique ids which affected registering handlers with unique ids. This pull request fixes that issue and renames ``_unique_id_cache`` to ``_unique_id_handlers`` to make it explicit that it is not a cache as opposed to a different representation of the handlers registered with respect to their unique id.

Fixes https://github.com/boto/botocore/issues/465

cc @jamesls @danielgtaylor


jamesls:
:shipit:

Issue #468, opened 2015-02-24
Clients sigv4 signing thread safety


kyleknap:
I was working on porting the AWS CLI ``s3`` commands over to clients, and I ran into occasional signing errors when performing large multi-threaded multi-part uploads that required sigv4 signing. The mismatch was because the timestamps was inconsistent in the signature. This is typically exhibited when there is a concurrency bug in our signing process. I believe it is because we do not lock the signing process for clients, which is needed because the timestamp state is shared across threads. I would like it though if we removed that timestamp shared state instead to avoid running into potential concurrency issues in the future.

PR #469, opened 2015-02-25
Make the sigv4 signer thread safe

kyleknap:
This includes making it thread safe for clients and operation objects. Also removed the locking for signing in operation objects since it is not needed anymore.

To do this, I removed the shared utcnow() state and added a ``context`` attribute to ``AWSRequest`` where a timestamp can be assigned when doing the signing. The timestamp attached to the request is set every time at the beginning of ``add_auth`` and other sigv4 signing helper functions then rely on the request's timestamp, that was set at the very beginning of the signing process.

I ran the integration tests for botocore and the CLI to make sure nothing is affected. I also uploaded/downloaded a couple of gigabytes to my ``eu-central-1`` s3 bucket using the s3 commands to further ensure nothing was affected.

Fixes https://github.com/boto/botocore/issues/468

cc @jamesls @danielgtaylor

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -149,8 +149,6 @@ class SigV4Auth(BaseSigner):
         # We initialize these value here so the unit tests can have
         # valid values.  But these will get overriden in ``add_auth``
         # later for real requests.
-        self._now = datetime.datetime.utcnow()
-        self.timestamp = self._now.strftime('%Y%m%dT%H%M%SZ')
         self._region_name = region_name
         self._service_name = service_name
 
@@ -265,17 +263,17 @@ class SigV4Auth(BaseSigner):
         normalized_path = quote(normalize_url_path(path), safe='/~')
         return normalized_path
 
-    def scope(self, args):
+    def scope(self, request):
         scope = [self.credentials.access_key]
-        scope.append(self.timestamp[0:8])
+        scope.append(request.context['timestamp'][0:8])
         scope.append(self._region_name)
         scope.append(self._service_name)
         scope.append('aws4_request')
         return '/'.join(scope)
 
-    def credential_scope(self, args):
+    def credential_scope(self, request):
         scope = []
-        scope.append(self.timestamp[0:8])
+        scope.append(request.context['timestamp'][0:8])
         scope.append(self._region_name)
         scope.append(self._service_name)
         scope.append('aws4_request')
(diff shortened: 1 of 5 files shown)


botocore/operation.py
```diff
@@ -151,10 +150,8 @@ def request_created(request, **kwargs):
             # a request has already been signed without needing
             # to acquire the lock.
             if not getattr(request, '_is_signed', False):
-                with self._lock:
```
jamesls: I think we still want the lock.  This was to prevent the same request from being signed multiple times because the operation isn't tied to an endpoint/request.  Even now that the signers are thread-safe against signing multiple requests concurrently, I don't think they're thread-safe against signing the same request multiple times, especially with things like body checksums on file like objects.

kyleknap:
@jamesls I re-added the lock as you pointed out. Is ready to get looked at once Travis is available  to test again.

jamesls:
:shipit: Thanks.

jamesls:
:shipit:

botocore/awsrequest.py
```diff
@@ -263,6 +263,7 @@ def __init__(self, *args, **kwargs):
             for key, value in self.headers.items():
                 headers[key] = value
         self.headers = headers
+        self.context = {}
```
danielgtaylor: It would be nice to add some documentation explaining what this new public attribute is and what gets set in it.

botocore/auth.py
```diff
@@ -344,14 +341,16 @@ def _set_necessary_date_headers(self, request):
         # header.  Otherwise we use the X-Amz-Date header.
         if 'Date' in request.headers:
             del request.headers['Date']
+            datetime_timestamp = datetime.datetime.strptime(
+                request.context['timestamp'], '%Y%m%dT%H%M%SZ')
```
danielgtaylor: Any reason to use `strptime` here on the string instead of just storing the datetime object in the context alongside the string? This seems like extra work without any real benefit unless I'm mistaken.

botocore/awsrequest.py
kyleknap: Will add some documentation

botocore/auth.py
kyleknap: @danielgtaylor I just did it to avoid bloating the ``context`` dictionary. The current timestamp is of the form ``'YYYYMMDDHHMMSSZ'``, which is what is the form most used in the signing code. Since this form is used once in the entire sigv4 signing process, I figured it was not worth adding a new entry to the context dictionary. Let me know what you think.

danielgtaylor: I suppose it's probably fine either way and is really a very minor point considering out bottleneck is usually the network. Go ahead and keep it if you like, but if you do then I suggest at the very least making the format string a constant, because having it in two places introduces an extra maintenance burden.

kyleknap:
@danielgtaylor I made changes based on your feedback.

danielgtaylor:
LGTM, thanks! :ship:-it!

PR #472, opened 2015-02-26
Support URL paths in an endpoint_url

### LINKED ISSUE #420: Not compatible with endpoints that have a url_path (regression from boto)

jamesls:
Fixes #420
Also fixes aws/aws-cli#961
Closes #460

This fixes the issue by extracting out the logic for converting
a request_dict that we get from a serializer to an AWSRequest object
that's used by requests.  This also lets have have more
specific tests for this functionality.

cc @kyleknap @danielgtaylor

### DIFF
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -23,8 +23,9 @@ from botocore.vendored import six
 
 from botocore.exceptions import UnknownEndpointError
 from botocore.awsrequest import AWSRequest
-from botocore.compat import urljoin, filter_ssl_san_warnings
-from botocore.utils import percent_encode_sequence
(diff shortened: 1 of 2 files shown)


danielgtaylor:
LGTM :ship:-it.

jamesls:
Not sure what's going on with travis.  I cancelled/restarted the job but not luck.  It runs clean on my system, I'm inclined to just merge this.  @kyleknap @danielgtaylor Any concerns?

danielgtaylor:
@jamesls I think it's probably fine if you run it through the different Python versions locally.

I also reached out to Travis CI to see if they can fix these builds not starting.

kyleknap:
That's fine. I am having the same issue with one of the PR's I'm working on for botocore.

PR #474, opened 2015-02-27
Add region name property to clients

jamesls:
As part of this change, I removed the `clone_client` method.  This ended up not being as useful as I thought, and gets harder to maintain as we add new features to the client.

For now, the region_name is just copied out from the endpoint.  I think eventually we're moving region name out of endpoint, but we can still maintain this interface.  That way, this should unblock any customizations that get the region name from the endpoint.  They can now just grab this value from clients.

cc @kyleknap @danielgtaylor


tests/unit/test_endpoint.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2012-2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2012-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
danielgtaylor: It doesn't look like there are any other changes to this file. Any reason to update the copyright year?

danielgtaylor:
LGTM and I'm :+1: on removing `clone_client` after talking with @kyleknap about it. :ship:-it!

tests/unit/test_endpoint.py
jamesls: This was part of a larger refactoring, which I rebased.  Looks like I didn't quite get them split cleanly, I'll change this back.

kyleknap:
Looks good to me. Its fine that we are using the ``endpoint.region_name`` as that is where ``region_name`` should be sourced from in the client until we move the ``region_name`` logic out of the endpoint and into the client. :ship:

PR #475, opened 2015-02-27
Move client methods

jamesls:
This is a non functional change that moves all the client specific methods into the BaseClient class.  This gives us two things:

1. Cleanly separates the client methods from the ClientCreator methods.  Makes it easier to follow the client creator code.
2. Allows us to test the client without having to go through a client creator.

Note that this PR builds off of the region name change PR (https://github.com/boto/botocore/pull/474), so part of that diff appears here.  
The new commit is just this one: https://github.com/boto/botocore/commit/d292abf1dd8a8f32c42c8636c6bec82b83737ac2

cc @kyleknap @danielgtaylor


jamesls:
Not a huge fan of that.  pep8's pretty clear about what to call these (https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments):

> Always use self for the first argument to instance methods.

 and I'd really only want to do that if it was absolutely necessary (e.g. in the inner function I needed to access the `self` in the outer scope).  It's also weird because it would inconsistent in that all the other methods in the class use `self` except for this one.

danielgtaylor:
I think this looks good, just had the couple minor comments above.

As an aside beyond the scope of this non-functional change, we may want to consider collisions between the output of `ClientCreator._create_methods(service_model)` and the methods on the BaseClient. Stuff like `meta`, `get_paginator`, `can_paginate`, and `get_waiter`.

danielgtaylor:
Okay makes sense. In Boto 3 I required access to both `factory_self` and `self` so it made a bit more sense there in a few places.

Anything this LGTM :ship:-it.

PR #476, opened 2015-03-02
Don't check for reason phrase in 100 continue response

jamesls:
As per the HTTP RFC, the reason phrase is intended for the
human user and the "client is not required to examine or
display the Reason-Phrase...they MAY be replaced by local
equivalents without affecting the protocol."

This isn't an issue for S3, but allows for better interop
for other services.

cc @kyleknap @danielgtaylor


kyleknap:
Looks good. :ship:

PR #477, opened 2015-03-03
Remove _path key when parsing INI config files

jamesls:
See https://github.com/boto/botocore/pull/461#issuecomment-77001875 for more info.

cc @kyleknap @danielgtaylor


kyleknap:
Sorry did not comment before. That's good with me. Looks good :ship:

PR #478, opened 2015-03-04
Show XML body when unable to parse XML

jamesls:
Right now you'll get a vague message about invalid
syntax if we can't parse the XML for whatever reason.

Now, we catch this error and raise a specific exception that
also includes the XML body so a user can see why we couldn't parse
the exception.

BTW, this happens when:

* You provide an invalid endpoint-url.  Sometimes you might get some
  boiler plate HTML from a site which will trigger this exception.
* There are some cases are XML parser can't handle when we get
  invalid XML.  While the end goal is to handle these cases, we
  should at least be providing a better error message when this
  isn't the case.

cc @kyleknap @danielgtaylor

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -98,7 +98,6 @@ import logging
 from pprint import pformat
 
 from botocore.compat import six, XMLParseError
-from six.moves import http_client
 
 from botocore.utils import parse_timestamp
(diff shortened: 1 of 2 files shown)


kyleknap:
Looks good. :ship:

PR #479, opened 2015-03-04
Validate that endpoint url has a valid hostname

jamesls:
Otherwise, we'll try to make a request with an invalid hostname
which will either fail at the http layer, or even worse, may
generate invalid signatures if the endpoint_url mistakenly has
an endpoint url.

This can also happen sometimes because of misconfigured region
values in the config file.

cc @kyleknap @danielgtaylor

### DIFF
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -26,6 +26,7 @@ from botocore.awsrequest import AWSRequest
 from botocore.compat import filter_ssl_san_warnings, urlsplit
 from botocore.compat import urlunsplit
 from botocore.utils import percent_encode_sequence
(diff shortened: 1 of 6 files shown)


kyleknap:
Looks good. One question for the tests was there any reference you used to determine invalid enpoint names or was it all based on knowledge and past experiences of what fails? :ship: otherwise.

jamesls:
http://tools.ietf.org/html/rfc1123#page-13 and http://tools.ietf.org/html/rfc952, with the hostname validation code lifted from stack overflow (http://stackoverflow.com/questions/2532053/validate-a-hostname-string).  I'll update with a comment about this.

PR #481, opened 2015-03-04
Give better error message for invalid regions

jamesls:
We need to be giving better error messages when a user provides
an invalid region instead of the ConnectionError/gaierror
info that's not helpful.  We also will display the endpoint
we were trying to connect to.

Fixes aws/aws-cli#902

cc @kyleknap @danielgtaylor 

### Before
```
$ aws s3 ls --region does-not-exist-foo

('Connection aborted.', gaierror(8, 'nodename nor servname provided, or not known'))
```

### After

```
$ aws s3 ls --region does-not-exist-foo

Could not connect to the endpoint "https://s3.does-not-exist-foo.amazonaws.com/", please verify your region value.
```

### DIFF
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -19,9 +19,11 @@ import threading
 
 from botocore.vendored.requests.sessions import Session
 from botocore.vendored.requests.utils import get_environ_proxies
+from botocore.vendored.requests.exceptions import ConnectionError
 from botocore.vendored import six
 
 from botocore.exceptions import UnknownEndpointError
+from botocore.exceptions import EndpointConnectionError
 from botocore.awsrequest import AWSRequest
 from botocore.compat import filter_ssl_san_warnings, urlsplit
 from botocore.compat import urlunsplit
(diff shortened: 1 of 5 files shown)


botocore/endpoint.py
```diff
@@ -227,6 +229,18 @@ def _get_response(self, request, operation_model, attempts):
                 request, verify=self.verify,
                 stream=operation_model.has_streaming_output,
                 proxies=self.proxies, timeout=self.timeout)
+        except ConnectionError as e:
+            # For a connection error, if it looks like it's a DNS
+            # lookup issue, 99% of the time this is due to a misconfigured
+            # region/endpoint so we'll raise a more specific error message
+            # to help users.
+            if self._looks_like_dns_error(e):
+                endpoint = e.request.url
```
danielgtaylor: Considering this is inside the `Endpoint` class, we may want to call this something more specific, like `endpoint_url` or `request_url`.

danielgtaylor:
Aside from the minor point above, this LGTM :+1:

botocore/endpoint.py
kyleknap: Yeah I think ``endpoint_url`` would be better here as well.

kyleknap:
Looks good to me. I agreed with Daniel's small comment. Otherwise, :ship:

botocore/endpoint.py
jamesls: Ok I'll get this updated.  I think I may send a PR to the CLI that catches this error and gives a better error message though.

The high level thing I'm after is that in the CLI, if a user types a wrong region name, e.g. `--region eu-central-typo`, we can give then some type of hint that tells them to verify their region name.  I agree that this part in botocore isn't the right place to have this error message.

jamesls:
Closing, merged in 4ca6eb938119f207b7e4731a9ce2a032fd5b5cf3

PR #483, opened 2015-03-05
Fix creating clients with unicode service name

danielgtaylor:
This is for a problem identified in boto/boto3#70. The workaround is
simply to call `str()` on the service name, as is done elsewhere in
Boto 3. This only affects Python 2.

cc @jamesls @kyleknap


jamesls:
+0 on this.  Especially given calls such as session.get_available_services() returns the service names as a str type, there's nothing that I'm aware of in our code base that actually returns unicode service names is there?  However, given it's such a small change, it's not a huge deal to get this in.


:shipit:

danielgtaylor:
@jamesls, loading config from UTF-8 JSON could give you unicode strings that get sent into `create_client`. I agree it's not super likely but we had a customer using Boto 3 run into this issue.

PR #484, opened 2015-03-10
Fix flaky unit test by fixing date

jamesls:
In some of the sigv4 tests we were signing the request twice
to see what to ensure we're handling the resign case properly.
However, the sigv4 tests were not patching out the current time
which meant that it was possible to have test failures if the time
between add_auth calls crossed the seconds boundary.

cc @danielgtaylor @kyleknap


kyleknap:
:ship:

PR #485, opened 2015-03-10
Fix build_full_result for clients

kyleknap:
Fixes ``build_full_result()`` for clients.

Before this would happen:
```
>>> import botocore.session
>>> client = session.create_client('s3')
>>> paginator = client.get_paginator('list_objects')
>>> pages = paginator.paginate(Bucket='mybucketfoo')
>>> pages.build_full_result()
  File "botocore/paginate.py", line 215, in build_full_result
    for _, page in self:
ValueError: too many values to unpack
```

Now this lists all of my objects.

Note that I only included that one unit test for clients to have a smoke test even though operation objects have much more tests. I was considering switching all of the operation object pagination test to clients, but I was hesitant to do that since the CLI still uses operation object pagination. Let me know what you think.

cc @jamesls @danielgtaylor

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -212,7 +212,15 @@ class PageIterator(object):
         for result_expression in self.result_keys:
             set_value_from_jmespath(complete_result,
                                     result_expression.expression, [])
-        for _, page in self:
+        for response in self:
+            page = response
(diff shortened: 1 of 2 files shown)


jamesls:
:shipit:

PR #486, opened 2015-03-10
Explicit profile overrides environment variables

danielgtaylor:
This changes the behavior of environment variable credential loading to
function as described in aws/aws-cli#113. Here is what this looks like
for both the CLI and Botocore/Boto 3:

CLI:
```bash
# Use the environment variables
$ AWS_ACCESS_KEY_ID=foo AWS_SECRET_ACCESS_KEY=bar aws s3 ls

# Ignore the environment variables and use a profile from the config instead
$ AWS_ACCESS_KEY_ID=foo AWS_SECRET_ACCESS_KEY=bar aws --profile dev s3 ls
```

Python:
```python
$ AWS_ACCESS_KEY_ID=foo AWS_SECRET_ACCESS_KEY=bar python
>>> import boto3
>>> boto3.setup_default_session(profile_name='dev')
>>> # The following will use the profile, not the env vars!
>>> s3dev = boto3.resource('s3')
```

Added a test and a new log message to ensure it's obvious what is happening.

cc @jamesls @kyleknap

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -46,10 +46,8 @@ def create_credential_resolver(session):
     config_file = session.get_config_variable('config_file')
     metadata_timeout = session.get_config_variable('metadata_service_timeout')
     num_attempts = session.get_config_variable('metadata_service_num_attempts')
+
     providers = [
-        # We use ``session.profile`` for EnvProvider rather than
-        # ``profile_name`` so that it can be ``None`` when unset.
-        EnvProvider(profile_name=session.profile),
         SharedCredentialProvider(
             creds_filename=credential_file,
             profile_name=profile_name
@@ -65,6 +63,18 @@ def create_credential_resolver(session):
                 num_attempts=num_attempts)
         )
     ]
+
+    # We use ``session.profile`` for EnvProvider rather than
+    # ``profile_name`` because it is ``None`` when unset.
+    if session.profile is None:
+        # No profile has been explicitly set, so we prepend the environment
+        # variable provider. That provider, in turn, may set a profile
+        # or credentials.
+        providers.insert(0, EnvProvider())
+    else:
+        logger.info('Skipping environment variable credential check'
+                    ' because profile name was explicitly set.')
+
     resolver = CredentialResolver(providers=providers)
     return resolver
 
@@ -277,7 +287,7 @@ class EnvProvider(CredentialProvider):
     # AWS_SESSION_TOKEN is what other AWS SDKs have standardized on.
     TOKENS = ['AWS_SECURITY_TOKEN', 'AWS_SESSION_TOKEN']
 
-    def __init__(self, environ=None, mapping=None, profile_name=None):
+    def __init__(self, environ=None, mapping=None):
         """
 
         :param environ: The environment variables (defaults to
@@ -291,7 +301,6 @@ class EnvProvider(CredentialProvider):
         if environ is None:
             environ = os.environ
         self.environ = environ
-        self._profile_name = profile_name
         self._mapping = self._build_mapping(mapping)
 
     def _build_mapping(self, mapping):
@@ -317,11 +326,6 @@ class EnvProvider(CredentialProvider):
         """
         Search for credentials in explicit environment variables.
         """
-        if self._profile_name is not None:
-            logger.info('Skipping environment variable credential check'
-                        ' because profile name was explicitly set')
-            return None
-
         if self._mapping['access_key'] in self.environ:
             logger.info('Found credentials in environment variables.')
             access_key, secret_key = self._extract_creds_from_mapping(
@@ -588,7 +592,7 @@ class CredentialResolver(object):
         # If we got here, no credentials could be found.
         # This feels like it should be an exception, but historically, ``None``
         # is returned.
-        # 
+        #
         # +1
         # -js
         return None
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -72,8 +72,8 @@ def create_credential_resolver(session):
         # or credentials.
         providers.insert(0, EnvProvider())
     else:
-        logger.info('Skipping environment variable credential check'
-                    ' because profile name was explicitly set.')
+        logger.debug('Skipping environment variable credential check'
+                     ' because profile name was explicitly set.')
 
     resolver = CredentialResolver(providers=providers)
     return resolver
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -47,7 +47,9 @@ def create_credential_resolver(session):
     metadata_timeout = session.get_config_variable('metadata_service_timeout')
     num_attempts = session.get_config_variable('metadata_service_num_attempts')
     providers = [
-        EnvProvider(),
+        # We use ``session.profile`` for EnvProvider rather than
+        # ``profile_name`` so that it can be ``None`` when unset.
+        EnvProvider(profile_name=session.profile),
         SharedCredentialProvider(
             creds_filename=credential_file,
             profile_name=profile_name
@@ -275,7 +277,7 @@ class EnvProvider(CredentialProvider):
     # AWS_SESSION_TOKEN is what other AWS SDKs have standardized on.
     TOKENS = ['AWS_SECURITY_TOKEN', 'AWS_SESSION_TOKEN']
 
-    def __init__(self, environ=None, mapping=None):
+    def __init__(self, environ=None, mapping=None, profile_name=None):
         """
 
         :param environ: The environment variables (defaults to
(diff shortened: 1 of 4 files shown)


botocore/credentials.py
```diff
@@ -275,7 +277,7 @@ class EnvProvider(CredentialProvider):
     # AWS_SESSION_TOKEN is what other AWS SDKs have standardized on.
     TOKENS = ['AWS_SECURITY_TOKEN', 'AWS_SESSION_TOKEN']
 
-    def __init__(self, environ=None, mapping=None):
+    def __init__(self, environ=None, mapping=None, profile_name=None):
```
jamesls: I think this is the wrong level of abstraction and couples profiles to retrieving environment variables when they don't need to be.  An env provider should just be a simple object that pulls credentials from environment variables and that's it.  If something up the call stack wants to either re-arrange cred providers or ignore what this class returns, that's another thing, but this class should just stay minimal.

It also gives this class weird semantics.  If you pass a profile_name to this class, then it won't do anything.

jamesls:
We also don't really clarify what precedence the `AWS_DEFAULT_PROFILE` env var has.  Should that be a higher precedence than the `AWS_ACCESS*`, `AWS_SECRET*` env vars?  I'm thinking it should, but it currently does not.

botocore/credentials.py
danielgtaylor: Would something along these lines be more like what you were thinking?

https://gist.github.com/danielgtaylor/075fcd07b381f1b73e78

jamesls: Yeah something along those lines seems better abstracted to me.

kyleknap:
@jamesls @danielgtaylor 

Here is the precedence leveling that I think we should be using:

| Location | CLI Representation | Boto3 Representation |
| ----------- | -------------------------- | ------------------------------ |
| Access Keys in Code | NA | ``client(access_key_id=..., secret_access_key_id=...)`` |
| Profile in Code | ``--profile <profilename>`` | ``client(profile_name=...)`` |
| Access Keys in Environment Variable | ``AWS_ACCESS_KEY_ID`` | ``AWS_ACCESS_KEY_ID``|
| Profile in Environment Variable | ``AWS_DEFAULT_PROFILE`` | ``AWS_DEFAULT_PROFILE`` |
| Shared Credentials File | ~/.aws/credentials | ~/.aws/credentials |
| AWS Config File | ~/.aws/config | ~/.aws/config |
| Instance Metadata | NA | NA |

The way I see it is that precedence should be based off of what is the most explicit. Similarly how access key in code is taking precedence of profile_name in code, access key in environment variable should take precedence over profile name in environment variable because the user is being more explicit and it is consistent with the code parallel of precedence.

I honestly don't think ``AWS_DEFAULT_PROFILE`` should affect precedence at all. It should just be switching what profile is default when we look into the shared credentials file and/or the config file. Or that is what I expect it would do given the fact that the name of the environment variable seems to imply changing what the default profile is, which is a little different than how it is described in our docs:

screen shot 2015-03-11 at 10 37 30 am

But yet again the docs is a little vague in that it really does not say where its precedence lies when it describes it as "name of the CLI profile to use".

Let me know what you guys think.

botocore/credentials.py
kyleknap: I agree. That would look better.

danielgtaylor:
Spoke with @kyleknap this morning about this and I think we both agree. To put this into more concrete examples, if these are set:

Code / arguments:
* `aws_access_key_id='abc123'`       <----- takes precedence
* `profile_name='dev'`

Environment
* `AWS_ACCESS_KEY_ID=abc123`     <------ takes precedence
* `AWS_DEFAULT_PROFILE=dev`

The case we **really** need to get working is this one:
* `AWS_ACCESS_KEY_ID=abc123`
* `--profile`                             <------ takes precedence

```bash
$ AWS_ACCESS_KEY_ID=abc123 aws --profile dev s3 ls
```

Thoughts?

jamesls:
Sorry, just catching up on this now.  To summarize, it sounds like we're saying:

For any given precedence level (i.e config file, env vars, cmdline args, etc), the akid/sak will always trump a profile value if all three are specified at the same precedence value.

A consequence of this is that in order to get a "profile" to take, you have to specify it at the next precedence level (or higher) of whatever is explicitly specifying the akid/skid.  So if akid/sak are in env vars, profile has to be as a cmd line option or in boto3 via a method arg.

In which case, I think that makes sense.  Thanks for looking into this.

I would suggest adding *integration* tests for these cases as well.

danielgtaylor:
@jamesls @kyleknap please take another look.

tests/integration/test_credentials.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class TestCredentialPrecedence(unittest.TestCase):
+    def setUp(self):
+        # Clean up any existing environment
+        for name in ['AWS_ACCESS_KEY_ID',
+                     'AWS_SECRET_ACCESS_KEY',
+                     'BOTO_DEFAULT_PROFILE']:
+            if name in os.environ:
+                del os.environ[name]
```
jamesls: This is general test case best practice advice:  Do *not* leave global state modified outside the scope of a test.  It's ok to mess with global state in a test if there's no better way to test what you need to test, but you *have* to put the state back to the way it initially was.

Here, if the vars you're deleting were defined in os.environ before the test, this setUp() is deleting these values and not restoring them after the test executes.  This can lead to all sorts of hard to track down problems where the test order may affect whether or not a test passes depending on if it passes before/after the global state modification.

That being, said, just use `BaseEnvVar` to handle the patching of env vars for you.

tests/integration/test_credentials.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # Set the config file to something that doesn't exist, then
+        # set the shared credential file to our test config.
+        os.environ['AWS_CONFIG_FILE'] = '~/.aws/config-missing'
+        Session.SessionVariables['credentials_file'] = (
```
jamesls: Same thing here.  You are modifying a class variable, which persists across tests.  Avoid doing this, as this is modifying global state and not putting it back.

tests/integration/test_credentials.py
```diff
@@ -0,0 +1,106 @@
...
+        os.environ['AWS_SECRET_ACCESS_KEY'] = 'env-secret'
+        os.environ['BOTO_DEFAULT_PROFILE'] = 'test'
+
+        s = Session()
+        credentials = s.get_credentials()
+
+        self.assertEqual(credentials.access_key, 'env')
+        self.assertEqual(credentials.secret_key, 'env-secret')
+
+    @mock.patch('botocore.credentials.Credentials')
```
jamesls: I'm not really a fan of unnecessary patching, *especially* when it's not necessary.  Why can't we just check the credential object directly?

jamesls:
I think your integ tests cover a good set of cases, I just have some comments on how the tests themselves are implemented.

Also, there are no unit tests for this.  While I think we want the integ tests to ensure we don't regress as we refactor some of the internals, I still think this change should have a unit test.  We should be able to call into `create_credential_resolver` and assert how the cred resolver is being created.

The actual code change itself looks good.

botocore/credentials.py
```diff
@@ -63,6 +63,18 @@ def create_credential_resolver(session):
...
+
+    # We use ``session.profile`` for EnvProvider rather than
+    # ``profile_name`` because it is ``None`` when unset.
+    if session.profile is None:
+        # No profile has been explicitly set, so we prepend the environment
+        # variable provider. That provider, in turn, may set a profile
+        # or credentials.
+        providers.insert(0, EnvProvider())
+    else:
+        logger.info('Skipping environment variable credential check'
```
jamesls: This seems more like a debug message.  We'll still log which provider ultimately gives creds, but the fact that we're disabling a check seems mostly of interest to a dev.

tests/integration/test_credentials.py
danielgtaylor: Unfortunately this isn't exposed anywhere in clients. There is no way to get the credentials using public interfaces once a client is created. I'm not opposed to adding a `meta.credentials` property or something on clients, but I figured there would be push-back if I add this just for a test.

danielgtaylor:
@jamesls please take another look.

tests/unit/test_credentials.py
```diff
@@ -587,10 +588,38 @@ def test_create_credential_resolver(self):
...
+
+        for provider in resolver.providers:
+            self.assertNotIsInstance(provider, credentials.EnvProvider)
+
+    def test_no_profile_checks_env_provider(self):
+        self.config['profile'] = None
+        self.session.profile = None
+        resolver = credentials.create_credential_resolver(self.session)
+
+        found = False
```
jamesls: Or to assert that all the providers are not an instance of the env provider, you can just use: `self.assertTrue(all(not isinstance(p, EnvProvider) for p in p.providers))`

jamesls:
:shipit:  Looks good, just had a small suggestion.  Thanks for updating.

PR #487, opened 2015-03-10
Add service_model property to operation_model

jamesls:
This makes it so that given a reference to an operation model, you can access its parent object.  Otherwise, existing code needs to track the service model when needed, despite the fact that it's accessible via an internal attribute on the operation model.

This is also needed as part of the client switchover work in the CLI.

cc @kyleknap @danielgtaylor

### DIFF
--- a/tests/unit/test_model.py
+++ b/tests/unit/test_model.py
@@ -122,6 +122,13 @@ class TestOperationModelFromService(unittest.TestCase):
         self.assertEqual(shape.name, 'OperationNameRequest')
         self.assertEqual(list(sorted(shape.members)), ['Arg1', 'Arg2'])
 
+    def test_service_model_available_from_operation_model(self):
+        service_model = model.ServiceModel(self.model)
+        operation = service_model.operation_model('OperationName')
+        self.assertEqual(
+            operation.service_model.service_name,
+            service_model.service_name)
+
     def test_operation_output_model(self):
         service_model = model.ServiceModel(self.model)
         operation = service_model.operation_model('OperationName')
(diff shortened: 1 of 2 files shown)


tests/unit/test_model.py
```diff
@@ -122,6 +122,13 @@ def test_operation_input_model(self):
         self.assertEqual(shape.name, 'OperationNameRequest')
         self.assertEqual(list(sorted(shape.members)), ['Arg1', 'Arg2'])
 
+    def test_service_model_available_from_operation_model(self):
+        service_model = model.ServiceModel(self.model)
+        operation = service_model.operation_model('OperationName')
+        self.assertEqual(
+            operation.service_model.service_name,
```
kyleknap: Or can we just:
```
self.assertEqual(operation.service_model, service_model)
```

Its not s big deal. Just curious.

kyleknap:
Looks good. Had one question/comment. Otherwise :ship:

tests/unit/test_model.py
jamesls: Yeah not opposed to changing this.  It's this way now mostly because the default implementation for object equality (given we don't override `__eq__`) is identity, i.e `a is b`, which is an overspecification here.  Given these are value objects, what we really should be doing is comparing equality based on the equality of the model attributes.  We don't technically care if `operation.service_model` and `service_model` aren't references to the exact same object, as long as the model objects describe the same model. We might even be able to get away with an `__eq__()` of `self.__dict__ == other.__dict__`, but I suspect not given we use `@CachedProperty` which messes with `self.__dict__`.

PR #488, opened 2015-03-11
Add service_name when using get_service_model

jamesls:
Clients did this, but not session.get_service_model.

Need this in the the CLI refactor, otherwise the model doesn't know the "service name" to use to create the client when endpoint_prefix != service_name.

cc @kyleknap @danielgtaylor


kyleknap:
:ship:

danielgtaylor:
LGTM :ship:-it.

PR #489, opened 2015-03-12
Add endpoint_url to client meta

jamesls:
I think eventually we'll move all the client args into the
client meta, but for now, we only need endpoint_url.

cc @kyleknap @danielgtaylor


kyleknap:
Thanks :ship:

PR #491, opened 2015-03-12
Added a PaginatorModel

kyleknap:
This pretty much provides a parallel to the ``WaiterModel`` in waiters.py and ``session.get_waiter_model()`` for paginators. It is pretty light-weight right now. We can add to this as we see needed.

cc @jamesls @danielgtaylor


jamesls:
:shipit:

PR #492, opened 2015-03-12
Add pending deprecations to old style interface

jamesls:
I only added them to the most common methods used in botocore
with the old interface.

Later we'll change these to deprecation warnings, but for now,
we'll use PendingDeprecationWarning so you have to opt in.

### Examples
```
$ python
>>> import botocore.session
>>> s.get_service('ec2')
Service(ec2)

# Running with warnings:
$ python -Wd
>>> import botocore.session
>>> ec2 = s.get_service('ec2')
session.py:559: PendingDeprecationWarning: get_service is deprecated and will be removed.  Use create_client instead.
  "Use create_client instead.", PendingDeprecationWarning)
service.py:230: PendingDeprecationWarning: get_service is deprecated and will be removed.  Use Session.create_client instead.
  PendingDeprecationWarning)
>>> endpoint = ec2.get_endpoint('us-east-1')
service.py:160: PendingDeprecationWarning: get_endpoint is deprecated and will be removed.  Use create_client instead.
  "Use create_client instead.", PendingDeprecationWarning)
```
cc @kyleknap @danielgtaylor

### DIFF
--- a/botocore/operation.py
+++ b/botocore/operation.py
@@ -15,6 +15,8 @@
 import functools
 import logging
 import threading
+import warnings
+
 from botocore.exceptions import MissingParametersError
 from botocore.exceptions import UnknownParameterError
 from botocore.exceptions import NoRegionError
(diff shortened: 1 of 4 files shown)


kyleknap:
Looks good. We should warn people in the readme or anywhere else that is visible to users about this deprecation. :ship:

danielgtaylor:
LGTM :ship:-it!

PR #493, opened 2015-03-16
Raise more specific exception when no region is configured

jamesls:
Currently, if no region is configured, you get a vague error message that isn't very clear:

```
In [1]: s.create_client('ec2')
UnknownEndpointError: Unable to construct an endpoint for ec2 in region None
```

Now, if the endpoint heuristics fail and there's no region name you'll get a `NoRegionError`:

```
In [1]: s.create_client('ec2')
NoRegionError: You must specify a region.
```

This also translates to better error messages in the CLI:

BEFORE:

```
$ aws ec2 describe-instances

Unable to construct an endpoint for ec2 in region None
```

AFTER:

```
$ aws ec2 describe-instances
You must specify a region. You can also configure your region by running "aws configure".
```

One of the other changes I made was to remove the `env_var` value from the NoRegionError.  From the commit message:

> First, we were using this inconsistently.  In the botocore
layer, this is actually BOTO_DEFAULT_REGION, and in some
places we were using AWS_DEFAULT_REGION.  Secondly, this value
is technically pluggable via session_vars in the Session
__init__.  And finally, you can actually set this value in the
config file so it's weird to only mention env vars.

I don't think there's any loss in helpfulness by removing this variable from the exception message.

cc @kyleknap @danielgtaylor

### DIFF
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -69,18 +69,6 @@ class PartialCredentialsError(BotoCoreError):
     fmt = 'Partial credentials found in {provider}, missing: {cred_var}'
 
 
-class NoRegionError(BotoCoreError):
-    """
-    No region was specified
-
-    :ivar env_var: The name of the environment variable to use to
-        specify the default region.
-    """
-    fmt = (
-        'You must specify a region or set the {env_var} environment variable.'
-    )
-
-
 class UnknownSignatureVersionError(BotoCoreError):
     """
     Requested Signature Version is not known.
@@ -100,7 +88,29 @@ class ServiceNotInRegionError(BotoCoreError):
     fmt = 'Service {service_name} not available in region {region_name}'
 
 
-class UnknownEndpointError(BotoCoreError):
+class BaseEndpointResolverError(BotoCoreError):
+    """Base error for endpoint resolving errors.
+
+    Should never be raised directly, but clients can catch
+    this exception if they want to generically handle any errors
+    during the endpoint resolution process.
+
+    """
+
+
+class NoRegionError(BaseEndpointResolverError):
+    """
+    No region was specified
+
+    :ivar env_var: The name of the environment variable to use to
+        specify the default region.
+    """
+    fmt = (
+        'You must specify a region or set the {env_var} environment variable.'
+    )
+
+
+class UnknownEndpointError(BaseEndpointResolverError):
     """
     Could not construct an endpoint.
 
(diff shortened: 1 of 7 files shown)


botocore/regions.py
```diff
@@ -63,8 +64,13 @@ def construct_endpoint(self, service_name, region_name, **kwargs):
...
         if endpoint is None:
-            raise UnknownEndpointError(service_name=service_name,
-                                       region_name=region_name)
+            if region_name is None:
+                # Raise a more specific error message that will give
+                # better guidance to the user what needs to happen.
+                raise NoRegionError()
+            else:
+                raise UnknownEndpointError(service_name=service_name,
+                                        region_name=region_name)
```
kyleknap: pep8 spacing

botocore/regions.py
```diff
@@ -63,8 +64,13 @@ def construct_endpoint(self, service_name, region_name, **kwargs):
...
 
         if endpoint is None:
-            raise UnknownEndpointError(service_name=service_name,
-                                       region_name=region_name)
+            if region_name is None:
+                # Raise a more specific error message that will give
+                # better guidance to the user what needs to happen.
+                raise NoRegionError()
+            else:
+                raise UnknownEndpointError(service_name=service_name,
```
kyleknap: Out of curisosity, how does the ``UnknownEndpointError`` get thrown now? It may be good to add a test for it if there is not one already.

kyleknap:
Looks good. I just had a couple of small comments. :ship:

botocore/exceptions.py
```diff
@@ -100,7 +88,27 @@ class ServiceNotInRegionError(BotoCoreError):
...
+
+    """
+
+
+class NoRegionError(BaseEndpointResolverError):
+    """
+    No region was specified
+
+    :ivar env_var: The name of the environment variable to use to
+        specify the default region.
```
danielgtaylor: We should probably remove the docs for `env_var` since it's no longer set or printed out.

botocore/regions.py
danielgtaylor: :+1:

jamesls: This is tested here: https://github.com/boto/botocore/pull/493/files#diff-e323551d51dbf48e6eb73d90a9d46809R338

You'll get this thrown if we don't resolve an endpoint, but we specified a region.  In practice this will happen if the heuristic does not have a catch all rule.

jamesls:
Feedback incorporated, feel free to take another look.

kyleknap:
Cool. Still :ship:

PR #496, opened 2015-03-20
Add api_name property to OperationModel

kyleknap:
It is needed when the name exposed in the service model is different from the operation model's name attribute in the JSON model so sometimes OperationModel.name did not match with the name used by the client. This will allow us to fix the cloudfront commands in the CLI.

cc @jamesls @danielgtaylor

### DIFF
--- a/botocore/model.py
+++ b/botocore/model.py
@@ -233,7 +233,7 @@ class ServiceModel(object):
             model = self._service_description['operations'][operation_name]
         except KeyError:
             raise OperationNotFoundError(operation_name)
-        return OperationModel(model, self)
+        return OperationModel(model, self, operation_name)
 
     @CachedProperty
     def documentation(self):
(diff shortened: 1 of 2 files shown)


danielgtaylor:
LGTM :ship:-it.

jamesls:
:shipit:  Can we add a docstring for `api_name`.  I've already forgot which one is which.  Also, should we not switch over to `api_name` anywhere else?  What do the event names look like for after-call/before-call?

It might actually be worth mapping `operation.name` to the name we want to expose to the user, and add something like `wire_name` which is only really needed in the serializer (and is a bit easier to remember IMO).  That way we can keep all the existing code as is using `operation_model.name`

Thoughts?

danielgtaylor:
I'm okay with `wire_name` :+1:

jamesls:
Ok I'll update shortly.

jamesls:
Ok, added here: https://github.com/boto/botocore/pull/497

Closing in favor of that.

PR #497, opened 2015-03-24
Add a wire_name property to op model

jamesls:
Continuation of https://github.com/boto/botocore/pull/496

cc @kyleknap @danielgtaylor


tests/unit/test_model.py
```diff
@@ -115,6 +115,31 @@ def setUp(self):
         }
         self.service_model = model.ServiceModel(self.model)
 
+    def test_api_name(self):
```
kyleknap: you may want to rename these tests since that we use ``wire_name`` and ``name`` now.

kyleknap:
Looks good. Had a comment about renaming the tests. Otherwise :ship:

tests/unit/test_model.py
jamesls: Ok I'll update.

PR #498, opened 2015-03-24
Fix copy_snapshot handler to work with clients

jamesls:
This handler still relied on operation objects, which no
longer works with the client switchover.

As part of this change, I switched over the test_ec2 integ
tests over to clients.

cc @kyleknap @danielgtaylor


kyleknap:
It looks good to me. I actually was working on porting the ec2 tests to clients. Your code looks similar to what I was working on. I also ran into the ``CopySnapshot`` issue where an endpoint was needed, but I had not figured out a great solution for it. I do not think there is a good solution such that we do not rely on endpoint till we have the presigner class. Otherwise, :ship:

PR #499, opened 2015-03-24
Remove Service and Operation Objects

kyleknap:
In order to completely remove service objects and operation objects from botocore a bunch of tests needed to be removed and updated. That is why the PR is so large. Here is a summary of what I did:
* Remove ``operation.py`` and ``service.py`` and their corresponding tests
* Remove the deprecated paginators and waiter and update their tests if the client version did not have a form of the test.
* Remove some tutorials that I found that use the service/operation interface.
* Remove the service specific unit test and converted them to protocol tests.
* Convert all unit and integration tests to the client/model interface such that they pass.

This will not be the final PR however, there are a few points that I want to address:
* There are a couple of integration that goes through the manual process of creating a presigned url for s3. This uses the old interface. I want to update these tests when we add the presigning class for clients. In the end, I decided to mark them as a skip because it is a lot of extra code to get a presign url and figured it is better off to wait till the presigning class as none of our features rely on an s3 presigner.

* Remove the response parsing tests and convert any coverage drops to the protocol tests. In the pull request, I updated the response parsing tests only because if I completely remove it, the coverage drops. I was going to save this for a subsequent PR as it potentially could be a fair amount of extra tests.

Let me know what you think.

cc @jamesls @danielgtaylor


docs/source/tutorial/index.rst
```diff
@@ -1,58 +0,0 @@
-*****************************
```
jamesls: Do we want to remove this file?  This uses clients in the tutorial.

kyleknap: Oh no we do not want to remove it. I did not check that file.

jamesls:
:shipit:  Looks good to me.  Jus had one question about the tutorial that uses clients.

danielgtaylor:
LGTM :ship:-it!

PR #500, opened 2015-03-25
Remove region name throughout endpoint

kyleknap:
There is now only one place where the region name is passed to any class in endpoint.py and that is with ``EndpointCreator``'s ``create_endpoint`` because if you want to create an endpoint from scratch you need the region to help resolve the host. However, any of the region deciding logic on how to sign the request is moved out of the ``Endpoint`` class and into the client class. The ``get_endpoint`` and ``get_endpoint_complex`` public methods were removed to encourage creation of endpoints through the endpoint creator.

cc @jamesls @danielgtaylor


kyleknap:
I updated the pull request to include ``region_name`` in the client config object. As to precedence, it takes precedence over environment variables and the config file, but not if ``region_name`` is specified in the ``create_client`` call.

Let me know what you think.

kyleknap:
Note that there is not much I can do about the small dip in coverage. All relevant lines that  I added is covered but I removed more relevant lines (which were all covered) than I added.

botocore/client.py
```diff
@@ -114,21 +115,50 @@ def _get_signature_version_and_region(self, service_model, region_name,
...
+        region_name = self._determine_region_name(
+            endpoint_config, region_name, endpoint_url)
+
         return signature_version, region_name
 
+    def _determine_region_name(self, endpoint_config, region_name=None,
+                               endpoint_url=None):
+        # This is a helper function to determine region name to use.
+        # It will take into account whether the user passes in a region
+        # name, whether their is a rule in the endpoint JSON, or
```
danielgtaylor: grammar: there

botocore/client.py
```diff
@@ -114,21 +115,50 @@ def _get_signature_version_and_region(self, service_model, region_name,
...
+        # We only support the credentialScope.region in the properties
+        # bag right now, so if it's available, it will override the
+        # provided region name.
+        region_name_override = endpoint_config['properties'].get(
+            'credentialScope', {}).get('region')
+
+        if endpoint_url is not None:
+            # If an endpoint_url is provided, do not use region name
+            # override if a region
+            # was provided by the user.
```
danielgtaylor: This comment looks split onto a new line kind of strange. It could just be two lines instead of three.

danielgtaylor:
Other than the nitpicks above this looks good. :ship:-it!

danielgtaylor:
LGTM :ship:-it

jamesls:
:shipit:  We should also document the precedence of `region_name` vs. `Config(region_name)` somewhere, either in the session's `create_client` or at least in the `ClientCreator`.

kyleknap:
Cool. I updated the session's doc string for the client config value. Merging.

PR #502, opened 2015-03-27
Allow using OrderedDict for structure builder.

danielgtaylor:
This change will make it much easier to test documentation, e.g. given
this structure expect this output. Order matters for this use case.

cc @jamesls, @kyleknap

### DIFF
--- a/botocore/model.py
+++ b/botocore/model.py
@@ -497,12 +497,19 @@ class DenormalizedStructureBuilder(object):
         }).build_model()
         # ``shape`` is now an instance of botocore.model.StructureShape
 
+    :type dict_type: class
+    :param dict_type: The dictionary type to use, allowing you to opt-in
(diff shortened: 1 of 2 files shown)


kyleknap:
Would it be possible to just make the default OrderedDict and not include the extra argument in the method definition? Are there any downsides to that? The code looks cleaner without the extra argument, but I am not sure how big of a performance hit it will take if any, if OrderedDict is the default.

Otherwise code looks fine.

@danielgtaylor @jamesls Thoughts?

jamesls:
I don't have a preference for the default.  Given this object is used more as a utility/helper class instead of being used in the main path of an API call via a client, I'm ok with changing the default to an ordered dict and I agree with @kyleknap that it will make the code less wordy.  It will be slower, but I don't think it will really matter, give its primary usage is in tests.

Otherwise, :shipit:

PR #503, opened 2015-03-27
Move User Agent from Endpoint

kyleknap:
Note that this PR is based off of this PR: https://github.com/boto/botocore/pull/500. So the actual changes are less than what is in the diff. So it would be good to get that one in first.

This PR does the following:
* Remove user agent from endpoint.
* Broke up the ``RequestCreator`` and moved it to utility functions in ``awsrequest.py``.
* Prepare ``request_dict`` in the client: The endpoint only receives a ``request_dict`` from the client that has been prepared by serializing the params of the operation and adding user agent and an url. The client then hands the ``request_dict`` off to the endpoint which does a simple translation of the ``request_dict`` to an ``AWSRequest`` object.
* Add user agent to the client config. Now that it is exposed in the client config, users can modify the user agent. I feel that we will want to change this, but if the user specifies a user agent, it completely overrides ours, as of now. We should discuss how we want to allow users to provide custom user agents.

This is the last PR that I need to get merged in order to submit a PR for presign feature, which I am getting very close to completeing depending on the status of these PRs.

cc @jamesls @danielgtaylor


jamesls:
:shipit:

PR #504, opened 2015-03-30
Generate url

kyleknap:
So this pull request is to expose an interface for supporting presigned url. Note that this PR does not include support for s3's presigned POST's. With this interface, in order to generate a presigned url. All you need is a ``RequestSigner``. With the ``RequestSigner`` you can call ``generate_url()`` method that takes in a request dictionary (from the serializer and prepare_request_dict) and will return a url that is presigned. Note that I considered calling the method ``presign_url()``, but I decided to name it how boto does. You also have the option to specify how many seconds till the url expires and if you want to sign with a different region name.

Also, this PR removes the last of the skips in the integration test for the clients-only branch.

cc @jamesls @danielgtaylor

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -460,10 +460,12 @@ class S3SigV4QueryAuth(SigV4QueryAuth):
 
 
 class S3SigV4PostAuth(SigV4Auth):
+    """
+    Presigns a s3 post
 
-    # Implementation doc here:
-    # http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-UsingHTTPPOST.html
-
+    Implementation doc here:
+    http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-UsingHTTPPOST.html
+    """
     def add_auth(self, request):
         datetime_now = datetime.datetime.utcnow()
         request.context['timestamp'] = datetime_now.strftime(SIGV4_TIMESTAMP)
@@ -672,6 +674,31 @@ class HmacV1QueryAuth(HmacV1Auth):
         request.url = urlunsplit(new_url_parts)
 
 
+class HmacV1PostAuth(HmacV1Auth):
+    """
+    Generates a presigned post for s3.
+
+    Spec from this document:
+
+    http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingHTTPPOST.html
+    """
+    def add_auth(self, request):
+        fields = request.context['s3-presign-post-fields']
+        policy = request.context['s3-presign-post-policy']
+        conditions = policy['conditions']
+
+        fields['AWSAccessKeyId'] = self.credentials.access_key
+
+        if self.credentials.token is not None:
+            fields['x-amz-security-token'] = self.credentials.token
+            conditions.append({'x-amz-security-token': self.credentials.token})
+
+        # Dump the base64 encoded policy into the fields dictionary.
+        fields['policy'] = base64.b64encode(json.dumps(policy))
+
+        fields['signature'] = self.sign_string(fields['policy'])
+
+
 # Defined at the bottom instead of the top of the module because the Auth
 # classes weren't defined yet.
 AUTH_TYPE_MAPS = {
@@ -682,6 +709,7 @@ AUTH_TYPE_MAPS = {
     'v3https': SigV3Auth,
     's3': HmacV1Auth,
     's3-query': HmacV1QueryAuth,
+    's3-presign-post': HmacV1PostAuth,
     's3v4': S3SigV4Auth,
     's3v4-query': S3SigV4QueryAuth,
     's3v4-presign-post': S3SigV4PostAuth,
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -488,7 +488,7 @@ class HmacV1Auth(BaseSigner):
         hoi = []
         if 'Date' in headers:
             del headers['Date']
-        headers['Date'] = formatdate(usegmt=True)
+        headers['Date'] = self._get_date()
         for ih in interesting_headers:
             found = False
             for key in headers:
@@ -501,6 +501,7 @@ class HmacV1Auth(BaseSigner):
         return '\n'.join(hoi)
 
     def canonical_custom_headers(self, headers):
+        hoi = []
         custom_headers = {}
         for key in headers:
             lk = key.lower()
@@ -509,7 +510,6 @@ class HmacV1Auth(BaseSigner):
                     custom_headers[lk] = ','.join(v.strip() for v in
                                                   headers.get_all(key))
         sorted_header_keys = sorted(custom_headers.keys())
-        hoi = []
         for key in sorted_header_keys:
             hoi.append("%s:%s" % (key, custom_headers[key]))
         return '\n'.join(hoi)
@@ -579,6 +579,12 @@ class HmacV1Auth(BaseSigner):
         signature = self.get_signature(request.method, split,
                                        request.headers,
                                        auth_path=request.auth_path)
+        self._inject_signature(request, signature)
+
+    def _get_date(self):
+        return formatdate(usegmt=True)
+
+    def _inject_signature(self, request, signature):
         if 'Authorization' in request.headers:
             # We have to do this because request.headers is not
             # normal dictionary.  It has the (unintuitive) behavior
@@ -591,6 +597,52 @@ class HmacV1Auth(BaseSigner):
             "AWS %s:%s" % (self.credentials.access_key, signature))
 
 
+class HmacV1QueryAuth(HmacV1Auth):
+    """
+    Generates a presigned request for s3.
+
+    Spec from this document:
+
+    http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html
+    #RESTAuthenticationQueryStringAuth
+
+    """
+    DEFAULT_EXPIRES = 3600
+
+    def __init__(self, credentials, expires=DEFAULT_EXPIRES):
+        self.credentials = credentials
+        self._expires = expires
+
+    def _get_date(self):
+        return str(int(time.time() + int(self._expires)))
+
+    def _inject_signature(self, request, signature):
+        query_dict = {}
+        query_dict['AWSAccessKeyId'] = self.credentials.access_key
+        query_dict['Signature'] = signature
+
+        for header_key in request.headers:
+            lk = header_key.lower()
+            # For query string requests, Expires is used instead of the
+            # Date header.
+            if header_key == 'Date':
+                query_dict['Expires'] = request.headers['Date']
+            # We only want to include relevant headers in the query string.
+            # These can be anything that starts with x-amz, is Content-MD5,
+            # or is Content-Type.
+            elif lk.startswith('x-amz-') or lk in ['content-md5',
+                                                   'content-type']:
+                query_dict[lk] = request.headers[lk]
+        # Combine all of the identified headers into an encoded
+        # query string
+        new_query_string = percent_encode_sequence(query_dict)
+
+        # Create a new url with the presigned url.
+        p = urlsplit(request.url)
+        new_url_parts = (p[0], p[1], p[2], new_query_string, p[4])
+        request.url = urlunsplit(new_url_parts)
+
+
 # Defined at the bottom instead of the top of the module because the Auth
 # classes weren't defined yet.
 AUTH_TYPE_MAPS = {
@@ -599,6 +651,7 @@ AUTH_TYPE_MAPS = {
     'v3': SigV3Auth,
     'v3https': SigV3Auth,
     's3': HmacV1Auth,
+    's3-query': HmacV1QueryAuth,
     's3v4': S3SigV4Auth,
     's3v4-query': S3SigV4QueryAuth,
     'v4-query': SigV4QueryAuth,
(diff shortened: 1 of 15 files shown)


botocore/auth.py
```diff
@@ -591,6 +594,52 @@ def add_auth(self, request):
...
+        for header_key in request.headers:
+            lk = header_key.lower()
+            # For query string requests, Expires is used instead of the
+            # Date header.
+            if header_key == 'Date':
+                query_dict['Expires'] = request.headers['Date']
+            # We only want to include relevant headers in the query string.
+            # These can be anything that starts with x-amz, is Content-MD5,
+            # or is Content-Type.
+            elif lk.startswith('x-amz-') or lk in ['content-md5',
```
kyleknap: I considered abstracting a public method out in HmacV1Auth, but I was trying to avoid touching existing code. The bottom line is that the auth class picks what headers it will sign for (much like the sigv4 signer), but it just does not have a public  ``headers_to_sign method``. I can add one though.

botocore/signers.py
```diff
@@ -130,3 +132,23 @@ def get_auth(self, signing_name, region_name, signature_version=None):
             auth = cls(**kwargs)
             self._cache[key] = auth
             return auth
+
+    def generate_url(self, request_dict, expires_in=None, region_name=None):
```
kyleknap: Note it may be a little weird that we use request dict instead of a request object, but once we refactor the auth class to accept request_dict, we can change the ``sign()`` method to the style of the ``generate_url()`` method.

kyleknap:
@jamels @danielgtaylor 

So the commits I recently pushed are for the ability to presign s3 post (both sigv2 and sigv4). The public method is ``build_post_form_args``. To get a sense on how the CLI and boto3 will use it, check out the integration tests in ``test_s3.py``. That has some examples of how you can leverage the presigner and ``requests`` library to make the post request.

One thing that would be great to get your opinions on is the interface in general. I feel that it is a little weird that the method is attached to the general signing class, but only applies to s3. Ideally for me, there would be some isolation between this method and the general signer class (because it only applies to s3 posts) but I have not been able to think of a better way. Maybe I can check that the service of the signer is ``s3``, and if not throw an error? Or rename it to have s3 in the name of the method? Also let me know what you think of the name ``build_post_form_args``. I picked it because that is the name boto uses, but the inputs and return values are a little different.

kyleknap:
In the latest commit, I exposed some client level methods, ``generate_url`` and ``build_s3_post_form_args``, in signers.py that will eventually be translated to actual methods of the client for the presigning methods. So I would pay more attention to the internals of the methods. When I do dangle them off the client, they will pretty much take the same arguments as they are currently minus the client argument.

botocore/signers.py
```diff
@@ -130,3 +134,275 @@ def get_auth(self, signing_name, region_name, signature_version=None):
...
+        }
+    """
+
+    if fields is None:
+        fields = {}
+
+    if conditions is None:
+        conditions = []
+
+    request_signer = client._request_signer
```
kyleknap: I realize that these are internal. However, we will get access to them when we add the appropriate event for adding methods to classes.

botocore/signers.py
```diff
@@ -130,3 +134,275 @@ def get_auth(self, signing_name, region_name, signature_version=None):
...
+    serializer = client._serializer
+
+    # We choose the CreateBucket operation model because its url gets
+    # serialized to what a presign post requires.
+    operation_model = client.meta.service_model.operation_model('CreateBucket')
```
kyleknap: I considered creating my own model for this. However, I felt ``CreateBucket`` is suitable. All I really need is the url when I serialize the request and the ``CreateBucket`` serialization is exactly how I need the url serialized. Let me know if you want me to change this.

tests/unit/auth/test_signers.py
```diff
@@ -516,8 +591,129 @@ def test_presign_with_security_token(self):
...
+        result_policy = json.loads(base64.b64decode(
+            result_fields['policy']).decode('utf-8'))
+        self.assertEqual(result_policy['expiration'],
+                         '2007-12-01T12:00:00.000Z')
+        self.assertEqual(
+            result_policy['conditions'],
+            [{"acl": "public-read"},
+             {"bucket": "mybucket"},
+             ["starts-with", "$key", "mykey"]])
+        self.assertIn('signature', result_fields)
```
kyleknap: I did not assert do assert equal because we dump a dictionary to json and there is no guarantee that gets dumped the same each time unless you order the json elements, which is not needed functionality-wise.

botocore/auth.py
```diff
@@ -23,6 +23,7 @@
 import functools
 import time
 import calendar
+import json
```
jamesls: from botocore.compat import json

botocore/auth.py
```diff
@@ -591,15 +629,91 @@ def add_auth(self, request):
...
+class HmacV1PostAuth(HmacV1Auth):
+    """
+    Generates a presigned post for s3.
+
+    Spec from this document:
+
+    http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingHTTPPOST.html
+    """
+    def add_auth(self, request):
+        fields = request.context['s3-presign-post-fields']
```
jamesls: This adds new preconditions to using the HmacV1PostAuth class that these keys are set in the request context.  Can we gracefully handle the case where these values are not set?

botocore/model.py
```diff
@@ -235,6 +236,13 @@ def operation_model(self, operation_name):
             raise OperationNotFoundError(operation_name)
         return OperationModel(model, self, operation_name)
 
+    def client_name_to_operation_name(self, name):
```
jamesls: Does this need to be in the models module?  Ideally the models should know nothing about clients, or even snake casing  Can we move this up to the client layer?

botocore/signers.py
```diff
@@ -130,3 +134,275 @@ def get_auth(self, signing_name, region_name, signature_version=None):
...
+        # Fix s3 host for s3 sigv2 bucket names
+        signature_type = signature_version.split('-', 1)[0]
+        fix_s3_host(request, signature_type, region_name)
+
+        auth.add_auth(request)
+        request.prepare()
+
+        return request.url
+
+    def build_post_form_args(self, request_dict, fields=None, conditions=None,
```
jamesls: I get that these are the names from boto, but I think we should be consistent in our naming, in one case we're using generate, in another we're using build.  I prefer generate.

botocore/signers.py
```diff
@@ -130,3 +134,275 @@ def get_auth(self, signing_name, region_name, signature_version=None):
             auth = cls(**kwargs)
             self._cache[key] = auth
             return auth
+
+    def generate_url(self, request_dict, expires_in=3600, region_name=None):
```
jamesls: What about generate_presigned_url, or something with presigned in the method name?  I think it makes it more clear exactly what kind of URL it's generating.

botocore/signers.py
```diff
@@ -130,3 +134,275 @@ def get_auth(self, signing_name, region_name, signature_version=None):
...
+        auth.add_auth(request)
+
+        # Fix s3 host for s3 sigv2 bucket names
+        signature_type = signature_version.split('-', 1)[0]
+        fix_s3_host(request, signature_type, region_name)
+        # Return the url and the fields for th form to post.
+        return {'url': request.url, 'fields': fields}
+
+
+def generate_url(client, client_method, client_kwargs=None, expires_in=3600):
```
jamesls: Would would you think about allowing the user to provide the HTTP method?

I could someone generating a presigned url for an operation that's normally a POST, but wanting to just curl the url with a GET.

botocore/signers.py
```diff
@@ -130,3 +134,275 @@ def get_auth(self, signing_name, region_name, signature_version=None):
...
+        signature_version = self._signature_version
+        if not signature_version.endswith(query_prefix):
+            signature_version += query_prefix
+
+        kwargs = {'signing_name': self._signing_name,
+                  'region_name': region_name,
+                  'signature_version': signature_version,
+                  'expires': expires_in}
+
+        auth = self.get_auth(**kwargs)
```
jamesls: Should we provide better error messages if you try to generate a presigned url for an unsupported signature version?  Right now you'll get something like`UnknownSignatureVersionError: Unknown Signature Version: v3https-query.` which may not be obvious to the caller that the real error is that this signature version does not have presigning url support.

jamesls:
Overall looks good, just some minor review feedback.  As far as the interface goes, I'd prefer to just have a separate class that does the s3 post forms stuff.  I agree that it doesn't feel right having it in the general signer class.

botocore/auth.py
kyleknap: Yeah I can make an empty dictionary for both.

botocore/model.py
kyleknap: @jamesls 
We can move it up to the client. After looking into dangling these presign methods onto the client, I learned that I actually do not need this method at all. I am inclined to remove it completely in that case. Thoughts?

botocore/signers.py
kyleknap: That is sounds good.

botocore/signers.py
kyleknap: Yeah that should not be hard to add. I think boto allows for it to.

botocore/signers.py
kyleknap: Yeah I can clear up the exception

botocore/signers.py
kyleknap: @jamesls 
So does something like ``generate_presigned_post()`` sound good?

kyleknap:
Alright in the very latest commit I did two things:
1) Update the code based on feedback
2) Dangle the presigning methods on clients as: ``client.generate_presigned_url``, ``client.generate_presigned_post``

With these additions presigning is officially available in boto3 as well.

Let me know what you think. @jamesls @danielgtaylor

botocore/model.py
```diff
@@ -1,6 +1,7 @@
 """Abstractions to interact with service models."""
 from collections import defaultdict
 
+from botocore import xform_name
```
danielgtaylor: Is this used anywhere?

botocore/signers.py
```diff
@@ -130,3 +148,320 @@ def get_auth(self, signing_name, region_name, signature_version=None):
...
+
+    :type Params: dict
+    :param Params: The parameters normally passed to
+        ``ClientMethod``.
+
+    :type ExpiresIn: int
+    :param expires_in: The number of seconds the presigned url is valid
+        for. By default it expires in an hour (3600 seconds)
+
+    :type HTTPMethod: string
```
danielgtaylor: Should we use `HttpMethod` here? I know services aren't that consistent but maybe we want to follow some rules here for our customizations?

botocore/utils.py
```diff
@@ -35,6 +35,11 @@
 # These are chars that do not need to be urlencoded.
 # Based on rfc2986, section 2.3
 SAFE_CHARS = '-._~'
+LABEL_RE = re.compile('[a-z0-9][a-z0-9\-]*[a-z0-9]')
```
danielgtaylor: I think this is equivalent and a little bit simpler: `[a-z0-9]-*[a-z0-9]`

tests/unit/test_signers.py
```diff
@@ -32,6 +35,9 @@ def setUp(self):
             'service_name', 'region_name', 'signing_name',
             'v4', self.credentials, self.emitter)
 
+
+class TestSigner(BaseSignerTest):
```
danielgtaylor: May want to add a test to make sure that the newly public properties can be accessed from the request signer (e.g. region name and such).

danielgtaylor:
Just a few minor comments. LGTM :+1:

botocore/model.py
jamesls: Is this still needed?

botocore/utils.py
jamesls: But isn't the `*` for the `[a-z0-9\-]` block?  That would mean something like aa-a wouldn't match anymore.

botocore/signers.py
jamesls: Yeah that sounds good.

jamesls:
Just some small stuff, otherwise looks good.  :shipit:

botocore/model.py
kyleknap: Nah. It did not get cleaned up after I removed the ``client_name_to_operation_name`` function I added to the model. Will remove

botocore/signers.py
kyleknap: Yeah that is fine with me. I do not have to big a preference for one of the other. I did it the way I did because acronyms are usually capitalized.

botocore/utils.py
kyleknap: I just copied and pasted this from handlers.py . I would be inclined not to change this to avoid accidental changes, if there is no bugs with it already.

tests/unit/test_signers.py
kyleknap: sounds good to me

kyleknap:
@danielgtaylor 
I updated the code based on your feedback. The only thing that I did not change was the regular expression as I was hesitant to change that as it was just a copy and paste from a different file. Let me know what you think.

botocore/utils.py
danielgtaylor: Ah right I wasn't thinking :fearful: Keep it as-is!

danielgtaylor:
Thanks, LGTM :ship:-it!

vlcinsky:
It would be great, if the method would allow also absolute expiration time.

Adding additional parameter `expires_in_absolute` with True of False would work (copying the way, `boto` does that.

PR #505, opened 2015-03-30
Ensure serialized REST-XML bodies are bytes.

### LINKED ISSUE #494: pyopenssl + strict handling of bytes
On trusty + python 2.7. Using a virtualenv. If you have pyopenssl+ndg-httpsclient+pyasn1 installed then botocore cannot call change_resource_record_sets. With:

    from botocore.session import get_session
    session = get_session()
    client = session.create_client("route53", "eu-west-1")
    client.change_resource_record_sets(
        HostedZoneId="1111",
        ChangeBatch={
            "Changes": [{"Action": "UPSERT", "ResourceRecordSet": {"Name": "www", "Type": "A"}}],
        },
    )

These are invalid parameters, but valid enough that the parameters are serialized and passed to requests. It looks like they serialize as UTF-8 because then i get:

      File "example.py", line 10, in <module>
    "Changes": [{"Action": "UPSERT", "ResourceRecordSet": {"Name": "www", "Type": "A"}}],
      File "botocore/client.py", line 315, in _api_call
      File "botocore/endpoint.py", line 110, in make_request
      File "botocore/endpoint.py", line 156, in _send_request
    success_response, exception):
      File "botocore/endpoint.py", line 207, in _needs_retry
      File "botocore/hooks.py", line 226, in emit
      File "botocore/hooks.py", line 209, in _emit
    response = handler(**kwargs)
      File "botocore/retryhandler.py", line 180, in __call__
      File "botocore/retryhandler.py", line 247, in __call__
    caught_exception)
      File "botocore/retryhandler.py", line 262, in _should_retry
      File "botocore/retryhandler.py", line 310, in __call__
    caught_exception)
      File "botocore/retryhandler.py", line 219, in __call__
      File "botocore/retryhandler.py", line 352, in _check_caught_exception
        raise caught_exception
    TypeError: data must be a byte string

This is actually coming from here: ``OpenSSL/SSL.py(947)send()``.

danielgtaylor:
This attempts to fix #494 and aws/aws-cli#1240 when someone has both
Botocore and PyOpenSSL installed in the same environment. It adds
a test to ensure that this continues to work.

cc @jamesls @kyleknap


jamesls:
What types do the other serializers produce as their body?

jamesls:
Also have you verified the integ tests and cli tests on this.  I'm a little concerned that this will cause breakage somewhere, we are changing the type and anything relying on that could break.

danielgtaylor:
@jamesls @kyleknap please have another look. I'll see about the coverage decrease, but let me know what you think of this new approach. Rather than do this for each individual serializer, this instead approaches the problem from another angle and ensures that the data sent into requests is always a `bytes` body if the input is one of the string types. It solves a few test failures and with the latest changes I'm able to run the Botocore, CLI, and Boto 3 integration tests.

jamesls:
So for my clarification, what are the allowed types for each of these layers?  Is this correct?

```
| Module       | Input Types   | Outputs   |
|--------------|---------------|-----------|
| serialize.py | str/bytes     | str/bytes |
| auth.py      | str/bytes     | str/bytes |
| endpoint.py  | str/bytes     | bytes     |
```

And then anything that works off a request object only has to handle the case where `request.body` is of type `bytes`.

Does this mean that we need to verify that each serializer and auth class needs to handle str/bytes?  Do we know if that's actually the case with our current code?  Is it possible to just pick one of the types and only support one type?

danielgtaylor:
@jamesls your chart is correct.

> Does this mean that we need to verify that each serializer and auth class needs to handle str/bytes? Do we know if that's actually the case with our current code?

Ideally with Boto 3 coming (and the ability to pass in either type) this should be the case already, but I'm not sure that it is. We probably do need to verify this.

> Is it possible to just pick one of the types and only support one type?

I think that if we do the conversion to bytes earlier (e.g. in the client operation methods) then we could standardize everything below that to use bytes. Would you prefer that?

jamesls:
I'm not sure about just converting to bytes as early as possible.  I usually see the opposite case, that is, keeping everything as str internally and converting down to bytes as late as possible.  While I think switching to bytes as early as possible *seems* like it would reduce a lot of redundant checks (as right now each layer potentially needs to type check for str vs. bytes), it also seems like switching everything over might take considerable effort.

Have you had a chance to audit the serializers?  If we're already handling the cases in the table above, then I'm fine with keeping it as is.  Switching everything to bytes internally should mostly be an internal change, so if we do decide that this keeps the internals cleaner we can always switch to it later.

danielgtaylor:
Module  | Inputs | Outputs
----------  | --------- | ----------
`auth.py` |  `str` **only** | `str`
`endpoint.py` | `str` or `bytes` | `str` or `bytes`
`serialize.py` | mostly `str`, but `bytes` work for blob types | `str`

The `auth.py` module simply doesn't support bytes as input. It makes too many assumptions about the input being a string. This PR changes that to allow support for byte bodies. The `endpoint.py` module passes things through as-is, but this PR is changing that to always send `bytes` to the HTTP library.

The `serialize.py` module is interesting. It only supports `bytes` in very specific cases and fails when they are passed elsewhere (e.g. a parameter of type `string` will fail to serialize when passed `b'hello'`). I'm not sure if we want to support this use-case.

Given all that, I still think this is a good change to make to support people with PyOpenSSL installed.

botocore/auth.py
```diff
@@ -241,7 +241,10 @@ def payload(self, request):
             request.body.seek(position)
             return hex_checksum
         elif request.body:
-            return sha256(request.body.encode('utf-8')).hexdigest()
+            body = request.body
```
jamesls: What about the first `if` condition?  Looks like it only supports bytes.  Do we need to worry about that case as well?

jamesls:
Closing, see https://github.com/boto/botocore/issues/494

PR #510, opened 2015-04-03
Add an event to hook into client class creation

jamesls:
This allows you to customize the client class creation in one of two ways:

1. Injecting directly into the class dict.
2. Adding a custom base class.

The integration test shows an example of how to do this.  Here are a few more examples:

```
import botocore.session


#### 1. Injecting a single function


def inject_single_method(class_attributes, **kwargs):
    class_attributes['get_user_agent'] = print_user_agent


def print_user_agent(self):
    return self._client_config.user_agent


s = botocore.session.get_session()
s.register('creating-client-class', inject_single_method)
s3 = s.create_client('s3')
print("==== Demo 1 ===")
print(s3.get_user_agent())


#### 2. Merging in a base class.
from botocore.client import BaseClient

class AddS3HighLevel(object):
    def upload_file(self, Filename, Bucket, Key):
        # Show that we have access to put_object.
        print("Fake uploading file with: %s" % self.put_object)
        # We can also factor our code into small units and helper methods.
        self._helper_method()

    def _helper_method(self):
        print("You called a helper method in: %s" % self)

    def download_file(self, Bucket, Key, Filename):
        print("Fake downloading file with: %s" % self.get_object)

def merge_base_class(base_classes, **kwargs):
    base_classes.insert(0, AddS3HighLevel)


s.register('creating-client-class.s3', merge_base_class)
s3 = s.create_client('s3')
print("\n\n==== Demo 2 ===")
s3.upload_file('filename', 'bucket', 'key')
s3.download_file('bucket', 'key', 'filename')


#### 3. How you can test the base class style customizations.

class HighLevel(object):
    def high_level_thing(self, name):
        self.put_object('bucket', 'key', name)


def merge_base_class2(base_classes, **kwargs):
    base_classes.insert(0, HighLevel)

s.register('creating-client-class.s3', merge_base_class2)

s3 = s.create_client('s3')
# Verify our "high_level_thing" method was added:
assert hasattr(s3, 'high_level_thing')

# And now here's an example of how to test this:


class StubTest(HighLevel):
    def put_object(self, *args, **kwargs):
        self.put_object_called = [args, kwargs]


def test_uploads():
    stubbed = StubTest()
    stubbed.high_level_thing('name')
    assert stubbed.put_object_called == [('bucket', 'key', 'name'), {}]


test_uploads()
```

cc @kyleknap @danielgtaylor


kyleknap:
Looks good. :ship:

jamesls:
Pushed to develop via 53a3074

PR #511, opened 2015-04-06
Serialize blobs as base64 strings for rest-json

kyleknap:
We were not serializing blobs for rest-json. We just dumped the entire json body into the body with no serialization. Now it follows the serialization code path for the JSON serializer.

cc @jamesls @danielgtaylor


jamesls:
:shipit:

PR #515, opened 2015-04-13
Standardize env vars to use AWS_ prefix

jamesls:
This standards the environment variables used in botocore.session to be consistent with other SDKs.  This includes two things:

1. Support `AWS_PROFILE` and `AWS_REGION` (the former is from aws/aws-cli#1281)
2. Change the default `BOTO_` env vars to use `AWS_` instead.

This is a backwards incompatible change, but there is a mitigation path is customers do in fact want the old mapping.  They can provide the old env var mapping to the `__init__` when creating a session.

cc @kyleknap @danielgtaylor

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -72,7 +72,7 @@ class Session(object):
     SessionVariables = {
         # logical:  config_file, env_var,        default_value
         'profile': (None, ['AWS_DEFAULT_PROFILE', 'AWS_PROFILE'], None),
-        'region': ('region', ['AWS_DEFAULT_REGION', 'AWS_REGION'], None),
+        'region': ('region', 'AWS_DEFAULT_REGION', None),
         'data_path': ('data_path', 'AWS_DATA_PATH', None),
         'config_file': (None, 'AWS_CONFIG_FILE', '~/.aws/config'),
         'provider': ('provider', 'BOTO_PROVIDER_NAME', 'aws'),
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -299,8 +299,9 @@ class Session(object):
                 config_name = None
             if logical_name == 'profile' and self._profile:
                 value = self._profile
-            elif 'env' in methods and envvar_name and envvar_name in os.environ:
-                value = os.environ[envvar_name]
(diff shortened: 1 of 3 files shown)


kyleknap:
The code looks good, but I had a general question related to the region environment variable.

I like the change for profile. Out of curiosity what was the reason for making the change for ``AWS_REGION``? Because the rest of the SDK's do not support ``AWS_REGION`` right? Was it just for consistency sake (It does look better if we need to support ``AWS_PROFILE``)?

If it is not supported by the other SDK's, I would be hesitant to make a change in the name now (even though it seems like the most logical naming convention if it is ever decided to be supported by the rest of the SDKs) and later down the road a different convention is chosen, which would force us to support that as well. This would mean that we support three different environment variables for region. But we could always influence the other SDKs to adopt our naming convention since we are the first ones to do it, if they do decide to support it later on.

Also we should make sure our getting started docs to state to use ``AWS_PROFILE`` as well.

jamesls:
Yeah it looks like we're being inconsistent.  I was under the impression the other SDKs support this.  As far as I can tell, javascript, ruby, and go support `AWS_REGION`.  Java/.NET/PHP do not.

Let me do some research on this, but I'm inclined to actually pull support for `AWS_REGION` until we have consensus on this.

jamesls:
@kyleknap I've updated the code to remove support for `AWS_REGION` for now.  We can always support it later once we discuss with the other SDK teams.  Let me know if there's any other feedback.

kyleknap:
Thanks! :ship:

krschultz:
Wanted to note that the [docs](http://docs.aws.amazon.com/cli/latest/topic/config-vars.html) are still out of date and specify AWS_DEFAULT_PROFILE. It works fine in the AWS CLI tool because of the backwards compatibility, but we were having trouble authorizing tools we had written since the AWS SDKs are only looking for AWS_PROFILE. Once we switched over to AWS_PROFILE everything worked much more seamlessly.

krschultz:
Also seems like [this page](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-environment) is out of date as well.

PR #516, opened 2015-04-13
Old style interface should warn to stderr by default

jamesls:
I noticed when I was trying out the old interface that the `DeprecationWarning` does *not* print to stderr.  We really should be at the point out where the error is front and center whenever you try to do anything with the old interface.  This change does that.

In order to help with migration I also sketched out a migration guide for going from the old interface to the new client interface so we have something we can point customer to when they run into issues.

I also manually verified the errors are on stderr now:

```
>>> import botocore.session
>>> s.get_service('ec2')
botocore/session.py:560: ImminentRemovalWarning: get_service is deprecated and will be removed.  Use create_client instead.
  "Use create_client instead.", ImminentRemovalWarning)
botocore/service.py:231: ImminentRemovalWarning: get_service is deprecated and will be removed.  Use Session.create_client instead.
  ImminentRemovalWarning)
Service(ec2)
```

cc @kyleknap @danielgtaylor

### DIFF
--- a/docs/source/client_upgrades.rst
+++ b/docs/source/client_upgrades.rst
@@ -1,0 +1,250 @@
+.. _client-upgrades:
+
+Upgrading to Clients
+====================
+
+Background
+----------
+
+Version 0.66.0 of botocore added support for clients (10/16/2014)
+added initial support for clients.  This provided an alternate
+interface to making AWS calls that provided a number of benefits over
+the existing interface.  At the time, both interfaces were added
+so that users could opt in to trying the new clients.
+
+Below is an example of the old interface:
+
+.. code-block:: python
+
+    import botocore.session
+    session = botocore.session.get_session()
+    s3 = session.get_service('s3')
+    endpoint = s3.get_endpoint('us-west-2')
+    list_objects = s3.get_operation('ListObjects')
+    http, response = list_objects.call(endpoint, Bucket='mybucket')
+    if http.status_code == 200:
+        print("Contents: %s" % response['Contents])
+    else:
+        print("API call failed, status code: %s, error: %s" % (
+            http.status_code, http.content))
+
+
+Here's an example of the newer (preferred) client interface:
+
+.. code-block:: python
+
+    import botocore.session
+    session = botocore.session.get_session()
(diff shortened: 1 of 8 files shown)


docs/source/client_upgrades.rst
```diff
@@ -0,0 +1,250 @@
...
+    ec2 = session.get_client('ec2', 'us-west-2')
+    parsed = ec2.describe_instances()
+    print("The status code is:",
+          parsed['ResponseMetadata']['HTTPStatusCode']))
+
+
+Error Handling
+--------------
+
+With clients, exceptions are now raised on any non 200 response.
```
kyleknap: I feel that more specifically it is any non 2xx response.

docs/source/client_upgrades.rst
```diff
@@ -0,0 +1,250 @@
...
+
+**New**
+
+.. code-block:: python
+
+    from botocore.exceptions import ClientError
+
+    ec2 = session.get_client('ec2', 'us-west-2')
+    try:
+        parsed = ec2.describe_instances()
```
kyleknap: Will this call throw a ResourceUnavailable error? From my experience, this call works for me without errors nearly all of the time.

kyleknap:
Looks good. I had a couple comments. This will help a lot if a user upgrades when we do the alpha release and has not switched to clients. :ship:

docs/source/client_upgrades.rst
jamesls: I would assume it's possible, even if unlikely.  If there's another call you'd prefer I can switch it.

docs/source/client_upgrades.rst
jamesls: Fixed.

docs/source/client_upgrades.rst
kyleknap: @jamesls 
Can we just make it a call to describe_instances using an instance-id of a non-existent instance?

jamesls: Sure, I'll update.

jamesls:
Cool, I believe all the feedback has been incorporated.

kyleknap:
Thanks! Good to merge.

PR #519, opened 2015-04-16
Add paginators / waiters for ECS

euank:
I had a go at writing paginators and waiters for ECS.
Currently I have only tested them in botocore via the aws cli, but all of them appear to work well there.


euank:
Thanks for adding them! Closing this as unneeded due to b2f4cee6af900cb74a2287e65b1ab162999a6f6c

PR #520, opened 2015-04-17
Add more helpful error message for positional args in clients

jamesls:
Before:

  TypeError: _api_call() takes exactly 1 argument (2 given)

After:

  TypeError: list_objects() only accepts keyword arguments.

From: https://github.com/boto/boto3/issues/35

cc @kyleknap


kyleknap:
:ship:

Issue #521, opened 2015-04-17
UserData should have a Blob shape


charleso:
As far as I can tell [RunInstances.userData](https://github.com/boto/botocore/blob/develop/botocore/data/aws/ec2/2014-10-01.normal.json#L10302) (and probably other cases) needs to be base64 encoded and (I think) therefore should have a shape of "Blob"? An example of a base64 encoded member that does have Blob is [PublicKeyMaterial](https://github.com/boto/botocore/blob/develop/botocore/data/aws/ec2/2014-10-01.normal.json#L6655).

I'm happy to raise a PR to fix the json, but it sounds like it's better to raise an issue. Thanks in advance.

kyleknap:
@charleso 

Yeah you are correct about the userData parameter needing to be modeled as a blob. To compensate, we actually have a handler: https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L383 that will base64 encode the user data for you. Are you seeing the data not getting base64 encoded from your side?

Otherwise thanks for the offer to fix it.

charleso:
https://github.com/boto/botocore/pull/522

kyleknap:
As I mention in my comment for https://github.com/boto/botocore/pull/522, we will look into updating the upstream source to use blobs as opposed to strings to model UserData.

charleso:
@kyleknap No worries. I misinterpreted the "thanks for the offer" part of your message. :) Apologies for the noise.

jamesls:
Closing out issue as per @kyleknap [comment](https://github.com/boto/botocore/issues/521#issuecomment-94861165)

PR #522, opened 2015-04-20
Issue #521: Fix UserData members to be Blobs for base64 encoding

charleso:
I'm not 100% sure whether I should change the `UserData` structure as well:

https://github.com/boto/botocore/blob/develop/botocore/data/aws/ec2/2014-10-01.normal.json#L11136

### DIFF
--- a/botocore/data/aws/ec2/2014-09-01.normal.json
+++ b/botocore/data/aws/ec2/2014-09-01.normal.json
@@ -6168,7 +6168,7 @@
           "locationName":"additionalInfo"
         },
         "UserData":{
-          "shape":"String",
+          "shape":"Blob",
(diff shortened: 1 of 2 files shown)


kyleknap:
@charleso 
It is my bad that I did not mention this in my comment on the issue you opened, but as per our [contributing guidelines](https://github.com/boto/botocore/blob/develop/CONTRIBUTING.rst), we do not accept pull request that update the ``.normal.json`` models because we pull these from an internal upstream source, and the change would need to happen at the source.

We will look into updating the shape for UserData upstream though and removing the custom handler we have to account for the wrong type. Thanks again though for pointing this out. Closing PR.

PR #523, opened 2015-04-21
Add integration tests for new services

jamesls:
This adds basic smoke tests for the services added in the last
6 months or so.

cc @kyleknap

### DIFF
--- a/tests/integration/test_smoke.py
+++ b/tests/integration/test_smoke.py
@@ -23,17 +23,21 @@ SMOKE_TESTS = {
                     'ListStacks': {}},
  'cloudfront': {'ListDistributions': {},
                 'ListStreamingDistributions': {}},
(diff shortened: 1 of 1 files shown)


kyleknap:
Two questions:
1) Are any of these services in preview (i.e. do you need to subscribe to them)? Just asking because sometimes for some of the newer services, you need to subscribe to them as they are in preview, or they will throw errors.
2) What is your thoughts on adding integration tests to the error smoke tests for these services?

jamesls:
1) Nope, AFAIK these are all generally available.  Verified these pass for me on multiple AWS accounts.

2) Not opposed.  We started that, but it looks like it's currently incomplete.  Would not be opposed to another PR that fleshes this out more.

kyleknap:
Awesome. :ship: It would be nice as principle for every service to have one success case and one error case for integration test. We can add a PR for that later down the road.

Issue #514, opened 2015-04-12
EC2 ImageAvailable Waiter contains an unreachable failure case


brendanhay:
The [ec2/2014-10-01.waiters.json](https://github.com/boto/botocore/blob/75be2f130e40d02f004f22b55f236d4ef5fd10f4/botocore/data/aws/ec2/2014-10-01.waiters.json#L140) file currently contains the expected match of `"failed"` for the failure state.

Is this actually correct? It appears that the  match should be `"deregistered"` to match the failure case against the actual `ImageState` [enumeration values](https://github.com/boto/botocore/blob/75be2f130e40d02f004f22b55f236d4ef5fd10f4/botocore/data/aws/ec2/2014-10-01.normal.json#L6454).

kyleknap:
It is weird that the current failure state is not listed. This may just be a documentation issue and we just need to update our model. As to adding ``"deregistered"`` as the failure case, I am not sure if we would want to add that because you could be trying to register a deregistered ami and you want to wait till the ami is available. So, if you fail out at a ``"deregistered"`` state you will not be able to successfully wait for an image to become available when registering it from a deregistered state.

Otherwise are you running into issues using this waiter?

brendanhay:
Then this is possibly just a request to please-educate. I'll outline my current understanding of how the waiter is configured and perhaps you can correct me:

1. The `DescribeImages` call is used to retrieve a response upon which the two acceptors for `success` and `failure` are checked against.
2. In the `state: success` case the `Images` list is checked using `matcher: pathAll` which succeeds when all `.State` accessors in the list have the expected value `available`.
3. In the `state: failure` case the `Images` list is checked using `matcher: pathAny` which
succeeds when any `.State` accessor in the list has the expected value `failed`.

If the above is true then there is no way `3.` can succeed (ie. the failure case is pointless) since `.State` can never have the value `failed` according to the enumeration for `ImageState`, which lists the possible values as `available` and `deregistered`.

Am I missing something obvious?

> I'm generating waiters from the botocore definitions for the Haskell SDK,  which in its current form expects the `expected` values to exist at the value-level _somewhere_, (ie. can be deserialised to a value as opposed to a stringly-typed check). which is where the mismatch between `expected: failed` and `Images[].State == (available|deregistered)` comes from.

kyleknap:
[...]

PR #526, opened 2015-04-24
Update ec2 model for docs and enumerated values

kyleknap:
This updates the enumerated values for ``ImageState`` and the docs are updated such that "for Linux" is appended to whenever "EC2 User Guide" is used.

Fixes: https://github.com/boto/botocore/issues/514

cc @jamesls


jamesls:
:shipit:

PR #527, opened 2015-04-24
Update and simplify botocore loaders to new dir layout

jamesls:
There are three main changes in this PR.  In order of significance:

* Convert the existing directory structure over to the new standard layout used by other AWS SDKs (php, ruby, etc.).  This structure is described [here](https://github.com/boto/botocore/pull/527/files#diff-f8c1b99f0b4538cf77c2052b498c9630R57)
* Update the existing code to work with this new structure.
* Refactor the loader.py code by removing unused features and making it more modular.  Part of this work is so that other libraries (such as boto3) can leverage the loader logic.

The first two points are pretty straightforward.  The third point (which is all isolated to the loaders.py module and associated tests) is more involved.

The biggest change conceptually in the loaders module is the introduction of a required "type name" in all the public APIs.  That is, it's not sufficient to just say `loader.list_available_services()`.  You have to tell it what type you're interested in.  For example:

```
>>> loader.list_available_services(type_name='service-2') # This is the new ".normal.json"
>>> loader.list_available_services(type_name='paginators-1') # All services that have pagination

# And in boto3, after adding to the .search_path, we can now say:

>>> loader.list_available_services(type_name='resource=1')  # All services with a resource definition
```

The other big thing is that I removed the "fuzzy API date" searching.  If you want to load a specific API version, you have to specify the *exact* API version.  To help with this, I've added a `list_api_versions` method.

cc @kyleknap


jamesls:
There's actually a few changes I'm going to need to make based on updating boto3, specifically with `list_available_services` and `determine_latest_version`.

The problem is that both of these assume that the directory structure is sufficient to extract meaningful data about what is or isn't available (in fact, I even comment on this in this PR: https://github.com/boto/botocore/pull/527/files#diff-f8c1b99f0b4538cf77c2052b498c9630R219).

We actually need to be able to specify the type name (`services-2`, `paginators-1`, `resources-1`) when asking about the latest available and the list of all services.  For example, the latest available service for `services-2` may not necessarily mean that there's a `resources-1` model for that same API version.  Similarly listing all the available resources may differ we only care about resources (i.e boto3).

A concrete example of breakage is doing `loader.load_service_model('ec2', 'resources-1')` in boto3.  It will try to find the latest API version for EC2 and assuming it can load a resource for it, which it currently cannot so it will raise an exception.

To fix this I'll be adding these changes:

* Modify determine_latest_version and list_available_services to require a ``type_name`` param, just like load_service_model.
* Update the file loader interface (currently only JSONFileLoader is implemented) to add an `exists()` method that will return True/False if the resource exists.

jamesls:
@kyleknap I've updated with the necessary changes for boto3 as well as the latest model updates from develop.  Should be ready now for another look.  Thanks.

jamesls:
https://github.com/boto/boto3/pull/104 is an example of how this new loader API can be used.  Now boto3 simply needs to add its internal directory to the [search path](https://github.com/boto/boto3/pull/104/files#diff-75fa8f276d7839b295fcc5e9b5ab596eR94) and then specify the [type name](https://github.com/boto/boto3/pull/104/files#diff-75fa8f276d7839b295fcc5e9b5ab596eR246) when dealing with services that must contain a resource definition.

botocore/loaders.py
```diff
@@ -1,425 +1,379 @@
...
+            # by searching for the corresponding type_name in each
+            # potential directory.
+            possible_services = [
+                d for d in os.listdir(possible_path)
+                if os.path.isdir(os.path.join(possible_path, d))]
+            for service_name in possible_services:
+                full_dirname = os.path.join(possible_path, service_name)
+                api_versions = os.listdir(full_dirname)
+                for api_version in api_versions:
+                    full_load_path = os.path.join(full_dirname, api_version, type_name)
```
kyleknap: line too long

botocore/loaders.py
```diff
@@ -1,425 +1,381 @@
...
+        This will traverse the search path and look for all known
+        services.
 
-        # We didn't find anything that matched on any path.
-        raise DataNotFoundError(data_path=data_path)
+        :type type_name: str
+        :param type_name: The type of the service (service-2,
+            paginators-1, waiters-2, etc).  This is needed because
+            the latest API version available can depend on the service
+            type.  For example, the latest API version available for
```
kyleknap: Docstring is a little weird here. Whether it is the latest api version should not matter. It is based on if the model type exists in any of the api versions.

tests/unit/test_loaders.py
```diff
@@ -23,157 +23,216 @@
...
-        loader = Loader(data_path=self.environ['BOTO_DATA_PATH'])
-        paths = loader.get_search_paths()[:-1]
-        self.assertEqual(paths, ['c:\\path1', 'c:\\path2'])
-
-
-if __name__ == "__main__":
-    unittest.main()
+        # Fake mapping of directory name
+        # to directory names.
+        fake_directories = {
```
kyleknap: I think that I would have actually set up a temporary directory with tempdir so I would not have to mock out some of the methods in the os library. Then set the data path to this that temporary directory.

For me either way is fine, just suggesting a different way in order that may simplify those two tests.

kyleknap:
It looks good. Looks like you updated the code after my pep8 comment through the pr-check script. Had a couple of comments about tests and docstrings, but nothing too significant :ship:

Looks like you also need to rebase on develop and pull in the latest model changes, which should only be dynamodb (not sure if you rebased after my ec2 pr).

tests/unit/test_loaders.py
jamesls: This was how the previous loader tests were setup and I found them hard to follow.  For me I find this harder to read because I have to manually hunt through the other files on disk systems and see exactly what the file system layout is in order to understand the test.  What I like about this approach is I can read the test and understand exactly how it's setup.

I do think that some of the boiler plate here can be cleaned up though.  Let me see what I can do.

jamesls:
The PR should have been rebased on develop as of the latest release a few hours ago.  I'll double check this is the case.

tests/unit/test_loaders.py
kyleknap: Yeah I can see that. If that is an issue you could also just make a helper function that generates directories and files based on the dictionary you pass to it. But at that point, I am not sure if it is worth the extra effort to add the utility. Mocking methods in ``os`` for me is fine as long as we do not plan to use any more of the filepath-handling related methods in ``os``.

kyleknap:
Hmm the merge button is grey right now. Usually means that I need to rebase.

jamesls:
Probably because I merged in the _retry bugfix PR.

jamesls:
@kyleknap pushed [a commit](https://github.com/jamesls/botocore/commit/4d08820408d61dd558504ada3f3542e3c35857ec) to extract out all the test boilerplate.  I think the tests are much easier to read now.

jamesls:
Also rebased against develop.  Looks like it's because of the _retry PR I merged, which makes sense.  Any changes to anything in botocore/data/ in develop will require a rebase to merge cleanly:

```
$ git rebase develop
First, rewinding head to replay your work on top of it...
Applying: Update and simplify botocore loaders to new dir layout
Applying: Move services to standard directory structure
Using index info to reconstruct a base tree...
M	botocore/data/aws/_retry.json
Falling back to patching base and 3-way merge...
Applying: Fix pr-check issues
Applying: Fix docstring
Applying: Move out boiler plate for dir patching to helper method
```

kyleknap:
Recent commits look good. Those tests look much cleaner. :ship: again.

PR #529, opened 2015-04-27
Simplify retry configuration

jamesls:
This changes two things:

* Move the ``throttling_exception`` rule up to the default errors to retry
* Move the ``throttling`` rule up to the default errors to retry

This came about because the ``workspaces`` service would receive
400/ThrottlingException errors that were not being retried.

The simple fix would be to add this missing condition to a new
``workspaces`` entry.

However, the error can occur for any json or rest-json service and
we'd need to keep this up to date every time we added support for a new
json/rest-json service.  Same thing with the 400/Throttling exception.
This will happen for any query service.  Moving this up to the default
set of errors to retry means we won't have to keep updating this as new
errors are added.

It also has the benefit that the remaining per-service conditions are
just the special cases.

cc @kyleknap

### DIFF
--- a/botocore/data/aws/_retry.json
+++ b/botocore/data/aws/_retry.json
@@ -55,133 +55,9 @@
           "general_socket_errors": {"$ref": "general_socket_errors"},
           "general_server_error": {"$ref": "general_server_error"},
           "service_unavailable": {"$ref": "service_unavailable"},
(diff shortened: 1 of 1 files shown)


kyleknap:
Makes sense. Looks good. :ship:

PR #530, opened 2015-04-29
Port user_agent_extra from session over to client config

jamesls:
This is the ability to append to an existing user agent.

Note that this branch is based off of the clients-only branch, which I just rebased against develop and force-pushed.

cc @kyleknap


kyleknap:
Looks good. :ship:

jamesls:
Merged into the clients-only branch via e770f6a

PR #531, opened 2015-04-30
Fix pagination issue

kyleknap:
The issue is we need to stop paginating if we recieve an empty string for
a next token marker. Before, if we did not have a check to see that the next
token was not the same as the previous token and throw an error, we would have
hit an infinite loop.

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -212,8 +212,12 @@ class PageIterator(object):
         next_tokens = []
         for token in self._output_token:
             next_token = token.search(parsed)
+            # We do not want to include any empty strings as actual tokens.
+            # Treat them as None.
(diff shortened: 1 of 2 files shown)


botocore/paginate.py
```diff
@@ -211,7 +211,9 @@ def _get_next_token(self, parsed):
                 return [None]
         next_tokens = []
         for token in self._output_token:
-            next_tokens.append(token.search(parsed))
+            next_token = token.search(parsed)
+            if next_token:
```
kyleknap: JmesPath expression will return None if the next token is not found. Should make consistent with having no token in response by appending None as opposed to just not appending anything to the next token list.

Also should add comment

PR #535, opened 2015-05-20
Add Documentation to Botocore

kyleknap:
This PR add documentation of the various services to botocore. This includes documentation for clients, waiters, and paginators.

Note that this PR will not pass tests till this bcdoc PR is merged: https://github.com/boto/bcdoc/pull/35

cc @jamesls


botocore/docs/utils.py
```diff
@@ -0,0 +1,614 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import inspect
+from collections import namedtuple
+
+
+def py_type_name(type_name):
+    """
+    Get the Python type name for a given model type.
```
jamesls: Summary should be on the first line of the docstring.

botocore/docs/utils.py
```diff
@@ -0,0 +1,614 @@
...
+    if short_name.startswith('Amazon'):
+        short_name = short_name[7:]
+    if short_name.startswith('AWS'):
+        short_name = short_name[4:]
+    if short_name and short_name.lower() not in official_name.lower():
+        official_name += ' ({0})'.format(short_name)
+    return official_name
+
+
+def get_instance_methods(instance):
```
jamesls: Can rename this method?  It's not technically true as it only returns the public methods.  Also the docstring says it returns variables but I don't think that's the case.

botocore/docs/utils.py
```diff
@@ -0,0 +1,614 @@
...
+        'blob': "b'bytes'",
+        'boolean': 'True|False',
+        'list': '[...]',
+        'map': '{...}',
+        'structure': '{...}',
+        'timestamp': 'datetime(2015, 1, 1)',
+    }.get(type_name, '...')
+
+
+def get_official_service_name(service_model):
```
jamesls: This seems like something that belongs in the service model.  I wouldn't be opposed to adding a property for this.

botocore/docs/utils.py
```diff
@@ -0,0 +1,614 @@
...
+class DocumentedShape (_DocumentedShape):
+    """Use this class to inject new shapes into a model for documentation"""
+    def __new__(cls, name, type_name, documentation, metadata=None):
+        if metadata is None:
+            metadata = []
+        return super(DocumentedShape, cls).__new__(
+            cls, name, type_name, documentation, metadata)
+
+
+class BaseMethodSignatureDocumentor(object):
```
jamesls: These classes seem out of place in a utils module.  These classes have a lot of the core documentation logic.  Seems like they should be in their own module.

botocore/docs/utils.py
```diff
@@ -0,0 +1,614 @@
...
+
+        :type include: Dictionary where keys are parameter names and
+            values are the shapes of the parameter names.
+        :param include: The parameter shapes to include in the documentation.
+
+        :type exclude: List of the names of the parameters to exclude.
+        :param exclude: The names of the parameters to exclude from
+            documentation.
+        """
+        self._document_signature(section, name, method_obj, include=include,
```
jamesls: I don't understand the point of this.  why not just use document_signature.

jamesls:
I'm still going through this, but I wanted to share some early EOD feedback so far.  This may change as I get more into the code:

* You have a typo in your class names: [documenter](https://www.google.com/search?q=documenter) instead of [documentor](https://www.google.com/search?q=documentor)

Also, can we consolidate these classes?  My initial take is that there seems to be numerous classes, several of which can be simplified or removed because they don't provide any benefit.  It would also simplify the code.  For example:

* There's a BaseMethodSignatureDocumenter with a single method that then delegates to an internal method with the same name (with the '_' prefix).

*  While I understand that the leaf classes make sense (RequestParams and ResponseParams, RequestExample and ResponseExample), having multiple levels of "base" classes makes the code hard to follow (RequestParams -> BaseParams, but BaseParams -> BaseShape).  Perhaps we can use composition instead of inheritance for the BaseParams -> BaseShape  (and BaseExample -> BaseShape) level?  All the base base class "BaseParams" is primarily handle the dispatch to specific shapes.

* BaseMethodSignatureDocumentor (and its subclasses) don't need to be classes.  They aren't really taking advantage of anything a class would give you.  There's no shared state across method calls (well there's only one method to begin with).  Polymorphism also isn't being leveraged in any consumers either, each subclass has a different code path (i.e in the client documenter CustomMethodSignatureDocumentor().document_signature is used for non model stuff and         ModelDrivenMethodDocumentor().document_method is used for methods from models).  So there's not really any benefit from them being classes.

Also, I'm not a huge fan of generating these rst files in the conf.py.  That's a hack that boto3/cli does and I'd prefer to not go down that route again. Primarily because we can't really test the code in conf.py.

Just some initial thoughts...I'll pick this up again tomorrow.

kyleknap:
@jamesls 

I updated the PR based on your initial feedback. Here is a summary of mainly what changed:
* Moved many of the functionality out of utils.py into its own separate modules.
* Converted some of the documenter classes to methods to simplify inheritance of the ParamDocumenters and the ExampleDocumenters
* Moved logic out of the conf.py into the botocore.docs module. It is literally only an import and call of a method in the conf.py.
* Add some more tests for some of the missing test cases.

Let me know what you think.

botocore/docs/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import os
+
+import botocore.session
+from botocore.compat import six
+from botocore.docs.service import ServiceDocumenter
+
+
+def generate_docs():
```
jamesls: I would suggest adding a parameter that lets you specify the output directory.  That way we don't depend on the fact that we can right to `<root>../docs` and we can also just have the docs/conf.py specify the directory botocore should generate documentation.  Also, we won't have to use a patch in test_docs.py.

botocore/docs/client.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def _add_client_method(self, section, method_name, method):
+        section = section.add_new_section(method_name)
+        if self._is_custom_method(method_name):
+            self._add_custom_method(section, method_name, method)
+        else:
+            self._add_model_driven_method(section, method_name)
+
+    def _is_custom_method(self, method_name):
+        return method_name not in self._client._PY_TO_OP_NAME
```
jamesls: `_PY_TO_OP_NAME` is used in a few places in this change.  Let's just expose a public API for this.  Perhaps we can add something to `client.meta` that allows you to get this mapping?

tests/unit/docs/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+        if is_required:
+            required_list = params_shape.get('required', [])
+            required_list.append(param_name)
+            params_shape['required'] = required_list
+
+    def assert_contains_line(self, line):
+        contents = self.doc_structure.flush_structure().decode('utf-8')
+        self.assertIn(line, contents)
+
+    def assert_contains_lines(self, lines):
```
jamesls: Can we update this to at assert the lines are in order?  Right now this will also pass with any order.

tests/unit/docs/test_params.py
```diff
@@ -0,0 +1,262 @@
...
+                'member': {'shape': 'String',
+                           'documentation': 'A string element'}}})
+        self.add_shape_to_params('Foo', 'List', 'This describes the list.')
+
+    def test_request_params(self):
+        self.request_params.document_params(
+            self.doc_structure, self.operation_model.input_shape)
+        self.assert_contains_lines([
+            ':type Foo: list',
+            ':param Foo: This describes the list.',
```
jamesls: Perhaps we could add some kind of transition text here.  I found this a bit hard to follow in the generated docs.  Something like "Each element of this list is a... <docs>".

jamesls:
Overall I think this is much better.  Just had a few small things.

kyleknap:
@jamesls 

I think I addressed all of your feedback. The most notable changes are the following:
* ``generate_docs()`` requires a root directory as an argument
* exposed an ``client.meta.method_to_api_mapping`` attribute that exposes ``_PY_TO_OP_NAME``
* Assert the order along with the contents for the tests
* Update the documentation to the latest botocore pagination interface.

jamesls:
:shipit:  Looks good.  Thanks for updating.

PR #536, opened 2015-05-20
Clients only

kyleknap:
This merges the clients only branch onto the develop branch.

There will be a subsequent PR to update the CLI as well. The clients-only branch switched around where some of our methods were and it affects how the CLI imports them.

cc @jamesls


kyleknap:
Here is the subsequent CLI PR: https://github.com/aws/aws-cli/pull/1336

jamesls:
:shipit:  Looks like travis is bogged down at the moment on just one python version.  I'm fine with merging as is.

kyleknap:
Cool. I have run the build twice now and it has passed on all python versions across those two builds when you pull the builds together.

PR #537, opened 2015-05-21
Add client-method-called event to botocore

kyleknap:
This event will help with the issue with the boto3 dynamodb customizations where if a dictionary gets passed to botocore, it gets modified.

cc @jamesls


jamesls:
:shipit:  My only feedback is that I'd prefer to call this event name something else.  Ideally this event name communicates the fact that this is the place to construct the api params that will be used for serialization.  Something like, `want-api-params`, `client-params-wanted`, etc.

kyleknap:
How about ``provide-client-params`` or ``client-params-provided``? I think for consistency the later is better based on verb tense

jamesls:
My vote would be the first choice.  I see the providing of the client params as something the event is asking the handlers to do, so the past tense doesn't make sense yet since the handlers haven't yet provided the client params.

kyleknap:
Sounds good. I will make the change.

jamesls:
:shipit:

PR #540, opened 2015-05-22
Improve repeated method calls in botocore

jamesls:
Speeds up botocore quite a bit when making repeated calls
to the same operation, e.g. dynamodb.get_item().

The main area of improvement was due to the fact that the operation models were not being cached so repeated calls to `get_item` would result in the op model being recreated each call.  This meant that even though the operation model itself would cache things like the input/output shape, that had no effect because we were throwing away the operation model afterwards.

The other thing that profiling showed was that the pformat() call was slowing things down.  This is still useful information so rather than `pformat()` we're just logging the header dict directly.

This gives performance that's on par with boto2.

cc @kyleknap

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -14,7 +14,6 @@ import re
 import logging
 import datetime
 import hashlib
-import math
 import binascii
 import functools
 
@@ -96,8 +95,8 @@ def remove_dot_segments(url):
 
 
 def validate_jmespath_for_set(expression):
-    # Validates a limited jmespath expression to determine if we can set a value
-    # based on it. Only works with dotted paths.
+    # Validates a limited jmespath expression to determine if we can set a
+    # value based on it. Only works with dotted paths.
     if not expression or expression == '.':
         raise InvalidExpressionError(expression=expression)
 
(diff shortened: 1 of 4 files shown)


tests/unit/test_utils.py
```diff
@@ -550,5 +551,35 @@ def test_dns_style_not_used_for_get_bucket_location(self):
...
+                return x + y
+
+            @instance_cache
+            def sub(self, x, y):
+                return x - y
+
+        adder = DummyClass()
+        self.assertEqual(adder.add(2, 1), 3)
+        # This one's from cache.
+        self.assertEqual(adder.add(2, 1), 3)
```
kyleknap: Can we assert that this value came from the cache?

botocore/utils.py
```diff
@@ -644,3 +645,34 @@ def _is_get_bucket_location_request(request):
...
+    will maintain their own cache.  In order to keep things simple,
+    this decorator requires that you provide an ``_instance_cache``
+    attribute on your instance.
+
+    """
+    func_name = func.__name__
+    @functools.wraps(func)
+    def _cache_guard(self, *args, **kwargs):
+        cache_key = (func_name, args)
+        if kwargs:
```
kyleknap: This if statement does not get hit. We can add a test for kwargs?

kyleknap:
Implementation looks good. Just had some comments to improve the tests.

tests/unit/test_utils.py
jamesls: There's a check at the end of this test that the cache has three items in it.  I can add the check after each assertEqual call to make it more explicit.  The one thing that I'd like to avoid is having to hardcode the cache key construction in this test, as ideally that should be able to change without affecting any unit tests.

jamesls:
Feedback incorporated.  Please take another look.

kyleknap:
Thanks. :ship:

PR #543, opened 2015-05-28
`aws route53 wait` now supports ResourceRecordSetsChanged.

quiver:
This will wait until JMESPath query ChangeInfo.Status returns INSYNC when polling with get-change.
Solves feature request aws/aws-cli#1106

## GetChange Response

http://docs.aws.amazon.com/Route53/latest/APIReference/API_GetChange.html

For valid change id, `ChangeInfo.Status` can return  `PENDING` or `INSYNC`.

For invalid change id, code `NoSuchChange` is returned.

```
<?xml version="1.0"?>
<ErrorResponse xmlns="https://route53.amazonaws.com/doc/2013-04-01/"><Error><Type>Sender</Type><Code>NoSuchChange</Code><Message>Could not find resource with ID: C15RYMEZ8RZJFM</Message></Error><RequestId>a721689b-007b-11e5-9fb7-cfa489e06a9c</RequestId></ErrorResponse>
```
I handled this case as success just like other waiters.

## Simple Usage

```
$ cat mx.json
{
  "Comment": "Adding a new MX record for the zone.",
  "Changes": [
    {
    "Action": "CREATE",
    "ResourceRecordSet": {
        "Name": "foo.com.",
        "Type": "MX",
        "TTL": 300,
        "ResourceRecords": [
            {
                "Value": "10 mail3.foo.com."
            }
        ]
      }
    }
  ]
}
$ aws route53 change-resource-record-sets --hosted-zone-id $ZONE_ID --change-batch file://mx.json
{
    "ChangeInfo": {
        "Status": "PENDING",
        "Comment": "Adding a new MX record for the zone.",
        "SubmittedAt": "2015-05-28T19:52:10.855Z",
        "Id": "/change/C15ON6EN6W25CD"
    }
}
$ aws route53 wait resource-record-sets-changed --id "/change/C15ON6EN6W25CD"
```

### DIFF
--- a/botocore/data/route53/2013-04-01/waiters-2.json
+++ b/botocore/data/route53/2013-04-01/waiters-2.json
@@ -1,0 +1,29 @@
+{
+  "version": 2,
+  "waiters": {
+    "ResourceRecordSetsChanged": {
+      "delay": 30,
+      "maxAttempts": 60,
+      "operation": "GetChange",
+      "acceptors": [
+        {
+          "matcher": "path",
+          "expected": "INSYNC",
+          "argument": "ChangeInfo.Status",
+          "state": "success"
+        },
+        {
+          "matcher": "path",
+          "expected": "PENDING",
+          "argument": "ChangeInfo.Status",
+          "state": "retry"
+        },
+        {
+          "expected": "NoSuchChange",
+          "matcher": "error",
+          "state": "success"
+        }
+      ]
+    }
+  }
+}


jamesls:
Sorry for the delay, looking at this now.  I would suggest two fixes:

* The second retry state isn't needed because the default behavior is to retry.  This is only needed when an exception would normally propogate up and stop the waiter from continuing.
* The last case that has the "error"/"success" doesn't seem correct.  The only waiters that do that are for the negative case, i.e wait until something is not available, in which case getting a "Resource does not exist" error is a success case, since we were waiting for the resource to not exist.

thoughts? cc @kyleknap @mtdowling

mtdowling:
Looks good. Agreed on your suggestions. Removing the last change is probably a good idea. Trying to wait on a change that does not exist should not be a successful wait. Instead, the user should see the corresponding error and know that they provided an invalid `--id`.

PR #548, opened 2015-06-03
Fixed Route53 ListResourceRecordSets API uri

coreymaher:
It looks like the Route 53 ListResourceRecordSets URI is incorrect. This pull request should fix it.

### DIFF
--- a/botocore/data/route53/2013-04-01/service-2.json
+++ b/botocore/data/route53/2013-04-01/service-2.json
@@ -861,7 +861,7 @@
       "name":"ListResourceRecordSets",
       "http":{
         "method":"GET",
-        "requestUri":"/2013-04-01/hostedzone/{Id}/rrset"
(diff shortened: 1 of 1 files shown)


kyleknap:
@coreymaher 

Thanks for the PR, but we will not be able to pull in this PR. Based on the model, the ``"HostedZoneId"`` member has a location ``"uri"`` and a locationName ``"Id"``. This means the value passed to "``HostedZoneId"`` will get mapped to the location in the uri labeled ``{Id}`` thus we cannot change ``{Id}`` to ``{HostedZoneId}``.

I also looked at the PR that you referenced. I am not entirely sure how the maintainer customizes the botocore JSON models, but I believe they will need to add more customization to the model or add logic to their sdk similar to what I explained above.

Let me know if you have any more questions.

coreymaher:
Ah I see. Thanks for taking a look.

PR #550, opened 2015-06-04
Add the ability to customize botocore docs

kyleknap:
I added the following in this PR:

* Two types of events: ``'docs.{section-in-documentation}.{service-name}.{operation-name}.{param-name}' `` that returns a section relevant to a specific parameter in the documentation and ``'docs.{section-in-documentation}.{service-name}.{operation-name}.complete-section' `` that returns a particular section for an entire method. 

* Add a customization that allows us to document auto-populated parameters. I tested this out with Glacier's accountId and checksum parameters. The auto-populated parameters are removed from the example, any required label is removed from them, and a note about the parameter being auto-populated is added to the parameter description.

* Add a functional test section to functionally test the docs for specific services. Pretty much adopted boto3's ci scripts.

I tried to make the section structures as consistent as possible, but there are some inconsistencies in structures due to formatting issues. Any suggestions on how to make these more consistent would be more than welcome.

Also this PR will not pass till this PR is merged: https://github.com/boto/bcdoc/pull/37

cc @jamesls @mtdowling


botocore/docs/utils.py
```diff
@@ -96,6 +96,33 @@ def __new__(cls, name, type_name, documentation, metadata=None,
...
+        """
+        if event_name.startswith('docs.request-params'):
+            if self.name in section.available_sections:
+                section = section.get_section(self.name)
+                if 'is-required' in section.available_sections:
+                    section.delete_section('is-required')
+                description_section = section.get_section(
+                    'param-documentation')
+                description_section.writeln(
+                    'Note this parameter is autopopulated. There is no '
```
jamesls: What about moving this up to a class constant and allowing a custom message to be passed in at instantiation?  I was thinking that for some of the other handlers that will use this, we might want to provide a slightly different doc message.

kyleknap: That's fine with me.

botocore/docs/utils.py
```diff
@@ -135,4 +162,18 @@ def traverse_and_document_shape(documenter, section, shape, history,
                     include=include, exclude=exclude,
                     is_top_level_param=is_top_level_param,
                     is_required=is_required)
+        if is_top_level_param:
```
jamesls: I know that this was extracted out before because of the deep inheritance structure, but given the most recent changes I think you can make a much stronger justification for bringing this into a base class as we're starting to copy/paste the `__init__`  of the various documenter classes.  What if we make a base class that contains only this function for shape traversal and have the other base classes share this.  I also suggest this because it's a bit odd to reach into an instance and call one of its collaborators (event_emitter).

tests/functional/docs/test_glacier.py
```diff
@@ -0,0 +1,22 @@
...
+# language governing permissions and limitations under the License.
+from tests.functional.docs import BaseDocsFunctionalTest
+
+class TestGlacierDocs(BaseDocsFunctionalTest):
+    def test_account_id(self):
+        self.check_autopopulated_param_doc(
+            'glacier', 'abort_multipart_upload', 'accountId')
+
+    def test_checksum(self):
+        self.check_autopopulated_param_doc(
```
jamesls: This is a vague method name that makes it hard to understand what's being tested here.

Is this asserting anything? (maybe it should start with assert_)?  What exactly about the autopopulated param doc are you checking.

Perhaps some combination of renaming the method and kwarg usage can make this more clear what's going on.

botocore/docs/example.py
```diff
@@ -15,6 +15,13 @@
 
 
 class BaseExampleDocumenter(object):
+    EVENT_NAME = ''
+
+    def __init__(self, service, operation, event_emitter):
+        self.service = service
```
jamesls: We've had an informal convention of not using the bare service/operation words and being more explicit about what about the service we're using: `service_name`, `service_model`, etc.  I'd prefer these to just be service_name/operation_name.

jamesls:
Overall I think this looks good.  I just had some minor suggestions.

kyleknap:
I updated the PR based on feedback.

botocore/docs/example.py
```diff
@@ -10,11 +10,18 @@
...
-from botocore.docs.utils import traverse_and_document_shape
+from botocore.docs.shape import ShapeDocumenter
 from botocore.docs.utils import py_default
 
 
-class BaseExampleDocumenter(object):
+class BaseExampleDocumenter(ShapeDocumenter):
+    EVENT_NAME = ''
+
+    def __init__(self, service_name, operation_name, event_emitter):
```
jamesls: Just curious, is there any reason not to move the `__init__` into the base class?  Both this class and the other subclass of `ShapeDocumenter` have the same `__init__`.

botocore/docs/example.py
```diff
@@ -10,11 +10,18 @@
...
+        self.service_name = service_name
+        self.operation_name = operation_name
+        self.event_emitter = event_emitter
```
jamesls: These don't need to be public anymore right?

jamesls:
:shipit: Just had some small suggestions, otherwise looks good.

scripts/ci/install
```diff
@@ -0,0 +1,32 @@
...
+    python_version = '.'.join([str(i) for i in sys.version_info[:2]])
+
+if python_version == '2.6':
+    run('pip install -r requirements26.txt')
+
+run('pip install -r requirements.txt')
+run('pip install coverage')
+if os.path.isdir('dist') and os.listdir('dist'):
+    shutil.rmtree('dist')
+run('python setup.py bdist_wheel')
```
jamesls: I was wondering how this worked without the conditional dependencies update via extras_requires.  But it's because the requirements26.txt duplicates the py26 deps.

Once you check this in, I'll go ahead and update requirements26.txt as I'm working on some .whl support changes.

kyleknap: I am not entirely sure. Did not look into it too closely as I was just copying and pasting from boto3. But I will merge it so you can mess with it.

PR #555, opened 2015-06-10
Sync with requests 2.7.0

jamesls:
Needed so we can pull in a few bug fixes related to packaging.

cc @kyleknap @mtdowling


kyleknap:
:ship:

PR #557, opened 2015-06-11
Adding documentation customizations.

mtdowling:
This commit adds two new customization classes:

1. HideParamFromOperations - hides a parameter from one or more
   operations.
2. AppendParamDocumentation - appends documentation to a parameter.

Documentation customizations have been added for each customization
listed in handlers.py.

@kyleknap @jamesls

### DIFF
--- a/botocore/docs/utils.py
+++ b/botocore/docs/utils.py
@@ -102,11 +102,12 @@ class AutoPopulatedParam(object):
         self.param_description = param_description
         if param_description is None:
             self.param_description = (
-                'Note this parameter is autopopulated. There is no '
-                'need to include in method call.\n')
+                'Please note that this parameter is automatically populated if'
+                'it is not provided. Including this parameter is not '
+                'required\n')
 
     def document_auto_populated_param(self, event_name, section, **kwargs):
-        """This documents the auto populated parameters
+        """Documents auto populated parameters
 
         It will remove any required marks for the parameter, remove the
         parameter from the example, and add a snippet about the parameter
@@ -124,3 +125,49 @@ class AutoPopulatedParam(object):
             section = section.get_section('structure-value')
             if self.name in section.available_sections:
                 section.delete_section(self.name)
+
+
+class HideParamFromOperations(object):
+    """Hides a single parameter from multiple operations.
+
+    This method will remove a parameter from documentation and from
+    examples. This method is typically used for things that are
+    automatically populated because a user would be unable to provide
+    a value (e.g., a checksum of a serialized XML request body)."""
+    def __init__(self, service, parameter_name, operation_names):
+        """
+        :type parameter_name: str
+        :param parameter_name: Name of the parameter to modify.
+
+        :type operation_names: list
+        :param operation_names: Operation names to modify.
+        """
+        self._parameter_name = parameter_name
+        self._event_list = {}
+        # Build up a hash of relevant event names.
+        param_template = 'docs.request-params.%s.%s.complete-section'
+        example_template = 'docs.request-example.%s.%s.complete-section'
+        for name in operation_names:
+            self._event_list[param_template % (service, name)] = 'params'
+            self._event_list[example_template % (service, name)] = 'example'
+
+    def hide_param(self, event_name, section, **kwargs):
+        if event_name in self._event_list:
+            if self._event_list[event_name] == 'example':
+                section = section.get_section('structure-value')
+            if self._parameter_name in section.available_sections:
+                section.delete_section(self._parameter_name)
+
+
+class AppendParamDocumentation(object):
+    """Appends documentation to a specific parameter"""
+    def __init__(self, parameter_name, doc_string):
+        self._parameter_name = parameter_name
+        self._doc_string = doc_string
+
+    def append_documentation(self, event_name, section, **kwargs):
+        if self._parameter_name in section.available_sections:
+            section = section.get_section(self._parameter_name)
+            description_section = section.get_section(
+                'param-documentation')
+            description_section.writeln(self._doc_string)
--- a/botocore/docs/utils.py
+++ b/botocore/docs/utils.py
@@ -134,8 +134,11 @@ class HideParamFromOperations(object):
     examples. This method is typically used for things that are
     automatically populated because a user would be unable to provide
     a value (e.g., a checksum of a serialized XML request body)."""
-    def __init__(self, service, parameter_name, operation_names):
+    def __init__(self, service_name, parameter_name, operation_names):
         """
+        :type service_name: str
+        :param service_name: Name of the service to modify.
+
         :type parameter_name: str
         :param parameter_name: Name of the parameter to modify.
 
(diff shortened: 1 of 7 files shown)


mtdowling:
@jamesls @kyleknap I fixed the Python 3 issue and the tests are now passing.

botocore/docs/utils.py
```diff
@@ -124,3 +125,49 @@ def document_auto_populated_param(self, event_name, section, **kwargs):
...
+
+
+class HideParamFromOperations(object):
+    """Hides a single parameter from multiple operations.
+
+    This method will remove a parameter from documentation and from
+    examples. This method is typically used for things that are
+    automatically populated because a user would be unable to provide
+    a value (e.g., a checksum of a serialized XML request body)."""
+    def __init__(self, service, parameter_name, operation_names):
```
kyleknap: I would prefer to stay consistent and change ``service`` to ``service_name`` in order to help remove ambiguity. Would help to have it included in the doc string for it as well.

botocore/docs/utils.py
```diff
@@ -124,3 +125,49 @@ def document_auto_populated_param(self, event_name, section, **kwargs):
...
+        """
+        :type parameter_name: str
+        :param parameter_name: Name of the parameter to modify.
+
+        :type operation_names: list
+        :param operation_names: Operation names to modify.
+        """
+        self._parameter_name = parameter_name
+        self._event_list = {}
+        # Build up a hash of relevant event names.
```
kyleknap: I do not think you need to build up a hash. It can just be a list of the event names. So something like:
```
self._event_list = []
self._event_list.append('docs.request-params.%s.%s.complete-section' % (service, name))
self._event_list.append('docs.request-example.%s.%s.complete-section' % (service, name))
```

Then you can check using the event_name:
```
if event_name in self._event_list:
    if event_name.startswith('docs.request-example'):
        # Do something
    if event_name.startswith('docs.request-params'):
        # Doc something
```

botocore/docs/utils.py
```diff
@@ -124,3 +125,49 @@ def document_auto_populated_param(self, event_name, section, **kwargs):
...
+        param_template = 'docs.request-params.%s.%s.complete-section'
+        example_template = 'docs.request-example.%s.%s.complete-section'
+        for name in operation_names:
+            self._event_list[param_template % (service, name)] = 'params'
+            self._event_list[example_template % (service, name)] = 'example'
+
+    def hide_param(self, event_name, section, **kwargs):
+        if event_name in self._event_list:
+            if self._event_list[event_name] == 'example':
+                section = section.get_section('structure-value')
```
kyleknap: What does this do? It is just getting the section, but it is not doing anything to the section.

botocore/handlers.py
```diff
@@ -441,8 +447,33 @@ def switch_host_with_param(request, param_name):
...
+     AutoPopulatedParam('PresignedUrl').document_auto_populated_param),
+    ('docs.*.ec2.CopySnapshot.complete-section',
+     AutoPopulatedParam('DestinationRegion').document_auto_populated_param),
+    # S3 SSE documentation modifications
+    ('docs.*.s3.*.complete-section',
+     AutoPopulatedParam('SSECustomerKeyMD5').document_auto_populated_param),
+    # The following S3 operations cannot actually accept a ContentMD5
+    ('docs.*.s3.*.complete-section',
+     HideParamFromOperations(
+         's3', 'ContentMD5',
```
kyleknap: I noticed that there is a similiar customization for SSECustomerKey and SSECustomerKeyMD5. Can we add documentation for that as well? I know the first time I tried using sse it took me a fair amount of time to realize a customization is being applied to the parameters.

botocore/handlers.py
```diff
@@ -441,8 +447,33 @@ def switch_host_with_param(request, param_name):
...
+         ['DeleteObjects', 'PutBucketAcl', 'PutBucketCors',
```
kyleknap: Some of these operations do not have an actual customization applied to them like PutBucketWebsite. Do we need to apply the handler there as well or should the ContentMD5 not be used in general (user should not add it and botocore should not add it)?

kyleknap:
Implementation looks good. Had some small comments and a couple questions.

botocore/docs/utils.py
mtdowling: This changes the section that's being checked to see if the parameter is present. If it's an example, then we need to get the structure-value. Otherwise we need to use the section directly.

botocore/handlers.py
mtdowling: SSECustomerKey and SSECustomerKeyMD5 are handled as an autopopulated parameter. We don't want to hide those as a customer could actually provide a real value for them.

botocore/docs/utils.py
kyleknap: ahhh right. I read the second if statement as an elif.

botocore/handlers.py
mtdowling: This customization removes ContentMD5 from the docs in places in which it's impossible for a user to supply them. These operations have ContentMD5 as a parameter. However, it's not possible for a user to supply a ContentMD5 because they do not supply the body of the request (because we serialize the request).

botocore/handlers.py
kyleknap: Can we document them as autopopulated then?

botocore/handlers.py
kyleknap: Cool. Sounds good.

botocore/docs/utils.py
mtdowling: I used a hash here because we already know exactly what type of event is is without the need for an additional string check. It's also faster (O(1)) to look in a hash for a specific event name. Considering there are a ton of documentation events, I thought this was a good technical reason for the hash with a precomputed event type ("params" vs "example").

Did you have a reason to not use a hash?

botocore/handlers.py
jamesls: I think was Kyle was saying was that these are values that a user should not provide but that botocore _should_ be filling in on behalf of the user.  However, looking at https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L410-L414, there are a few of these operations where this auto populate content md5 should be enabled but actually is not.  I think we should just go ahead and update them as part of this PR, or feel free to send another PR, but I think it's something that needs to get addressed.

mtdowling: Ah gotcha. I can add those to this PR.

botocore/docs/utils.py
kyleknap: I suggested it because the code looks a little cleaner if you use a list. Also you named it as ``event_list`` which implies it is a list. I do not think the list will take long to traverse as its size is limited to 2*(the number of parameters specified in the constructor).

Honestly, I am fine with either. I just suggested it because it took me a little long to understand what ``_event_list`` was representing. Maybe rename the variable to something more specific to what it represents like ``_event_type_dict``?

botocore/handlers.py
mtdowling: The following operations do not have a documented Content-MD5 header:

- PutBucketAcl: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTacl.html
- PutBucketLogging: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTlogging.html
- PutBucketNotification: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGETnotification.html
- PutBucketPolicy: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTpolicy.html
- PutBucketRequestPayment: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTrequestPaymentPUT.html
- PutBucketVersioning: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTVersioningStatus.html
- PutBucketWebsite: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTwebsite.html
- PutObjectAcl: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUTacl.html

These operations can, however, take a Content-MD5 header as described in S3's common request headers documentation: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonRequestHeaders.html.

Thing is, you don't need to add a Content-MD5 check for these operations when using SignatureV4 because the message integrity check is built-in to the sigv4 signature. I think this means that we should either not do anything for these operations (other than hide them in docs) or add a new handler that only adds a Content-MD5 header if the signature version is not sigv4.

Adding a new handler would be a new handler and a separate PR I think.

Thoughts?

kyleknap: Most users use non-sigv4 so I would prefer to add a handler for this for non-sigv4. We can address this in another pull request.

kyleknap:
Thanks for looking into this. Looks good. :ship:

botocore/handlers.py
jamesls: Yeah I agree, we can address this in another pull request later.

Issue #395, opened 2014-12-01
range() integer end argument expected, got str


hltbra:
I found this using aws-cli, and investigating I found out this is a bug in botocore:
* https://github.com/boto/botocore/blob/d4094511f7c5f084133578bb4d5a30befa7ccbe8/botocore/credentials.py#L47
* https://github.com/boto/botocore/blob/42b49b103662048fed22fc9debfbde78e11d5686/botocore/utils.py#L140

This is an easy fix, just apply `int()` to the values that must be integers - `metadata_service_num_attempts` is only one of them.

I just reproduced in my laptop:
```
$ aws ec2 describe-instances

range() integer end argument expected, got str.

$ cat ~/.aws/config
[default]
metadata_service_num_attempts = 10

$ pip freeze | egrep '(aws|boto)'
awscli==1.6.5
botocore==0.76.0
```

anthcourtney:
+1

PR #559, opened 2015-06-15
Allow config variables to provide conversion functions

jamesls:
This is primarily to address #395, where the `metadata_*` variables need to be converted to integers.

To fix this, the session var mappings now have an additional field, which is a conversion function, which if provided, will be called before returning the value in `get_config_variable`.

The main change is provided in https://github.com/boto/botocore/commit/b0ea7bc62550665488a69ca0beeb9a240f249c50.  I've also included a number of commits for code cleanup to the session module.

Additionally, I've made one change to the signature of `get_config_variable` which is to remove the `default` argument.  This simplifies the code logic, and nothing in botocore, the cli, nor boto3 used this functionality.  I'd prefer to remove this until there's a strong case for adding this back.  It also simplifies the code logic.

This will also require a corresponding change to the CLI.

cc @kyleknap @mtdowling

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -20,17 +20,12 @@ import copy
 import logging
 import os
 import platform
-import shlex
-import warnings
-from collections import namedtuple
 
 from botocore import __version__
 import botocore.config
 import botocore.credentials
 import botocore.client
-from botocore.endpoint import EndpointCreator
 from botocore.exceptions import EventNotFound, ConfigNotFound, ProfileNotFound
-from botocore.exceptions import ImminentRemovalWarning
 from botocore import handlers
 from botocore.hooks import HierarchicalEmitter, first_non_none_response
 from botocore.loaders import create_loader
@@ -206,7 +201,7 @@ class Session(object):
                 event_name, handler, register_type = spec
                 if register_type is handlers.REGISTER_FIRST:
                     self._events.register_first(event_name, handler)
-                elif register_first is handlers.REGISTER_LAST:
+                elif register_type is handlers.REGISTER_LAST:
                     self._events.register_last(event_name, handler)
 
     @property
@@ -336,7 +331,6 @@ class Session(object):
 
         """
         self._session_instance_vars[logical_name] = value
-
 
     def get_scoped_config(self):
         """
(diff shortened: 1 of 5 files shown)


botocore/session.py
```diff
@@ -53,7 +49,9 @@ class Session(object):
     :ivar profile: The current profile.
     """
 
-    AllEvents = {
+    #: A dictionary where each key is an event name and the value
+    #: is the formatting string used to construct a new event.
+    ALL_EVENTS = {
```
kyleknap: The event names look out of date as well.

docs/source/index.rst
```diff
@@ -35,6 +35,14 @@ Contents:
 Upgrade Notes
 =============
 
+Upgrading to 1.0.0rc1
```
kyleknap: I forgot that we started this. There may be some things that I forgot to add like the update to the pagination interface, but I am not entirely sure when it was added.

kyleknap:
LGTM. Had a couple comments that were sort of unrelated to the PR, but :ship: otherwise.

docs/source/index.rst
kyleknap: Should we backport them in another PR?

jamesls: Yeah, it might be worth backporting these in another PR.

botocore/session.py
jamesls: Yeah, this is interesting because as we started to refactor code to be less dependent on a session, and instead just take event_emitters where necessary, they now no longer have this list of available events accessible.

The original idea behind this was to a) see what events were available and b) avoid typos in `emit()/register()` calls when referencing event names.

What do you think we should do with these now?

kyleknap: I do not even sure what they are used for as they are inaccurate. If the functionality is not really used anymore, I would be in favor of removing the global dictionary and any methods that use the dictionary.

jamesls: Ok, I'll remove them.

jamesls:
@kyleknap Feedback incorporated, mind taking another look?

kyleknap:
@jamesls 
Looks like you need to rebase. You removed an import that is needed for another test.

kyleknap:
Otherwise, once the test pass. You are good to merge. :ship:

PR #560, opened 2015-06-15
Scope sign request to service name

jamesls:
Even though botocore clients get a fresh copy of their event
emitter, we don't necessarily want to assume that non service
specific (or other service) events can be emitted from this
event emitter.

cc @kyleknap @mtdowling


mtdowling:
LGTM

kyleknap:
:ship:

PR #561, opened 2015-06-16
Fix issue with machinelearning predict

kyleknap:
Fixes the issue brought up in the forums:
https://forums.aws.amazon.com/thread.jspa?threadID=182689&tstart=0 

We have tests for machinelearning in the CLI but that mocks out the ``make_request`` method of the endpoint so the handler for machine learning was never called.

cc @jamesls @mtdowling


jamesls:
:shipit:

PR #563, opened 2015-06-16
Use more random random names

jamesls:
Fixes occasional integ test failures.

Removed all calls to `random.random` in integ tests.

cc @kyleknap @mtdowling


mtdowling:
Nice. :ship:

PR #564, opened 2015-06-16
[WIP] Conditionally add Content-MD5 in S3 when not sigv4

mtdowling:
This PR conditionally adds a Content-MD5 to several S3 operations when the following constraints are met:

1. Not using signature version 4 (because a Content-SHA256 header is already used).
2. When a Content-MD5 is not already present.

I ran into a few issues with this commit because a file is being passed into the "body" parameter of a request dict, but `six.b` requires a buffer or string: `md5.update(six.b(params['body']))`


jamesls:
Superseded by https://github.com/boto/botocore/pull/583

PR #566, opened 2015-06-17
Port bcdoc

kyleknap:
This pull request pulls bcdoc into botocore. I also removed bcdoc as a dependency. I installed botocore in a clean virtualenv and the test pass fine. I will also have subsequent pull requests for Boto3 and the CLI.

cc @jamesls @mtdowling


jamesls:
:shipit:

PR #567, opened 2015-06-17
Change use of example to syntax in docs

kyleknap:
Update headers in methods to be ``Request Syntax`` and ``Response Syntax`` as opposed to ``Request Example`` and ``Response Example``.

cc @jamesls @mtdowling


jamesls:
:shipit:

PR #568, opened 2015-06-17
Fix paragraph indentation

kyleknap:
There was an issue where if text was written directly after a paragraph
it would be misaligned. This specifically fixes issues with the
route53 and route53domains docs.

Here is an example of this misalignment currently:
screen shot 2015-06-17 at 2 22 42 pm

Here is how it looks like with this pr:
screen shot 2015-06-17 at 2 24 08 pm

Notice that the period is still there but it is aligned properly.

cc @jamesls @mtdowling


jamesls:
:shipit:

PR #569, opened 2015-06-17
Add functionality for shapes with no members

kyleknap:
This fixed the formatting issues where an empty method call/structure had
the starting and ending parenthesis/bracket on separate lines.

It currently looks like this:
screen shot 2015-06-17 at 2 38 18 pm

The PR changes it to be like this:
screen shot 2015-06-17 at 2 41 55 pm

cc @jamesls @mtdowling

### DIFF
--- a/botocore/docs/bcdoc/restdoc.py
+++ b/botocore/docs/bcdoc/restdoc.py
@@ -197,3 +197,6 @@ class DocumentStructure(ReSTDocument):
 
     def getvalue(self):
         return ''.join(self._writes).encode('utf-8')
+
+    def clear_text(self):
+        self._writes = []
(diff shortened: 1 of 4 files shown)


jamesls:
:shipit:

Although technically the response object contains a "ResponseMetadata" key that will have things like the request id and HTTPStatusCode.  Do we document that fact anywhere?

kyleknap:
We do not document that. I did not document this at first because documenting implies that it will not change. I can look into adding that in another PR. If that sounds good? I could add it to the reference pages and/or a new handwritten document that talks about it.

jamesls:
Yeah, I don't envision that key changing if that's the concern.  I think it's fine to address this in a separate PR, just want to ensure that we do actually document the response metadata key somewhere in our docs.  I don't really have a strong opinion about where this is documented right now.

kyleknap:
Sounds good. I opened up an issue so we would not forget: https://github.com/boto/botocore/issues/574

PR #570, opened 2015-06-17
Fix custom method documentation

kyleknap:
Also, extracted logic to new method and fixed a doc string in the presigner methods.

The old documentation looked like this:
screen shot 2015-06-17 at 2 53 41 pm

The new documentation looks like this:
screen shot 2015-06-17 at 2 56 41 pm

The main difference being the docstring is fixed on the returns and there is no more paragraph block surrounding the summary of the doc string.

cc @jamesls @mtdowling

### DIFF
--- a/botocore/docs/client.py
+++ b/botocore/docs/client.py
@@ -13,7 +13,7 @@
 import inspect
 
 from botocore.docs.utils import get_official_service_name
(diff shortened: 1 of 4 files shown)


jamesls:
:shipit:

PR #571, opened 2015-06-17
Fix put_object when body is a bytes

jamesls:
This is more general change to handle how we handle blob types in general.

There's 4 possible types that a user could give us for a blob type that would pass validation:

* text type (str/unicode)
* bytes type (bytes())
* binary file obj
* text file obj

This PR updates botocore such that the first three cases are handled.  The fourth case could be something we support down the road, but it's easy enough for now to just open the file as binary.
It's also a somewhat difficult because we can't reliably tell if a file like object is opened as text.

To accomplish this, I picked up the work down a few months ago to standardize on the input types for serialization such that the *request dict body is always `bytes()`*.

cc @kyleknap @mtdowling

### DIFF
--- a/botocore/serialize.py
+++ b/botocore/serialize.py
@@ -302,7 +302,7 @@ class JSONSerializer(Serializer):
         input_shape = operation_model.input_shape
         if input_shape is not None:
             self._serialize(body, parameters, input_shape)
-        serialized['body'] = json.dumps(body).encode('utf-8')
+        serialized['body'] = json.dumps(body).encode(self.DEFAULT_ENCODING)
         return serialized
 
     def _serialize(self, serialized, value, shape, key=None):
@@ -513,7 +513,7 @@ class RestJSONSerializer(BaseRestSerializer, JSONSerializer):
     def _serialize_body_params(self, params, shape):
         serialized_body = self.MAP_TYPE()
         self._serialize(serialized_body, params, shape)
-        return json.dumps(serialized_body).encode('utf-8')
+        return json.dumps(serialized_body).encode(self.DEFAULT_ENCODING)
 
 
 class RestXMLSerializer(BaseRestSerializer):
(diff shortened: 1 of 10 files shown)


botocore/serialize.py
```diff
@@ -453,6 +455,11 @@ def _serialize_payload(self, partitioned, parameters,
             serialized['body'] = self._serialize_body_params(
                 partitioned['body_kwargs'], shape)
 
+    def _encode_payload(self, body):
+        if isinstance(body, six.text_type):
+            return body.encode(self.DEFAULT_ENCODING)
```
kyleknap: Out of curiosity, is the reason you used ``DEFAULT_ENCODING`` is because it is in a base class? It seems that in the more specific serializers this variable (whose value is ``utf-8``) is not used and straight ``utf-8`` is used.

tests/integration/test_s3.py
```diff
@@ -871,5 +872,53 @@ def test_can_set_utf_8_headers(self):
                         Bucket=bucket_name, Key="foo.txt")
 
 
+class TestSupportedPutObjectBodyTypes(BaseS3ClientTest):
+    def setUp(self):
+        super(TestSupportedPutObjectBodyTypes, self).setUp()
+        self.region = 'us-east-1'
+        self.client = self.create_client()
+        self.bucket_name = self.create_bucket()
+        self.tempdir = tempfile.mkdtemp()
```
kyleknap: Can we just move it to the ``BaseS3ClientTest`` class. The tempfile logic has been used twice now.

Also I think you can subclass from ``TestS3BaseWithBucket`` and get the same logic for the setUp that you implement above.

kyleknap: We could also move ``create_client`` method to ``BaseS3ClientTest`` class and update its ``setUp`` method.

kyleknap:
Implementation looks fine. Had a couple questions/comments. I guess one question I have is for the fourth scenario is what are the drawbacks of not supporting it (and instead reading the text file as binary)? :ship: otherwise.

botocore/serialize.py
jamesls: I think we should probably update these to all use `self.DEFAULT_ENCODING`.  At least that way it's configurable if someone needs to modify this.

tests/integration/test_s3.py
jamesls: Sounds good, I'll refactor this module and clean it up a bit.

jamesls:
After thinking about it, I'm not even sure if this is something we'd want to support.  What happens in python3 if you don't open the file in binary would be:

```
<some encoded file> ---> python3 will decode to str() --> http.client sends data using iso-8859-1.
```

This is almost never what you would want and has potential to break.  We could also just say that if we detect you've given us a text file, we'll encode it to utf-8 before sending it over to the http client, but it seems to make more sense to just require the user to be explicit and give us a binary file.  That way they have complete control.

jamesls:

kyleknap:
Thanks. :ship:

PR #573, opened 2015-06-18
Only quote path component of url for copy source

### LINKED ISSUE #509: Unable to pass version to s3:CopyObject
I'm trying to copy a specific version of an object from one s3 bucket to another by adding ?versionId=someversion to the CopySource parameter of s3.copy_object. It looks like the code assumes that the value should be urlencoded, so it mangles the value.

```
        args = {
            'CopySource': '/{}/{}?versionId={}'.format(source_bucket, source_key, version),
            'Bucket': dest_bucket,
            'Key': dest_key,
        }
        session.create_client('s3').copy_object(**args)
```

```
DEBUG:botocore.endpoint:Making request for <botocore.model.OperationModel object at 0x108220a50> (verify_ssl=True) with params: {'query_string': {}, 'headers': {u'x-amz-copy-source': '/source_bucket/source_key%3FversionId%3DabCdeCghIjkLmnOp'}, 'url_path': u'/dest_bucket/dest_key', 'body': '', 'method': u'PUT'}
...
botocore.exceptions.ClientError: An error occurred (NoSuchKey) when calling the CopyObject operation: The specified key does not exist.
```

Is there another way to do this?

jamesls:
Fixes #509 

Fix is in https://github.com/boto/botocore/commit/5e568c5fa53e9bb8ec9b8809e1b53de271485511, there's an additional commit that cleans up some of the flake8 issues.

cc @kyleknap @mtdowling

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -207,8 +207,13 @@ def add_expect_header(model, params, **kwargs):
 def quote_source_header(params, **kwargs):
     if params['headers'] and 'x-amz-copy-source' in params['headers']:
         value = params['headers']['x-amz-copy-source']
-        params['headers']['x-amz-copy-source'] = quote(
-            value.encode('utf-8'), '/~')
+        p = urlsplit(value)
+        # We only want to quote the path.  If the user specified
+        # extra parts, say '?versionId=myversionid' then that part
+        # should not be quoted.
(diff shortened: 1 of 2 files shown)


kyleknap:
LGTM. :ship:

Issue #565, opened 2015-06-17
Add proper class names for clients


jamesls:
We use the service names when creating class names for clients, which have class names that don't match up with the standard CamelCase conventions used for python classes.

Additionally, this causes issues with documentation, as sphinx won't think this is a proper python class.

An example class name:

```
In [1]: s.create_client('cognito-identity')
Out[1]: <botocore.client.cognito-identity at 0x10718ba50>

In [2]: s3 = _

In [3]: client = s.create_client('cognito-identity')

In [4]: client.__class__.__name__
Out[4]: 'cognito-identity'
```

Ideally, the `client.__class__.__name__` should be `CognitoIdentity`.

kyleknap:
The related PR is now merged. Closing.

PR #575, opened 2015-06-18
Rename client class

kyleknap:
This PR updates the client names to be more pythonic (i.e. camel case as opposed to just using the endpoint prefix). It updates the documentation as well. Addresses: https://github.com/boto/botocore/issues/565

You can get the updated name by:
```
>>> import botocore.session
>>> client = session.create_client('cognito-identity', 'us-east-1')
>>> client.__class__.__name__
'CognitoIdentity'
```

This is how a client class name looked before:
screen shot 2015-06-17 at 5 23 58 pm

This is how a client class name looks now:
screen shot 2015-06-17 at 5 23 36 pm

cc @jamesls @mtdowling


botocore/client.py
```diff
@@ -77,6 +79,15 @@ def _load_service_model(self, service_name, api_version=None):
         self._register_retries(service_model)
         return service_model
 
+    def _get_client_class_name(self, service_model):
+        name = service_model.metadata.get(
+            'serviceAbbreviation',
+            service_model.metadata.get('serviceFullName'))
```
jamesls: Should we just fall back to `.service_name` if there's no serviceAbbreviation or serviceFullName?  It's is technically possible for this statement to evaluate to `None`.

kyleknap: Yeah makes sense. I should not be a reason for a hard run-time fail.

jamesls:
:shipit:

PR #576, opened 2015-06-18
Use **kwargs for method signatures

kyleknap:
Ensures if a model driven parameter requires parameters, it is represented as ``**kwargs`` in the method signature. The purpose of doing this is to make the method signatures less cluttered. Note that non-data driven methods still use their method signature.

Here is what it looked before:
screen shot 2015-06-17 at 7 01 00 pm

Here is what it looks after:
screen shot 2015-06-17 at 7 01 08 pm

cc @jamesls @mtdowling


jamesls:
:shipit:

PR #578, opened 2015-06-19
Shorten naming when listing client methods

kyleknap:
This shortens the names when the client methods are listed. Instead of writing out the full name of the method (e.g. showing the class name as well), it just lists the method name.

Here is what it looked before:
screen shot 2015-06-19 at 6 14 31 am

Here is what it looks like now:
screen shot 2015-06-19 at 6 16 48 am

cc @jamesls @mtdowling


jamesls:
:shipit:

PR #579, opened 2015-06-19
Remove loader param from session

jamesls:
This param isn't actually used anymore.

This will require an update to the CLI tests, as there are a few
places where it's still passing a loader to the session.

cc @kyleknap @mtdowling


kyleknap:
:ship:

PR #580, opened 2015-06-19
Allow profile to be provided when session is created

jamesls:
This also makes the profile property read only.

This will require changes in the CLI to account for this, as it currently has this code: https://github.com/aws/aws-cli/blob/develop/awscli/clidriver.py#L221-L222

Same deal with boto3.

The only other thing that might need to be updated is whether or not we should allow `set_config_variable` to support `profile`.  You can technically still do this:

```
s = Session(...)
s.set_config_variable('profile', 'myprofile')
s.create_client(...)
```

thoughts?  cc @kyleknap @mtdowling

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -102,7 +102,7 @@ class Session(object):
     LOG_FORMAT = '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
 
     def __init__(self, session_vars=None, event_hooks=None,
-                 include_builtin_handlers=True, loader=None):
+                 include_builtin_handlers=True, profile=None):
         """
         Create a new Session object.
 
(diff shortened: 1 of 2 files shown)


kyleknap:
Looks good. Does it run into the same issue when the credentials are not reset when you call ``set_config_variable`` on the profile? I think I am fine with the use of it there. :ship: otherwise.

jamesls:
`set_config_variable()` does run into that same issue.  Not really sure what the best solution is other than to completely disallow setting the profile via `set_config_variable()`.  This would have interesting implications in the CLI.  So instead of: https://github.com/aws/aws-cli/pull/1403

I've have to probably do something like:

```
def create_clidriver():
   if '--profile' in sys.argv:  <then get the profile name>
   s = Session(profile=profile_name)
   CLIDriver(session=s)
   ...
```

kyleknap:
Yeah nothing is coming to mind for me either for me. I am not sure if we would want to do that in the CLI. What are some of the other variables you can set  via ``set_config_variable``?

jamesls:
All of the vars in the session var map class variable of the session can be set via set_config_variable.

jamesls:
Merged via 6bd1149

PR #582, opened 2015-06-19
Boto3 doc update

kyleknap:
This pull request is needed for the impending boto3 pull request. It pretty much involves adding a few more methods/attributes to the ``DocumentStructure`` class. The functionality added includes:

* a ``remove_all_sections()`` method that removes all subsections
* a ``clear_text()`` method that clears all of the text for a section, but does not erase the text in subsections.
* a ``meta`` attribute that allows you to store extra data about a section.
* a way to document python attributes in bcdoc.

There is also some logic in recording the shape that represents a section via the new ``meta`` attribute. This is useful for documenting the dynamodb customizations in boto3.

I tried to extract out the commits the best I can to give you a better sense of every change that is made. Might be better to review each commit one at a time.

cc @jamesls @mtdowling

### DIFF
--- a/botocore/docs/bcdoc/restdoc.py
+++ b/botocore/docs/bcdoc/restdoc.py
@@ -103,7 +103,7 @@ class ReSTDocument(object):
 
 
 class DocumentStructure(ReSTDocument):
-    def __init__(self, name, section_names=None, target='man', meta=None):
+    def __init__(self, name, section_names=None, target='man', context=None):
         """Provides a Hierarichial structure to a ReSTDocument
 
         You can write to it similiar to as you can to a ReSTDocument but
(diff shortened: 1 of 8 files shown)


botocore/docs/bcdoc/restdoc.py
```diff
@@ -103,7 +103,7 @@ def remove_last_doc_string(self):
 
 
 class DocumentStructure(ReSTDocument):
-    def __init__(self, name, section_names=None, target='man'):
+    def __init__(self, name, section_names=None, target='man', meta=None):
```
jamesls: Given botocore/boto3 already has this concept of `meta` which is an actual object, it would be less confusing is this was maybe called `metadata` or `context` or something besides `meta`.

jamesls:
:shipit: Looks good.

botocore/docs/bcdoc/restdoc.py
kyleknap: sounds good

PR #583, opened 2015-06-19
Conditionally add Content-MD5 in S3 when not sigv4

jamesls:
This fleshes out the rest of https://github.com/boto/botocore/pull/564 and adds support for calculating content md5 headers when the body is a file like object.

cc @kyleknap @mtdowling

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -116,6 +116,13 @@ def calculate_md5(params, **kwargs):
         params['headers']['Content-MD5'] = value
 
 
+def conditionally_calculate_md5(params, **kwargs):
+    """Only add a Content-MD5 when not using sigv4"""
+    signer = kwargs['request_signer']
+    if signer.signature_version != 'v4':
+        calculate_md5(params, **kwargs)
+
+
 def sse_md5(params, **kwargs):
     """
     S3 server-side encryption requires the encryption key to be sent to the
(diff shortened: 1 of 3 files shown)


jamesls:
Tests fixed.  I'll also need to send a corresponding PR to the CLI to remove the etag checking in the s3 customizations.

botocore/handlers.py
```diff
@@ -110,10 +110,34 @@ def json_decode_template_body(parsed, **kwargs):
...
-        md5 = hashlib.md5()
-        md5.update(six.b(params['body']))
-        value = base64.b64encode(md5.digest()).decode('utf-8')
-        params['headers']['Content-MD5'] = value
+        body = request_dict['body']
+        if isinstance(body, bytes):
+            binary_md5 = _calculate_md5_from_bytes(body)
+        else:
+            binary_md5 = _calculate_md5_from_file(body)
+        base64_md5 = base64.b64encode(binary_md5).decode('ascii')
```
kyleknap: why ascii now? It looks like it used to be decoded with utf-8. I do not think it will make a difference though since it is just an md5.

botocore/handlers.py
```diff
@@ -423,6 +447,17 @@ def switch_host_with_param(request, param_name):
     ('before-call.s3.PutBucketCors', calculate_md5),
     ('before-call.s3.DeleteObjects', calculate_md5),
     ('before-call.s3.PutBucketReplication', calculate_md5),
```
kyleknap: Should the ``calculate_md5`` be changed to ``conditionall_calculate_md5`` now?

tests/unit/test_handlers.py
```diff
@@ -443,6 +443,49 @@ def test_switch_host_with_param(self):
...
+    def test_adds_md5_with_file_like_body(self):
+        request_dict = {
+            'body': six.BytesIO(b'foobar'),
+            'headers': {}
+        }
+        handlers.calculate_md5(request_dict)
+        self.assertEqual(request_dict['headers']['Content-MD5'],
+                         'OFj2IjCsPJFfMAxmQxLGPw==')
+
+    def test_adds_md6_with_bytes_object(self):
```
kyleknap: change ``md6`` to ``md5``

kyleknap:
Looks good. Had a couple comments. Just checking so if there is a mismatch in the SHA256 for sigv4 and MD5 for non sigv4, it will be retried right? Also I would suggest running the s3 integration test once you are done with the CLI PR to make sure everything still passes. Otherwise :ship:

botocore/handlers.py
jamesls: No, I believe these are all correct.  For example, PutBucketCORS documents Content-md5 as being a required parameter: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTcors.html

jamesls:
PR updated to retry content md5 failures.  I actually need to pull this in to also fix the failing CLI travis jobs so once I verify the CLI integ tests I'll go ahead and merge.

botocore/handlers.py
jamesls: Yeah no other reason other than base64 is always ascii chars.

jamesls:
Merged via 4843789

PR #584, opened 2015-06-19
Fall back to get_config_variable() in profile property

jamesls:
So we can continue to keep boto3 code simple such as: https://github.com/boto/boto3/blob/develop/boto3/session.py#L84-L89

cc @kyleknap @mtdowling


kyleknap:
:ship:

Issue #586, opened 2015-06-23
botocore 1.0.0 reports being in beta


vlcinsky:
On [PyPi botocore packege page](https://pypi.python.org/pypi/botocore) I see version 1.0.0 but the classifiers do report `Development Status :: 4 - Beta`

[Botocore Readthedocs](http://botocore.readthedocs.org/en/latest/tutorial/index.html) it claims:

> Botocore is currently under a developer preview, and its API is subject to change prior to a GA (1.0) release. Until botocore reaches a 1.0 release, backwards compatibility is not guaranteed.

I guess, `botocore` now got out of beta and shall be considered having stable API. If that is true, it would be nice to change the classifier on PyPi to report real development status.

jamesls:
Thanks for reporting.  We'll get that updated.  Looks like we missed removing the dev preview disclaimer from the tutorial doc.

jamesls:
Fixed in #587

PR #587, opened 2015-06-23
Remove all beta/dev preview warnings

jamesls:
There's a few additional things we need to change:

* Change trove classifiers to stable
* Remove dev preview warning from tutorial.

The tutorial/getting started guide could use some sprucing up, but we can address that in a separate PR.

cc @kyleknap @mtdowling

### DIFF
--- a/docs/source/tutorial/index.rst
+++ b/docs/source/tutorial/index.rst
@@ -3,16 +3,6 @@ Getting Started With botocore
 *****************************
 
 
-.. warning::
-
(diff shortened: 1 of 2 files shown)


jamesls:
Closes https://github.com/boto/botocore/issues/586

kyleknap:
:ship: We should also fix the setup.py for boto3 as well:
https://github.com/boto/boto3/blob/develop/setup.py#L55

jamesls:
Yeah I'll send a PR for boto3.

PR #589, opened 2015-06-24
Use module names for service titles

kyleknap:
This makes the same title changes as the boto3 doc update: https://github.com/boto/boto3/pull/149

cc @jamesls @mtdowling


mtdowling:
:ship:

AMeng:
:+1: Thanks!

PR #590, opened 2015-06-24
Making UserData docs more clear about base64

### LINKED ISSUE #585: UserData docs for EC2 should be updated
The docs currently say:

image

We should clarify this.  We *always* encode the user data to base64 if it's provided.  I think we should remove the append doc handler and just completely replace the docs with a doc string that says that this is a string and that you don't need to base64 encode the value.

mtdowling:
This commit updates the UserData docs to be more clear that you should not base64 encode the value when providing the parameter. The text is now bold to make it more pronounced.

Also added a minor typo fix for the default AutoPopulatedParam description.

Closes #585 

@jamesls @kyleknap

### DIFF
--- a/botocore/docs/utils.py
+++ b/botocore/docs/utils.py
@@ -102,8 +102,8 @@ class AutoPopulatedParam(object):
         self.param_description = param_description
         if param_description is None:
             self.param_description = (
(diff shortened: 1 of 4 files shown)


jamesls:
:shipit:

PR #595, opened 2015-06-29
Updating theme version

mtdowling:
This commit updates to the latest version of the Guzzle sphinx theme.

Theme changes:

1. Removed some kerning rules that were not noticeable but caused a significant rendering penalty. This change reduces the time to finish rendering the EC2 page from about 4.6 seconds to about 1.7 seconds.
2. Lightened the left nav border.
3. Added a border between each method and class to better visually distinguish them.

This commit is only really necessary to get readthedocs to update their local package cache.

@kyleknap


kyleknap:
:ship:

PR #596, opened 2015-07-01
Making the use of session sound less optional

mtdowling:
The current wording of getting started with botocore makes it seem like there are other ways that developers could use when interacting with botocore. While it's possible to use botocore by creating clients manually without a session, I think we should update the documentation to be more clear that using a session is needed for almost every use case, and using a session is the public API of botocore.


kyleknap:
:ship:

PR #600, opened 2015-07-07
Add ability to set location of credentials file with AWS_CREDENTIALS_…

freimer:
…FILE environment variable

Sorry, I don't know how to add to the unit tests...

Signed-off-by: Fred Reimer <Fred.Reimer@jmfamily.com>

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -82,7 +82,7 @@ class Session(object):
         # These variables are intended for internal use so don't have any
         # user settable values.
         # This is the shared credentials file amongst sdks.
-        'credentials_file': (None, None, '~/.aws/credentials', None),
+        'credentials_file': (None, 'AWS_CREDENTIALS_FILE', '~/.aws/credentials', None),
 
         # These variables only exist in the config file.
 


jamesls:
I agree we should add an env var for this value.

However, that env var name is already used to support pre-existing CLI tools (https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L347) which is a different file with a different format.  We'll need to choose another env name.

The java sdk uses `AWS_CREDENTIAL_PROFILES_FILE` (https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/ProfilesConfigFile.java#L73), so perhaps we can just use that env var to be consistent.

jamesls:
Sorry for the delay, I do like `AWS_CREDENTIALS_FILE` but as I mentioned it's already taken.  I discussed this with the team and we'll just go with `AWS_SHARED_CREDENTIALS_FILE`.

jamesls:
Merged in your pull request and renamed the env var to `AWS_SHARED_CREDENTIALS_FILE`.  Tracking here: https://github.com/boto/botocore/pull/623

PR #608, opened 2015-07-20
Adding a StringShape and enum attribute

mtdowling:
This commit updates the shape builders to utilize a new concrete shape
type, StringShape. StringShape exposes a property that makes it easier
to inspect the enum attribute of a string shape. StringShape is the only
concrete shape type to expose this property because, at present, the only
shapes that have the `enum` property are string shapes.

In addition to this feature, the unused `SHAPE_CLASSES` constant has been
deleted from ServiceModel. If removing this constant is not acceptable, I can
put it back and rebase.

### DIFF
--- a/botocore/model.py
+++ b/botocore/model.py
@@ -191,18 +191,18 @@ class MapShape(Shape):
         return self._resolve_shape_ref(self._shape_model['value'])
 
 
+class StringShape(Shape):
+    @CachedProperty
+    def enum(self):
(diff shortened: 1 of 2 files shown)


jamesls:
:shipit:  Looks good to me.  I think it's safe to remove the `SHAPE_CLASSES` class attr.  I think that's just an artifact from a previous refactoring where that attr wasn't removed from the old location after being moved to a different class.

PR #609, opened 2015-07-21
Use env vars to set metadata service config

laurrentt:
I think there's no reason to limit the users from setting `metadata_service_timeout` and `metadata_service_num_attempts` config from  environment variables.


spg:
10/10 would merge

PR #610, opened 2015-07-25
Address issue https://github.com/boto/boto3/issues/186

rayluo:
Since we don't have control to field names in upstream services, let's play safe by not assuming any field's existence.


jamesls:
I believe this can be closed now.

PR #612, opened 2015-07-29
add docs to sdist

alunduil:
When packaging for distributions we sometimes want to include
documentation in the installation (especially on Gentoo).  This adds the
documentation to the sdist so we can use the PyPi distribution and build
the documentation for our users.


jamesls:
Seems relatively harmless to include, not opposed, especially if it helps package maintainers.  Boto2 also includes docs in the sdist as well (https://github.com/boto/boto/blob/develop/MANIFEST.in#L12).  Any objections?  cc @kyleknap @mtdowling @rayluo

PR #615, opened 2015-07-30
RestJSONParser now parses "code" from "__type" field

rayluo:
According to issue boto/boto3#186, this happens in cloudsearchdomain service and some other services too. This fix is generic for all RestJSON responses.

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -487,21 +487,6 @@ class BaseJSONParser(ResponseParser):
     def _handle_timestamp(self, shape, value):
         return self._timestamp_parser(value)
 
-
-class JSONParser(BaseJSONParser):
-    """Response parse for the "json" protocol."""
-    def _do_parse(self, response, shape):
-        # The json.loads() gives us the primitive JSON types,
-        # but we need to traverse the parsed JSON data to convert
-        # to richer types (blobs, timestamps, etc.
-        parsed = {}
-        if shape is not None:
-            body = response['body'].decode(self.DEFAULT_ENCODING)
-            original_parsed = json.loads(body)
-            parsed = self._parse_shape(shape, original_parsed)
-        self._inject_response_metadata(parsed, response['headers'])
-        return parsed
-
     def _do_error_parse(self, response, shape):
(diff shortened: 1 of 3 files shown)


botocore/parsers.py
```diff
@@ -632,7 +632,7 @@ def _initial_body_parse(self, body_contents):
 
     def _do_error_parse(self, response, shape):
         body = self._initial_body_parse(response['body'])
-        error = {'Error': {}, 'ResponseMetadata': {}}
+        error = BaseJSONParser._do_error_parse(self, response, shape)
```
jamesls: Was there a reason not to use `super()`?

tests/unit/response_parsing/json/expected/elastictranscoder-delete-pipeline.json
```diff
@@ -1,5 +1,5 @@
 {
-    "ResponseMetadata": {},
+    "ResponseMetadata": {"RequestId": "1234"},
```
jamesls: Why was this added?

jamesls:
Just had a couple questions, but overall looks good.

botocore/parsers.py
rayluo: I should use super(). Will update the PR soon.

tests/unit/response_parsing/json/expected/elastictranscoder-delete-pipeline.json
rayluo: As a (harmless or even desirable?) side effect, RestJSONParser now inherits the parsing RequestId ability from the adjusted JSONParser.

I actually mentioned it in the commit log as below.

    Besides, RestJSONParser will now grab x-amzn-requestid header
    and populate the ResponseMetadata.
    An old JSON response test case is updated accordingly.

jamesls:
Looks good :shipit:.  Please rebase and edit your commit message before merging.  Here's some guidelines that we follow: http://git-scm.com/docs/git-commit#_discussion, you can also look through the commit message for the CLI repo to get a feel for how the commit message are formatted.

PR #617, opened 2015-07-31
Generate shared examples for API documentation

DanielNi:
Adds functionality for generating shared documentation examples from the language-agnostic JSON format. These new examples will live alongside the JSON model files.

The code in this PR creates a new loader for loading examples and passes them through to the session, service and client until they are documented in the document_model_driven_method function. From there, the SharedExampleBuilder class parses the dict and creates Python examples at the end of each operation in the client documentation.

Unit tests to be added in a separate PR.

### DIFF
--- a/botocore/docs/method.py
+++ b/botocore/docs/method.py
@@ -16,6 +16,7 @@ from botocore.docs.params import RequestParamsDocumenter
 from botocore.docs.params import ResponseParamsDocumenter
 from botocore.docs.example import ResponseExampleDocumenter
 from botocore.docs.example import RequestExampleDocumenter
+from botocore.docs.sharedexample import SharedExampleBuilder
 
 
 def get_instance_public_methods(instance):
@@ -121,10 +122,10 @@ def document_custom_method(section, method_name, method):
 
 
 def document_model_driven_method(section, method_name, operation_model,
-                                 event_emitter, method_description=None,
-                                 example_prefix=None, include_input=None,
-                                 include_output=None, exclude_input=None,
-                                 exclude_output=None,
+                                 event_emitter, method_description=None, 
+                                 example_prefix=None, shared_examples=None,
+                                 include_input=None, include_output=None, 
+                                 exclude_input=None, exclude_output=None,
                                  document_output=True):
     """Documents an individual method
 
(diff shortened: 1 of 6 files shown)


botocore/docs/method.py
```diff
@@ -231,3 +234,45 @@ def document_model_driven_method(section, method_name, operation_model,
                 include=include_output, exclude=exclude_output)
     else:
         return_section.write(':returns: None')
+
+    # Add the shared examples
```
jamesls: Let's extract this to a separate method.

botocore/docs/sharedexample.py
```diff
@@ -0,0 +1,82 @@
...
+    def example_code(self, prefix=''):
+        if self._is_input:
+            return prefix + self._visit(self._params, '', [], is_param=True)
+        else:
+            return self._visit(self._params, '', [], is_param=False)
+
+    def _visit(self, value, indent, path, is_param=False):
+        if is_param:
+            return self._visit_param(value, indent, path)
+        if isinstance(value, dict):
```
jamesls: I'd suggest using `getattr` to implement visitors.

botocore/docs/sharedexample.py
```diff
@@ -0,0 +1,82 @@
...
+            shape_val = self._visit(val, '    ' + indent, path)
+            lines.append("%s    '%s': %s, %s" %
+                (indent, key, shape_val, comment))
+            path.pop()
+        lines.append('%s}' % indent)
+        return '\n'.join(lines)
+
+    def _visit_list(self, value, indent, path):
+        lines = ['[']
+        for ind, val in enumerate(value):
```
jamesls: full variable names: `index`

botocore/loaders.py
```diff
@@ -363,6 +364,35 @@ def load_data(self, name):
         # We didn't find anything that matched on any path.
         raise DataNotFoundError(data_path=name)
 
+    @instance_cache
+    def load_examples(self, service_name, api_version=None):
```
jamesls: Not a fan of adding new methods to the loader interface.  Note how there's no methods for paginators/waiters.  I'd prefer to have a separate object that takes a loader and uses `load_service_model`.  This keeps the loader interface small.

jamesls:
Thanks for the pull request.  I think this will be really useful.  I've made some suggestions and as you mentioned we need tests before this can be merged (as per our contributing guide https://github.com/boto/botocore/blob/develop/CONTRIBUTING.rst).

I'm going to go ahead and close this out for now, but we'll definitely pick this work up again.

PR #620, opened 2015-08-06
Serialize Uri for no input shape

kyleknap:
There was an issue where if the operation had no input shape, we would not use the modeled URI and just use ``'/'`` because that is the default.

### DIFF
--- a/tests/unit/protocols/input/rest-xml.json
+++ b/tests/unit/protocols/input/rest-xml.json
@@ -73,16 +73,16 @@
       {
         "given": {
           "http": {
-            "method": "PUT",
+            "method": "GET",
             "requestUri": "/2014-01-01/hostedzone"
           },
           "name": "OperationName"
         },
         "params": {},
         "serialized": {
-          "method": "PUT",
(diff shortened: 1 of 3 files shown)


tests/unit/protocols/input/rest-xml.json
```diff
@@ -73,16 +73,16 @@
       {
         "given": {
           "http": {
-            "method": "PUT",
+            "method": "GET",
```
jamesls: Why change this instead of adding a new test?  If the PUT case no longer valid if you send an empty body?

kyleknap: @jamesls 
I was the one to originally added this test case. I added it to help improve coverage on our serializers. The PUT method was just used because of me just copying and pasting from previous examples. The test case was to cover if there was no input shape for the operation and so I believe the expected value is incorrect so it needed changing. I also cannot think of an example of a ``PUT`` request with no input parameters.

jamesls:
:shipit:  Looks good.

PR #621, opened 2015-08-07
Add content length for seekable files

kyleknap:
For some cases like an extracted file from a tar file were not able to
be uploaded because requests did not calculate the content-length despite
the fact that it was seekable. Added logic to calculate and add the content
length header for these types of files.

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -419,11 +419,6 @@ class AWSPreparedRequest(models.PreparedRequest):
         """Prepares the given HTTP body data."""
         super(AWSPreparedRequest, self).prepare_body(data, files, json)
 
-        # Transfer-Encoding is not supported for AWS Services so remove it
-        # if it is added.
-        if 'Transfer-Encoding' in self.headers:
-            self.headers.pop('Transfer-Encoding')
-
         # Calculate the Content-Length by trying to seek the file as
         # requests cannot determine content length for some seekable file-like
         # objects.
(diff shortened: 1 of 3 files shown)


botocore/awsrequest.py
```diff
@@ -414,6 +415,25 @@ def reset_stream(self):
...
+        # Transfer-Encoding is not supported for AWS Services so remove it
+        # if it is added.
+        if 'Transfer-Encoding' in self.headers:
+            self.headers.pop('Transfer-Encoding')
+
+        # Calculate the Content-Length by trying to seek the file as
+        # requests cannot determine content length for some seekable file-like
+        # objects.
+        if 'Content-Length' not in self.headers:
+            if hasattr(data, 'seek') and hasattr(data, 'tell'):
```
jamesls: Shouldn't we have an error if the data isn't seekable/tellable?  Otherwise we popped the transfer encoding header and didn't replace it with anything.

jamesls:
Just had a quick question about the error cases.

botocore/awsrequest.py
kyleknap: I did not add the error because I assumed that the seekable/tellable part would got caught a lot in earlier in the process of sending a request (like in the signing part for sigv4 or Content-MD5 header addition for s3). I think what I am going to do is pop off the Transfer-Encoding if I am about to add the Content-Length header.

kyleknap:
@jamesls I updated based on feedback. I decided to remove the ``'Transfer-Encoding'`` header if *we* add the ``'Content-Length'`` header ourselves. This help makes the logic have less side effects.

tests/unit/test_awsrequest.py
```diff
@@ -72,6 +72,31 @@ def seek(self, offset, whence):
...
+    file-like objects such as StringIO's getvalue() and file object's fileno
+    property. If the file-like object does not have either of these attributes
+    requests will not calculate the content length even though it is still
+    possible to calculate it.
+    """
+    def __init__(self, stream):
+        self._stream = stream
+
+    def __iter__(self):
+        yield self._stream.__iter__()
```
jamesls: Generally see this written as `yield iter(self._stream)`.

jamesls: Actually on second thought, this doesn't look right.  Wouldn't you either want to just return the `__iter__()` of the proxied object instead or yield the values from the proxy objects as you iterate over `self._stream`?

Not a huge deal since this doesn't appear to be used by the tests directly.

jamesls:
:shipit:  Small comment on the `__iter__` method in a test utility class.

tests/unit/test_awsrequest.py
kyleknap: Yeah that's right I need to return the iterator. Was not paying too closely to it. I only need this to make the test object look like an iterable.

kyleknap:
I updated the utility test class. Merging.

GaretJax:
Thanks a lot @kyleknap!

Issue #599, opened 2015-07-07
Add AWS_CREDENTIALS_FILE for alternate credentials file location


freimer:
The credentials file is hard-coded for ~/.aws/credentials with no environment variable to specify an alternate location.  For many organizations, keeping credentials on an unencrypted volume is not allowed.  Simple one-line change to be in a forthcoming pull request to add this capability.

jamesls:
Linking pull request: https://github.com/boto/botocore/pull/600

jamesls:
Closing out issue.  This was fixed in #623

PR #623, opened 2015-08-11
Support AWS_SHARED_CREDENTIALS_FILE

jamesls:
This pulls in https://github.com/boto/botocore/pull/600, and renames the env var to `AWS_SHARED_CREDENTIALS_FILE`.  It also adds an integration test to verify behavior.

Note that there's no actual new code.  It's just changing a config value from None to `AWS_SHARED_CREDENTIALS_FILE`, and that mechanism for the env var->config file->default lookup is already tested.  So it seems reasonable to add an integration test to verify that this specific value was being picked up.

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -82,7 +82,7 @@ class Session(object):
         # These variables are intended for internal use so don't have any
         # user settable values.
         # This is the shared credentials file amongst sdks.
-        'credentials_file': (None, None, '~/.aws/credentials', None),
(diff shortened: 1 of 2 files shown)


jamesls:
Linking original issue: https://github.com/boto/botocore/issues/599

kyleknap:
Looks good. :ship: It would also be nice to update any related documentation. I know there is some documentation in the CLI that will need to be updated like in the README and possibly a topic guide.

jamesls:
Yeah I checked the botocore docs and didn't see anything.  I'll send a PR updating the CLI docs shortly.

Issue #613, opened 2015-07-29
Paginators are needed for logstream items


eldondevcg:
I think the following items in the cloudwatch "logs" resource need paginators:
describe_log_groups
describe_log_streams
describe_metrics_filters
describe_subscription_filters
filter_log_events
get_log_events

kyleknap:
Marking as feature request.

mjschultz:
I have a [PR](https://github.com/boto/botocore/pull/627) that adds paginators for *most* of those except `get_log_events`.

`get_log_events` is a bit weird because it returns both a `nextForwardToken` and a `nextBackwardToken` and the one you'd want to use depends on if you set `startFromHead` or not. The token itself is also different from other APIs in that the `next*Token` will be a repeat on the last page (and the `events` list will be empty). So I'm not sure of the best way to handle that.

All that said, I'd rather have the other paginations in with that one missing then follow up with the `get_log_events` down the road.

jamesls:
Closing out issue, #627 has been merged.  We'll need to address get_log_events later on down the road.

PR #627, opened 2015-08-12
Add paginators for CloudWatch Logs

mjschultz:
This adds some paginators requested in https://github.com/boto/botocore/issues/613. I've verified them to the best of my ability with the services we use, but we don't have any subscription filters.

GetLogEvents is also strange because, it seems, that on the "last" page the `nextForwardToken` is intentially a self reference which causes botocore to raise `botocore.exceptions.PaginationError: Error during pagination: The same next token was received twice: [u'f/...']` exception.

Let me know what you think!

### DIFF
--- a/botocore/data/logs/2014-03-28/paginators-1.json
+++ b/botocore/data/logs/2014-03-28/paginators-1.json
@@ -29,12 +29,6 @@
       "output_token": "nextToken",
       "limit_key": "Limit",
       "result_key": "Events"
-    },
-    "GetLogEvents": {
-      "input_token": "nextToken",
-      "output_token": "nextForwardToken",
-      "limit_key": "Limit",
-      "result_key": "Events"
     }
   }
 }
--- a/botocore/data/logs/2014-03-28/paginators-1.json
+++ b/botocore/data/logs/2014-03-28/paginators-1.json
@@ -3,32 +3,32 @@
     "DescribeLogGroups": {
       "input_token": "nextToken",
       "output_token": "nextToken",
-      "limit_key": "Limit",
-      "result_key": "LogGroups"
+      "limit_key": "limit",
+      "result_key": "logGroups"
     },
     "DescribeLogStreams": {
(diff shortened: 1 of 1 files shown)


mjschultz:
I elected to remove the `GetLogEvents` paginator for now since it's behavior is strange compared to most other APIs.

jamesls:
Thanks.  Do you know if FilterLogEvents has the same issue as GetLogEvents?

mjschultz:
It does not appear to have the same issue. The [AWS docs](http://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_FilterLogEvents.html) don't mention two tokens or have a `startFromHead` parameter.

eldondevcg:
:+1: I have not tested this change myself, but it looks good. I didn't see verbose information about the string pattern in FilterLogEvents. Is it an exact string match? Does a 0-length pattern match all log lines?

mjschultz:
@eldondevcg It looks like the `filterPattern` parameter takes inputs as documented in the [Filter and Pattern Syntax](http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/FilterAndPatternSyntax.html) document.

Not providing the parameter or having it `filterPattern=''` both return the same number of results (for me).

I'm using this for VPC Flow Logs so if I provide a `filterPattern` like `'[version, account, eni, source, destination, srcip, destip, protocol="6", packets, bytes, windowstart, windowend, action="ACCEPT", flowlogstatus]'` it returns only the results where the protocol (8th space delimited field) is 6 and the action (13th field) is ACCEPT.

jamesls:
Thanks!

jstewmon:
I am getting `PaginationError` `Error during pagination: The same next token was received twice: ` with a `filter_log_events` paginator using boto3==1.2.3 and botocore==1.3.18, so it looks like the previously dismissed concern is actually a problem.

PR #628, opened 2015-08-13
(Updated) Append signature to HmacV1Query querystring instead of replace

jamesls:
This pulls in https://github.com/boto/botocore/pull/597 and adds a few unit tests and an integration test.


kyleknap:
LGTM. :ship:

PR #632, opened 2015-08-20
Support client params from a single positional arg

jamesls:
This is an alternative to those that don't prefer to use
**kwargs for client calls.

I'd like to get some feedback on this before merging.  We've had some concern about using ``**kwargs``.   We'd also heard that using the `CamelCasing` in our ``**kwargs`` annoys tools such as pylint.  This adds an option to make calls via:

```
client = session.create_client('dynamodb')
client.describe_table({"TableName": "foo"})
```

thoughts?  cc @kyleknap @mtdowling @rayluo


kyleknap:
Looks good. One thing that would be nice to check is that boto3, still works as expected. Maybe run the tests to make sure nothing is unexpectedly broken with the the resource interface. :ship:

jamesls:
Ran boto3 tests against this change, they passed.

kyleknap:
Cool. Good with me to merge then.

rayluo:
Nice! Ship it! :shipit:

wt:
FWIW, the following should already work without this change:

    client = session.create_client('dynamodb')
    client.describe_table(**{"TableName": "foo"})

Why not just rely on that?

wt:
There seems to be an agreement with the original author of this PR (@jamesis) that it should not be merged (see #230). Perhaps this PR should be closed?

jamesls:
Yeah, I think fi we don't get feedback additional feedback by tomorrow, we can go ahead and pass on this pull request for the time being.

jamesls:
Closing due to lack of interest (https://github.com/boto/boto3/issues/230).  We can look at this again in the future if there's any future interest.

PR #633, opened 2015-08-20
Dynamically add docstrings to client methods

kyleknap:
So now if you are in the python repl you can access a client's method docstrings with ``__doc__`` and ``help()``. These docstrings are lazily loaded so they do not actually get generated until they are accessed, as it is really slow to generate all of docstrings for a client upfront. So they should not affect the speed to generate clients or make method calls. I did some performance testing to make sure. Also note that these automatically get picked up by boto3 clients which is nice. Here is the script I used:

```py
import boto3
import time

for i in range(5):
    before = time.time()
    boto3.client('ec2')
    after = time.time()
    delta = after - before
    print('For ec2 client %s, it took %s seconds to create' % (i, delta))

print('')
client = boto3.client('ec2')
for i in range(5):
    before = time.time()
    response = client.describe_regions()
    after = time.time()
    delta = after - before
    print('For ec2 client %s, it took %s seconds to DescribeRegions' % (
        i, delta))
```
Here is the current speed on the develop branch:
```
For ec2 client 0, it took 0.10999083519 seconds to create
For ec2 client 1, it took 0.00277781486511 seconds to create
For ec2 client 2, it took 0.00267887115479 seconds to create
For ec2 client 3, it took 0.00264096260071 seconds to create
For ec2 client 4, it took 0.00269412994385 seconds to create

For ec2 client 0, it took 0.341856002808 seconds to DescribeRegions
For ec2 client 1, it took 0.115956783295 seconds to DescribeRegions
For ec2 client 2, it took 0.106163024902 seconds to DescribeRegions
For ec2 client 3, it took 0.0971441268921 seconds to DescribeRegions
For ec2 client 4, it took 0.094202041626 seconds to DescribeRegions
```

For this pull request, here is the output:
```
For ec2 client 0, it took 0.110929012299 seconds to create
For ec2 client 1, it took 0.00621104240417 seconds to create
For ec2 client 2, it took 0.00612497329712 seconds to create
For ec2 client 3, it took 0.00794386863708 seconds to create
For ec2 client 4, it took 0.00695705413818 seconds to create

For ec2 client 0, it took 0.333263874054 seconds to DescribeRegions
For ec2 client 1, it took 0.112825870514 seconds to DescribeRegions
For ec2 client 2, it took 0.113487005234 seconds to DescribeRegions
For ec2 client 3, it took 0.0971641540527 seconds to DescribeRegions
For ec2 client 4, it took 0.0956180095673 seconds to DescribeRegions
```


jamesls:
While the change looks good overall, I'm surprised that the 2-N times for creating a client is 3x slower.  I think we'll have to address this in the future.

Otherwise, :shipit:

kyleknap:
Yeah that is something that we can look into. I am not too sure why it happened. Maybe some more iterations can be done to see if it is consistent.

PR #634, opened 2015-08-25
Feature customizable timeout

rayluo:
So far Amazon SWF suggests to use 70 seconds read_timeout (which is 10 seconds longer than the service timeout 60 seconds).

There was no way to customize this. This PR provides a couple new attributes in botocore.client.Config to do that. You can now do it in both botocore and boto3 clients.

    import boto3

    # Both timeouts default to 60, you can customize them, seperately
    config = Config(connect_timeout=50, read_timeout=70)

    session = Session()

    # There will be a line of debug log for this
    session.set_debug_logger()

    # Botocore with default config
    swf = session.create_client('swf')

    # Botocore with customized config
    swf = session.create_client('swf', config=config)

    # Boto3 with customized config
    swf = boto3.client('swf', config=config)


botocore/client.py
```diff
@@ -183,10 +183,13 @@ def _get_client_args(self, service_model, region_name, is_secure,
 
         endpoint_creator = EndpointCreator(self._endpoint_resolver,
                                            region_name, event_emitter)
+        timeout = (client_config.connect_timeout, client_config.read_timeout) \
```
jamesls: If you're going to break this into multiple lines, let's just write this as a standard `if` statement.  The main appeal of these conditional expressions is putting them on a  single line.

botocore/client.py
```diff
@@ -183,10 +183,13 @@ def _get_client_args(self, service_model, region_name, is_secure,
...
+            if client_config else Config.DEFAULT_TIMEOUT
```
jamesls: `if client_config is None`

botocore/client.py
```diff
@@ -523,9 +526,15 @@ class Config(object):
         * User agent extra
 
     """
+
+    DEFAULT_TIMEOUT = 60
+
     def __init__(self, region_name=None, signature_version=None,
-                 user_agent=None, user_agent_extra=None):
+            user_agent=None, user_agent_extra=None,
```
jamesls: pep8, please line these up to the parens on the previous line.

botocore/client.py
```diff
@@ -523,9 +526,15 @@ class Config(object):
         * User agent extra
 
     """
+
```
jamesls: Would you mind updating the `Config` doc string?

tests/unit/test_client.py
```diff
@@ -467,6 +467,14 @@ def test_client_validates_params(self):
             # Missing required 'Foo' param.
             service_client.test_operation(Bar='two')
 
+    def test_client_with_custom_timeout(self):
+        self.create_client_creator().create_client('myservice', 'us-west-2',
+            client_config=client.Config(connect_timeout=123, read_timeout=234))
```
jamesls: This is testing when different `connect/read` timeouts are provided.  Can we add a test for when only a read_timeout (or connect_timeout) is specified?

botocore/client.py
```diff
@@ -523,9 +526,15 @@ class Config(object):
...
+    DEFAULT_TIMEOUT = 60
```
jamesls: We should probably not have this defined twice (here and in endpoints.py).  Let's consolidate this so the default timeout value lives in a single place.

jamesls:
Just had some small comments.

botocore/endpoint.py
```diff
@@ -247,16 +248,17 @@ def create_endpoint(self, service_model, region_name=None, is_secure=True,
...
-                                          response_parser_factory)
+        proxies = self._get_proxies(final_endpoint_url)
+        verify_value = self._get_verify_value(verify)
+        return Endpoint(
+            final_endpoint_url,
+            endpoint_prefix=service_model.endpoint_prefix,
+            event_emitter=self._event_emitter,
+            proxies=proxies,
+            verify=verify_value,
+            timeout=timeout,
```
mtdowling: Does timeout stuff need to be added to endpoints?

botocore/client.py
rayluo: Good point. Originally I was thinking that the client.py is now in the driver's seat anyway, so it seemed reasonable to have its own definition. Also the endpoint.py's DEFAULT_TIMEOUT will always be overridden and arguably should be removed, but I still kept it for backward compatibility (just in case some customers directly use the Endpoint class constructor).

But I have no problem at all to change the client.py to import DEFAULT_TIMEOUT from endpoint.py.

And all other suggestions are also good, I will incorporate them.

botocore/endpoint.py
rayluo: Probably not. They are now constructed in one place which is ClientCreator. The Endpoint, thanks to its underlying infrastructure "requests", supports timeout=None or timeout=integer or timeout=(connect_timeout, read_timeout) by nature.

botocore/client.py
rayluo: I \
don't \
like the "\" either. :-)

I would like to take this chance to hear your comment on the following alternatives. They all pass PEP8.

        # Option 1
        if client_config is None:
            timeout = DEFAULT_TIMEOUT
        else:
            timeout = (client_config.connect_timeout,
                       client_config.read_timeout)

        # Option 2
        timeout = (
            client_config.connect_timeout, client_config.read_timeout
            ) if client_config is not None else Config.DEFAULT_TIMEOUT

        # Option 3: Skip the timeout variable definition, just calculate it and use it on-the-fly
        endpoint = endpoint_creator.create_endpoint(
            ...,
            timeout=DEFAULT_TIMEOUT if client_config is None else (
                client_config.connect_timeout, client_config.read_timeout),
            )

The difference is subtle and probably a matter of taste. Just want to know the team's style so that we are in tune.

jamesls: Bare bones, standard if statements are preferred.  They are the simplest and easiest to read, so that would be Option 1.

jamesls:
The `DEFAULT_TIMEOUT` is still used in two places, `_get_client_args` and the `__init__` of the Config class.  I would expected either:

* We use the values from the Config object directly in `_get_client_args`, since they have default values if they're not provided.  This will require some restructuring of the internals of `_get_client_args`.
* The `DEFAULT_TIMEOUT` is used in `_get_client_args` and has no default in `Config`.

I think I slightly prefer option 1 because I *think* the implementation ends up being simpler.

rayluo:
So the only way to achieve option 1 in _get_client_args() would be to create a Config instance just for using its default value:

        if client_config is None:
            client_config = Config()
        timeout = (client_config.connect_timeout, client_config.read_timeout)

The code does look simpler, which can justify the minor performance penalty of building an extra instance.

New change is coming soon!

jamesls:
@rayluo A few lines below (L228) we create a copy of the client config.  Let's consolidate all that so we're not creating multiple config objects.  That's what I was implying when I mentioned "restructuring some of the internals of `_get_client_args`.  Move the endpoint creation till after we've created the copy of the config, and just use whatever the timeouts happen to be in that config object to create the endpoint.

rayluo:
OK then. We will have to conditionally initialize the both timeout parameters. The ending part of _get_client_args() will look like this:

        # Create a new client config to be passed to the client based
        # on final values. ...
        config_kwargs = dict(
            region_name=region_name, signature_version=signature_version,
            user_agent=user_agent)
        if client_config is not None:
            config_kwargs.update(
                connect_timeout=client_config.connect_timeout,
                read_timeout=client_config.read_timeout)
        new_config = Config(**config_kwargs)

        endpoint_creator = EndpointCreator(...)
        endpoint = endpoint_creator.create_endpoint(
            ...,
            timeout=(new_config.connect_timeout, new_config.read_timeout),
            )

        return {...}

I am refactoring some test cases.

jamesls:
Just had a quick question on one of the unit tests from your latest changes.

jamesls:
Just one small change to the unit tests, otherwise looks good.  :shipit: after updating that one test case.

mtdowling:
I'm still curious why timeouts need to be associated with the Endpoint. @jamesls do you have an opinion on that?

jamesls:
To clarify, the timeout's being associated with an endpoint was [pre-existing](https://github.com/boto/botocore/blob/develop/botocore/endpoint.py#L89).  This PR just plumbs that up to the client configs so users can specify the values.

The endpoint in botocore conceptually represents a remote AWS endpoint you'd send and receive requests to.  So you'd create a request dict, and then use `endpoint.make_request` to send the request to an AWS service.  It also abstracts away (for the most part) an HTTP client.  So given it's the part of botocore responsible for sending/receiving requests (of which a timeout is a property of sending a request), I'm ok with keeping timeout associated with an endpoint object.  I'd be open to hearing alternatives though.

jamesls:
:shipit: Thanks for incorporating all the feedback.

balihoo-gens:
This works well for setting the read_timeout to 70, for use with SWF, but it does not for low values. I run SWF poll_for_decision_task from AWS Lambda, and want to long-poll for no more than 10 seconds. When I set my read_timeout to 10, I do get the expected Exception, but not until after 56 seconds:

```
python2.7/site-packages/botocore/retryhandler.py", line 356, in _check_caught_exception                                                                                                                                                                                   
    raise caught_exception
botocore.vendored.requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='swf.us-east-1.amazonaws.com', port=443): Read timed out. (read timeout=10)

real    0m56.130s
user    0m0.244s
sys     0m0.020s

```
using botocore-1.4.31

Is there yet another socket timeout? I tried setting socket.setdefaulttimeout, but it had no effect.
(obviously my preference would be to just be able to set a timeout on the long poll)

erikbern:
I'm also seeing some weird behavior. If I set read_timeout to 1.0, it still takes about 13-14s to timeout. If I set it to 10.0s it takes about 60s to timeout. Does boto3 do multiple requests under the hood or something?

drmorr0:
@balihoo-gens @erikbern Yes, botocore will do [multiple requests](https://github.com/boto/botocore/blob/develop/botocore/endpoint.py#L171).  I *believe* it does exponential backoff, but I couldn't immediately find the code for that.

ludwigschwardt:
In a unit test that deals with no connectivity I managed to whittle the timeout down to 0.1 seconds by using `botocore.config.Config(connect_timeout=0.1, retries={'max_attempts': 0})`.

Adriayala:
I'm trying to deal with no internet connectivity too, and setting: 
`botocore.config.Config(connect_timeout=0.1, retries={'max_attempts': 0})`
does not reduce the timeout to anything less than 40-45s
I'm using the latest botocore==1.8.40 and boto3==1.5.26

Is there any news on this issue? ETAs or work in progress been made?

jonathanwcrane:
@AdrianSanchezLopez you need to also set the retries/attempts value to 0 as @ludwigschwardt notes.

PR #635, opened 2015-08-27
Implement helpers for CloudFront url signer

rayluo:
setup.py and requirements.txt are also updated to include new dependency rsa.

### DIFF
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -11,6 +11,15 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import datetime
+import logging
+import json
+import base64
+try:
+    maketrans = bytes.maketrans  # For Python 3
+except:
+    from string import maketrans
+
+import rsa
 
 import botocore
 import botocore.auth
@@ -19,6 +28,9 @@ from botocore.exceptions import UnknownSignatureVersionError
 from botocore.exceptions import UnknownClientMethodError
 from botocore.exceptions import UnsupportedSignatureVersionError
 from botocore.utils import fix_s3_host
+
+
+LOG = logging.getLogger(__name__)
 
 
 class RequestSigner(object):
@@ -194,6 +206,65 @@ class RequestSigner(object):
         request.prepare()
 
         return request.url
+
+
+class CloudFrontSigner(object):
+
+    @staticmethod
+    def b64encode(content):
+        return base64.b64encode(content).translate(maketrans(b'+=/', b'-_~'))
+
+    def __init__(self, key_id, private_key):
+        """Create a CloudFrontSigner.
+
+        :type key_id: str
+        :param key_id: The CloudFront Key Pair ID
+
+        :type private_key: str
+        :param private_key: The private key string used for signing.
+        """
+        self.key_id = key_id
+        self.rsa_pk = rsa.PrivateKey.load_pkcs1(private_key.encode('utf8'))
+
+    def sign(self, url, expires, starts=None, ip_address=None):
+        """Sign a given url.
+
+        :type url: str
+        :param url: The URL of the protected object.
+
+        :type expires: int
+        :param expires: The expiry time of the URL. Format is a timestamp.
+
+        :type starts: int
+        :param starts: The start time of the URL. Format is a timestamp.
+
+        :type ip_address: str
+        :param ip_address: Use 'x.x.x.x' for an IP, or 'x.x.x.x/x' for a subnet
+
+        :rtype: str
+        :return: The signed URL.
+        """
+        # SEE: http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html
+        condition = {"DateLessThan": {"AWS:EpochTime": int(expires)}}
+        if ip_address:
+            if '/' not in ip_address:
+                ip_address += '/32'
+            condition["IpAddress"] = {"AWS:SourceIp": ip_address}
+        if starts:
+            condition["DateGreaterThan"] = {"AWS:EpochTime": int(starts)}
+        custom_policy = {
+            "Statement": [{"Resource": url, "Condition": condition}]}
+        compact_policy = json.dumps(
+            custom_policy,
+            sort_keys=True,  # Otherwise test cases become unstable on Python3
+            separators=(',', ':')).encode('utf8')
+        LOG.debug("Policy = %s", compact_policy)
+        signed_policy = rsa.sign(compact_policy, self.rsa_pk, 'SHA-1')
+        return url + ('&' if '?' in url else '?') + '&'.join([
+            'Policy=%s' % self.b64encode(compact_policy).decode('utf8'),
+            'Signature=%s' % self.b64encode(signed_policy).decode('utf8'),
+            'Key-Pair-Id=%s' % self.key_id,
+            ])
 
 
 class S3PostPresigner(object):
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -364,6 +364,23 @@ def parse_to_aware_datetime(value):
     return datetime_obj
 
 
+def datetime2timestamp(dt):
+    """Calculate the timestamp based on the given datetime instance.
+
+    Naive datetime input will be treated as UTC time rather than local time.
+    Note that this behavior is different from datetime.timestamp() in Python 3.
+    (We want unambiguous UTC to avoid test cases failing on different locale.)
+    """
+    epoch = datetime.datetime(1970, 1, 1)
+    if dt.tzinfo:
+        d = dt.replace(tzinfo=None) - dt.utcoffset() - epoch
+    else:
+        d = dt - epoch
+    if hasattr(d, "total_seconds"):
+        return d.total_seconds()  # Works in Python 2.7+
+    return (d.microseconds + (d.seconds + d.days * 24 * 3600) * 10**6) / 10**6
+
+
 def calculate_sha256(body, as_hex=False):
     """Calculate a sha256 checksum.
 
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@ from setuptools import setup, find_packages
 
 requires = ['jmespath==0.7.1',
             'python-dateutil>=2.1,<3.0.0',
+            'rsa>=3.1.2,<=3.1.4',
             'docutils>=0.10']
 
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@ tox==1.4
 python-dateutil>=2.1,<3.0.0
 nose==1.3.0
 mock==1.0.1
+rsa>=3.1.2,<=3.1.4
 wheel==0.24.0
 docutils>=0.10
 -e git://github.com/boto/jmespath.git@develop#egg=jmespath
--- a/tests/unit/test_signers.py
+++ b/tests/unit/test_signers.py
@@ -21,7 +21,7 @@ from botocore.credentials import Credentials
 from botocore.exceptions import NoRegionError, UnknownSignatureVersionError
 from botocore.exceptions import UnknownClientMethodError, ParamValidationError
 from botocore.exceptions import UnsupportedSignatureVersionError
-from botocore.signers import RequestSigner, S3PostPresigner
+from botocore.signers import RequestSigner, S3PostPresigner, CloudFrontSigner
 
 from tests import unittest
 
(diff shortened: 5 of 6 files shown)


botocore/signers.py
```diff
@@ -11,6 +11,15 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import datetime
+import logging
+import json
+import base64
+try:
```
kyleknap: Lets move this to ``compat.py``. That is where we put all of these type of conversions and ensures we never have to repeat the logic. All you have to do is import it from ``botocore.compat``

botocore/signers.py
```diff
@@ -196,6 +208,65 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
         return request.url
 
 
+class CloudFrontSigner(object):
```
kyleknap: Lets be more specific here and call the class CloudFrontPresigner.

botocore/signers.py
```diff
@@ -196,6 +208,65 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+
+    @staticmethod
```
kyleknap: What is the reason behind making this a static method? I would have probably made it an internal method

botocore/signers.py
```diff
@@ -196,6 +208,65 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+        """Create a CloudFrontSigner.
+
+        :type key_id: str
+        :param key_id: The CloudFront Key Pair ID
+
+        :type private_key: str
+        :param private_key: The private key string used for signing.
+        """
+        self.key_id = key_id
+        self.rsa_pk = rsa.PrivateKey.load_pkcs1(private_key.encode('utf8'))
```
kyleknap: We should ensure that the type passed into private_key is a string. i.e. reject Binary types because the encode would fail if a bytes object is passed in.

botocore/signers.py
```diff
@@ -196,6 +208,65 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+
+    def sign(self, url, expires, starts=None, ip_address=None):
```
kyleknap: It would nice to be consistent here with the other presigner and call it ``generate_presigned_url``

botocore/signers.py
```diff
@@ -196,6 +208,65 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+        """Sign a given url.
+
+        :type url: str
+        :param url: The URL of the protected object.
+
+        :type expires: int
```
kyleknap: Can we name these something like ``expires_on`` and ``starts_on``? It is not completely obvious that expires represents at what time the url will expire or how long it will take to expire.

botocore/signers.py
```diff
@@ -196,6 +208,65 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+        :param starts: The start time of the URL. Format is a timestamp.
+
+        :type ip_address: str
+        :param ip_address: Use 'x.x.x.x' for an IP, or 'x.x.x.x/x' for a subnet
+
+        :rtype: str
+        :return: The signed URL.
+        """
+        # SEE: http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html
+        condition = {"DateLessThan": {"AWS:EpochTime": int(expires)}}
```
kyleknap: If we are passing specific dates. Would it be a better interface if we used datetimes and convert these to timestamps?

botocore/signers.py
```diff
@@ -196,6 +208,65 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+        if ip_address:
+            if '/' not in ip_address:
+                ip_address += '/32'
```
kyleknap: We should document the default for the subnet mask in the docstring,

botocore/signers.py
```diff
@@ -196,6 +208,65 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+            condition["IpAddress"] = {"AWS:SourceIp": ip_address}
+        if starts:
+            condition["DateGreaterThan"] = {"AWS:EpochTime": int(starts)}
+        custom_policy = {
+            "Statement": [{"Resource": url, "Condition": condition}]}
+        compact_policy = json.dumps(
+            custom_policy,
+            sort_keys=True,  # Otherwise test cases become unstable on Python3
+            separators=(',', ':')).encode('utf8')
+        LOG.debug("Policy = %s", compact_policy)
```
kyleknap: We typically use ``:`` instead of ``=`` in botocore debug log messages.

botocore/utils.py
```diff
@@ -364,6 +364,23 @@ def parse_to_aware_datetime(value):
...
+def datetime2timestamp(dt):
+    """Calculate the timestamp based on the given datetime instance.
+
+    Naive datetime input will be treated as UTC time rather than local time.
+    Note that this behavior is different from datetime.timestamp() in Python 3.
+    (We want unambiguous UTC to avoid test cases failing on different locale.)
+    """
+    epoch = datetime.datetime(1970, 1, 1)
+    if dt.tzinfo:
+        d = dt.replace(tzinfo=None) - dt.utcoffset() - epoch
```
kyleknap: Why not just replace it with a utc timezone instead of none?

botocore/utils.py
```diff
@@ -364,6 +364,23 @@ def parse_to_aware_datetime(value):
...
+    else:
+        d = dt - epoch
+    if hasattr(d, "total_seconds"):
```
kyleknap: We already have a method for ``total_seconds`` in ``botocore.compat``

kyleknap:
Looks good. Just had some questions and comments.

botocore/signers.py
rayluo: But I thought "pre-signed" is a naming convention only for S3. For example, in [S3 document, there are 19 occurrence of "pre-signed" and they don't use "signed"](http://docs.aws.amazon.com/AmazonS3/latest/dev/PresignedUrlUploadObject.html). On the contrary, [in CloudFront document, the naming convention is always saying "signed", never "pre-signed"](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html). So, shall we keep the current "CloudFrontSigner" as-is?

botocore/signers.py
rayluo: Well, it does not need anything else from other class member, but it still tightly couples with the functionality of this class, so I put it in this class, as a static method. By the way, there is a [similar function in Boto, also organized as a static method](https://github.com/boto/boto/blob/develop/boto/cloudfront/distribution.py#L680-L690).

For the "why not internal" part, I think it doesn't harm to also provide this helper for outsider, for what it's worth. If you prefer to name it as _b64encode() or even __b64encode(), just let me know.

botocore/signers.py
mtdowling: I'd prefer CloudFrontUrlSigner considering we are only signing URLs.

botocore/signers.py
rayluo: Probably not. This method is intended to be a lower level implementation closely matching to [the data type defined in the wire protocol](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html). Otherwise the logic here would become unnecessarily complicated and bring extra trouble, should we ever need to debug a scenario "oops, signed url doesn't work, which part goes wrong?".

On the other hand, I do implement a datetime2timestamp() helper in this PR, but it is not really used in botocore. I plan to use it to support flexible datetime formats in awscli, after this PR gets merged.

botocore/signers.py
rayluo: I thought I mentioned that as "**Use 'x.x.x.x' for an IP**, or 'x.x.x.x/x' for a subnet" already.  Feel free to suggest a different wording. Thanks!

botocore/utils.py
rayluo: Correct me if I am wrong, I thought there is no UTC TZ defined in builtin datetime in Python. So I always tend to write it in a way to avoid the dependency to "from dateutil.tz import tzutc".

Besides, the current logic is kinda documented in the standard python document too. Just search ``converted to naive UTC datetimes first. The result is (a.replace(tzinfo=None) - a.utcoffset())`` in [this page](https://docs.python.org/2/library/datetime.html).

botocore/signers.py
mtdowling: I think it makes sense to remove this method and merge it into the signing method. This method is not just base64 encoding a string, but also transforming the string to conform with the unique requirements of CloudFront urls. This is not really something that needs to be exposed publicly.

botocore/signers.py
mtdowling: CloudFront supports canned policies and custom policies. The way this is currently implemented does not support custom policies. I think you should only use a canned policy if the user does not supply a custom policy. Perhaps there could be two methods -- one that signs with a custom policy, and one that accepts a few abstracted parameters like expires, creates a policy, and then calls the other function that signs custom policies. Here's how this same functionality is implemented in the PHP SDK: https://github.com/aws/aws-sdk-php/blob/master/src/CloudFront/UrlSigner.php#L61

botocore/signers.py
```diff
@@ -196,6 +208,65 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+        :param url: The URL of the protected object.
+
+        :type expires: int
+        :param expires: The expiry time of the URL. Format is a timestamp.
+
+        :type starts: int
+        :param starts: The start time of the URL. Format is a timestamp.
+
+        :type ip_address: str
+        :param ip_address: Use 'x.x.x.x' for an IP, or 'x.x.x.x/x' for a subnet
```
mtdowling: Instead of trying to build up a custom policy for the user, I'd get rid of all of the extra parameters and just have users pass in a policy represented as a string.

botocore/signers.py
```diff
@@ -196,6 +208,65 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+            condition["DateGreaterThan"] = {"AWS:EpochTime": int(starts)}
+        custom_policy = {
+            "Statement": [{"Resource": url, "Condition": condition}]}
+        compact_policy = json.dumps(
+            custom_policy,
+            sort_keys=True,  # Otherwise test cases become unstable on Python3
+            separators=(',', ':')).encode('utf8')
+        LOG.debug("Policy = %s", compact_policy)
+        signed_policy = rsa.sign(compact_policy, self.rsa_pk, 'SHA-1')
+        return url + ('&' if '?' in url else '?') + '&'.join([
```
mtdowling: This might be better as its own function. You could move your b64encode function body into this new function as well.

botocore/signers.py
rayluo: But please don't forget that CloudFront not only allows signing URLs, they also supports [signing cookies](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-signed-cookies.html). I envision we may have a generate_signed_cookie() into this CloudFrontSigner class in (far?) future.

mtdowling:
There are a few things missing from this PR:

1. Support for canned policies.
2. Support for truly custom policies.
3. Explicit support for rtmp URLs. rtmp urls are serialized differently than http/https URLs. See the docs and https://github.com/aws/aws-sdk-php/blob/master/src/CloudFront/UrlSigner.php for examples.

botocore/signers.py
mtdowling: I think most of this code should be removed actually, and instead just allow users to pass in custom policies. If we really wanted to and had the resources, we could create a better abstraction for creating custom policies.

botocore/signers.py
rayluo: I would recommend to NOT remove it. When this [special base64 may be as simple as a one-liner in other language such as PHP](https://github.com/aws/aws-sdk-php/blob/master/src/CloudFront/UrlSigner.php#L95), in Python it needs special care to handle different strings in Python 2 and Python 3. It bit me, and I ended up developing dedicated test case for it. So I think it deserve to become a separated method.

I will adopt your suggestion to demote it to an internal function _b64encode(). Thanks!

botocore/signers.py
rayluo: But, given that the main point of [using custom policy is to customize the combination of those 3 parameters](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html#private-content-custom-policy-statement), wouldn't it be more convenient that we allow caller to simply provide some of those parameters and then we build a correct custom policy for them?

mtdowling: Building a policy vs signing a policy are different things, so I think that should be a separate abstraction and not part of this method (e.g., as policies evolve and become more complicated, so would this method). Instead, I think it's better to have a separate method for building up custom policies, but I don't think that level of abstraction is necessary in this PR and could be added in later. I think this class should be able to sign completely custom policies (represented as a string) and be able to sign/create canned policies for a user.

botocore/signers.py
mtdowling: Ah that's true. Will you be able to reuse much code such that it makes sense for them to be the same class?

botocore/signers.py
rayluo: Thanks @mtdowling, I did not know you implemented the CloudFront in PHP. Good job! We could have talked more about this topic beforehand.

You have several similar comments in this page. Please allow me to respond them here in one place. Yes I am aware of the "canned policy" and "custom policy" difference. In my understanding, the name "custom policy" is misleading in the first place. Customer can not really customize an arbitrary policy (just like what they can do in IAM policies), and expects the API to just sign it. [CloudFront only accepts "custom policy" with combinations of predefined 3 parameters in one format](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html#private-content-custom-policy-statement). So I can not see a scenario that customer would need more flexibility to build the policy by themselves and call a separated sign_custom_policy() function.

In other words, despite of the subtle output difference of so-called "canned policy" and "custom policy", their input are all driven by a subset of same parameters. There won't be a need to expose a public sign_custom_policy() for a "customized" policy.

kyleknap: Let me chime in on this one as I was not super familiar with cloudfront signing before I looked at the implementation. I also think we should just only allow for the custom/canned policy in the signature and then build abstractions like policy building on top of the signer. Here are the reasons why I think it is appropriate design choice:

1) Sure there are only 3 parameter combinations that cloudfront allows currently, but what stops them from adding more? Everytime they add a new parameter, we will have to go back and update the signer class because it involves the building of the custom policy, and we should not need to because that class's job is strictly to generate urls. This causes to be really a maintenance burden, and has been a pain for us in the past.

2) Botocore is the low-level library. The underlying classes and methods like the signers are not about being super user friendly. They are about being very robust, constant, and reliable. This allows us to build abstractions on top of these classes in other libraries like the CLI and boto3 that are much more user friendly (such as building the policy for the user).

I would recommend taking any of the policy building and move it to a CloudFrontPolicyBuilder class that exists in another module in botocore or lives in a library like the CLI, boto3, or a even new package and leave all of the signing logic in your signer class.

rayluo: OK then. Incorporating your suggestions, how about this?

1. Let's change the major method to require a policy as input. This one will be future-proof, in a sense that caller can always custom their policy to adapt to future CloudFront requirement.

        def generate_signed_url(self, url, policy):
            pass

2. And let's also have an OPTIONAL policy builder sitting inside this Signer as a helper. The (current) implementation is merely a dozen lines of code, so separating it into another module or even another repo seems like an overkill.

        def build_policy(self, url, expires, starts=None, ip_address=None):
            pass

Eventually the usage pattern would become:

    signer = CloudFrontSigner(key_id, private_key)

    # The following is optional.
    policy = signer.build_policy(url, expires, starts=foo, ip_address=bar)
    # You can always skip above and provide your own.
    policy = {...}

    signed_url = signer.generate_signed_url(url, policy)

I hope this can give us the best of both worlds. I will start working on this one, and also incorporate other review feedbacks. Thanks!

botocore/signers.py
jamesls: Call it something else please to make it clear this isn't just a basic base64 encoding.  Looks like it's an encoding that also makes it valid in a url.

I'd also be in favor of removing the staticmethod and just making it internal.

botocore/signers.py
jamesls: Can we also be more explicit in our docs about this.  We should exactly what format we expect the private key to be in.

botocore/signers.py
jamesls: I'm just catching up on all this, but where did you all land on the canned acls?  I'm not seeing it in any of the proposed actions.

It's something we should be supporting isn't it?

Also, I would suggest not worrying about the policy builder in this PR.  We can add it later.

botocore/signers.py
jamesls: I agree with Kyle on this.  I don't think anyone thinks in terms of an epoch timestamp.  It's much simpler to work with datetimes, especially when dealing with absolute times.

setup.py
```diff
@@ -7,6 +7,7 @@
 
 requires = ['jmespath==0.7.1',
             'python-dateutil>=2.1,<3.0.0',
+            'rsa>=3.1.2,<=3.1.4',
```
jamesls: This is weird.  Why are we locking to 3 patch version?  Either lock to a minor version or a major version.

jamesls:
I agree with most of the comments here.  The only thing I would add in addition is that I would improve the test cases.  It's good to have a few overall tests that verify the entire expected presigned url, which is what you currently have, but these tests have bad error locality on failures.  Take a look at some of the other signer unit tests.  They're much more granular, e.g "When I provide this param, then this specific part is in the URL."  Makes it much easier to follow and provides better error messages on failure.  You also have no tests for any error cases.

setup.py
rayluo: Quick feedback on this one. This line is currently copied from its counterpart in aws-cli. Later I did test the latest version of rsa 3.2 and it works. So I will lock to >=3.2,<3.3 in next PR.

But then it will become inconsistent with aws-cli. We may want to also upgrade rsa in aws-cli too. Coincidentally, there is a recent issue in our aws-cli repo suggesting an upgrade.

botocore/signers.py
rayluo: Sure I can change that.

While I have no problem with removing the staticmethod part, I will catch up with you guys to understand the criteria so that I will make confident choice in future. At this moment I see some other staticmethod internal methods in our code base. Not sure when they would be appropriate.

By the way, just share some anecdote in this context for your weekend fun. In my very first unpublished prototype, this function was originally a completely separated (and therefore static) function named ``cloudfront_b64encode()``, because it is neither a ``standard_b64encode()`` nor the another Python builtin ``urlsafe_b64encode()``. Later it was grouped into this CloudFrontSigner class, so I removed that seemingly excessive prefix.

So it will be a normal internal method as ``_cloudfront_b64encode(self, content)`` in next PR.

jamesls:
Closing out PR for now.  We can reopen once we take up this work again.

PR #637, opened 2015-08-29
add tests to MANIFEST.in

alunduil:
When packaging botocore it's nice to be able to run the tests from the
source distribution to ensure that the packaging has been done
correctly.  This includes the tests directoy in the sdist without
installing it to the system.  Piggy-backed is a change to the docs
inclusion that uses graft rather than recursive-include.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,5 +3,6 @@ include LICENSE.txt
 include requirements.txt
 include botocore/vendored/requests/cacert.pem
 recursive-include botocore/data *.json
-recursive-include docs *
+graft docs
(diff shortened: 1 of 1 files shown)


jamesls:
From reading the distutils docs, am I correct in assuming that `graft docs` is identical to `recursive-include docs *`?  The main benefit in using `graft` is that it's slightly shorter?

alunduil:
Yep.  That's why I piggy-backed it.  Let me know if you want that in its own commit.

jamesls:
Nope, this is fine.  Thanks for the pull request.

PR #638, opened 2015-09-02
Add resource leak tests

jamesls:
This adds several functional and integration tests for verifying that we don't leak memory when performing various botocore operations.  This includes:

* Creating multiple clients.  Verify creating/deleting multiple clients does not cause undesired memory growth.
* Performing various AWS operation does not have undesired memory growth.
* Performing streaming S3 uploads/downloads has constant memory utilization during (and after) the transfers.

The best description of how this works is in the tests/cmd-runner module docstring.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -176,7 +176,7 @@ class ClientRunner(object):
 
     def start(self):
         """Start up the command runner process."""
-        self._popen = Popen(['python', self.CLIENT_SERVER],
+        self._popen = Popen([sys.executable, self.CLIENT_SERVER],
                             stdout=PIPE, stdin=PIPE)
 
     def stop(self):
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -203,7 +203,8 @@ class ClientRunner(object):
 
         """
         cmd_str = ' '.join(cmd) + '\n'
-        self._popen.stdin.write(cmd_str)
+        cmd_bytes = cmd_str.encode('utf-8')
+        self._popen.stdin.write(cmd_bytes)
         self._popen.stdin.flush()
 
     def is_cmd_finished(self):
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -20,6 +20,9 @@ import shutil
 import contextlib
 import tempfile
 import binascii
+import platform
+import select
+from subprocess import Popen, PIPE
 
 
 # The unittest module got a significant overhaul
@@ -34,6 +37,12 @@ else:
 import botocore.loaders
 import botocore.session
 _LOADER = botocore.loaders.Loader()
+
+
+def requires_memory_collection(cls):
+    if platform.system() not in ['Darwin', 'Linux']:
+        return unittest.skip('Memory tests only supported on mac/linux.')(cls)
+    return cls
 
 
 def random_chars(num_chars):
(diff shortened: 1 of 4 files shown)


tests/__init__.py
```diff
@@ -114,3 +123,108 @@ def setUp(self, **environ):
...
+
+    def stop(self):
+        """Shutdown the command runner process."""
+        self.cmd('exit')
+        self._popen.wait()
+
+    def send_cmd(self, *cmd):
+        """Send a command and return immediately.
+
+        This is a lower level method thatn cmd().
```
kyleknap: thatn --> than

tests/cmd-runner
```diff
@@ -0,0 +1,173 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""Driver for client scripts.
+
+How it Works
+============
+
+This script is part of the infrastructure used for resource leak tests.
+At a high level, we're trying to varify that specific types of operations
```
mtdowling: varify -> verify

tests/__init__.py
```diff
@@ -114,3 +123,108 @@ def setUp(self, **environ):
...
+        command_list = ['ps', 'u', '-p']
+        command_list.append(str(pid))
+        p = Popen(command_list, stdout=PIPE)
+        stdout = p.communicate()[0]
+        if not p.returncode == 0:
+            raise RuntimeError("Could not retrieve memory")
+        else:
+            # Get the RSS from output that looks like this:
+            # USER       PID  %CPU %MEM      VSZ    RSS   TT  STAT STARTED      TIME COMMAND
+            # user     47102   0.0  0.1  2437000   4496 s002  S+    7:04PM   0:00.12 python2.6
```
mtdowling: Would `ps -p ... -o rss` work? It will just return the following:
```
   RSS
242144
```

tests/__init__.py
```diff
@@ -36,6 +39,12 @@
 _LOADER = botocore.loaders.Loader()
 
 
+def requires_memory_collection(cls):
```
kyleknap: Can we indicate that this is a skip if decorator? Maybe rename it to ``skip_if_requires_memory_collection``

tests/__init__.py
```diff
@@ -114,3 +123,108 @@ def setUp(self, **environ):
...
+        self.runner.record_memory()
+
+    def memory_samples(self):
+        return self.runner.memory_samples()
+
+    def tearDown(self):
+        self.runner.stop()
+
+
+class ClientRunner(object):
```
kyleknap: The naming convention is a little confusing to me based on the description in ``cmd-runner``. The fact that the code for the "driver" here uses a ``ClientRunner`` while the other file that represents the runner uses the name ``CommandRunner`` makes it hard to differentiate between the classes purposes based on looking at the class name.

Can we name it something like ``CommandDriver`` because it serves as the driver and all of its methods are something like ``send_cmd()`` or ``cmd()``?

tests/__init__.py
```diff
@@ -114,3 +123,108 @@ def setUp(self, **environ):
...
+        the memory during the upload process, you can use
+        send_cmd() instead of cmd().
+
+        """
+        cmd_str = ' '.join(cmd) + '\n'
+        cmd_bytes = cmd_str.encode('utf-8')
+        self._popen.stdin.write(cmd_bytes)
+        self._popen.stdin.flush()
+
+    def is_cmd_finished(self):
```
kyleknap: So this will be used for the s3 file upload commands?

tests/cmd-runner
```diff
@@ -0,0 +1,173 @@
...
+        num_clients = args[0]
+        client_args = args[1:]
+        for _ in range(int(num_clients)):
+            self._clients.append(self._session.create_client(*client_args))
+
+    def _do_free_clients(self, args):
+        # Frees the memory associated with all clients.
+        self._clients = []
+
+    def _do_gc_collect(self, args):
```
kyleknap: When will this be used?

tests/functional/leak/test_resource_leaks.py
```diff
@@ -0,0 +1,42 @@
...
+        self.record_memory()
+        for _ in range(100):
+            self.cmd('create_client', 's3')
+            self.cmd('free_clients')
+        self.record_memory()
+        start, end = self.runner.memory_samples
+        self.assertTrue((end - start) < self.MAX_GROWTH_BYTES, (end - start))
+
+    def test_create_memory_clients_in_loop(self):
+        self.cmd('create_multiple_clients', '200', 's3')
```
kyleknap: What is the reason for having to create a bunch of clients ahead of time? Is it because of python's garbage collector? A comment here about why and in the previous test when you do something similar would be nice.

kyleknap:
It looks good. I had a couple of minor comments and questions that may get answered once the s3 tests are added.

One general thought I had though was that if we decide to ever add a lot of different types of performance tests the ``CommandRunner`` will get pretty bloated. What would be really cool is if you could send an arbitrary function or set of code to the ``CommandRunner`` and have the ``CommandRunner`` run that. That would help the ``CommandRunner`` from being too bloated and you would not have to go to the ``cmd-runner script`` to figure out what code is actually being run. That being said I do not think this is an issue given we do not plan to add too many of these resource leak tests.  Thoughts?

tests/__init__.py
jamesls: Not opposed to renaming it.  Technically, this is skipping the test if the system doesn't support memory collection, but that would be pretty long: `skip_unless_has_memory_collection`.  You ok with that?

tests/__init__.py
jamesls: Yeah, I'll rename.

tests/__init__.py
jamesls: Yeah, I'll update.

tests/__init__.py
kyleknap: Sounds good to me.

tests/__init__.py
jamesls: Yes.

tests/cmd-runner
jamesls: Not used, I'll remove.  I was initially using this because botocore has some cyclic references we can clean up, but we're not using it now so I'll remove it.

tests/functional/leak/test_resource_leaks.py
jamesls: Good point.  This is specifically because of how python allocate's memory and doesn't give it back.  I'll add an explanation here.

jamesls:
@kyleknap @mtdowling Feedback incorporated.  Let me know if I missed anything.

mtdowling:
:ship:

jamesls:
@kyleknap Sorry, didn't see your overall comment.

Being able to perform you're asking would require a change to the communication protocol, but would be doable.

However, we'd only need to add commands if the tasks we'd need to perform change.  For example, if we wanted to add timing/perf tests for creating clients or S3 uploads or making an AWS API call, we wouldn't have to make any changes to `cmd-runner`, we'd only have to subclass `BaseClientDriverTest` and implement timing using the existing commands.

I don't envision us adding a whole lot of new tasks, but I do see us adding more tests that utilize the existing tasks.

Either way, I think we can definitely address that when we start adding new tests.

kyleknap:
Looks good. :ship:

tests/__init__.py
jamesls: Oh wow, now you're comment makes more sense.  I never actually included the integration tests that test s3 file uploads!  Sending now.

PR #639, opened 2015-09-02
Add back paginator-1.json and waiters-2.json for cloudfront

rayluo:
These 2 files are now copied from previous version, the 2014-11-06 directory.

Also add a couple test cases to prevent this from happening in future.

### DIFF
--- a/botocore/data/cloudfront/2015-04-17/paginators-1.json
+++ b/botocore/data/cloudfront/2015-04-17/paginators-1.json
@@ -1,0 +1,32 @@
+{
+  "pagination": {
+    "ListCloudFrontOriginAccessIdentities": {
+      "input_token": "Marker",
+      "output_token": "CloudFrontOriginAccessIdentityList.NextMarker",
+      "limit_key": "MaxItems",
+      "more_results": "CloudFrontOriginAccessIdentityList.IsTruncated",
+      "result_key": "CloudFrontOriginAccessIdentityList.Items"
+    },
+    "ListDistributions": {
+      "input_token": "Marker",
+      "output_token": "DistributionList.NextMarker",
+      "limit_key": "MaxItems",
+      "more_results": "DistributionList.IsTruncated",
+      "result_key": "DistributionList.Items"
+    },
+    "ListInvalidations": {
+      "input_token": "Marker",
+      "output_token": "InvalidationList.NextMarker",
+      "limit_key": "MaxItems",
+      "more_results": "InvalidationList.IsTruncated",
+      "result_key": "InvalidationList.Items"
+    },
+    "ListStreamingDistributions": {
+      "input_token": "Marker",
+      "output_token": "StreamingDistributionList.NextMarker",
+      "limit_key": "MaxItems",
+      "more_results": "StreamingDistributionList.IsTruncated",
+      "result_key": "StreamingDistributionList.Items"
+    }
+  }
+}
--- a/botocore/data/cloudfront/2015-04-17/waiters-2.json
+++ b/botocore/data/cloudfront/2015-04-17/waiters-2.json
@@ -1,0 +1,47 @@
+{
+  "version": 2,
+  "waiters": {
+    "DistributionDeployed": {
+      "delay": 60,
+      "operation": "GetDistribution",
+      "maxAttempts": 25,
+      "description": "Wait until a distribution is deployed.",
+      "acceptors": [
+        {
+          "expected": "Deployed",
+          "matcher": "path",
+          "state": "success",
+          "argument": "Distribution.Status"
+        }
+      ]
+    },
+    "InvalidationCompleted": {
+      "delay": 20,
+      "operation": "GetInvalidation",
+      "maxAttempts": 60,
+      "description": "Wait until an invalidation has completed.",
+      "acceptors": [
+        {
+          "expected": "Completed",
+          "matcher": "path",
+          "state": "success",
+          "argument": "Invalidation.Status"
+        }
+      ]
+    },
+    "StreamingDistributionDeployed": {
+      "delay": 60,
+      "operation": "GetStreamingDistribution",
+      "maxAttempts": 25,
+      "description": "Wait until a streaming distribution is deployed.",
+      "acceptors": [
+        {
+          "expected": "Deployed",
+          "matcher": "path",
+          "state": "success",
+          "argument": "StreamingDistribution.Status"
+        }
+      ]
+    }
+  }
+}
--- a/tests/functional/test_model_completeness.py
+++ b/tests/functional/test_model_completeness.py
@@ -1,0 +1,42 @@
+# Copyright 2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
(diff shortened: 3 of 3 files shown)


tests/functional/test_data_completeness.py
```diff
@@ -0,0 +1,50 @@
...
+from botocore.exceptions import DataNotFoundError
+
+
+class LoaderWrapper(Loader):
+
+    def load_data(self, name):
+        try:
+            return super(LoaderWrapper, self).load_data(name)
+        except DataNotFoundError:
+            return None  # So we map the exception to None
```
jamesls: Why don't we just catch the DataNotFoundError in the test?  I'm not really a fan of changing the semantics of the Loader's `load_data` in this test.

tests/functional/test_data_completeness.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class TestDataTypes(unittest.TestCase):
+
+    def _test_data_type_is_not_lost_in_new_version(self, type_name):
+        # Make sure if a paginator and/or waiter exists in previous version,
+        # there will be a successor existing in latest version.
+        loader = LoaderWrapper()
+        for service_name in Session().get_available_services():
```
jamesls: Use test generators (search in the tests/ directory for `test_*` methods that `yield`).

Right now, this will stop as soon as a single service fails this check and we won't know if other services are susceptible to this issue.

tests/functional/test_data_completeness.py
rayluo: Actually I tried that in my previous version (which I did not push to here). Without the help of this wrapper, the main logic in the test_blah() function became arguably less obvious. Personally I don't really have a preference here. I will provide a no-wrapper version in next push.

tests/functional/test_data_completeness.py
rayluo: Good point. Will do.

mtdowling:
LGTM

tests/functional/test_data_completeness.py
```diff
@@ -0,0 +1,44 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from botocore.session import Session
+from botocore.loaders import Loader
+from botocore.exceptions import DataNotFoundError
+
+
+class TestDataTypes(object):
```
jamesls: We generally just use functions for test generators.  Given that there's no instance variables for this class, what are the advantages for using a class in this situation?

rayluo: It is encapsulation, one of the 3 benefits of OO.

TL;DR.

The pattern of such test generator always include a generator and one (or more) helper. The helper(s) is tightly coupled with the generator, and is unlikely to be reused by others. The class encapsulation makes it explicit that these 2 guys always sit together. The conceptual benefit may not be obvious here but, should this file evolve to contain more test cases in future, I believe cases grouped by several classes would look better than [a bunch of flattened functions like this](https://github.com/boto/botocore/blob/develop/tests/integration/test_smoke.py#L181-L249).

jamesls: We should be evaluating this approach in the context of "Given this change is inconsistent with how all of the other botocore test generators are written, are there significant and immediately tangible benefits that make such inconsistencies worth it?"  And from what I'm seeing, my opinion is no, it's not worth it.

As you mention "should this file evolve to contain more test cases in the
future...".  So we don't actually need this right now.  When (*if*) we reach
such a scenario, then sure let's adapt and switch to this class based approach.

Until then, I prefer to keep it simple and consistent with the rest of botocore.

rayluo: I am open and flexible.  Just want to make sure we are on the same page about the pros and cons.

If I understand correctly, we agree that the class-based approach is (slightly) better; should the need arises, we can adapt to it in spite of inconsistency.

Deduction 1: In the long term, inconsistency will eventually be there anyway, how about start to adapt to the future-proof approach now?

Deduction 2: At a hindsight, if we rewind back to day 1, perhaps we would write the test cases in class-based flavor, so there wouldn't be such inconsistency in the first place. (I suspect that the flatten-function flavor is influenced by the [nose 101 document](http://nose.readthedocs.org/en/latest/writing_tests.html), but 101 document typically focuses only on features, it does not necessarily guide us how to organize the code.)

Last but not the least, if we stick with current flatten-function flavor (which is fine), chances are, we won't bother to refactor old runnable test cases when we need to add more test cases in future because "If it ain't broken, why fix it?", and then even the new test cases are less likely to choose the class-based flavor to violate the convention in the same file.

By the way, in this specific case, both functions use a loader instance independently. We could choose to make that loader instance as a member in this class to be shared by the 2 methods. That should make the test cases run a bit faster. Will that be better reason alone, on top of all above discussion?

kyleknap: +1 on keeping it consistent with botocore. All generator-driven tests so far have been expressed as functions and not methods of classes. Furthermore, the class you created does not inherit from ``unittest.TestCase`` so you are not taking advantage of any of the classes testing methods like ``setUp``, ``tearDown`` or any of the assertions. So, there is no *current* reason to wrap the tests in a class.

tests/functional/test_data_completeness.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def _test_data_type_is_not_lost(self, service_name, type_name,
```
kyleknap: Can we switch ``data_type`` naming convention to ``model``. When I think of ``data_type`` I think of the actual data type passed to a parameter and the not the model type.

kyleknap:
Looks good. Only had a couple minor comments. Thanks for adding the tests they will be a good safety check.

tests/functional/test_data_completeness.py
jamesls: @rayluo That's not what I'm saying.  I don't think you can make a sweeping generalization like "a class-based approach is better."  If you're taking advantage of things that a class gives you (inheritance, instance state, etc) then yes it's better.  If you're not using those things (like the situation here) then it's not any better than a function.

If we ever run into a scenario where a class based approach is *actually* better, and provides immediate benefit, I don't think you'll get any push back from us.

At any rate, here's what I suggest:

* Take @kyleknap and my suggestion.  Be simple and consistent.  Use functions.  This PR is fixing a bug whose fix (which we all agree on) should not be blocked on a discussion on changing how we write test generators.

* Continue this discussion independently of this bug fix.  If you're willing to put in the effort to rework how we write test generators at a more holistic level, then that's certainly something we can discuss.

Sound good?

tests/functional/test_data_completeness.py
rayluo: No problem. In that case, I will also change the file name as well.

tests/functional/test_data_completeness.py
rayluo: I can not say I 100% agree, but it is time to "disagree and commit". :-) Will commit my new PR soon.

kyleknap:
Thanks for updating. Looks good. :ship:

PR #640, opened 2015-09-02
Add bucket name validation to S3 calls

jamesls:
Fixes boto/boto3#181

Based on:

http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#non-dns-compliant-bucketname-challenges

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -21,6 +21,7 @@ import hashlib
 import logging
 import xml.etree.cElementTree
 import copy
+import re
 import string
 
 from botocore.compat import urlsplit, urlunsplit, unquote, json, quote, six
@@ -42,7 +43,12 @@ logger = logging.getLogger(__name__)
 
 REGISTER_FIRST = object()
 REGISTER_LAST = object()
-VALID_BUCKET_CHARS = string.ascii_letters + string.digits + '.-_'
+# From the S3 docs:
+# The rules for bucket names in the US Standard region allow bucket names
+# to be as long as 255 characters, and bucket names can contain any
+# combination of uppercase letters, lowercase letters, numbers, periods
+# (.), hyphens (-), and underscores (_).
+VALID_BUCKET = re.compile('^[a-zA-Z0-9.\-_]{1,255}$')
 
 
 def check_for_200_error(response, **kwargs):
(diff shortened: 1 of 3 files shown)


botocore/handlers.py
```diff
@@ -140,6 +143,28 @@ def conditionally_calculate_md5(params, **kwargs):
...
+    # (.), hyphens (-), and underscores (_).
+    if 'Bucket' not in params:
+        return
+    bucket = params['Bucket']
+    if len(bucket) > 255:
+        error_msg = (
+            'Invalid bucket name "%s": Bucket name cannot be '
+            'longer than 255 characters.' % bucket)
+        raise ParamValidationError(report=error_msg)
+    for char in bucket:
```
mtdowling: Any reason to not use a regex here instead? Did you benchmark it and find it to be slower than checking against a list of valid characters?

Alternatively it might be better to use a set for VALID_BUCKET_CHARS so that you don't have to iterate over all of the characters each time your determining if the bucket string is valid. Considering this is going to be used for every S3 request that involves a Bucket parameter (almost all), I think this micro-optimization is reasonable.

mtdowling:
I just had a question about how your checking the string, but other than that, looks good.

botocore/handlers.py
jamesls: The main reason for not using the regex was just to have more specific error messages for the two cases.  Otherwise we'd just have to say it's not valid, and maybe just show the regex it has to match.

mtdowling: I think it would be acceptable to have a generic error message for any case in which the regex does not match the provided bucket. Right now it's just the difference between too long and invalid character cases. I think it would be fairly obvious that their bucket is too long if you are including the provided bucket in the error message. Even still, you could keep the length check and then use the regex check to have different error messages if desired.

jamesls: Using:

```
$ python -m timeit -s 'import botocore.session; from botocore.handlers import validate_bucket_name; params={"Bucket": "my-bucket-name"}' 'validate_bucket_name(params)'
$ python -m timeit -s 'import botocore.session; from botocore.handlers import validate_bucket_name; params={"Bucket": "z9_.-" * 51}' 'validate_bucket_name(params)'
```

I get these results:

| Algo    | Bucket Name    | Time                |
|---------|----------------|---------------------|
| current | my-bucket-name | 1.79 usec per loop  |
| current | 'z9_.-' * 51   | 25.8 usec per loop  |
| set     | my-bucket-name | 1.67 usec per loop  |
| set     | 'z9_.-' * 51   | 16.3 usec per loop  |
| regex   | my-bucket-name | 0.971 usec per loop |
| regex   | 'z9_.-' * 51   | 2.67 usec per loop  |

So not much of a difference between set() and the current approach, but the regex version is faster for the worse case scenario.

I'll switch it over to a regex, and just include the regex in the error msg.

jamesls:
@mtdowling Feedback incorporated.  I think the error message is still pretty reasonable:

```
session.create_client('s3').list_objects(Bucket='nota/valid/bucket')
Invalid bucket name "nota/valid/bucket": Bucket name must match the regex "^[a-zA-Z0-9.\-_]{1,255}$"
```

mtdowling:
:ship:

PR #641, opened 2015-09-02
Remove GetLogEvents paginator

kyleknap:
This paginator returns the same token twice to indicate pagination is
done. Botocore treats this as an error, but really we should stop paginating.
So remove this paginator for now until support is added in botocore.


kyleknap:
Fixes https://github.com/aws/aws-cli/issues/1485

jamesls:
:shipit:

PR #642, opened 2015-09-03
Some moderate refactoring for boto3 docstrings

kyleknap:
Includes extracting out the logic that creates the client classes into a public function. This is needed because resource factories in boto3 do not have access to the client's class, and thus its name, when creating the resource class.

There was also some moving around some of the internals of the docstring class to make it easier to inherit from for example:
```py
class ActionDocstring(LazyLoadedDocstring):
    def __init__(self, *args, **kwargs):
        super(ActionDocstring, self).__init__(*args, **kwargs)
        self._docstring_writer = document_action
```

### DIFF
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -26,6 +26,7 @@ from botocore.model import ServiceModel
 from botocore.paginate import Paginator
 from botocore.signers import RequestSigner
 from botocore.utils import CachedProperty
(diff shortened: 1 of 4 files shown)


mtdowling:
:ship:

rayluo:
:ship:

PR #643, opened 2015-09-03
Waiter docstrings

kyleknap:
This adds docstrings to waiter's ``wait`` method. It also renames the waiter class returned from client to a class that falls in line with what is in the documentation e.g. ``S3.Waiter.bucket_exists``.
```py

In [1]: import botocore

In [2]: session = botocore.session.get_session()

In [3]: client = session.create_client('s3')

In [4]: waiter = client.get_waiter('bucket_exists')

In [5]: waiter.__class__.__name__
Out[5]: 'S3.Waiter.bucket_exists'

In [6]: help(waiter.wait)
```

Note that it includes: https://github.com/boto/botocore/pull/642 so that can be ignored in the diff
I will update this PR once it gets merged so that it will get excluded.


botocore/docs/docstring.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        The args and kwargs are the same as the underlying document
+        generation function. These just get proxied to the underlying
+        function.
+        """
+        super(LazyLoadedDocstring, self).__init__()
+        self._gen_args = args
+        self._gen_kwargs = kwargs
+        self._docstring = None
+        self._docstring_writer = self._raise_docstring_not_implemented
```
mtdowling: Why could you not just implement the `def _docstring_writer` method directly instead of using this layer of indirection though `_raise_docstring_not_implemented`?

botocore/docs/waiter.py
```diff
@@ -50,26 +51,51 @@ def _add_single_waiter(self, section, waiter_name):
...
+
+    :param include_signature: Whether or not to include the signature.
+        It is useful for generating docstrings.
+    """
+    waiter_model = service_waiter_model.get_waiter(waiter_name)
+    operation_model = service_model.operation_model(
+        waiter_model.operation)
+
+    wait_description = (
+        'This polls :py:meth:`{0}.Client.{1}` every {2} '
```
mtdowling: Can you either remove "This" or change this to "This waiter"?

botocore/waiter.py
```diff
@@ -43,7 +45,32 @@ def create_waiter_with_client(waiter_name, waiter_model, client):
...
+        event_emitter=client.meta.events,
+        service_model=client.meta.service_model,
+        service_waiter_model=waiter_model,
+        include_signature=False
+    )
+
+    # Rename the waiter class based on the type of waiter.
+    waiter_class_name = str('%s.Waiter.%s' % (
+        get_service_module_name(client.meta.service_model),
+        xform_name(waiter_name)))
```
jamesls: Remind me again why we document the class name as snake_cased instead of CamelCase (which is what I'd expect)?

botocore/docs/docstring.py
jamesls: I agree.  I think it's just more straightforward to have each subclass override self._docstring_writer and call out to the appropriate `document_*` function.

botocore/docs/docstring.py
```diff
@@ -0,0 +1,93 @@
...
+
+        So this is a big hack in order to get lazy loaded docstring work
+        for the ``help()``. In the ``help()`` function, ``pydoc`` and
+        ``inspect`` are used. At some point the ``inspect.cleandoc``
+        method is called. To clean the docs ``expandtabs`` is called
+        and that is where we override the method to generate and return the
+        docstrings.
+        """
+        if self._docstring is None:
+            self._generate().expandtabs(tabsize)
```
jamesls: Is the `.expantabs()` call after `_generate()` needed?  It doesn't seem like it.

botocore/waiter.py
```diff
@@ -43,7 +45,32 @@ def create_waiter_with_client(waiter_name, waiter_model, client):
...
+
+    # Create the new waiter class
+    documented_waiter_cls = type(
```
jamesls: This will create a new class object every time `client.get_waiter` is called.  Can we avoid creating new class objects?  Are we doing this only so the class name will be different?

jamesls:
Just curious, what are the perf differences with this change?

Also, I think it would be worth adding a resource leak test to ensure that we don't leak memory when creating many waiters in a loop.  Especially now that we're generating docs/classes on demand.

botocore/docs/docstring.py
kyleknap: We can remove that

botocore/waiter.py
kyleknap: It is because a client names can be confusing. Such as ``ds`` for ``DirectoryService``. Affects discoverablility. Also it inconsistent from the other SDKs.

kyleknap: Wait do you mean ``directory_service`` instead of ``DirectoryService``? Or did you mean the client names?

kyleknap: We talked about this. I am good with making the waiter class names CamelCase

botocore/waiter.py
kyleknap: It is mainly to override the docstring. You cannot do it on the method of an instantiated client. The class name change is only because why not since I am already doing it for the docstring.

botocore/docs/waiter.py
kyleknap: I will remove ``This``

botocore/docs/docstring.py
kyleknap: I am a little lost here. Comments seem to differ.

@mtdowling I set the ``_raise_docstring_not_implemented`` method to be informative when developing or building subclasses on top of this base class. I could always set ``_docstring_writer`` to ``None`` and that will lead to a ``None`` object is not callable error. I just want the class to fail to generate documentation if the subclass does not specify a ``_docstring_writer`` to help with development and avoiding bugs as there are going to be a fair amount of Docstring class that will subclass from it in the future. What would you suggest?

@jamesls
It sounds like I am doing what you suggest. For example take the [ClientMethodDocstring](https://github.com/boto/botocore/pull/643/files#diff-5989425b25925188aeb4541546e0e61bR84) class. How does it differ from what you are suggesting?

kyleknap:
Good point on the performance test. I will do some testing. And add some tests to perf test suite no matter what I find.

kyleknap:
Here are some initial results based on speed. Using this script:
```py
import time
import botocore.session

ec2 = session.create_client('ec2')

total = 0
iterations = 5
for i in range(iterations):
    start_time = time.time()
    ec2.get_waiter('instance_running')
    end_time = time.time()
    diff = end_time - start_time
    print('%s round instantiation: %s seconds' % (i, diff))
    total += diff

average = total/iterations
print('Average %f seconds' % average)
```

Here are the results of the develop branch:
```
0 round instantiation: 0.00259804725647 seconds
1 round instantiation: 6.00814819336e-05 seconds
2 round instantiation: 4.38690185547e-05 seconds
3 round instantiation: 4.29153442383e-05 seconds
4 round instantiation: 4.19616699219e-05 seconds
Average 0.000557 seconds
```

Here are the results of this PR:
```
0 round instantiation: 0.00261998176575 seconds
1 round instantiation: 0.000185012817383 seconds
2 round instantiation: 8.58306884766e-05 seconds
3 round instantiation: 8.58306884766e-05 seconds
4 round instantiation: 7.9870223999e-05 seconds
Average 0.000611 seconds
```

So it is a little slower, but not substantial due to relative change is on the scale of 10ths or 100ths of milliseconds. Next, I am going to test for memory leaks.

kyleknap:
@jamesls 
So I ran the two resource leak tests that I added and printed out the memory leak comparisons. Note that I did this for ec2 InstanceRunning waiter, but here are my results.

I ran each test five times and took the average difference between the end memory usage and the beginning memory usage (i.e. ``end - start``):

develop ``test_create_single_waiter_memory_constant``: 8.35 MB
develop ``test_create_memory_waiters_in_loop``: 3.53 MB

PR ``test_create_single_waiter_memory_constant``: 8.50 MB
PR ``test_create_memory_waiters_in_loop``: 4.10 MB

So it looks like there was a hit but it does not look really large. Not sure warrants caching of classes. Thoughts?

botocore/docs/docstring.py
jamesls: What I (and I believe Michael) are both saying is that you don't need the level of indirection where an instance variable has a reference to a function to call, especially because these *aren't* dynamic.  That is, each  docstring writer is tied to a specific subclass.  So you can just have:

```
class BaseClass:
  def _docstring_writer(self, *args, **kwargs):
      raise NotImplementedError("Subclass must implement _docstring_writer.")

class ClientMethodDocString(BaseClass):
  def _docstring_writer(self, ...):
    document_model_drive_method(...)
```

jamesls:
Yeah, I'm, convinced now.  We can hold off on the caching.

I replied to your other comment about the implementation of the template method.  I don't feel strongly about it, but I still think it's simpler to add a new abstract method and have each subclass implement it.

botocore/docs/docstring.py
kyleknap: Ahh that makes it much cleaner.

kyleknap:
@jamesls @mtdowling I think I addressed all of the feedback. It should be good to be looked at again.

jamesls:
:shipit: Looks good.

PR #646, opened 2015-09-10
Make the StreamingBody wrapper more file-like with a close() method

dstufft:
This will solve pypa/warehouse#676 which is essentially caused by trying to treat the StreamingBody like a file-like object.

### DIFF
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -83,6 +83,9 @@ class StreamingBody(object):
                 actual_bytes=self._amount_read,
                 expected_bytes=int(self._content_length))
 
+    def close(self):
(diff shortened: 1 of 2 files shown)


rayluo:
Sounds good. Can you also provide test case for it? Thanks!

dstufft:
Done!

jamesls:
:shipit:  Thanks!

dstufft:
Awesome, how long do you think it'll be until this is released? Not sure what your schedule looks like.

rayluo:
It will be release later today.

dstufft:
Thanks!

PR #647, opened 2015-09-10
Added error code and message to unexpected WaiterError

JordonPhillips:
Fixes boto/boto3#233

In that specific case, the result now looks like:

    WaiterError: Waiter InstanceStopped failed: Request would have succeeded, but DryRun flag is set.

### DIFF
--- a/botocore/waiter.py
+++ b/botocore/waiter.py
@@ -272,8 +272,9 @@ class Waiter(object):
                 if 'Error' in response:
                     # Transition to the failure state, which we can
                     # just handle here by raising an exception.
-                    raise WaiterError(name=self.name,
-                                      reason='Unexpected error encountered.')
+                    raise WaiterError(
+                        name=self.name,
+                        reason=response['Error'].get('Message', 'Unknown'))
             if current_state == 'success':
                 logger.debug("Waiting complete, waiter matched the "
                              "success state.")
--- a/tests/unit/test_waiters.py
+++ b/tests/unit/test_waiters.py
@@ -342,6 +342,26 @@ class TestWaitersObjects(unittest.TestCase):
         with self.assertRaises(WaiterError):
             waiter.wait()
 
+    def test_unspecified_errors_propagate_error_code(self):
+        # If a waiter receives an error response, then the
+        # waiter should pass along the error code
+        config = self.create_waiter_config()
+        operation_method = mock.Mock()
+        error_code = 'error_message'
+        error_message = 'error_message'
+        self.client_responses_are(
+            # This is an unknown error that's not called out
+            # in any of the waiter config, so when the
+            # waiter encounters this response it will transition
+            # to the failure state.
+            {'Error': {'Code': error_code, 'Message': error_message}},
+            for_operation=operation_method
+        )
+        waiter = Waiter('MyWaiter', config, operation_method)
+
+        with self.assertRaisesRegexp(WaiterError, error_message):
+            waiter.wait()
+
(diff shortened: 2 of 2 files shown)


botocore/waiter.py
```diff
@@ -272,8 +272,12 @@ def wait(self, **kwargs):
                 if 'Error' in response:
                     # Transition to the failure state, which we can
                     # just handle here by raising an exception.
+                    reason = 'Unexpected error encountered'
+                    if 'Code' in response['Error']:
+                        reason += ': ' + response['Error']['Code']
+
                     raise WaiterError(name=self.name,
-                                      reason='Unexpected error encountered.')
+                                      reason=reason)
```
jamesls: We should probably include the error message.  As your code suggests, there may not always be a code, but there's always a message.  Maybe have it be `"An error {code} occurred: {message}"` where code can get `response['Error'].get('Code', 'Unknown')`

jamesls:
Looks good, but what do you think about incorporating the code somewhere in the error message in exceptions.py?

```
    fmt = 'Waiter {name} failed: An error {code} occured: {reason}'
```

or something along those lines?

JordonPhillips:
Most WaiterErrors don't have a code associated, so unless we want to add some in then it wouldn't make sense to change the string to favor a single error.

jamesls:
Ok.  The only other option would be to subclass from WaiterError to have something specific for this scenario (that included a code in its error message format), which I'm ok with not doing for now.

:shipit:

JordonPhillips:
The more I think about it, the more I think the subclassing might be a better idea. With what I have now, the customer would have to call `e.kwargs.get('code')` which doesn't seem like a great interface. If I subclass then I'll be able to add the response as an attribute, so something like this:

```python
class WaiterUnknownError(WaiterError):
    """Waiter failed for an unknown reason"""
    fmt = 'Waiter {name} failed: An error {code} ocurred: {message}'

    def __init__(self, response, **kwargs):
        super(WaiterUnknownError, self).__init__(**kwargs)
        self.response = response
```

Though [I'm not sure that's the best name for it](https://twitter.com/codinghorror/status/506010907021828096)

kyleknap:
@JordonPhillips I would not worry too much about saving the response in the error. The rest of the exceptions other than a ClientError does not have the response. I think the point of this PR was to make the waiter error message clear and not have a bunch of vague messaging around it.

Given that we do not even use the code in the message to the error, I would be in favor of removing the code argument in the WaiterError's constructor and test that the error thrown simply includes the more detailed message.

tests/unit/test_waiters.py
```diff
@@ -342,6 +342,34 @@ def test_unspecified_errors_stops_waiter(self):
...
+            # to the failure state.
+            {'Error': {'Code': error_code, 'Message': error_message}},
+            for_operation=operation_method
+        )
+        waiter = Waiter('MyWaiter', config, operation_method)
+
+        exception = None
+        try:
+            waiter.wait()
+        except WaiterError as e:
```
kyleknap: Oh for this type of logic, I like using [assertRaisesRegexp](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaisesRegexp). It makes the assertion much shorter and nicer.

kyleknap:
Awesome. Looks better. Just one last small comment on your test assertion.

JordonPhillips:
:+1:

kyleknap:
LGTM :ship:

jamesls:
:shipit:

PR #648, opened 2015-09-10
Added validation to ensure that service names passed in conform to those exposed by the service model

### LINKED ISSUE #611: Make clients case sensitive
Following up from https://github.com/boto/boto3/issues/171, this should *not* be allowed:

```
>>> import botocore.session
>>> session.create_client('S3')
<botocore.client.S3 object at 0x10368e2b0>
```

Our reference docs has all the services in lower case, and all our examples use lower case, so we should enforce that this is what you specify.  It's also what you get back from `session.get_available_services()`.

JordonPhillips:
Fixes #611

### DIFF
--- a/botocore/loaders.py
+++ b/botocore/loaders.py
@@ -98,7 +98,7 @@ import os
 from botocore import BOTOCORE_ROOT
 from botocore.compat import json
 from botocore.compat import OrderedDict
-from botocore.exceptions import DataNotFoundError
+from botocore.exceptions import DataNotFoundError, ValidationError
 
 
 def instance_cache(func):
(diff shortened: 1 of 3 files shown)


jamesls:
I don't think we can really validate that all services are always lower cased, that's just our convention.  I think this is a bug in our loader, where it will incorrectly give you a service model for `iam` even though you asked for `iAm` or `IaM`.  The loader should only give you a service model if it's *exactly* the same name you asked for.

JordonPhillips:
That's a trickier view, though I understand the reasoning. I found a working solution is put a check in `list_api_versions` enforcing the case by calling `listdir` to make sure the case-correct directory exists. I have a feeling there's some edge cases that might not take care of even though this passes the tests.

jamesls:
:shipit:  Looks good.  Just remove the client.py comments.

kyleknap:
Looks good. :ship: You can merge it once the tests pass.

jamesls:
:shipit:

PR #649, opened 2015-09-15
Disable signing for 3 more public apis

rayluo:


rayluo:
The Python 2.6 test fails.  I will need to install Python 2.6 on my laptop to test it again tomorrow. Other than that, I believe the code itself should be straightforward.

mtdowling:
Looks good.

Something that would be cool in the future would be a way to designate signers per/operation in our service descriptions, where a signer name could be "anonymous" to disable signing. We could then remove these customizations entirely in the future and just rely on a generic approach.

If/when that happens (something I really hope does happen), I could see all of the related unit tests getting removed. Perhaps it would be a good idea to also add functional tests for these operations that ensures they are not signed. That way we can ensure we don't regress here if we fix this at the model level.

JordonPhillips:
looks good

jamesls:
I agree with Michael.  I'd rather see functional/integration tests for this that actually verify the generated request is unsigned, rather than functional tests that verify emits are emitted properly (which we already have tests for in a more general manner).

tests/unit/test_handlers.py
```diff
@@ -562,3 +562,34 @@ def test_s3_special_case_is_before_other_retry(self):
...
+
+def _test_disable_signing_is_triggered(event):
+    responses = botocore.session.Session().emit(event)
+    if not any(handler==handlers.disable_signing for handler, _ in responses):
+        raise AssertionError("Need disable_signing() for %s" % event)
+
+def test_disable_signing_for_public_apis():
+    public_apis = []
+
+    # See http://docs.aws.amazon.com/ja_jp/cognitoidentity/latest/APIReference/API_Operations.html
```
quiver: minor nit.
In the url, `ja_jp` part is not needed.

tests/unit/test_handlers.py
```diff
@@ -562,3 +562,34 @@ def test_s3_special_case_is_before_other_retry(self):
...
+        'GetCredentialsForIdentity',
+        'GetId',
+        'GetOpenIdToken',
+        'UnlinkIdentity',
+        ]
+    public_apis.extend([
+        'choose-signer.cognito-identity.{0}'.format(action)
+        for action in cognito_identity_public_apis])
+
+    # See http://docs.aws.amazon.com/ja_jp/STS/latest/APIReference/API_Operations.html
```
quiver: minor nit.
In the url, `ja_jp` part is not needed.

tests/functional/test_public_apis.py
```diff
@@ -0,0 +1,62 @@
...
+def _test_public_apis_will_not_be_signed(func, kwargs):
+    with mock.patch('botocore.endpoint.Session.send') as _send:
+        _send.side_effect = EarlyExit("we don't care about response here")
+        try:
+            func(**kwargs)
+        except EarlyExit:
+            pass
+        except NoCredentialsError:
+            assert False, "NoCredentialsError should not be triggered"
+        request = _send.call_args[0][0]
```
jamesls: pep8 casing please.

tests/functional/test_public_apis.py
```diff
@@ -0,0 +1,62 @@
...
+        SigV2_disabled = 'SignatureVersion=2' not in request.url
+        SigV3_disabled = 'X-Amzn-Authorization' not in request.headers
+        SigV4_disabled = 'Authorization' not in request.headers
+        assert SigV2_disabled and SigV3_disabled and SigV4_disabled
```
jamesls: I think it would be helpful to know which specific case failed.  I'd prefer having an assertion for each of the above cases.

tests/functional/test_public_apis.py
```diff
@@ -0,0 +1,62 @@
...
+
+def test_cognito_identity_public_apis():
+    session = Session()
+    session.get_credentials = mock.Mock(return_value=None)
+    client = session.create_client('cognito-identity', 'us-east-1')
+    tasks = {
```
jamesls: Rather than have duplicated for each service, why not use the same approach we do for our smoke tests: https://github.com/boto/botocore/blob/develop/tests/integration/test_smoke.py#L18

I'm not really seeing a strong reason why this should deviate from that approach.  I'd prefer to just have one way of doing these types of tests (given a list of service/operation/params, do something and make some assertions about each combo).

tests/unit/test_handlers.py
```diff
@@ -562,3 +562,34 @@ def test_s3_special_case_is_before_other_retry(self):
...
                         "S3 200 error handler was supposed to be before "
                         "the general retry handler, but it was not.")
+
+
+def _test_disable_signing_is_triggered(event):
+    responses = botocore.session.Session().emit(event)
+    if not any(handler==handlers.disable_signing for handler, _ in responses):
+        raise AssertionError("Need disable_signing() for %s" % event)
+
+def test_disable_signing_for_public_apis():
```
jamesls: Let's remove this.  These don't really add any value IMO.  This is really testing he event emitter, which we already have tests for.

tests/functional/test_public_apis.py
kyleknap: +1 specifically wording of what it was signed with would be awesome.

tests/functional/test_public_apis.py
rayluo: Sorry I ran pep8 and flake8 just now, and they suggest to add one more blank line between each function.  Other than that, there is no warning.  @jamesls Do you see something else that we need to fix?

Meanwhile, I'll work on other comments in this PR. Thanks!

jamesls: Sounds like this won't matter because you're going to be adding assertions for each case so you won't need to assign them to vars, but in general, python variable names are always snake_cased

tests/functional/test_public_apis.py
rayluo: Will be changed into:

        SigV2_disabled = 'SignatureVersion=2' not in request.url
        assert SigV2_disabled, "SigV2 is incorrectly enabled"
        SigV3_disabled = 'X-Amzn-Authorization' not in request.headers
        assert SigV3_disabled, "SigV3 is incorrectly enabled"
        SigV4_disabled = 'Authorization' not in request.headers
        assert SigV4_disabled, "SigV4 is incorrectly enabled"

tests/functional/test_public_apis.py
rayluo: Understood!

jamesls:
:shipit: Thanks for incorporating the feedback.

kyleknap:
Same :ship:

PR #650, opened 2015-09-15
Tolerate missing code or message in error response

rayluo:
Since we don't have control to field names in upstream services, let's play safe by not assuming any field's existence.

PS: This is a successor of the closed PR https://github.com/boto/botocore/pull/610.
That one was originally developed to address another issue with similar symptom, but then we used another approach to solve the root cause for that one.
Now this one seems still useful as part of the fix for boto/boto3#249. (Similar fix can be found here too. borrows the concept in https://github.com/boto/botocore/pull/647)

### DIFF
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -296,8 +296,8 @@ class ClientError(Exception):
 
     def __init__(self, error_response, operation_name):
         msg = self.MSG_TEMPLATE.format(
-            error_code=error_response['Error']['Code'],
-            error_message=error_response['Error']['Message'],
+            error_code=error_response['Error'].get('Code', 'Unknown'),
+            error_message=error_response['Error'].get('Message', 'Unknown'),
             operation_name=operation_name)
         super(ClientError, self).__init__(msg)
         self.response = error_response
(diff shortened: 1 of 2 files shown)


mtdowling:
LGTM

JordonPhillips:
Looks good

botocore/exceptions.py
```diff
@@ -295,9 +295,11 @@ class ClientError(Exception):
         'operation: {error_message}')
 
     def __init__(self, error_response, operation_name):
+        code = error_response.get('Error', {}).get('Code', 'Unknown')
```
jamesls: It makes sense to handle the case where 'Code' can be None, but is there ever a case where `Error` will be None?  I don't think that' possible, this is a key we create in our response parser, not a value we get from the service.

rayluo: Good point. Will change back to error_response['Error'].get('Code', 'Unknown') in next PR. That way, shoud the 'Error' be missing in future, it deserves a lousy error trace. :)

jamesls:
:shipit:

PR #652, opened 2015-09-18
Account for Output not in response

jamesls:
This can happen in the case of an error.

Fixes boto/boto3#262.


kyleknap:
LGTM :ship:

PR #653, opened 2015-09-22
Remove hardcoded uris in endpoint heuristics

jamesls:
As a general principle, anything that has a startsWith
constraint should not have a hard coded region name.

Note that there's already existing tests for this: https://github.com/boto/botocore/blob/develop/tests/unit/test_regions.py#L27

### DIFF
--- a/botocore/data/_endpoints.json
+++ b/botocore/data/_endpoints.json
@@ -18,7 +18,7 @@
   ],
   "iam":[
     {
-      "uri":"https://{service}.cn-north-1.amazonaws.com.cn",
+      "uri":"https://{service}.{region}.amazonaws.com.cn",
       "constraints":[
         ["region", "startsWith", "cn-"]
       ]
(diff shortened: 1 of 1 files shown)


rayluo:
This rule of thumb makes sense.

Considering the current test_regions.py will not necessarily always cover future new regions, do we need to add a dedicate test case for the "{region} must exist if it has a startsWith rule" behavior?  For example, at this moment, there is another case violates this rule, which we may or may not need to fix. https://github.com/jamesls/botocore/blob/update-regions/botocore/data/_endpoints.json#L29

jamesls:
@rayluo I don't think it's necessary.  The rule you reference is not a hardcoded region name so it doesn't violate this guideline.  The actual region names are "us-gov-west-1": http://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html.

kyleknap:
That one needs to be hard coded: http://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html. If you put region there it will get mapped to iam.us-gov-west-1.amazonaws.com which is incorrect. Otherwise looks good :ship:

rayluo:
Thanks for the information. That one is different, it needs to be hard coded.

:ship:

PR #654, opened 2015-09-22
Added GLOBAL_ENDPOINT for private S3 endpoints

david-dever-23-box:
Exposed `global_endpoint` value within **fix_s3_host()** as `GLOBAL_ENDPOINT`, for use with private S3 clouds.
Sample usage with override for signature type follows (`HmacV1QueryAuth` signer for `s3-query` on `http://mydomain.com`):
```
import boto3
import botocore

# botocore / boto3 require explicit config and credentials files,
# in order to manually override the default values for signature version
AWS_CONFIG_FILE = '~/.aws/config'

HTTP_PROTOCOL = 'http'
ENDPOINT = 'mydomain.com'
SIGNATURE_TYPE = 's3-query'

# boto S3 session setup
connection_data = {
    'config_file': (None, 'AWS_CONFIG_FILE', None, None),
}
s3_session = botocore.session.get_session(connection_data)

# overrides values from AWS_CONFIG_FILE at '~/.aws/config'
config = botocore.client.Config(signature_version=SIGNATURE_TYPE)

# overrides GLOBAL_ENDPOINT value
botocore.utils.GLOBAL_ENDPOINT = ENDPOINT

# create the client
conn_client = s3_session.create_client(
    's3',
    endpoint_url=HTTP_PROTOCOL + '://' + ENDPOINT,
    config=config)
```
If DNS-compatible buckets are used, the calling format will be re-mapped to `subdomain` for all requests except `GET Service (List Buckets)` and `GET Bucket ?location`.

Make sure to run **aws configure** beforehand, which generates the `~/.aws/config` and `~/.aws/credentials` files required to override the default signature type (sample config shown):
```
[default]
s3 =
    signature_version = s3v4
```

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -40,6 +40,7 @@ RESTRICTED_REGIONS = [
     'us-gov-west-1',
     'fips-us-gov-west-1',
 ]
+GLOBAL_ENDPOINT = 's3.amazonaws.com'
 
 
 class _RetriesExceededError(Exception):
@@ -649,8 +650,7 @@ def fix_s3_host(request, signature_version, region_name, **kwargs):
                 if request.auth_path[-1] != '/':
                     request.auth_path += '/'
             path_parts.remove(bucket_name)
-            global_endpoint = 's3.amazonaws.com'
-            host = bucket_name + '.' + global_endpoint
+            host = bucket_name + '.' + GLOBAL_ENDPOINT
             new_tuple = (parts.scheme, host, '/'.join(path_parts),
                          parts.query, '')
             new_uri = urlunsplit(new_tuple)


smira:
+1

kyleknap:
Thanks for the pull request. This was something that we wanted to implement as well. I recently submitted a pull request that achieves the same functionality: https://github.com/boto/botocore/pull/655. 

Would this work for your use case? All you would have to do is specify a virtual host or possibly path format in the Config object that you use to instantiate the client or in the .aws/config file. The reason we went with this approach was that it would allow users to take advantage of the functionality in the AWS CLI as well.

Let me know what you think.

david-dever-23-box:
I'm not sure that the scopes are ultimately the same - this strictly addresses the need to expose the global endpoint, and not address (no pun intended) anything relating to calling format.

kyleknap:
So there is no default endpoint global. The functionality is that if you manually pick the addressing_style to say virtual, it will always use the endpoint_url you passed in to instantiating the clients to create the virtual style address. You can see that in the tests that I added where it uses ``foo.amazonaws.com`` as opposed to the current default ``s3.amazonaws.com``.

david-dever-23-box:
I would prefer to be able to inject a base endpoint, then manipulate calling format atop that. So far, it appears that your approach requires that one assemble and inject a complete URL - am I reading that intention correctly?

kyleknap:
Yeah so you can inject the endpoint url upon creation of the client, but you can also specify it in the ``~/.aws/config`` file. So if you set you're ``~/.aws/config`` file to something like this:
```
[default]
s3 =
    addressing_style = virtual
```
you will not have to modify any of your code assuming you specify an endpoint somewhere when you create the client. Or if you want to have the workflow that you initially wrote, you can modify it to get the same functionality:
```py
import boto3
import botocore

# botocore / boto3 require explicit config and credentials files,
# in order to manually override the default values for signature version
AWS_CONFIG_FILE = '~/.aws/config'

HTTP_PROTOCOL = 'http'
ENDPOINT = 'mydomain.com'
SIGNATURE_TYPE = 's3-query'

# boto S3 session setup
connection_data = {
    'config_file': (None, 'AWS_CONFIG_FILE', None, None),
}
s3_session = botocore.session.get_session(connection_data)

# overrides values from AWS_CONFIG_FILE at '~/.aws/config'
config = botocore.client.Config(
    s3={'addressing_style': 'virtual'},
    signature_version=SIGNATURE_TYPE)

# create the client
conn_client = s3_session.create_client(
    's3',
    endpoint_url=HTTP_PROTOCOL + '://' + ENDPOINT,
    config=config)
```
We also recently merged some logic into botocore so that we have a canonical source for endpoint resolution that we are really trying to stick with and are trying to avoid having too many edge cases. Given that you can make calls to a custom endpoint by setting the addressing style, I do not think we need to add the option override a default s3 endpoint.

PR #655, opened 2015-09-23
Add option to force s3 virtual host addressing

kyleknap:
You can force virtual style addressing through the config file:
```
s3 =
    addressing_style = virtual_host
```
You can also specify it in the Config object passed in when a client is instantiated. The virtual host addressing no longer defaults to the s3 global endpoint. Instead, it will use the endpoint url determined from the region or if the endpoint url that is specified upon instantiation of a client:
```py
s3 = session.create_client('s3', config=Config(s3_addressing_style='virtual'))
```
If you force the virtual style addressing and the bucket name on an operation you call is not DNS compatible it will throw an error. Another important note to emphasize is that if you force the virtual style addressing, it will use the endpoint of the region you specified so you may hit PermanentRedirect errors. Otherwise if it is not forced, the requests will still be sent to the the global s3 endpoint.

As for testing, the s3 integration tests pass for both botocore and the CLI. I even tried running the integration tests with virtual style forced and the botocore tests passed and most of the CLI test passed (their were a few tests that randomly generated '.' in their name which would cause an error to be thrown as expected because the name is not DNS compatible).

Also, thanks to testing it in this fashion I fixed some bugs where if virtual style is forced and it is sigv4, we would get signature errors. We did not run into these bugs before because we would turn off virtual style if sigv4 is being used.


jamesls:
Looking at this now I have some initial high level feedback before digging into this:
Given the scoped config is `{'s3': {'addressing_style': 'virtual_host'}}`, I think we should map as closely to that in the client config, so something like `Config(s3={'addressing_style': 'virtual_host'})`.  It also makes it a little easier to see all the s3 specific config options rather than having to iterate over `config.__dict__` and check if it starts with `s3_`.

kyleknap:
That sounds good. It would provide a precedence to prevent the Config file from becoming to bloated. Maybe name it to something more explicit like ``s3_options`` or ``s3_configuration``?

botocore/client.py
```diff
@@ -218,6 +221,28 @@ def _get_client_args(self, service_model, region_name, is_secure,
             config_kwargs.update(
                 connect_timeout=client_config.connect_timeout,
                 read_timeout=client_config.read_timeout)
+
+        # Determine if the user specified the use of virtual host addressing
+        # style for s3.
+        s3_addressing_style = 'default'
```
jamesls: Can you move this out to a separate method?  This method is getting pretty long.

botocore/client.py
```diff
@@ -554,10 +601,12 @@ class Config(object):
     def __init__(self, region_name=None, signature_version=None,
                  user_agent=None, user_agent_extra=None,
                  connect_timeout=DEFAULT_TIMEOUT,
-                 read_timeout=DEFAULT_TIMEOUT):
+                 read_timeout=DEFAULT_TIMEOUT,
+                 s3_addressing_style=None):
```
jamesls: We should document the valid values.  It's not clear what values we support.

I'd propose at least supporting these three:

* auto (the current behavior.  Try to use virtual host when we can, path style when we can't).
* virtual_host - bucket in the hostname
* path - bucket in the URI

botocore/client.py
```diff
@@ -307,9 +332,27 @@ def __init__(self, serializer, endpoint, response_parser,
         self.meta = ClientMeta(event_emitter, self._client_config,
                                endpoint.host, service_model,
                                self._PY_TO_OP_NAME)
+        self._validate_arguments()
+        self._register_handlers()
+
+    def _validate_arguments(self):
+        if self.meta.s3_addressing_style not in ['virtual', 'default']:
```
jamesls: I think the validation should live elsewhere.  What do you think about the Config owning the validation of the valid values?  It doesn't really make sense to create a Config() object with invalid values, so maybe the Config() object should be validating this when it's created.

botocore/client.py
```diff
@@ -307,9 +332,27 @@ def __init__(self, serializer, endpoint, response_parser,
...
         self.meta.events.register('request-created.%s' %
-                                  service_model.endpoint_prefix,
+                                  self.meta.service_model.endpoint_prefix,
                                   self._sign_request)
+        # If the virtual host addressing style is being forced,
+        # switch the default fix_s3_host handler for the more general
+        # switch_to_virtual_host_style handler that does not have opt out
+        # cases (other than throwing an error if the name is DNS incompatible)
+        if self.meta.s3_addressing_style == 'virtual':
+            self.meta.events.unregister('before-sign.s3', fix_s3_host)
```
jamesls: Does it make sense to just move the default registration out of handlers.py?

```
if we're virtual hosted:
   register switch_to_virtual_host_tyle
if we're auto:
   register fix_s3_host (maybe rename this now)
else:
   ..do nothing
```

botocore/client.py
```diff
@@ -537,6 +580,10 @@ def config(self):
     def method_to_api_mapping(self):
         return self._method_to_api_mapping
 
+    @property
+    def s3_addressing_style(self):
```
jamesls: Are we planning on exposing  every single config option going forward?  Why not use the config property here?

botocore/exceptions.py
```diff
@@ -305,3 +305,18 @@ def __init__(self, error_response, operation_name):
 
 class ImminentRemovalWarning(Warning):
     pass
+
+
+class InvalidDNSNameError(BotoCoreError):
+    """Error when virtual host path is forced on a non-DNS compatible bucket"""
+    fmt = (
+        'Bucket named {bucket_name} is not DNS compatible. Virtual '
+        'hosted-style access cannot be used.'
```
jamesls: Should we provide a pointer to not configuring your client to use virtual hosted buckets?

botocore/utils.py
```diff
@@ -621,6 +622,26 @@ def fix_s3_host(request, signature_version, region_name, **kwargs):
...
+    try:
+        switch_to_virtual_host_style(
+            request, signature_version, region_name, 's3.amazonaws.com')
+    except InvalidDNSNameError as e:
+        bucket_name = e.kwargs['bucket_name']
+        logger.debug('Not changing URI, bucket is not DNS compatible: %s',
+                     bucket_name)
+
+
+def switch_to_virtual_host_style(request, signature_version, region_name,
```
jamesls: Can you document the params?  Not sure what default_endpoint_url is suppose to be.

jamesls:
Looks good, just had some small feedback.

botocore/client.py
david-dever-23-box: or use `subdomain` vs `ordinary`

jamesls:
@kyleknap I would prefer s3 instead of s3_configuration.  I think the fact that it's in the `Config` object makes it clear that it's configuration data.

kyleknap:
Ok that's fair.

botocore/client.py
kyleknap: yeah that's fine

botocore/client.py
kyleknap: I think we should stick with virtual and path because it is consistent with the s3 docs. I fine with ``auto`` as well. Will need to do some more testing to make sure we can support path without signing errors.

botocore/client.py
kyleknap: I was thinking that as well. I was concerned that you can also supply it through the config file it would not hit this validation, but  we currently always create a new config object when creating a new client.

botocore/client.py
kyleknap: That's fine. It would help for supporting path as well

botocore/utils.py
kyleknap: yep

botocore/client.py
kyleknap: Ah forgot that we expose the config property. Yeah lets use that.

kyleknap:
@jamesls 
I updated based on feedback. I ended up adding support for ``'path'``. I also set it and ran it on the s3 CLI integration tests and all of those tests passed. So there should not be any auth issues.

botocore/client.py
```diff
@@ -164,6 +167,21 @@ def _determine_region_name(self, endpoint_config, region_name=None,
...
+    def _inject_s3_configuration(self, config_kwargs, scoped_config,
+                                 client_config):
+        s3_configuration = None
+
+        # Check the scoped config first
+        if scoped_config is not None:
+            s3_configuration = scoped_config.get('s3')
+
+        # Next the client config value takes precedence
+        if client_config is not None:
```
jamesls: That's an interesting point.  Should we maybe be merging the configs if a user has both scoped config and client config?  As we add more s3 options, it would mean that a customer would typically have to grab the s3 config from the session, merge their options on top, and then create a Config object.

jamesls:
Changes look good.  Just wondering what we should be doing about merging config.  It makes it slightly more complicated for the user once they have various S3 options in their config and they want to override only specific options, but I suppose it's not that hard for them to manually merge the s3 config dicts before creating a client.  What do you think?

kyleknap:
@jamesls 
We talked about this. Going with the merging of dictionaries approach. Should be good to look at again.

jamesls:
:shipit:

kyleknap:
This broke some CLI integration tests reverting.

PR #657, opened 2015-09-23
Add integration leak tests

jamesls:
Verifies:

* s3 uploads use O(1) memory
* making repeated AWS calls does not leak memory

This was suppose to be included with #638.


kyleknap:
Looks good. :ship:

PR #661, opened 2015-09-27
Empty response is now properly translated to list.

rayluo:
This fixes issue aws/aws-cli#1365

I also manually test it with this command line to make sure it solves the issue.

    aws rds download-db-log-file-portion --db-instance-identifier mydbinstance --log-file-name error/mysql-error-running.log --starting-token 1000

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -191,6 +191,8 @@ class PageIterator(object):
         # the truncated amount.
         starting_truncation = self._parse_starting_token()[1]
         all_data = primary_result_key.search(parsed)
+        if all_data is None:
+            all_data = []
         set_value_from_jmespath(
             parsed,
             primary_result_key.expression,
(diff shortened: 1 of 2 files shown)


kyleknap:
We talked about this in person. Looks like changes needed for this PR: https://github.com/boto/botocore/pull/662 is needed for this one to get the correct behavior.

rayluo:
On a second thought, this seems still necessary, even if we accept PR #662 's successor, PR #689.

botocore/paginate.py
```diff
@@ -191,6 +191,8 @@ def _handle_first_request(self, parsed, primary_result_key,
         # the truncated amount.
         starting_truncation = self._parse_starting_token()[1]
         all_data = primary_result_key.search(parsed)
+        if all_data is None:
+            all_data = []
```
jamesls: Where did we land on the default value for this?

In this scenario the result key actually evaluates to a string, so defaulting to an empty list does not seem like the correct behavior.

What's our plan to address this?

rayluo: Yes I already have a clear idea about what should be adjusted here in order to adapt to the upcoming string support.  I even have a prototype ready.  But it will only work after we merge PR #689 so ideally let's try to merge that one first.

rayluo:
Replaced by PR #700

PR #664, opened 2015-09-28
Sets PolicyDocument return type to dict in generated documentation.

JordonPhillips:
Addresses boto/boto3#229
Accepting the dict for put is beyond the scope of this commit.

### DIFF
--- a/tests/functional/docs/test_iam.py
+++ b/tests/functional/docs/test_iam.py
@@ -1,0 +1,49 @@
+# Copyright 2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from tests.functional.docs import BaseDocsFunctionalTest
+from botocore.docs.service import ServiceDocumenter
+
+
+class TestIAMDocs(BaseDocsFunctionalTest):
+    def test_changes_policy_doc_to_dict(self):
+        modified_methods = ['create_policy', 'get_role_policy']
+        service_contents = ServiceDocumenter('iam', self._session)\
+            .document_service()
+        for method_name in modified_methods:
+            method_contents = self.get_method_document_block(
+                method_name, service_contents)
+            self.assertNotIn(b"Document': 'string'", method_contents)
+            self.assertNotIn(b"Document='string", method_contents)
+            self.assertTrue(
+                (b"Document': {}" in method_contents) or
+                (b"Document={}" in method_contents))
+            self.assertNotIn(b"Document** *(string)", method_contents)
+            self.assertNotIn(b"Document: string", method_contents)
+            self.assertTrue(
+                (b"Document** *(dict)" in method_contents) or
+                (b"Document: dict" in method_contents))
+
+    def test_changes_policy_doc_list_to_dict(self):
+        modified_methods = ['get_context_keys_for_custom_policy',
+                            'simulate_custom_policy']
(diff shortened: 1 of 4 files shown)


kyleknap:
Let's get the tests to pass first before reviewing it. I am not sure about the root cause for the test error, but the test failure is a python3 thing where bytes types cannot be equal to string types.

JordonPhillips:
There were going to be some merging conflicts, so I rebased to fix them.

kyleknap:
This will fix the issue that was linked but I do not think it will fix all of the possible cases. Based on this [handler](https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L483), all members with the ``policyDocumentType`` shape will have the customization applied. To get the most complete fix, you will have to traverse the shape/document searching for ``policyDocumentType`` and replacing it where needed. Fortunately, I implemented the functionality in boto3: https://github.com/boto/boto3/blob/develop/boto3/docs/utils.py#L93. I would suggest looking to see if you can use the class in boto3 in botocore.

jamesls:
It would also be good to add a functional test here.  The doc internals have a tendency change so it would be great to have a high level functional test to ensure the docs are updated.  I would even suggest just picking a single operation and ensuring the return type is correct.

JordonPhillips:
Updated to use the more re-usable modification code from Boto3 and added a functional test.

tests/functional/docs/test_iam.py
```diff
@@ -0,0 +1,68 @@
...
+                            'list_instance_profiles_for_role',
+                            'list_policy_versions',
+                            'list_roles',
+                            'put_group_policy',
+                            'put_role_policy',
+                            'put_user_policy',
+                            'update_assume_role_policy']
+        service_contents = ServiceDocumenter('iam', self._session)\
+            .document_service()
+        for method_name in modified_methods:
```
kyleknap: It may be worth breaking this test out to rely on yield for when this test is ran. If this test ever fails, it will be hard to determine where exactly in the test it failed because it is just a big for loop. By using yield it will create a bunch of tests that will be easier to track down and if one operation fails, all of the operations will still be checked.

Otherwise, I would also be fine with just checking just two arbitrary operations (one for the input and one for the output). Checking all of the operations seem a little overkill because they all same the same logic especially since we now have unit test for the customization. Thoughts?

tests/functional/docs/test_iam.py
```diff
@@ -0,0 +1,68 @@
...
+        modified_methods = ['get_context_keys_for_custom_policy',
+                            'get_context_keys_for_principal_policy',
+                            'simulate_custom_policy',
+                            'simulate_principal_policy']
+        service_contents = ServiceDocumenter('iam', self._session)\
+            .document_service()
+        for method_name in modified_methods:
+            method_contents = self.get_method_document_block(
+                method_name, service_contents)
+            self.assertNotIn(
```
kyleknap: I would be hesitant to have newlines in the output. It may cause failures on windows because carriage returns are sometimes added to the end. I am not completely sure about that though. I would prefer using ``get_parameter_document_block`` and then ``assert_contains_lines_in_order`` and ``assert_not_contains_line``

tests/unit/docs/test_utils.py
```diff
@@ -219,3 +220,50 @@ def test_appends_documentation(self):
...
+        modifier.replace_documentation_for_matching_shape(
+            'docs.request-params', self.doc_structure)
+        self.assert_contains_line(self.type)
+        self.assert_not_contains_line(old_documentation)
+        self.assert_contains_line(self.description)
+
+    def test_modifies_example(self):
+        context = {'shape': self.name}
+        section = self.doc_structure.add_new_section(self.name, context)
+        old_example = 'gets_delted'
```
kyleknap: gets_delted --> gets_deleted

kyleknap:
Looks good. Just had a few comments about the tests.

tests/functional/docs/test_iam.py
JordonPhillips: Since we can't generate tests with `yield` under `unittest.TestCase`, I'll do the arbitrary two

kyleknap: Yeah I like the arbitrary two option.

JordonPhillips:
:+1:

JordonPhillips:
Closing while a more general solution is investigated

PR #666, opened 2015-09-28
Override python deps version in setup.cfg

jamesls:
Similar fix from: https://github.com/boto/boto3/pull/282


kyleknap:
:ship:

PR #667, opened 2015-09-29
Update serializers/parsers with latest protocol tests

jamesls:
This fixed three issues with protocol serialization/parsing:

1. If a struct is returned as `null` across the wire in `json` we should parse that as `None`, not `{}`.
2. If we're serializing to a query string and the value is a map we need to ignore the top level key and pull the map keys up to top level query string params.
3. If a query string *value* is a sequence, we need to serialize it as: `{"a": ["b", "c"]}` -> `a=b&a=c`.
4. Honor locationName in query protocol for non flattened lists.

See commit messages for more details.

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -465,6 +465,11 @@ class BaseJSONParser(ResponseParser):
 
     def _handle_structure(self, shape, value):
         member_shapes = shape.members
+        if value is None:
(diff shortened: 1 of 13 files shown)


kyleknap:
:ship:

JordonPhillips:
:ship:

PR #668, opened 2015-09-30
Update service-2.json to be clearer about nextToken element

jonathanwcrane:
Update more visible part of documentation to make it clearer (and less likely to miss this tidbit).

### DIFF
--- a/botocore/data/logs/2014-03-28/service-2.json
+++ b/botocore/data/logs/2014-03-28/service-2.json
@@ -1350,7 +1350,7 @@
         },
         "nextToken":{
           "shape":"NextToken",
-          "documentation":"<p>A pagination token obtained from a <code class=\"code\">FilterLogEvents</code> response to continue paginating the FilterLogEvents results.</p>"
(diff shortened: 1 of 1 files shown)


rayluo:
A side note. This PR is the effort from @jonathanwcrane to address the issue boto/boto3#281

We will figure out how to incorporate this request.

jonathanwcrane:
Thanks, @rayluo

jamesls:
Sorry about the confusion.  As per our [contributing guidelines](https://github.com/boto/botocore/blob/develop/CONTRIBUTING.rst) we aren't able to accept pull requests to these JSON files as they are auto generated from an upstream source.  There would get lost the next time we sync with that source.  I will follow up with the doc team to the source updated (http://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_FilterLogEvents.html#CWL-FilterLogEvents-request-NextToken).

jonathanwcrane:
Interesting.  So there's an upstream from the upstream (this is upstream from boto3), but that upstream is not open source?

jamesls:
There is an upstream source that the SDKs pull from where this change will need to be made.  I've followed up with the necessary team to get this documentation updated.  I'll update this once we pull the new JSON model.

jonathanwcrane:
Thanks @jamesls

PR #670, opened 2015-10-01
Run all services if AWS_SMOKE_TEST_SERVICES not provided

jamesls:
This fixes a bug where we'd default to a set of an empty
string if the AWS_SMOKE_TEST_SERVICES was not provided.

Verified all smoke tests are run and AWS_SMOKE_TEST_SERVICES
still works.


kyleknap:
:ship:

PR #672, opened 2015-10-02
Fix circular reference in docs

kyleknap:
Removed dependency of session from docs package

For example, you no longer get a circular reference from doing this:
```py
```


jamesls:
:shipit:

PR #673, opened 2015-10-02
Unrevert s3 config

kyleknap:
This puts back in place the s3 configuration options that I reverted because CLI integration tests were failing. Fixed the tests by putting back the ``fix_s3_host`` registering to the default section so when the event for unregistering is called in the CLI with ``--endpoint-url``, the registered handler does exist so it does get unregistered. This seemed like the safest way to ensure there was no breaking behavior in the CLI. 

Looked into trying to modify the session in the CLI, but that would have required making the session more mutable and adding more functionality to the session, which is not necessarily desired. 

cc @jamels @mtdowling @rayluo @JordonPhillips


botocore/client.py
```diff
@@ -541,23 +593,60 @@ def method_to_api_mapping(self):
...
+    :type read_timeout: int
+    :param read_timeout: The time in seconds till a timeout exception is
+        thrown when attempting to read from a connection.
+
+    :type s3: dict
+    :param s3: A dictionary of s3 specific configurations.
+        Valid keys are:
+            * 'addressing_style' -- Refers to the style in which to address
+              s3 endpoints. Values must be a string that equals:
+                  * auto -- Addressing style is chosen for user. Depending
```
jamesls: It would be worth mentioning this is the default value.

botocore/utils.py
```diff
@@ -636,10 +664,17 @@ def fix_s3_host(request, signature_version, region_name, **kwargs):
...
+
     if len(path_parts) > 1:
         bucket_name = path_parts[1]
+        if not bucket_name:
+            # If the bucket name is empty we should not be checking for
+            # dns compatibility.
+            return
         logger.debug('Checking for DNS compatible bucket for: %s',
                      request.url)
         if check_dns_name(bucket_name) and _allowed_region(region_name):
```
jamesls: Do we still want the _allowed_region() call?  I thought the idea was that if you specified virtual host addressing style we do it unless it's impossible.

jamesls:
Just had some small feedback.

botocore/client.py
kyleknap: Yep

botocore/utils.py
kyleknap: I did it for the sake of trying to avoid breaking any current behavior. I believe the only restricted regions is US GovCloud. There is some mutations that happen prior to hitting this if statement (mainly setting the request.auth_path) in the previous logic. I will need to be careful not to cause any regressions if we want to make the logic completely independent of the region.

jamesls: This is an issue because we're trying to share code between the existing "auto" behavior and "virtual" right?  Because to be clear, I'm talking specifically about the new behavior for "virtual".  So if i have addressing=virtual and a --region of us-gov-west-1, I would expect a DNS compatible bucket to go to:  `bucket.s3-us-gov-west-1.amazonaws.com`.  If the addressing mode is "auto" I'd expect no change in the current behavior.

kyleknap: Exactly. It is because we are trying to share as much logic as we can. The behavior you outlined is correct. I think I found a way to get the appropriate behavior without having to worry about any potential regressions (specifically with respect to signing).

kyleknap: I just tested it out. There should not be any issues, and it makes sense because when path style addressing is enabled the request.auth_path is never set so there should be no issues here. I am going to hoist the ``_allowed_region`` call to ``fix_s3()``

kyleknap:
@jamels I updated based on feedback

jamesls:
@kyleknap Can we get a test added for https://github.com/kyleknap/botocore/commit/4c043070a9c7af7e057e6da43ece9641113fa9c8?

kyleknap:
@jamesls We already have one: https://github.com/boto/botocore/blob/develop/tests/unit/test_s3_addressing.py#L87. Or did you want a test for if it us-gov-west-1 is used but a path style like DNS is used?

jamesls:
I didn't see a case for the scenario we've been discussing:

>  So if i have addressing=virtual and a --region of us-gov-west-1, I would expect a DNS compatible bucket to go to: bucket.s3-us-gov-west-1.amazonaws.com.

Is there a test for this?

kyleknap:
Oh no I can add one for that then.

kyleknap:
@jamesls Test case added.

jamesls:
:shipit:  Looks good.

PR #675, opened 2015-10-05
Add console-output-available waiter

jamesls:
The "aws ec2 get-console-output" does not immediately
return the console output when you launch a new
EC2 instance.  This adds a waiter so you can block
until this data is available.

This is similar to the PasswordDataAvailable, with the
exception that in this API call, the "Output" element
isn't in the response when there is no output.  So
the JMESPath expression needed to be "Output || ''" to avoid
type errors when calling length().


kyleknap:
:ship:

PR #677, opened 2015-10-08
Make error parsing more robust

kyleknap:
For the rest-xml error parser, the body of the response had headers of a
200 response even though the http status code was 500. This blew up
of the error parser because it was invalid xml. For these situations, we should
at least respect the error code.

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -674,9 +674,14 @@ class RestXMLParser(BaseRestParser, BaseXMLResponseParser):
         #   <RequestId>request-id</RequestId>
         # </ErrorResponse>
         if response['body']:
-            return self._parse_error_from_body(response)
-        else:
-            return self._parse_error_from_http_status(response)
+            # If the body ends up being invalid xml, the xml parser should not
+            # blow up. It should at least try to pull information about the
+            # the error response from other sources like the HTTP status code.
+            try:
+                return self._parse_error_from_body(response)
+            except ResponseParserError:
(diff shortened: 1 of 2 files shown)


JordonPhillips:
LGTM
:ship:

rayluo:
:ship:

botocore/parsers.py
```diff
@@ -674,9 +674,14 @@ def _do_error_parse(self, response, shape):
...
-            return self._parse_error_from_body(response)
-        else:
-            return self._parse_error_from_http_status(response)
+            # If the body ends up being invalid xml, the xml parser should not
+            # blow up. It should at least try to pull information about the
+            # the error response from other sources like the HTTP status code.
+            try:
+                return self._parse_error_from_body(response)
+            except ResponseParserError:
+                pass
```
jamesls: We should probably debug log this when it happens (with the traceback included).

kyleknap:
@jamesls 
I agree with adding a debug log for the caught exception. Updated the PR. Ready to be looked at again.

kyleknap:
@jamesls Updated debug log statement.

mtdowling:
:shipit: this is similar to what some of the other SDKs are doing when parsing errors.

jamesls:
:shipit:

kyleknap:
I updated the comment in the test as well to reflect the unknown cause of the issue and the fact that it may be fixed in the future.

PR #679, opened 2015-10-09
Use utf8 encoding to open model file

rayluo:
We encountered a test case failure on one of the test machine, when running Python 3.x.

      File "C:\Python33\Lib\unittest\case.py", line 385, in _executeTestPart
        function()
      File "C:\Jenkins\shiningpanda\jobs\263e0464\virtualenvs\d41d8cd9\lib\site-packages\nose\failure.py", line 38, in runTest
        raise self.exc_val.with_traceback(self.tb)
      File "C:\Jenkins\shiningpanda\jobs\263e0464\virtualenvs\d41d8cd9\lib\site-packages\nose\loader.py", line 254, in generate
        for test in g():
      File "C:\Jenkins\workspace\botocore-integ\PYTHON\CPython-3.3\tests\integration\test_utils.py", line 33, in test_can_generate_all_inputs
        service_model = session.get_service_model(service_name)
      File "C:\Jenkins\workspace\botocore-integ\PYTHON\CPython-3.3\botocore\session.py", line 469, in get_service_model
        service_description = self.get_service_data(service_name, api_version)
      File "C:\Jenkins\workspace\botocore-integ\PYTHON\CPython-3.3\botocore\session.py", line 492, in get_service_data
        api_version=api_version
      File "C:\Jenkins\workspace\botocore-integ\PYTHON\CPython-3.3\botocore\loaders.py", line 119, in _wrapper
        data = func(self, *args, **kwargs)
      File "C:\Jenkins\workspace\botocore-integ\PYTHON\CPython-3.3\botocore\loaders.py", line 343, in load_service_model
        return self.load_data(full_path)
      File "C:\Jenkins\workspace\botocore-integ\PYTHON\CPython-3.3\botocore\loaders.py", line 119, in _wrapper
        data = func(self, *args, **kwargs)
      File "C:\Jenkins\workspace\botocore-integ\PYTHON\CPython-3.3\botocore\loaders.py", line 363, in load_data
      File "C:\Jenkins\workspace\botocore-integ\PYTHON\CPython-3.3\botocore\loaders.py", line 157, in load_file
        return json.load(fp, object_pairs_hook=OrderedDict)
      File "C:\Python33\Lib\json\__init__.py", line 261, in load
        return loads(fp.read(),
      File "C:\Jenkins\shiningpanda\jobs\263e0464\virtualenvs\d41d8cd9\lib\encodings\cp1252.py", line 23, in decode
        return codecs.charmap_decode(input,self.errors,decoding_table)[0]
    UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 78208: character maps to <undefined>

It is because some non-ascii characters in one of our latest documentation model file can not be decoded by the locale encoding on our test machine. Note that you may or may not observe similar symptom on your own machine because your locale may be different.

This fix specifies utf8 encoding.

### DIFF
--- a/botocore/loaders.py
+++ b/botocore/loaders.py
@@ -94,6 +94,7 @@ version to use.
 
 """
 import os
+import io
 
 from botocore import BOTOCORE_ROOT
 from botocore.compat import json
@@ -153,7 +154,10 @@ class JSONFileLoader(object):
         full_path = file_path + '.json'
         if not os.path.isfile(full_path):
             return
-        with open(full_path) as fp:
+
+        # By default the file will be opened with locale encoding on Python 3.
+        # We specify "utf8" here to ensure the correct behavior.
+        with io.open(full_path, encoding='utf8') as fp:
             return json.load(fp, object_pairs_hook=OrderedDict)
 
 
--- a/botocore/loaders.py
+++ b/botocore/loaders.py
@@ -94,7 +94,6 @@ version to use.
 
 """
 import os
-import io
 
 from botocore import BOTOCORE_ROOT
 from botocore.compat import json
@@ -157,8 +156,9 @@ class JSONFileLoader(object):
 
         # By default the file will be opened with locale encoding on Python 3.
         # We specify "utf8" here to ensure the correct behavior.
-        with io.open(full_path, encoding='utf8') as fp:
-            return json.load(fp, object_pairs_hook=OrderedDict)
+        with open(full_path, 'rb') as fp:
+            payload = fp.read().decode('utf-8')
+            return json.loads(payload, object_pairs_hook=OrderedDict)
 
 
 def create_loader(search_path_string=None):
--- a/tests/unit/data/non_ascii.json
+++ b/tests/unit/data/non_ascii.json
@@ -1,0 +1,11 @@
+{
+    "Note": "The following special characters show up in doc occasionally",
+
+    "dash_in_ascii": "-",
+    "dash_0x93": "–",
+    "dash_0x94": "—",
+
+    "quotation_mark_0x9d": "”",
+
+    "normal": "ascii"
+}
--- a/tests/unit/test_loaders.py
+++ b/tests/unit/test_loaders.py
@@ -55,6 +55,13 @@ class TestJSONFileLoader(BaseEnvVar):
         self.assertFalse(self.file_loader.exists(
             os.path.join(self.data_path, 'does', 'not', 'exist')))
 
+    def test_file_with_non_ascii(self):
+        try:
+            filename = os.path.join(self.data_path, 'non_ascii')
+            self.assertTrue(self.file_loader.load_file(filename) is not None)
+        except UnicodeDecodeError:
+            self.fail('Fail to handle data file with non-ascii characters')
+
 
 class TestLoader(BaseEnvVar):
 


rayluo:
Although this fix is supposed to fix the symptom discovered in our latest build, the root cause (i.e. not opening file in utf8 on Python 3) exists for months.

We probably need to audit and apply similar adjustment to some other `open(filename, 'r')` or `open(filename)` function call, especially to those dealing with the model data, which tends to accidentally contain non-ascii in documentation strings. (All the non-ascii characters in the test case of this PR come from actual model files.)

Some of the possibly affected files are:

* https://github.com/boto/botocore/blob/develop/botocore/translate.py#L46
* https://github.com/boto/botocore/blob/develop/scripts/get-model-filename#L136

We may also want to have similar check for Boto 3 and CLI.

Thoughts?

JordonPhillips:
Looks good. `io.open` is the way to go if we want to have consistent behavior across our supported versions. It will slow down people running on python 2 a bit, but I think it is an acceptable tradeoff. We really should be enforcing unicode.

:ship:

JordonPhillips:
Looking through old issues, our lack of unicode support seems to be causing a lot of problems. We should probably prioritize support.

botocore/loaders.py
```diff
@@ -153,7 +154,10 @@ def load_file(self, file_path):
         full_path = file_path + '.json'
         if not os.path.isfile(full_path):
             return
-        with open(full_path) as fp:
+
+        # By default the file will be opened with locale encoding on Python 3.
+        # We specify "utf8" here to ensure the correct behavior.
+        with io.open(full_path, encoding='utf8') as fp:
```
kyleknap: Usually in the code we use ``utf-8`` instead of ``utf8``. It would be nice to stay consistent. I do not think there is any drawbacks in using ``utf8`` over ``utf-8`` though.

kyleknap:
Looks fine. Had a small comment. :ship:

jamesls:
1. We should be fixing these upstream in the models.
2. io.open is pretty slow.  Why not add just add the encoding arg to JSON?

jamesls:
Also @JordonPhillips can you link to the other issues that seem to be because of unicode support?  Maybe there's a more general solution we're missing.

kyleknap:
@jamesls out of curiosity do you know how much slower is ``io.open`` than ``open``?

jamesls:
This was a popular point of discussion in the python community when the `io` module came out years ago.  The big reason was that in py2 the io code was not written in C, but instead pure python, at least in py2.

I would in general not recommend using io in python2.  io and codecs have a tendency to have subtle differences in behavior between py2/py3.

I'd still recommend a) fixing this upstream and/or b) fixing this in the JSON decode (which is the thing that actually cares about the utf-8 data).

This has happened before, and we've historically pushed these changes back upstream.

I don't have a whole lot of context here, but was the issue because of something like a smart quote char?  If so, this was likely accidental and not something we should be working around in our code.

kyleknap:
Makes sense. In that case, I vote to do both option a and b. We need to be both robust and make sure it fixed upstream so that it does not affect other SDKs.

rayluo:
Heading to the aforementioned both options. A quick update here. Since the json.load() or json.loads() in Python 3 expects an input as string rather than bytes, and also [ignores the encoding parameter](https://docs.python.org/3/library/json.html#basic-usage), so it seems we will need to address this before json.load(). Something like this.

    stream = open('service-2.json', 'rb')
    payload = stream.read().decode('utf-8')
    j = json.loads(payload)
    print('Read %d key-value pairs' % len(j))

jamesls:
:shipit: Thanks for updating the feedback.

rayluo:
OK. Our build is back to normal now.

For what it's worth, the code snippet below is a small script to detect such errors in affected data files.

    import sys
    import string


    if len(sys.argv) < 2:
        sys.exit('Usage: %s filename.xml' % sys.argv[0])

    payload = open(sys.argv[1], 'rb').read().decode('utf-8')
    ascii = set(string.printable)

    for row, line in enumerate(payload.splitlines()):
        non_ascii = set(line) - ascii
        for c in non_ascii:
            print('Found non-ascii character %s in line %d' % (c, row+1))

PR #682, opened 2015-10-13
Add the retry for Route53 when PriorRequestNotComplete happens

### LINKED ISSUE #91: better to add the route53 request limits in _retry.json
better to add route53 request limits here:
https://github.com/boto/botocore/blob/develop/botocore/data/aws/_retry.json

Details can be found in the issue #1618  in boto:
https://github.com/boto/boto/issues/1618

henrysher:
Fix #91 accroding to the document (http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html)

### DIFF
--- a/botocore/data/_retry.json
+++ b/botocore/data/_retry.json
@@ -150,6 +150,20 @@
           }
         }
       }
+    },
+    "route53": {
+      "__default__": {
+        "policies": {
+          "request_limit_exceeded": {
+            "applies_when": {
+              "response": {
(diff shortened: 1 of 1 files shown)


jamesls:
Thanks.  I'll also add the Throttling exception code from their docs as well:

>  If you submit more than five requests per second, Amazon Route 53 returns an HTTP 400 error (Bad request). The response header also includes a Code element with a value of Throttling and a Message element with a value of Rate exceeded.

PR #684, opened 2015-10-13
Add util for checking if operation is paginatable

kyleknap:
Going to use this to cross check pagination interfaces for resources in boto3.


mtdowling:
What's the motivation for this?

kyleknap:
I was going to have scripts in botocore and tests on boto3 rely on it. Now I am just using it in a test for boto3: https://github.com/boto/boto3/pull/302. No longer really need it in botocore.

PR #686, opened 2015-10-14
Accept PolicyDocuments as both Dict and String

JordonPhillips:
This will close boto/boto3#229, but is blocking on #664. If the user provides a dict when a policy document string is expected, it will automagically be converted to a string now. With #664, dicts will be the documented input type.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -300,6 +300,17 @@ def copy_snapshot_encrypted(params, request_signer, **kwargs):
     params['PresignedUrl'] = presigned_url
 
 
+def json_encode_policies(params, model, **kwargs):
+    input_shape = model.input_shape
+    if input_shape is not None:
+        for member_name, member_shape in input_shape.members.items():
+            param = params.get(member_name)
+            if member_shape.name == 'policyDocumentType' and \
+                    param is not None and \
+                    isinstance(param, dict):
+                params[member_name] = json.dumps(param)
+
+
 def json_decode_policies(parsed, model, **kwargs):
     # Any time an IAM operation returns a policy document
     # it is a string that is json that has been urlencoded,
@@ -538,6 +549,7 @@ BUILTIN_HANDLERS = [
      base64_encode_user_data),
     ('before-parameter-build.route53', fix_route53_ids),
     ('before-parameter-build.glacier', inject_account_id),
+    ('before-parameter-build.iam', json_encode_policies),
 
     # Glacier documentation customizations
     ('docs.*.glacier.*.complete-section',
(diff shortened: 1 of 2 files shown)


botocore/handlers.py
```diff
@@ -300,6 +300,17 @@ def copy_snapshot_encrypted(params, request_signer, **kwargs):
     params['PresignedUrl'] = presigned_url
 
 
+def json_encode_policies(params, model, **kwargs):
+    input_shape = model.input_shape
+    if input_shape is not None:
+        for member_name, member_shape in input_shape.members.items():
+            param = params.get(member_name)
+            if member_shape.name == 'policyDocumentType' and \
+                    param is not None and \
```
kyleknap: Is the check of if it is ``None`` needed because you are already checking it it an instance of dict?

tests/unit/test_handlers.py
```diff
@@ -332,6 +332,50 @@ def test_get_template_has_error_response(self):
...
         self.assertEqual(original, handler_input)
 
+    def test_encode_json_policy(self):
+        params = {
+            'Document': {
+                'foo': 'bar'
+            },
+            'Other': 'baz'
+        }
+        service_def = {
```
kyleknap: You should try to use the [DenormalizedStructureBuilder](https://github.com/boto/botocore/blob/develop/botocore/model.py#L487). It is a lot easier for creating shapes.

kyleknap:
Had a couple small comments, but I had a couple of larger comments as well:

1) I was looking at the output handler for the decoding json and it looks like it actually crawls even nested structures. Do you know if there are any nested json documents shapes for input? This may affect your logic as you only encode top level parameter members.
2) It would be nice if we can have an integration or functional test to make sure that the handler was applied appropriately from end to end.

botocore/handlers.py
JordonPhillips: Thought `None` might not jive with `isinstance`. Could have sworn it was that way in Java, but I just searched and even that isn't the case. Thanks for the tip! :+1:

tests/unit/test_handlers.py
JordonPhillips: I can see where it might be simpler in some areas, but it doesn't look like it will be all that much help in this case since I'm building a ServiceModel, not just a StructureShape. FWIW I'm basing this test on the test immediately below.

JordonPhillips:
There are cases where there is nested input, yes. For example, there is [`get_context_keys_for_custom_policy`](http://botocore.readthedocs.org/en/latest/reference/services/iam.html#IAM.Client.get_context_keys_for_custom_policy) which takes a list of policy documents. That is as deeply nested as it gets though, so maybe I could just look for that list explicitly.

kyleknap:
I would prefer it be general because if we do not make it general there will be extra overhead. If IAM ever adds a new member or operation that is nested, we will have to go back and update the customization to pick up the new member, and that is hard to detect usually (it usually takes a user to report it which is what we do not want to happen).

JordonPhillips:
Closing while a more general solution is investigated

PR #687, opened 2015-10-15
Use HTTP POST to send Cloudsearchdomain search operation

rayluo:
This can enable sending a potentially large request to Cloudsearchdomain.
This fixes boto/boto3#211

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -472,6 +472,15 @@ def check_openssl_supports_tls_version_1_2(**kwargs):
         pass
 
 
+def change_get_to_post(request, **kwargs):
+    # This is useful when we need to change a potentially large GET request
+    # into a POST with x-www-form-urlencoded encoding.
+    if request.method == 'GET' and '?' in request.url:
+        request.headers['Content-Type'] = 'application/x-www-form-urlencoded'
+        request.method = 'POST'
+        request.url, request.data = request.url.split('?', 1)
+
+
 # This is a list of (event_name, handler).
 # When a Session is created, everything in this list will be
 # automatically registered with that Session.
@@ -539,6 +548,9 @@ BUILTIN_HANDLERS = [
     ('before-parameter-build.route53', fix_route53_ids),
     ('before-parameter-build.glacier', inject_account_id),
 
+    # Cloudsearchdomain search operation will be sent by HTTP POST
+    ('request-created.cloudsearchdomain.Search',
+     change_get_to_post),
     # Glacier documentation customizations
     ('docs.*.glacier.*.complete-section',
      AutoPopulatedParam('accountId', 'Note: this parameter is set to "-" by \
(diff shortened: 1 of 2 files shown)


botocore/handlers.py
```diff
@@ -472,6 +472,13 @@ def check_openssl_supports_tls_version_1_2(**kwargs):
         pass
 
 
+def cloudsearchdomain_search_post(request, operation_name, **kwargs):
```
kyleknap: The name is sort of ambiguous is weird here. The functionality is not cloudsearchdomain specific. It could be used in other places possibly. Be nice if we can be explicit in the function name of what you are doing. Something a long the lines of switching the get to post and putting the querysting into the body. Also I do not think you need ``operation_name`` in the signature.

tests/functional/test_cloudsearchdomain.py
```diff
@@ -0,0 +1,43 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import mock
+
+from tests import BaseSessionTest
+
+
+class EarlyExit(BaseException):
```
kyleknap: Ah I see this a fair amount.  The following usually does the trick for me: https://github.com/boto/botocore/blob/develop/tests/functional/test_s3.py#L44. I would rather mock out the response so we do not cover up any other potential errors being thrown.

kyleknap:
Looks good. Just had a couple comments.

botocore/handlers.py
mtdowling: I could see this named as something like `convert_query_request_to_post`.

kyleknap:
Instead of ``self.assertTrue('q=foo' in sent_request.body)``, you can use ``self.assertIn('q=foo', sent_request.body)``. Otherwise, looks good :ship:

JordonPhillips:
:+1: :ship:

botocore/handlers.py
```diff
@@ -472,6 +472,15 @@ def check_openssl_supports_tls_version_1_2(**kwargs):
         pass
 
 
+def change_get_to_post(request, **kwargs):
+    # This is useful when we need to change a potentially large GET request
+    # into a POST with x-www-form-urlencoded encoding.
+    if request.method == 'GET' and '?' in request.url:
+        request.headers['Content-Type'] = 'application/x-www-form-urlencoded'
+        request.method = 'POST'
+        request.url, request.data = request.url.split('?', 2)
```
kyleknap: Shouldn't this be 1?
```py
In [4]: 'url?q=something?else'.split('?', 2)
Out[4]: ['url', 'q=something', 'else']

In [5]: 'url?q=something?else'.split('?', 1)
Out[5]: ['url', 'q=something?else']
```

rayluo: Arrr. My bad.

kyleknap:
:ship:

jamesls:
:shipit:

PR #688, opened 2015-10-15
Log traceback on ConnectionErrors

jamesls:
We log the traceback for the `except Exception` clause but not for `ConnectionError`s.  Even though the retry handler will reraise this exception, we miss the original stack trace so we can't see where in requests/urllib3 the ConnectionError originated from in python2.


kyleknap:
:ship:

PR #662, opened 2015-09-27
Feature paginator sum

rayluo:
A new sum-up behavior is added into paginator.  It will be triggered by a new setting in paginators-1.json named `sum_keys` (new naming suggestions are welcomed!).

PR #689, opened 2015-10-16
Adds the count ability to paginator

rayluo:
Paginator's result_key used to expect list value only.
Now, if a result_key's value is an integer (or a float),
paginator can aggregate them.

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -253,10 +253,6 @@ class PageIterator(object):
 
     def build_full_result(self):
         complete_result = {}
-        # Prepopulate the result keys with an empty list.
-        for result_expression in self.result_keys:
-            set_value_from_jmespath(complete_result,
-                                    result_expression.expression, [])
         for response in self:
             page = response
             # We want to try to catch operation object pagination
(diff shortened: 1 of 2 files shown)


rayluo:
This PR is trying to use a different approach to replace PR #662.

So far, the test case only covers the new behavior, but no paginator is using this ability yet. If this one is accepted, we will need to move the [relevant counters](https://github.com/boto/botocore/blob/develop/botocore/data/dynamodb/2012-08-10/paginators-1.json#L15-L16) from non_aggregate_keys to result_key to use the new ability and eventually solve aws/aws-cli#1423

jamesls:
Is this supersedes #662, any reason why #662 is still open?

jamesls:
You're missing a test for when a result key is missing from a page other than the first page.

This technically won't create keys in the returned dict that never show up in any page of results, but I *think* this is ok.  Previously we were always creating empty lists for every result key, which with non-list result keys is technically wrong now anyways.  I think this is an ok change.

Let's get that missing test case added.  Otherwise looks good.

rayluo:
To clarify, if "a result key is missing from a page other than the first page", it will still collect all available content into the final result.  On the other hand, if a result key "never show up in any page of results", yes it will then be absent from final result. I agree with you, this is an ok change.

These 2 scenarios are now included in the test cases.

jamesls:
:shipit:  Looks good.

PR #691, opened 2015-10-16
Add back old cloudsearch version

kyleknap:
This is adds an old api version of cloudsearch. You need this to use old cloudsearch domains. I pulled this file from the javascript SDK. It is also formatted as expected.

Fixes https://github.com/boto/boto3/issues/203


JordonPhillips:
Nice, great job! :ship:

rayluo:
Good job in finding the root cause of that issue! :ship:

PR #692, opened 2015-10-18
Fix the document for checksum in Glacier

rayluo:
The checksum in complete_multipart_upload() was incorrectly documented.
This fix provides proper documentation with a code snippet as example.

This fixes boto/boto3#251


jamesls:
To capture offline discussion, let's just call out the two operations where we calculate this value rather than complicating the AutoPopulatedParam class.  There's only two operations that need these docs.  This is also how the actual calculation of tree hash is called out anyways (https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L521-L522) so I think it makes sense to be consistent.

rayluo:
I believe I did reasonably well in the latest commit. That test failure happens on Python 2.6 only, at this test case "tests.functional.leak.test_resource_leaks.TestDoesNotLeakMemory". Interestingly, I rerun the same test on my local laptop and it passes.

Shall we increase that memory leak threshold (to as much as some 25MB?) in another PR and then I'll rebase this PR?

jamesls:
I tried rerunning one of my PRs for botocore I have out for review and I have all passing tests: https://travis-ci.org/boto/botocore/builds/86497828

I'd like to understand why this test is failing before bumping up resource limits.  If this was a systemic issue, I would expect to see all of our latest PRs failing this test case on travis, but it appears that only this PR fails.  Let's figure out why.

rayluo:
So I did some experiment here (and I will clean them up soon).  The root cause is in the memory leak checker. In short, deleting a (bunch of) object does not necessarily free their memory immediately. In the sensitive situation like this, we will need a gc.collect() to avoid false positive (see that successful [experiment 1](https://github.com/rayluo/botocore/commit/fc6eae2f3ae4130dc67e1ee037033a8d2ada9d14)). I'll come up with a complete adjustment on memory leak checker in a separate PR, and then rebase this one.

jamesls:
I'm not sure this is the right approach.  This seems to suggest a memory leak with our doc handlers.  Comparing the memory usage for the test in question:

* develop branch - 1.28515625 MB
* your PR - 22.625 MB

This is a pretty large growth that I don't really feel comfortable with `gc.collect()`ing away unless you can explain *exactly* why a 21MB growth is unavoidable.

Are you sure there's no way to fix this in our doc handlers?

botocore/handlers.py
```diff
@@ -556,8 +562,13 @@ def change_get_to_post(request, **kwargs):
      AutoPopulatedParam('accountId', 'Note: this parameter is set to "-" by \
                          default if no value is not specified.')
      .document_auto_populated_param),
-    ('docs.*.glacier.*.complete-section',
+    ('docs.*.glacier.UploadArchive.complete-section',
+     AutoPopulatedParam('checksum').document_auto_populated_param),
+    ('docs.*.glacier.UploadMultipartPart.complete-section',
      AutoPopulatedParam('checksum').document_auto_populated_param),
+    ('docs.request-params.glacier.CompleteMultipartUpload.complete-section',
+     document_glacier_tree_hash_checksum()),
```
kyleknap: Why are you calling this function when registering the handler? That is probably why we are seeing leak test issues.

botocore/handlers.py
```diff
@@ -556,8 +562,13 @@ def change_get_to_post(request, **kwargs):
...
+
     # UserData base64 encoding documentation customizations
     ('docs.*.ec2.RunInstances.complete-section', document_base64_encoding()),
```
kyleknap: I just happened to notice the same pattern of calling the function here as well when registering it. We should not be calling these.

tests/unit/test_client.py
```diff
@@ -445,7 +445,7 @@ def test_client_signs_call(self, signer_mock):
             'request-created.myservice.test_operation', request=request,
             operation_name='test_operation')
 
-        signer_mock.return_value.sign.assert_called_with(
+        signer_mock.assert_called_with(
```
rayluo: FYI: This pre-existing test case is adjusted a little bit to reflect the latest code structure, but the observable behavior of sign() is NOT changed, so this should not be considered as a breaking change.

botocore/signers.py
```diff
@@ -56,7 +57,11 @@ def __init__(self, service_name, region_name, signing_name,
         self._signing_name = signing_name
         self._signature_version = signature_version
         self._credentials = credentials
-        self._event_emitter = event_emitter
+
+        self._event_emitter = None
+        if event_emitter is not None:
```
jamesls: Quick question: are there cases when the passed in `event_emitter` can be `None`?  I don't think this is possible is it?

rayluo: Off-line discussion: True. There shouldn't be a case when emitter=None. We will update 4 test cases who violate this rule.

botocore/handlers.py
```diff
@@ -432,6 +432,24 @@ def add_glacier_checksums(params, **kwargs):
...
+        This is a required field.
+        You can use botocore.util.calculate_tree_hash() to calculate it
+        like this::
+
+            from botocore.utils import calculate_tree_hash
+
+            ...
+
+            client.complete_multipart_upload(
+                checksum=calculate_tree_hash(open('your_file.txt', 'rb')),
```
jamesls: FWIW, I don't think this is typically how a tree is calculated because you have to scan the file twice, which may be a fairly large file.

It's not technically wrong, but I typically see code that keeps the individual tree hashes as you're uploading parts and then aggregate the individual tree hashes all into a final tree hash (this is what boto does in its multipart archive uploader).

Might be worth pointing that out.

jamesls:
It would be great to get the docs updated with the comment I made above, but otherwise looks good.  Thanks for digging into the memory leak issue. :shipit:

jamesls:
:shipit:  Looks good.  Thanks for all your work on this.

PR #694, opened 2015-10-20
Profile should override env var creds

### LINKED ISSUE #663: Recent commit breaks --profile functionality in 1.2.5.
Commit 3c09b00e6593bb565edfa4d214b8362d6dd9cebb breaks the --profile overriding ENV functionality which was recently introduced in revision 1.2.5.

Once line 70 of botocore/credentials.py is adjusted to `if session.profile is None:`, everything continues to function as expected.

I'd submit a PR, but I'm unsure how this breaks your test cases.

Thanks.

jamesls:
Profile should override env var creds

Unless the profile and creds are both specified
as env vars.

In order to fix this regression, I had to make a
small change to session.  This makes the 'profile'
config var less special cased than it was previously.

Now, we always check the _session_instance_vars for
a profile, and simple use _profile as a placeholder
for the cached lookup value of 'profile'.

The only place this gets weird is we need to scope
down our call to ".get_config_variable" in the
credential provider creation to only check the
session instance variables for a profile to determine
whether or not we need to add the EnvProvider to our
lookup chain.

One of the reasons this regressed is we didn't have
sufficient functional tests to cover this behavior.
I've added functional tests that ensure that regardless
of internals, we won't regress on this in the future.

There will also be a pull request to the AWS CLI
to verify this behavior to ensure the CLI doesn't regress
on this behavior.

Fixes #663.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -67,7 +67,7 @@ def create_credential_resolver(session):
     # We use ``session.profile`` for EnvProvider rather than
     # ``profile_name`` because it is ``None`` when unset.
     # TODO: Remove use of internal var.  We'll need to rethink this.
-    if session._profile is None:
+    if session.profile is None:
         # No profile has been explicitly set, so we prepend the environment
         # variable provider. That provider, in turn, may set a profile
         # or credentials.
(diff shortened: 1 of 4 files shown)


kyleknap:
Looks fine. :ship:

JordonPhillips:
Looks good :ship:

PR #695, opened 2015-10-20
Port ca_bundle from AWS CLI into botocore session

jamesls:
This takes the work from aws/aws-cli#1379 and moves
it into the botocore session.

The benefit in doing this is that botocore/boto3
can now support the ca_bundle option from
~/.aws/config, which gives better interop between
the AWS CLI and the python SDK.

There will also need to be a corresponding PR to
aws/aws-cli that removes this entry from the env vars
passed to the session created in CLIDriver.


mtdowling:
LGTM

botocore/client.py
```diff
@@ -191,6 +191,14 @@ def _inject_s3_configuration(self, config_kwargs, scoped_config,
...
         config_kwargs['s3'] = s3_configuration
 
+    def _resolve_verify_value(self, provided_verify_value, scoped_config):
+        # If a user provides a value for "verify" then we don't do anything.
+        # We always use explicit values provided.
+        if provided_verify_value is not None:
+            return provided_verify_value
+        elif scoped_config is not None:
+            return scoped_config.get('ca_bundle')
+
```
kyleknap: Should we return ``provide_verify_value`` at the end if none of these if statements are hit as we are overriding the ``verify`` value in the ``_get_client_args``? I do not think this affects functionality right now because the default is ``None`` anyways.

kyleknap:
It looks good. I just had one question that I would like answered.

botocore/client.py
jamesls: You're correct, this doesn't affect anything because we always return immediately if the user passed in some value for `verify`.

This would also never change because we can't change the default value of the `verify` param from `None` as that would be a backwards incompatible change.

kyleknap:
Ok good just checking. :ship:

PR #700, opened 2015-10-23
Now can resume from an empty first page

rayluo:
This is built on top of PR #689, and fixes issue aws/aws-cli#1365

I also manually test it with this command line to make sure it solves the issue.

    aws rds download-db-log-file-portion --db-instance-identifier mydbinstance --log-file-name error/mysql-error-running.log --starting-token 1000

A couple previous test cases have also been adjusted a little bit. Their previous default value as list is no longer considered as proper anyway.

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -189,14 +189,14 @@ class PageIterator(object):
 
     def _handle_first_request(self, parsed, primary_result_key,
                               starting_truncation):
-        # First we need to slice into the array and only return
+        # If the payload is an array, we need to slice into it and only return
         # the truncated amount.
         starting_truncation = self._parse_starting_token()[1]
-        all_data = primary_result_key.search(parsed)
+        data = primary_result_key.search(parsed)
         set_value_from_jmespath(
             parsed,
             primary_result_key.expression,
-            all_data[starting_truncation:]
+            data[starting_truncation:] if isinstance(data, list) else None
         )
         # We also need to truncate any secondary result keys
         # because they were not truncated in the previous last
@@ -204,7 +204,7 @@ class PageIterator(object):
         for token in self.result_keys:
             if token == primary_result_key:
                 continue
-            set_value_from_jmespath(parsed, token.expression, [])
+            set_value_from_jmespath(parsed, token.expression, None)
         return starting_truncation
 
     def _truncate_response(self, parsed, primary_result_key, truncate_amount,
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -189,14 +189,18 @@ class PageIterator(object):
 
     def _handle_first_request(self, parsed, primary_result_key,
                               starting_truncation):
-        # If the payload is an array, we need to slice into it and only return
-        # the truncated amount.
+        # If the payload is an array or string, we need to slice into it
+        # and only return the truncated amount.
         starting_truncation = self._parse_starting_token()[1]
-        data = primary_result_key.search(parsed)
+        all_data = primary_result_key.search(parsed)
+        if isinstance(all_data, (list, string_types)):
+            data = all_data[starting_truncation:]
+        else:
+            data = None
         set_value_from_jmespath(
             parsed,
             primary_result_key.expression,
-            data[starting_truncation:] if isinstance(data, list) else None
+            data
         )
         # We also need to truncate any secondary result keys
         # because they were not truncated in the previous last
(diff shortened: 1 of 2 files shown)


tests/unit/test_paginate.py
```diff
@@ -613,12 +613,18 @@ def test_resume_with_multiple_input_keys(self):
         complete = pages.build_full_result()
         self.assertEqual(complete,
                          {"Users": ['User4'],
-                          "Groups": [],
```
jamesls: Isn't this technically a breaking change?

rayluo: In this specific test case, yes. But a robust paginator will need to deal with possible absence of keys in any page anyway, and our paginator.build_full_result() will handle it correctly, so it should be fine.

botocore/paginate.py
```diff
@@ -189,14 +189,18 @@ def _inject_token_into_kwargs(self, op_kwargs, next_token):
 
     def _handle_first_request(self, parsed, primary_result_key,
                               starting_truncation):
-        # First we need to slice into the array and only return
-        # the truncated amount.
+        # If the payload is an array or string, we need to slice into it
+        # and only return the truncated amount.
         starting_truncation = self._parse_starting_token()[1]
         all_data = primary_result_key.search(parsed)
+        if isinstance(all_data, (list, string_types)):
```
jamesls: Is this right?  Isn't this equivalent to: `isinstance(all_data, (list, (basestring,)))` in python2?  Shouldn't it be `isinstance(all_data, (list, bastring))`?

botocore/paginate.py
```diff
@@ -204,7 +208,16 @@ def _handle_first_request(self, parsed, primary_result_key,
         for token in self.result_keys:
             if token == primary_result_key:
                 continue
-            set_value_from_jmespath(parsed, token.expression, [])
+            sample = token.search(parsed)
+            if isinstance(sample, list):
+                empty_value = []
+            elif isinstance(sample, string_types):
+                empty_value = ''
```
jamesls: Is this tested?  I don't see a test for this?

botocore/paginate.py
```diff
@@ -204,7 +208,16 @@ def _handle_first_request(self, parsed, primary_result_key,
...
+            elif isinstance(sample, (int, float)):
+                empty_value = 0
```
jamesls: Is this tested?  I don't see a test for this.

jamesls:
Change looks good.  I think there's just a few missing tests I'm seeing:

```
$ nosetests tests/unit/ --with-coverage --cover-package botocore.paginate --cover-erase
Name                Stmts   Miss Branch BrMiss  Cover   Missing
---------------------------------------------------------------
botocore.paginate     258      7    110     10    95%   187, 215, 217, 227, 279, 316, 447
Ran 1560 tests in 6.639s
```

I believe lines 215/217 are part of the conditionals you've added.

botocore/paginate.py
rayluo: While I don't have any particular objection for adding those 2 test cases (and I've updated this PR already), I would like to take this chance to say, I am worrying those new test cases and (especially) our tendency of "never break existing test cases", will further limit our future options on a potential paginator refactoring.

To better understand the background, let's imagine that we were still in day 1 of designing the paginator, it would be a matter of taste to decide whether an absent (primary or secondary) result_key should show up as the corresponding empty type (i.e. my previous [commit](https://github.com/rayluo/botocore/commit/ccdbcfc778ded1060e3db68f6f9e9c529afa0e14)), or not show up at all (i.e. my first [commit](https://github.com/rayluo/botocore/commit/a3e70de5e76f9f77806c959bb04ffe0ee8a04a3f)). In other words, these behavior was an implementation detail. And, implementation detail "should not impact API" in the first place. (There is a good explanation in slide 15 of this keynote http://lcsd05.cs.tamu.edu/slides/keynote.pdf and BTW, this is the webcast http://www.infoq.com/presentations/effective-api-design). If now it becomes a rigid part of our api contract and then limits our flexibility of refactoring, hence unnoticeably increases the complexity (thus the workload) on subsequent bug fix/new feature development, that doesn't feel right.

I'm NOT advocating it is OK to break any existing test cases. I'm just hoping we either: (a) do not put implementation detail into test cases, or; (b) or allow us to adjust those implementation level test cases when needed. Besides, we can make it clear that, all those response examples in documentation are merely demonstrating the format of each element, it does NOT hint each element will show up.

PS: This can be considered as a separated conversation. It should not delay or block the review of this PR.

jamesls:
:shipit:

PR #701, opened 2015-10-24
Unify tox/travis tests

catlee:
I was having some problems running tests locally that weren't evident on travis, e.g.

```
ERROR: test_generate_presigned_post (tests.unit.test_signers.TestS3PostPresigner)
  File "/home/catlee/src/botocore/tests/unit/test_signers.py", line 276, in test_generate_presi
gned_post
    self.add_auth.assert_called_once()
  File "/home/catlee/src/botocore/.tox/py27/local/lib/python2.7/site-packages/mock/mock.py", li
ne 721, in __getattr__
    raise AttributeError(name)
AttributeError: assert_called_once
```

The problem turned out to be due to travis using code scripts/ci/ to execute tests, while tox.ini was used locally.

I've updated .travis.yml to use tox to reproduce the same behavior as the old scripts (I hope!)

I also cleaned up some package requirements,  fixed usage of a non-existent mock assertion.

Finally, I adjusted the signing tests to stop trying to contact the special AWS metadata url to get credentials. This sped up tests locally from 24s to 13s, since we don't need to wait 1s per test.

Any feedback on this would be appreciated! I'm not sure if code coverage still works, and I've removed the integration test script which didn't seem to be used. The same tests are still runnable via tox if you pass the explicit environment name.


tests/unit/test_signers.py
```diff
@@ -273,7 +273,7 @@ def test_generate_presigned_post(self):
         self.auth.assert_called_with(
             credentials=self.credentials, region_name='region_name',
             service_name='signing_name')
-        self.add_auth.assert_called_once()
+        self.add_auth.call_count == 1
```
mtdowling: Are these changes necessary for this pull request, or are they accidentally included due to the related PR? Can you please remove these changes if they are not necessary.

catlee: They are necessary if we upgrade mock. assert_called_once() isn't actually a supported method on any version of mock. In mock 1.0, it will just act like any other mock method and return a new Mock object.

mock 1.1 added an 'unsafe' flag (defaults to False), which raises assertions if you try and call non-existent assert_ methods.

kyleknap:
So I recently updated the ``tox.ini`` to use the script that we use for travis. I think we want to go with that route because that is the primary way our team tests the various projects we work on and the ``tox.ini`` file is often out of date because of that. Ultimately, that will ensure that the ``tox.ini`` file stays current with how we test.

Thanks for the pull request though. Closing issue. Let us know if there is anything you like to see different with the ``tox.ini`` file that we have now.

PR #703, opened 2015-10-26
Generalize special casing of pluralized acronyms

jamesls:
Given that this is in the critical path for making API calls,
I did some benchmarking to ensure that this doesn't cause any
serious performance issues.  Short answer is that this seems
acceptable.

I timed how long it took to xform every single operation and
top level member name against this branch and the develop
branch.

Given 3034 xform's, I'm seeing:

```
dev branch   - 0.0930128 seconds averaged over 10 runs of 3034
fix-arn-name - 0.0944845 seconds averaged over 10 runs of 3034
```

Or in terms of ops/sec:

```
before - 32619.2 ops/sec
after  - 32111.1 ops/sec
```

The reason I think this should be generalized is that some
services use this pattern for particular types of resources
(mostly ACLs and ARNs), and we'll need to possibly update
our special cases every time we pull in a service update.

With this change, we won't need to do this.  I will need
to make a corresponding PR to the AWS CLI to put in
backward compat shims so we can still support both.

Using [this script](https://gist.github.com/jamesls/0d1005415d30d2f9f44a),
the delta between the develop branch and this PR is:

```
Model Name                                          xform name

-OpenIdConnectProviderARNs                          open_id_connect_provider_ar_ns
+OpenIdConnectProviderARNs                          open_id_connect_provider_arns

-TapeARNs                                           tape_ar_ns
+TapeARNs                                           tape_arns

-VTLDeviceARNs                                      vtl_device_ar_ns
+VTLDeviceARNs                                      vtl_device_arns

-VolumeARNs                                         volume_ar_ns
+VolumeARNs                                         volume_arns

```

### DIFF
--- a/botocore/__init__.py
+++ b/botocore/__init__.py
@@ -31,11 +31,12 @@ log.addHandler(NullHandler())
 _first_cap_regex = re.compile('(.)([A-Z][a-z]+)')
 _number_cap_regex = re.compile('([a-z])([0-9]+)')
 _end_cap_regex = re.compile('([a-z0-9])([A-Z])')
+# The regex below handles the special case where some acryonym
+# name is pluralized, e.g GatewayARNs, ListWebACLs, SomeCNAMEs.
+_special_case_transform = re.compile('[A-Z][A-Z][A-Z]+s$')
 # Prepopulate the cache with special cases that don't match
 # our regular transformation.
 _xform_cache = {
-    ('SwapEnvironmentCNAMEs', '_'): 'swap_environment_cnames',
-    ('SwapEnvironmentCNAMEs', '-'): 'swap-environment-cnames',
     ('CreateCachediSCSIVolume', '_'): 'create_cached_iscsi_volume',
     ('CreateCachediSCSIVolume', '-'): 'create-cached-iscsi-volume',
     ('DescribeCachediSCSIVolumes', '_'): 'describe_cached_iscsi_volumes',
@@ -44,10 +45,6 @@ _xform_cache = {
     ('DescribeStorediSCSIVolumes', '-'): 'describe-stored-iscsi-volumes',
     ('CreateStorediSCSIVolume', '_'): 'create_stored_iscsi_volume',
     ('CreateStorediSCSIVolume', '-'): 'create-stored-iscsi-volume',
-    ('NotificationARNs', '_'): 'notification_arns',
-    ('NotificationARNs', '-'): 'notification-arns',
-    ('ListWebACLs', '_'): 'list_web_acls',
-    ('ListWebACLs', '-'): 'list-web-acls',
 }
 ScalarTypes = ('string', 'integer', 'boolean', 'timestamp', 'float', 'double')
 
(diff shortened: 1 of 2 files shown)


botocore/__init__.py
```diff
@@ -70,6 +67,11 @@ def xform_name(name, sep='_', _xform_cache=_xform_cache):
         return name
     key = (name, sep)
     if key not in _xform_cache:
+        if _special_case_transform.search(name) is not None:
+            is_special = _special_case_transform.search(name)
```
rayluo: How about switch these two lines like this? It seems cleaner, and probably slightly faster.

    is_special = _special_case_transform.search(name)
    if is_special:
        blah_blah

jamesls: It's actually not faster for the common case, which is why I reversed the order (you aren't storing a local var).  Definitely a microoptimization though...

botocore/__init__.py
```diff
@@ -31,11 +31,12 @@ def emit(self, record):
 _first_cap_regex = re.compile('(.)([A-Z][a-z]+)')
 _number_cap_regex = re.compile('([a-z])([0-9]+)')
 _end_cap_regex = re.compile('([a-z0-9])([A-Z])')
+# The regex below handles the special case where some acryonym
+# name is pluralized, e.g GatewayARNs, ListWebACLs, SomeCNAMEs.
+_special_case_transform = re.compile('[A-Z][A-Z][A-Z]+s$')
```
mtdowling: Is this basically "match strings that end in 3 or more uppercase characters followed by 's'"? If so, I think this expression is more clear and possibly more efficient: `[A-Z]{2,}s$`

```
>>> re.compile('[A-Z]{2,}s$', re.DEBUG)
max_repeat 2 4294967295
  in
    range (65, 90)
literal 115
at at_end
```

Compared to

```
>>> re.compile('[A-Z][A-Z][A-Z]+s$', re.DEBUG)
in
  range (65, 90)
in
  range (65, 90)
max_repeat 1 4294967295
  in
    range (65, 90)
literal 115
at at_end
```

I haven't benchmarked this to see if it makes a difference, but I think using `[A-Z]{2,}s$` is easier to read.

jamesls: Updated the code (but I think you meant `{3,}`).

botocore/__init__.py
rayluo: Interesting. Makes sense. :+1: 

In this case, perhaps we would like to add an extra line of comment as something like "# The following two lines are indeed faster in common case". This comment suggestion is definitely optional though.

botocore/__init__.py
mtdowling: doh. Yep that's what I meant.

mtdowling:
LGTM

botocore/__init__.py
```diff
@@ -44,10 +45,6 @@ def emit(self, record):
     ('DescribeStorediSCSIVolumes', '-'): 'describe-stored-iscsi-volumes',
     ('CreateStorediSCSIVolume', '_'): 'create_stored_iscsi_volume',
     ('CreateStorediSCSIVolume', '-'): 'create-stored-iscsi-volume',
```
rayluo: Just a side note. I happened to notice that our Ruby SDK uses a slightly different approach by [defining the special keywords "iSCSI" (and some others) only once](https://github.com/aws/aws-sdk-ruby/blob/d9e649c2b858647b68dcc80e38dff8f3ab1e752c/aws-sdk-core/lib/aws-sdk-core.rb#L4-L12), and then [preprocess them](https://github.com/aws/aws-sdk-ruby/blob/65fc6a6ea69792668763611af57c61db2c87a51d/aws-sdk-core/lib/seahorse/util.rb#L17-L20). Do we also need to handle "Ec2" and "ElastiCache" cases here?

jamesls: Let's address these as separate PRs if needed.  iSCI I think is too special cased to have a keyword for, ElastiCache isn't a concern for us (not used for opname/inputs), but Ec2 I believe we'll need to fix for opsworks.RegisterVolume.

rayluo:
LGTM.

PR #706, opened 2015-10-29
Respect locationName for JSON Protocols

kyleknap:
We were not respecting ``locationName`` for JSON protocols, which caused issues with API Gateway's ``PutIntegration`` operation's [integrationHttpMethod](https://github.com/boto/botocore/blob/develop/botocore/data/apigateway/2015-07-09/service-2.json#L4133) which uses ``locationName``.

Fixes https://github.com/aws/aws-cli/issues/1605


JordonPhillips:
:ship:

PR #707, opened 2015-10-30
Wrap the BadStatusLine with a clear description

rayluo:
New test case is added. Also simplify the existing test cases a little bit.

### DIFF
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -53,14 +53,14 @@ class EndpointConnectionError(BotoCoreError):
         'Could not connect to the endpoint URL: "{endpoint_url}"')
 
 
-class BadStatusLineError(ConnectionError):
+class ConnectionClosedError(ConnectionError):
     fmt = (
-        'Bad status line occurred on endpoint URL: "{endpoint_url}". '
-        'You may consider increasing retry attempts.')
+        'Connection was closed before we received a valid response '
+        'from endpoint URL: "{endpoint_url}".')
     def __init__(self, **kwargs):
         msg = self.fmt.format(**kwargs)
         kwargs.pop('endpoint_url')
-        super(BadStatusLineError, self).__init__(msg, **kwargs)
+        super(ConnectionClosedError, self).__init__(msg, **kwargs)
 
 
 class NoCredentialsError(BotoCoreError):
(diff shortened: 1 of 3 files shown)


tests/unit/test_endpoint.py
```diff
@@ -99,12 +100,6 @@ def test_make_request_with_no_auth(self):
         self.assertNotIn('Authorization', prepared_request.headers)
 
     def test_make_request_no_signature_version(self):
-        self.endpoint = Endpoint(
```
jamesls: Were the lines removed because they were unnecessary?  Just curious.

botocore/exceptions.py
```diff
@@ -52,6 +53,16 @@ class EndpointConnectionError(BotoCoreError):
         'Could not connect to the endpoint URL: "{endpoint_url}"')
 
 
+class BadStatusLineError(ConnectionError):
+    fmt = (
+        'Bad status line occurred on endpoint URL: "{endpoint_url}". '
```
jamesls: I think the intent was to completely hide anything about a bad status line from the customer.  This is confusing and misleading.  I think instead we should say something along the lines that it appears that the server closed the connection before we received a response.

We also can't really say you should increase retry attempts until we have a (supported) way for them to actually do so.

tests/unit/test_endpoint.py
rayluo: True. Those lines are duplicated in the base class. In my first prototype, I copy-and-paste them into my new test case, and then noticed then-3-copies of such snippets in subclass. Bad smell.

botocore/exceptions.py
rayluo: OK. The tricky thing here is that, the server did not purposely close connection before it send back a proper response; it is just somehow the response (or the connection) was lost during the transmission. So I think we would need some neutral wording.

How about this?

    class ConnectionLost(ConnectionError):
        fmt = 'Connection is lost before we received a valid response from endpoint URL: "{endpoint_url}".'

jamesls: I think "lost" is confusing, it's not something I commonly hear with regard to connections.

What are the *specific* conditions in which we would get an empty string for the status line and not have it be because the server/remote-end closed the connection?

I'm not aware of when else this would happen.

rayluo: Well then. I'll go back to your suggestion of "closed", but use a passive form.

    class ConnectionClosed(ConnectionError):
        fmt = 'Connection is closed before we received a valid response from endpoint URL: "{endpoint_url}".'

jamesls: s/is/was/ and it looks good.  Thanks.

jamesls:
:shipit: Looks good.

PR #709, opened 2015-11-04
Autopopulate CopySourceSSECustomerKeyMD5 for s3

kyleknap:
This is implementing logic used for this PR: https://github.com/aws/aws-cli/pull/1487. We have it for SSECustomer members but not the copy source one which may be needed when calling CopyObject.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -170,21 +170,37 @@ def sse_md5(params, **kwargs):
     encryption key. This handler does both if the MD5 has not been set by
     the caller.
     """
-    if not _needs_s3_sse_customization(params):
+    _sse_md5(params, 'SSECustomer')
+
+
+def copy_source_sse_md5(params, **kwargs):
+    """
+    S3 server-side encryption requires the encryption key to be sent to the
+    server base64 encoded, as well as a base64-encoded MD5 hash of the
+    encryption key. This handler does both if the MD5 has not been set by
+    the caller specifically if the parameter is for the copy-source sse-c key.
+    """
+    _sse_md5(params, 'CopySourceSSECustomer')
+
+
+def _sse_md5(params, sse_member_prefix='SSECustomer'):
+    if not _needs_s3_sse_customization(params, sse_member_prefix):
         return
-    key_as_bytes = params['SSECustomerKey']
+    sse_key_member = sse_member_prefix + 'Key'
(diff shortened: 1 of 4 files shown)


JordonPhillips:
Looks good to me :ship:

botocore/handlers.py
```diff
@@ -170,21 +170,37 @@ def sse_md5(params, **kwargs):
...
-    params['SSECustomerKeyMD5'] = key_md5_str
+    params[sse_key_member] = key_b64_encoded
+    params[sse_md5_member] = key_md5_str
 
 
-def _needs_s3_sse_customization(params):
-    return (params.get('SSECustomerKey') is not None and
-            'SSECustomerKeyMD5' not in params)
+def _needs_s3_sse_customization(params, sse_member_prefix='SSECustomer'):
+    return (params.get(sse_member_prefix+'Key') is not None and
```
mtdowling: I think we typically include spaces between operators

mtdowling:
:ship:

botocore/handlers.py
```diff
@@ -170,21 +170,37 @@ def sse_md5(params, **kwargs):
...
-    params['SSECustomerKey'] = key_b64_encoded
-    params['SSECustomerKeyMD5'] = key_md5_str
+    params[sse_key_member] = key_b64_encoded
+    params[sse_md5_member] = key_md5_str
 
 
-def _needs_s3_sse_customization(params):
-    return (params.get('SSECustomerKey') is not None and
-            'SSECustomerKeyMD5' not in params)
+def _needs_s3_sse_customization(params, sse_member_prefix='SSECustomer'):
```
jamesls: Any reason this needs to have a default value if we're always providing it when we call `_needs_s3_sse_customization`?

botocore/handlers.py
kyleknap: hmm flake8 did not catch that. I am changing it, but for future references, there are some nuances to that rule in pep8 if you have multiple operators on the same line: https://www.python.org/dev/peps/pep-0008/#other-recommendations.

botocore/handlers.py
kyleknap: Nah we do not need a default. Just forgot to remove it upon implementation.

kyleknap:
@jamesls I updated. Let me know what you think.

jamesls:
:shipit:

PR #710, opened 2015-11-04
Generate shared examples for API documentation

JordonPhillips:
This is an update of #617, addressing some comments there and adding in unit tests.


mtdowling:
Can you attach a screenshot of what this would look like in the docs?

JordonPhillips:
Will do. Looks like I'll need to make some other changes to fix the tests too.

JordonPhillips:
example

botocore/docs/service.py
```diff
@@ -56,7 +56,13 @@ def table_of_contents(self, section):
         section.style.table_of_contents(title='Table of Contents', depth=2)
 
     def client_api(self, section):
-        ClientDocumenter(self._client).document_client(section)
+        examples = None
+        try:
+            examples = self._session.get_examples(self._service_name)
```
jamesls: Can we just move `get_examples` into this class rather than on the `session`?  I'd really like to only add methods to the `session` when necessary and when it's something a customer could conceivably call.  I don't really see a customer needing to call `session.get_examples`.

botocore/docs/sharedexample.py
```diff
@@ -0,0 +1,142 @@
...
+            comments=comments.get('input')).build_example_code(
+            prefix=example_prefix).split('\n')
+        shared_example_section.style.start_codeblock()
+        shared_example_section.write(request.pop(0) + '\n')
+        for line in request:
+            shared_example_section.writeln(line)
+
+        if 'output' in example:
+            shared_example_section.style.new_line()
+            shared_example_section.write('# Print out the response.\n')
```
jamesls: I don't think this comment is necessary.

tests/unit/docs/__init__.py
```diff
@@ -135,6 +135,30 @@ def _setup_models(self):
             }
         }
 
+        self.examples_json_model = {
```
jamesls: I'd prefer this just move to the test module where it's used.  Otherwise you have to keep jumping between both files to understand the test.

botocore/docs/sharedexample.py
```diff
@@ -0,0 +1,142 @@
...
+    def _visit(self, value, indent, path, is_param=False):
+        if is_param:
+            return self._visit_param(value, indent, path)
+        if isinstance(value, dict):
+            return self._visit_dict(value, indent, path)
+        if isinstance(value, list):
+            return self._visit_list(value, indent, path)
+
+        try:
+            if isinstance(value, basestring):
```
jamesls: This will fail in python3.

jamesls:
What do you think about adding the `ResponseMetadata` key in the output?  It's going to be in every client response, and I think it makes sense to have output that matches what the customer would see.

botocore/docs/sharedexample.py
JordonPhillips: Yes, which is why it is surrounded by a try/except block.

jamesls: Oh totally missed catching NameError.  The preferred way to handle py2/py3 differences is to pull from compat.py.  You should be able to use `compat.six.string_types` which works for py2/py3.

JordonPhillips:
@jamesls 
I think that's probably a good idea, but would any services send back excess information on a 200? If so, it could be confusing.

botocore/docs/sharedexample.py
JordonPhillips: Would I pull from vendored or compat? I get `six` import errors when I try `botocore.compat.six`

kyleknap: Agree with James. Either should work. I would prefer compat. Strange that you got import errors though.

botocore/session.py
```diff
@@ -35,6 +35,7 @@
 from botocore import paginate
 from botocore import waiter
 from botocore import retryhandler, translate
+from botocore.docs import sharedexample
```
kyleknap: Is this import needed?

tests/unit/docs/test_sharedexample.py
```diff
@@ -0,0 +1,121 @@
...
+        builder = SharedExampleBuilder(
+            params={'foo': 'bar'},
+            operation_name='SampleOperation',
+            is_input=True
+        )
+        example = builder.build_example_code()
+        self.assertIn("(\n    foo='bar', \n)", example)
+
+        example = builder.build_example_code(prefix='sample_operation')
+        self.assertIn("sample_operation(\n    foo='bar', \n)", example)
```
kyleknap: I would prefer using ``assert_lines_in_order``. It is easier to read.

tests/unit/docs/test_sharedexample.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def test_dict_example(self):
+        builder = SharedExampleBuilder(
+            params={'foo': {'bar': 'baz'}},
+            operation_name='SampleOperation',
+            is_input=False
+        )
+        example = builder.build_example_code()
+        self.assertIn("'foo': {\n        'bar': 'baz', \n    },", example)
+        print('hello')
```
kyleknap: Remove these print statements. Same for the rest of the ``assertIn`` statements.

kyleknap:
Looks like you have a good idea on how to get this working. I had a few small comments, but I have one relatively larger comment which is in ``SharedExampleBuilder`` there is a lot of logic (in terms of traversing the parameters and writing out examples) is it possible you could share this logic with the [BaseExampleDocumenter](https://github.com/boto/botocore/blob/develop/botocore/docs/example.py#L17)? This class is able to handle some of the smaller edge cases you may be missing such as proper handling of indentation (as opposed to hardcoding spaces), handling of doubly nested structures, and actually writing it into the DocumentStructure as opposed to generating it all upfront and then writing it all at once to the DocumentStructure.

botocore/docs/sharedexample.py
JordonPhillips: Will there be any issues pulling from vendored? I can't get it to work from compat. No tests fail, but maybe there's something I'm not seeing.

botocore/session.py
JordonPhillips: Good catch, that was leftover from previous commits

botocore/docs/sharedexample.py
kyleknap: It should be fine. I am curious why we cannot pull it from compat. There must be a circular dependency somewhere since compat pulls it from vendored.

tests/unit/docs/test_sharedexample.py
JordonPhillips: That function wouldn't apply here, this is before it ever gets into the doc structure.

tests/unit/docs/test_sharedexample.py
JordonPhillips: Get rid of the `assertIn` statements? Why?

JordonPhillips:
I looked at `BaseExampleDocumenter`, but it seems strongly tied to shapes which aren't in use here. I would have to try to create shapes on the fly to get it to work at all, and then I would need to re-write everything anyway because there's no way to inject comments.

tests/unit/docs/test_sharedexample.py
kyleknap: oh not the ``assertIn``. I meant the ``print`` statement is showing up after every other ``assertIn``

jamesls:
@JordonPhillips Sometimes there are additional keys in the ResponseMetadata (I believe with S3).  What if we just include the fact that the key will be there without calling out its contents?  Maybe something like:

```
"ResponseMetadata": {...}
```

JordonPhillips:
I went and re-wrote it to address some concerns and to be more consistent with how we generate some other documentation.

botocore/docs/method.py
```diff
@@ -235,3 +238,8 @@ def document_model_driven_method(section, method_name, operation_model,
                 include=include_output, exclude=exclude_output)
     else:
         return_section.write(':returns: None')
+
+    # Add the shared examples
+    if shared_examples:
```
kyleknap: I wonder if we should add this here because this could affect other (non-client) methods as this gets used for waiters, paginators, and resources if shared_examples gets set. I think it would be better if we just call this ``document_shared_examples`` method in the client as the examples do not pertain to the general case, just the client. I think we can bump it down to this method though once we get examples for waiters, paginators etc.

botocore/docs/sharedexample.py
```diff
@@ -0,0 +1,166 @@
...
+        self._document_dict(output_section, params, comments, [], True)
+        closing_section = output_section.add_new_section('output-close')
+        closing_section.style.end_codeblock()
+
+    def _document(self, section, value, comments, path):
+        if isinstance(value, dict):
+            self._document_dict(section, value, comments, path)
+        if isinstance(value, list):
+            self._document_list(section, value, comments, path)
+        if isinstance(value, string_types):
```
kyleknap: What if the value is not a string (like an integer or datetime). How does that get handled?

botocore/docs/sharedexample.py
```diff
@@ -0,0 +1,166 @@
...
+            path.pop()
+        list_section_end = list_section.add_new_section('ending-bracket')
+        self._end_nested_value(list_section_end, '],')
+
+    def _document_str(self, section, value, comments, path):
+        comment = self._get_comment(path, comments)
+        section.write("'%s',%s" % (value, comment))
+
+    def _get_comment(self, path, comments):
+        key = re.sub('^\.', '', ''.join(path))
```
kyleknap: I was curious on what the reason for the path is for because it is passed around for all of the methods. May be good to add comments in the code to explain why you are appending to the list.

botocore/loaders.py
```diff
@@ -94,6 +94,7 @@
 
 """
 import os
+import glob
```
kyleknap: unused import

tests/unit/docs/test_sharedexample.py
```diff
@@ -0,0 +1,170 @@
+# Copyright 2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kyleknap: It would be nice if we can add a test in either ``test_client.py`` or ``test_service.py`` to ensure that the example get plummed into the documentation for a client.

kyleknap:
Thanks for making the change to use sections. It looks much better. I just had a few more comments/questions.

botocore/docs/sharedexample.py
JordonPhillips: It wouldn't. Aside from those two types, what else would I need to handle? Would you be supportive of treating unknown types as strings? It wouldn't be ideal, but we wouldn't drop them on the floor at least.

kyleknap: Here is the complete list: https://github.com/boto/botocore/blob/develop/botocore/docs/utils.py#L16. I would be in support of treating unknown types as strings.

botocore/docs/method.py
```diff
@@ -125,7 +125,7 @@ def document_model_driven_method(section, method_name, operation_model,
                                  example_prefix=None, include_input=None,
                                  include_output=None, exclude_input=None,
                                  exclude_output=None, document_output=True,
-                                 include_signature=True):
+                                 include_signature=True, shared_examples=None):
```
kyleknap: ``shared_examples`` are no longer needed.

botocore/docs/sharedexample.py
```diff
@@ -0,0 +1,190 @@
...
+        closing_section.style.new_line()
+        closing_section.write('print(response)')
+        closing_section.style.end_codeblock()
+
+    def document_output(self, section, example):
+        output_section = section.add_new_section('output')
+        output_section.writeln('Expected Output:')
+        output_section.style.start_codeblock()
+
+        if 'output' in example:
```
kyleknap: this could be simplified to ``params = example.get('output', {})``

botocore/docs/sharedexample.py
```diff
@@ -0,0 +1,190 @@
...
+                     parameters. This is used to find the equivalent location
+                     in the comments dictionary.
+        """
+        if isinstance(value, dict):
+            self._document_dict(section, value, comments, path)
+        elif isinstance(value, list):
+            self._document_list(section, value, comments, path)
+        elif isinstance(value, numbers.Number):
+            self._document_number(section, value, comments, path)
+        elif isinstance(value, datetime):
```
kyleknap: Will this value be datetime? I am assuming this value is coming straight from the loaded json and it does not convert to datetime. Do you know what the indicator in the example spec on if a value is a timestamp?

botocore/docs/sharedexample.py
```diff
@@ -0,0 +1,190 @@
...
+            self._document_datetime(section, value, comments, path)
+        elif isinstance(value, string_types):
+            self._document_str(section, value, comments, path)
```
kyleknap: Yeah let's just fallback to string if nothing gets caught in this if statement tree so we do not leave it blank if the value is missed.

kyleknap:
Looks better. So we talked about this with @mtdowling about how to handle timestamps by traversing the model as we traverse the example. I would recommend looking at the shape traverser in the ``example.py`` and ``params.py`` or in the validation part of the serializer works as well.

botocore/docs/sharedexample.py
```diff
@@ -0,0 +1,207 @@
...
+class SharedExampleDocumenter(object):
+    def document_shared_example(self, example, prefix, section,
+                                operation_model):
+        """Documents a single shared example based on its definition.
+
+        :param example: The model of the example
+
+        :param prefix: The prefix to use in the method example.
+
+        :param section: The section to write to.
```
kyleknap: Add ``operation_model`` in documentation

botocore/docs/sharedexample.py
```diff
@@ -0,0 +1,207 @@
...
+        :param section: The section to add the docs to.
+        :param value: The input / output values representing the parameters that
+                      are included in the example.
+        :param comments: The dictionary containing all the comments to be
+                         applied to the example.
+        :param path: A list describing where the documenter is in traversing the
+                     parameters. This is used to find the equivalent location
+                     in the comments dictionary.
+        """
+        if isinstance(value, dict):
```
kyleknap: Based on the logic you assume that all dictionaries will be ``structures``. However, dictionaries can be maps as well, which will affect how you parse the model. I think you will have to update the if statements to account for maps.

kyleknap:
Looks good. I think you just need to account for ``map`` types and this should be good to get merged.

tests/unit/docs/test_sharedexample.py
```diff
@@ -0,0 +1,246 @@
...
+            "foo.bar(",
+            "    foo='bar'",
+            ")"
+        ])
+        self.remove_param('foo')
+
+    def test_dict_example(self):
+        self.add_shape({
+            'bar': {
+                "type": "structure",
```
kyleknap: Let's get a test for map shapes.

kyleknap:
Almost there. Let's just lets get a test for maps as well.

JordonPhillips:
:+1:

botocore/docs/sharedexample.py
```diff
@@ -0,0 +1,212 @@
...
+from datetime import datetime
+
+
+class SharedExampleDocumenter(object):
+    def document_shared_example(self, example, prefix, section,
+                                operation_model):
+        """Documents a single shared example based on its definition.
+
+        :param example: The model of the example
+
```
mtdowling: Any reason some of your docs are double-spaced and others are not?

JordonPhillips: They were auto generated by my IDE with single space in that case, fixing

mtdowling:
It's not a big deal, but I think some of the comments could end up being pretty long, which makes me wonder if they should instead be placed on the previous line rather than after the comma.

Otherwise, :shipit:

kyleknap:
:ship: Looks good.

PR #711, opened 2015-11-04
Detect non-ascii in credential

rayluo:
Also provides better message to customer. This will resolve aws/aws-cli#708

Did not patch RefreshableCredentials, because it is less error-prone.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -124,9 +124,9 @@ class Credentials(object):
             method = 'explicit'
         self.method = method
 
-        self.normalize()
+        self._normalize()
 
-    def normalize(self):
+    def _normalize(self):
         # Keys would sometimes (accidentally) contain non-ascii characters.
         # It would cause a confusing UnicodeDecodeError in Python 2.
         # We explicitly convert them into unicode to avoid such error.
@@ -164,9 +164,9 @@ class RefreshableCredentials(Credentials):
         self._expiry_time = expiry_time
         self._time_fetcher = time_fetcher
         self.method = method
-        self.normalize()
+        self._normalize()
 
-    def normalize(self):
+    def _normalize(self):
         self._access_key = botocore.compat.unicode(self._access_key, 'utf-8')
         self._secret_key = botocore.compat.unicode(self._secret_key, 'utf-8')
 
@@ -245,7 +245,7 @@ class RefreshableCredentials(Credentials):
         self.token = data['token']
         self._expiry_time = parse(data['expiry_time'])
         logger.debug("Retrieved credentials will expire at: %s", self._expiry_time)
-        self.normalize()
+        self._normalize()
 
 
 class CredentialProvider(object):
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -23,6 +23,7 @@ from dateutil.parser import parse
 from dateutil.tz import tzlocal
 
 import botocore.config
+import botocore.compat
 from botocore.compat import total_seconds
 from botocore.exceptions import UnknownCredentialError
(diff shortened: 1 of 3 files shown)


botocore/credentials.py
```diff
@@ -123,6 +126,15 @@ def __init__(self, access_key, secret_key, token=None,
...
         self.method = method
 
+    def _assert_ascii(self, s):
+        # Possible Non-ASCII in input, which presumably came from an incorrect
+        # copy-and-paste, would cause confusing UnicodeDecodeError later.
+        # We intercept it here, and provide a meaningful message to customer.
+        try:
+            s.decode('ascii')
+        except UnicodeDecodeError:
+            raise ValueError('Non-ASCII character found in credential: ' + s)
```
mtdowling: I wonder if putting the credentials in the exception message has any security implications? If we think most of the errors here will be copy paste, then the provided credentials should be valid credentials + some invalid characters. Are you concerned that we would potentially be dumping out plain-text credentials to logfiles?

jamesls:
I would prefer if we just fix this at the source, rather than patching a single concrete class.  Customer are free to provide their own subclasses of `Credentials` and it seems like an unnecessary burden we're placing on them.   Other possible options:

* Check when we read the akid/sak from the config file.  Can we just decode at that point?
* Can we update auth to handle the decode?  Rather than forcing the implicit decode to ascii, just decode to utf-8.

rayluo:
Yep. Actually I noticed the previous attempt did not work well at Python 3 either. The new PR still tries to solve it in the Credentials class because:

* It would be difficult to patch all the different Auth methods in auth.py (and who knows when we will have new Auth5?)
* Now the "fix" in Credentials class looks more natural anyway. We simply ensure it stores unicode.

jamesls:
Looked into the problem in more detail as well as you're latest changes. I'm still not sure this is the right approach.

If I understand the issue properly it's essentially because we assume the access/secret keys are values we can `.encode('utf-8')`.  The issue is that reading a value from the config file returns the data already utf-8 encoded (or more specifically it doesn't decode the data for you).

If that's the case, isn't the right approach to ensure that when we read the values from the config file that we're reading them as the correct unicode/str type?  That way any consuming code won't have to perform this conversion.

rayluo:
To be precise, the issue is caused by Python 2 trying to auto-cast the raw credential string into unicode by an implicit `.decode('ascii')`, which fails when the raw string accidentally contains non-ascii content.

There are multiple Credential consumers in auth.py, and multiple Credential providers in credentials.py. And there may be more in future which may not pay attention to this detail. So it seems reasonable to have the microsurgery in the single critical point, the Credential class. I can add more safely logic though, to ensure we do not have extra `.decode("utf-8")` when the raw input happens to be unicode already, so that it would be future-proof. Will that helps?

botocore/credentials.py
```diff
@@ -151,6 +164,11 @@ def __init__(self, access_key, secret_key, token,
         self._expiry_time = expiry_time
         self._time_fetcher = time_fetcher
         self.method = method
+        self.normalize()
+
+    def normalize(self):
+        self._access_key = botocore.compat.unicode(self._access_key, 'utf-8')
+        self._secret_key = botocore.compat.unicode(self._secret_key, 'utf-8')
```
rayluo: FYI. In case you may ask, why do we have to override the same name function in the base class, instead of reuse it. The reason is, super(...).normalize() will trigger the setter function of access_key and secret_key, which trigger _refresh() DURING initialization,  which will cause some existing test cases to fail.

Reimplementing it here is not too difficult. So I choose this route.

jamesls:
:shipit:  Let's just make `normalize()` internal and call it good.

PR #714, opened 2015-11-09
Skip convertion if the input is already unicode

rayluo:
It happens when the credential comes from iam-role.

### DIFF
--- a/botocore/compat.py
+++ b/botocore/compat.py
@@ -112,7 +112,7 @@ else:
         return inspect.getargspec(func)[2]
 
     def ensure_unicode(s, encoding='utf-8', errors='strict'):
-        if type(s) == six.text_type:
+        if isinstance(s, six.text_type):
             return s
         return unicode(s, encoding, errors)
 
--- a/botocore/compat.py
+++ b/botocore/compat.py
@@ -60,7 +60,7 @@ if six.PY3:
         # changes when using getargspec with functools.partials.
         return inspect.getfullargspec(func)[2]
 
-    def unicode(s, encoding=None, errors=None):
+    def ensure_unicode(s, encoding=None, errors=None):
         # NOOP in Python 3, because every string is already unicode
         return s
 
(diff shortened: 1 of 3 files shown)


tests/unit/test_credentials.py
```diff
@@ -54,10 +54,19 @@ def path(filename):
...
-    def test_detect_nonascii_character(self):
-        c = credentials.Credentials('foo\xe2\x80\x99', 'bar\xe2\x80\x99')
+    def _ensure_credential_is_normalized_as_unicode(self, access, secret):
+        c = credentials.Credentials(access, secret)
         self.assertTrue(isinstance(c.access_key, type(u'u')))
         self.assertTrue(isinstance(c.secret_key, type(u'u')))
+        return True
+
+    def test_detect_nonascii_character(self):
+        self.assertTrue(self._ensure_credential_is_normalized_as_unicode(
```
jamesls: I'm also ok with not having the `assertTrue` call and just calling the _ensure method in this test..  The `_ensure_credential_is_normalized_as_unicode` will raise an AssertionError if it fails anways so it simplifies the test a bit.

botocore/compat.py
```diff
@@ -111,7 +111,11 @@ def set_socket_timeout(http_response, timeout):
     def accepts_kwargs(func):
         return inspect.getargspec(func)[2]
 
-    unicode = unicode
+    _unicode = unicode
+    def unicode(s, encoding='utf-8', errors='strict'):
+        if type(s) == six.text_type:
```
jamesls: isinstance(s, six.text_type).

botocore/compat.py
```diff
@@ -111,7 +111,11 @@ def set_socket_timeout(http_response, timeout):
     def accepts_kwargs(func):
         return inspect.getargspec(func)[2]
 
-    unicode = unicode
+    _unicode = unicode
+    def unicode(s, encoding='utf-8', errors='strict'):
```
jamesls: I'd prefer ensure_unicode instead of unicode given it doesn't quite behave the same as the built in unicode now.

jamesls:
Looks good, just some small feedback.  Thanks for fixing.

jamesls:
:shipit:  Thanks!

PR #718, opened 2015-11-12
Document the Body parameter properly. This is an FAQ.

rayluo:
This will add a new sentence at the end of the current documentation for that Body parameter.

This fixes boto/boto3#330

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -592,6 +592,11 @@ BUILTIN_HANDLERS = [
     # Cloudsearchdomain search operation will be sent by HTTP POST
     ('request-created.cloudsearchdomain.Search',
      change_get_to_post),
+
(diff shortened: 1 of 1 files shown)


kyleknap:
I would prefer that this is built-in to our documentation as opposed to it being a customization. S3 ``GetObject`` is not the only operation that returns a file-like object to read from. I would double check but I believe the condition is if the parameter is a streaming output payload then it becomes a file like object. You can check with ``operation_model.has_streaming_output``.

jamesls:
I agree with @kyleknap .  This would make it more general purpose and potentially reduce customer confusion in the future.

rayluo:
Incorporating feedback and coming up with a totally different approach. So this PR is abandoned. Please review #723 instead.

PR #723, opened 2015-11-19
Document streaming body

rayluo:
Automatically document all operations with a streaming body. Previously they were incorrectly documented as bytes.

This fixes boto/boto3#330

### DIFF
--- a/botocore/docs/example.py
+++ b/botocore/docs/example.py
@@ -16,8 +16,6 @@ from botocore.model import get_streaming_body
 
 
 class BaseExampleDocumenter(ShapeDocumenter):
-    streaming_shape = None
-
     def document_example(self, section, shape, prefix=None, include=None,
                          exclude=None):
         """Generates an example based on a shape
@@ -36,7 +34,7 @@ class BaseExampleDocumenter(ShapeDocumenter):
         :param exclude: The names of the parameters to exclude from
             documentation.
         """
-        self.streaming_shape = get_streaming_body(shape)
+        self._context['streaming_shape'] = get_streaming_body(shape)
         history = []
         section.style.new_line()
         section.style.start_codeblock()
@@ -52,7 +50,7 @@ class BaseExampleDocumenter(ShapeDocumenter):
     def document_shape_default(self, section, shape, history, include=None,
                                exclude=None, **kwargs):
         py_type = py_default(shape.type_name)
-        if self.streaming_shape == shape:
+        if self._context.get('streaming_shape') == shape:
             py_type = 'StreamingBody()'
         section.write(py_type)
 
--- a/botocore/docs/example.py
+++ b/botocore/docs/example.py
@@ -12,9 +12,12 @@
 # language governing permissions and limitations under the License.
 from botocore.docs.shape import ShapeDocumenter
 from botocore.docs.utils import py_default
+from botocore.model import get_streaming_body
 
 
 class BaseExampleDocumenter(ShapeDocumenter):
+    streaming_shape = None
+
     def document_example(self, section, shape, prefix=None, include=None,
                          exclude=None):
         """Generates an example based on a shape
@@ -33,6 +36,7 @@ class BaseExampleDocumenter(ShapeDocumenter):
         :param exclude: The names of the parameters to exclude from
             documentation.
         """
+        self.streaming_shape = get_streaming_body(shape)
         history = []
         section.style.new_line()
         section.style.start_codeblock()
@@ -48,8 +52,7 @@ class BaseExampleDocumenter(ShapeDocumenter):
     def document_shape_default(self, section, shape, history, include=None,
                                exclude=None, **kwargs):
         py_type = py_default(shape.type_name)
-        if (self._operation_model and
-                self._operation_model.streaming_output_shape == shape):
+        if self.streaming_shape == shape:
             py_type = 'StreamingBody()'
         section.write(py_type)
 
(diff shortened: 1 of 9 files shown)


jamesls:
High level:

1. Can you show an example of the rendered docs?
2. I see there's functional tests, but I feel we should still have unit tests for this.

docs/source/reference/response.rst
```diff
@@ -0,0 +1,11 @@
+.. _ref-response:
+
+==================
+Response Reference
+==================
+
+botocore.response
+-----------------
+
+.. automodule:: botocore.response
```
jamesls: Won't this document any additional functions we add to this module?  I'd prefer we just call out StreamingBody for now.

botocore/response.py
```diff
@@ -67,6 +67,11 @@ def set_socket_timeout(self, timeout):
             raise
 
     def read(self, amt=None):
+        """
+        Read at most amt bytes from the stream.
```
jamesls: This should be on the first line of the docstring, line 70 (pep257).

rayluo: Thanks! Will do.

tests/unit/docs/test_method.py
```diff
@@ -301,3 +301,16 @@ def test_exclude_output(self):
...
+        self.json_model['shapes']['Blob'] = {'type': 'blob'}
+        self.json_model['shapes']['SampleOperationInputOutput']['payload'] = \
+            'Body'
+        document_model_driven_method(
+            self.doc_structure, 'foo', self.operation_model,
+            event_emitter=self.event_emitter,
+            method_description='This describes the foo method.',
+            example_prefix='response = client.foo'
+        )
+        self.assert_contains_line('StreamingBody')
```
jamesls: Can we make this assertion check that it's associated with the right param name as well.  Similar to what the tests above this one do.

kyleknap: +1

botocore/docs/shape.py
```diff
@@ -19,10 +19,12 @@
 class ShapeDocumenter(object):
     EVENT_NAME = ''
 
-    def __init__(self, service_name, operation_name, event_emitter):
+    def __init__(self, service_name, operation_name, event_emitter,
+                 operation_model=None):
```
jamesls: Is there a reason this is an optional param?

kyleknap:
Looks fine. It is sort of weird that we pass the ``operation_model`` to the constructor of ``ShapeDocumenter`` when we access properties of ``operation_model`` to pass to the constructor. But I think we can wait on that refactoring this. :ship:

botocore/docs/example.py
```diff
@@ -12,9 +12,12 @@
 # language governing permissions and limitations under the License.
 from botocore.docs.shape import ShapeDocumenter
 from botocore.docs.utils import py_default
+from botocore.model import get_streaming_body
 
 
 class BaseExampleDocumenter(ShapeDocumenter):
+    streaming_shape = None
```
kyleknap: I would prefer this is not an attribute of the class. Much rather have it live in the constructor as an attribute of a ``ShapeDocumenter`` instance. If you are worried that ``streaming_shape ``is too specific, you can create a context attribute that is a dictionary that is used to store a bunch of this sort of metadata.

kyleknap:
I like this approach more. Just had one comment.

botocore/model.py
```diff
@@ -321,6 +321,20 @@ def signature_version(self, value):
         self._signature_version = value
 
 
+def get_streaming_body(shape):
```
jamesls: Not a fan of having this be a module function.  This seems like a method that should live on a shape class.

botocore/docs/example.py
```diff
@@ -48,6 +50,8 @@ def document_recursive_shape(self, section, shape, **kwargs):
     def document_shape_default(self, section, shape, history, include=None,
                                exclude=None, **kwargs):
         py_type = py_default(shape.type_name)
+        if self._context.get('streaming_shape') == shape:
```
mtdowling: Is the caching of `streaming_shape` inside of a `self._context` value necessary? Do we need to figure out if it's streaming or not multiple times for a single shape?

rayluo: You are right. It is only for documentation, and we won't call it too often. I have removed the cache in later commit.

mtdowling:
This seems fine. :ship:

JordonPhillips:
looks good :shipit:

botocore/docs/method.py
```diff
@@ -209,6 +209,8 @@ def document_model_driven_method(section, method_name, operation_model,
         return_section.style.indent()
         return_section.style.new_line()
 
+        context = {'streaming_shape': operation_model.get_streaming_output()}
```
kyleknap: I like this actually more than what I originally suggested. The only possible issue that I could see is if a streaming output shape is shared elsewhere in the output shape but is not actually a streaming output, but I am having trouble seeing this happening.

kyleknap:
:ship: on the assumption that a streaming output shape will never be reused in the output shape as a non streaming output.

botocore/docs/method.py
rayluo: Now I remembered that I checked that during my preliminary research, and I did a manual check just now to double confirm that, for the following sample model definition where the 2 members share the same `blobType` shape, they will end up as 2 different shape instances despite they are with the same type. So there would be no problem when we later use `if self._context.get('streaming_shape') == shape`.

            'shapes': {
                'OperationNameResponse': {
                    'type': 'structure',
                    'members': {
                        "Blob": {'shape': 'blobType'},
                        "Body": {'shape': 'blobType'}
                    },
                    'payload': 'Body'
                },
                'blobType': {
                    'type': 'blob'
                }
            }

So, no need to worry. It is a good question. Thanks!

PR #724, opened 2015-11-19
Add support for default client config on session

kyleknap:
Exposed a set_default_client_config() and get_default_client_config() that will be used as the config for when clients are created. If a config is also passed to create_client() the config will be merged into the default client config and only override default client config values with non-default values provided to the client config. Also added a merge() method on client config to help facilitate the merging of configs.

This is really going to help with the unification of model-driven CLI commands and custom commands in terms of handling arguments that need to be passed to the creation of clients. For example, it will allow me to remove all of the logic that I added to the s3 commands for this CLI PR: https://github.com/aws/aws-cli/pull/1649


botocore/client.py
```diff
@@ -655,3 +655,32 @@ def _validate_s3_configuration(self, s3):
...
+        :returns: A config object built from the merged values of both
+            config objects.
+        """
+        # Make a copy of the current attributes in the config object.
+        config_properties = copy.copy(vars(self))
+
+        # Merge the config object's attributes with the other config
+        # object's attributes.
+        for property_key, property_value in vars(other_config).items():
+            if property_key in ['connect_timeout', 'read_timeout']:
```
jamesls: Why does this need to be special cased?  If there no way to write this in a generic way?

botocore/session.py
```diff
@@ -727,15 +748,28 @@ def create_client(self, service_name, region_name=None, api_version=None,
...
+            default config with the config provided to this call.
 
         :rtype: botocore.client.BaseClient
         :return: A botocore client instance
 
         """
+        default_client_config = self.get_default_client_config()
+        # If a config is provided and a default config is set, then
+        # use the config resulting from merging the two.
+        if config and default_client_config:
```
jamesls: `is not None`.

botocore/session.py
```diff
@@ -727,15 +748,28 @@ def create_client(self, service_name, region_name=None, api_version=None,
...
+            config = default_client_config.merge(config)
+        # If a config was not provided then use the default
+        # client config from the session
+        elif default_client_config:
```
jamesls: `is not None`

botocore/client.py
kyleknap: What would be generic? In terms of searching for timeout in the name instead of searching for if the key is in the list? Or how all config values are handled? The most general logic I could have is:
```py
for property_key, property_value in vars(other_config).items():
    if property_value not in [None, DEFAULT_TIMEOUT]:
        config_properties[property_key] = property_value
```
But that pretty much black lists any config option that uses the value ``60``. Given none of the other config options use integers, I think I would work for now. I was just trying to prevent a possible future bug if an integer property gets added to the config later on.

tests/unit/test_client.py
```diff
@@ -1106,3 +1106,61 @@ def test_verify_takes_precedence_over_scoped_config(self):
...
+        new_config = config.merge(other_config)
+        self.assertEqual(new_config.connect_timeout, 55)
+
+    def test_merge_override_connect_timeout(self):
+        config = client.Config(connect_timeout=55)
+        other_config = client.Config(connect_timeout=75)
+        new_config = config.merge(other_config)
+        self.assertEqual(new_config.connect_timeout, 75)
+
+    def test_merge_default_does_not_override_read_timeout(self):
```
jamesls: Why is this test needed given `test_merge_default_does_not_override_connect_timeout`?

tests/unit/test_client.py
```diff
@@ -1106,3 +1106,61 @@ def test_verify_takes_precedence_over_scoped_config(self):
                               scoped_config={'ca_bundle': 'mycerts.pem'})
         call_kwargs = self.endpoint_creator.create_endpoint.call_args[1]
         self.assertEqual(call_kwargs['verify'], 'verify.pem')
+
+
+class TestConfig(unittest.TestCase):
```
jamesls: I didn't see any tests where a config value doesn't exist in the base config but does exist in the `other_config`.

botocore/client.py
jamesls: Generic in the sense that I would not expect a general config merging method to have to special case specific config values such as `connect_timeout` and `read_timeout`.

tests/unit/test_client.py
kyleknap: There was one already. I added a check though to make sure that there is no default set prior to sending the config object

kyleknap:
@jamesls I updated the PR to make the merge implementation more general. It required me storing the user provided options and make the signature use ``*args`` and ``**kwargs``. I added some validation as well to help ensure user constructs the config correctly. Let me know if I am missing any checks for these.

tests/unit/test_client.py
```diff
@@ -1106,3 +1108,72 @@ def test_verify_takes_precedence_over_scoped_config(self):
...
+        # Ensure that the values all stayed the same in the new config
+        self.assertEqual(config_properties, vars(new_config))
+
+    def test_merge_overrides_values(self):
+        config = client.Config(region_name='us-east-1')
+        other_config = client.Config(region_name='us-west-2')
+        new_config = config.merge(other_config)
+        self.assertEqual(new_config.region_name, 'us-west-2')
+
+    def test_merge_overrides_values_even_when_using_default(self):
```
jamesls: Can we also add this same test but for default integer values?  Would be nice to have a test that catches that we're not relying on `None` or a `False` value.  e.g original_timeout:30, new_timout:60, merged_value: 60.

jamesls:
:shipit:  Looks good, I think `merge()` looks much cleaner now.  Just had one small request for a test.

kyleknap:
I added a test for default timeouts. Merging.

PR #725, opened 2015-11-23
Remove error message strings from tests

jamesls:
This fixes one of the TODOs in the acceptance test runner.

We just need to check that the error response contains
a message.  We don't need to check the contents of that error message.

Relevant lines: https://github.com/boto/botocore/pull/725/files#diff-82cf6007372edb07a2f866f77827bbe3R102


kyleknap:
:ship:

mtdowling:
:ship:

PR #726, opened 2015-11-23
Automatically set EncodingType to url

JordonPhillips:
Some unicode values can break the XML parser, so we have s3 url encode the keys and decode them on our end.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -519,16 +519,20 @@ def change_get_to_post(request, **kwargs):
         request.url, request.data = request.url.split('?', 1)
 
 
-def set_list_objects_encoding_type_url(params, **kwargs):
+def set_list_objects_encoding_type_url(params, context, **kwargs):
     if 'EncodingType' not in params:
+        # We set this context so that we know it wasn't the customer that
+        # requested the encoding.
+        context['EncodingTypeAutoSet'] = True
         params['EncodingType'] = 'url'
 
 
-def decode_list_object(parsed, **kwargs):
+def decode_list_object(parsed, context, **kwargs):
     # This is needed because we are passing url as the encoding type. Since the
     # paginator is based on the key, we need to handle it before it can be
     # round tripped.
-    if 'Contents' in parsed and parsed.get('EncodingType') == 'url':
+    if 'Contents' in parsed and parsed.get('EncodingType') == 'url' and \
+                    context.get('EncodingTypeAutoSet'):
         for content in parsed['Contents']:
             content['Key'] = unquote_str(content['Key'])
 
@@ -549,7 +553,8 @@ BUILTIN_HANDLERS = [
 
     ('before-parameter-build.s3', validate_bucket_name),
 
-    ('before-parameter-build.s3.ListObjects', set_list_objects_encoding_type_url),
+    ('before-parameter-build.s3.ListObjects',
+     set_list_objects_encoding_type_url),
     ('before-call.s3.PutBucketTagging', calculate_md5),
     ('before-call.s3.PutBucketLifecycle', calculate_md5),
     ('before-call.s3.PutBucketLifecycleConfiguration', calculate_md5),
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -22,10 +22,10 @@ import logging
 import xml.etree.cElementTree
 import copy
 import re
-import string
 import warnings
 
-from botocore.compat import urlsplit, urlunsplit, unquote, json, quote, six
+from botocore.compat import urlsplit, urlunsplit, unquote, \
+    json, quote, six, unquote_str
 from botocore.docs.utils import AutoPopulatedParam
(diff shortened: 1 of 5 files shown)


botocore/handlers.py
```diff
@@ -519,6 +519,20 @@ def change_get_to_post(request, **kwargs):
...
+def set_list_objects_encoding_type_url(params, **kwargs):
+    if 'EncodingType' not in params:
+        params['EncodingType'] = 'url'
+
+
+def decode_list_object(parsed, **kwargs):
+    # This is needed because we are passing url as the encoding type. Since the
+    # paginator is based on the key, we need to handle it before it can be
+    # round tripped.
+    if 'Contents' in parsed and parsed.get('EncodingType') == 'url':
```
mtdowling: I wonder if we should only automatically decode the URI encoded values if the EncodingType was not explicitly set by the user? For example, if a user actually looks at the S3 API and sees that you can set and EncodingType, then they might reasonably expect that the results they receive are encoded... In fact, that's probably the behavior they are expecting right now. If we automatically start decoding these, it could be a breaking change.

JordonPhillips: There's not really any way to know that though.

mtdowling: Yeah, I think we'd have to add support for that in order to make it backwards compatible.

tests/integration/test_s3.py
```diff
@@ -176,12 +176,12 @@ def test_can_delete_urlencoded_object(self):
         bucket_contents = self.client.list_objects(
             Bucket=self.bucket_name)['Contents']
         self.assertEqual(len(bucket_contents), 1)
-        self.assertEqual(bucket_contents[0]['Key'], 'a+b/foo')
+        self.assertEqual(bucket_contents[0]['Key'], u'a+b/foo')
```
kyleknap: Why do you need to add the unicode prefix to these? I feel like these should still pass with the unicode prefixed removed.

kyleknap:
Implementation looks fine. It would be nice to have an integration test for this by adding a key that would normally break the xml parser and show how it is now fixed with autopopulation of encoding type and also show that it is still url encoded (not automatically decoded) if the user specified it.

tests/integration/test_s3.py
```diff
@@ -333,6 +333,15 @@ def test_unicode_key_put_list(self):
...
         self.assertEqual(parsed['Body'].read().decode('utf-8'), 'foo')
 
+    def test_unicode_system_character(self):
+        # Verify we can use a unicode system character which would normally
+        # break the xml parser
+        key_name = 'foo\x08'
+        self.create_object(key_name)
+        parsed = self.client.list_objects(Bucket=self.bucket_name)
+        self.assertEqual(len(parsed['Contents']), 1)
+        self.assertEqual(parsed['Contents'][0]['Key'], key_name)
```
kyleknap: Could we also do a second call where we specify ``EncodingType=url`` and show that it is not being url decoded? You could just add it to this test.

kyleknap:
Just one more thing. It should be good to merge. Also make sure you have a PR for the CLI to update it so we can merge this and the corresponding CLI PR to avoid test failures when we just merge this PR.

JordonPhillips:
:+1:

kyleknap:
Thanks :ship:. We can merge this when the CLI PR is made to update the CLI based off of these changes.

mtdowling:
:shipit:

botocore/handlers.py
```diff
@@ -519,6 +519,24 @@ def change_get_to_post(request, **kwargs):
...
+        kwargs['context']['EncodingTypeAutoSet'] = True
+        params['EncodingType'] = 'url'
+
+
+def decode_list_object(parsed, **kwargs):
+    # This is needed because we are passing url as the encoding type. Since the
+    # paginator is based on the key, we need to handle it before it can be
+    # round tripped.
+    if 'Contents' in parsed and parsed.get('EncodingType') == 'url' and \
+                    kwargs['context'].get('EncodingTypeAutoSet') == True:
```
jamesls: `and context.get('EncodingTypeAutoSet', False):`   or just:

 `and context.get('EncodingTypeAutoSet')`

botocore/handlers.py
```diff
@@ -519,6 +519,24 @@ def change_get_to_post(request, **kwargs):
         request.url, request.data = request.url.split('?', 1)
 
 
+def set_list_objects_encoding_type_url(params, **kwargs):
+    if 'EncodingType' not in params:
+        # We set this context so that we know it wasn't the customer that
+        # requested the encoding.
+        kwargs['context']['EncodingTypeAutoSet'] = True
```
jamesls: You can also just accept `context` as a function param so you don't have to pull it from `**kwargs`.  `**kwargs` was really just for kwargs you don't care about and to future proof event handlers.

jamesls:
:shipit:  Just had a few small comments.

PR #727, opened 2015-11-24
Skip sts/importexport acceptance tests

jamesls:
We have the same behavior in our integration smoke tests.

See: https://github.com/boto/botocore/blob/develop/tests/integration/test_smoke.py#L86-L90


kyleknap:
:ship:

PR #728, opened 2015-11-24
Content of .zip file bytestring needs not be base64-encoded

grisha:
There is a mistake in the documentation, there is no need to base64-encode the content of the zip file to be uploaded, botocore will take care of that.


mtdowling:
These files are sourced from an upstream, AWS internal repository, and these changes would need to be applied at that level. I'll forward this on to our documentation team to try to address the issue.

PR #730, opened 2015-11-25
Update Bucket Lifecycle Rule Transition shape

DenverJ:
A bucket with multiple transitions in a single lifecycle rule was not correctly returning. Instead it was only returning the first transition.

Eg. Before
`[{u'Status': 'Enabled', u'Prefix': '', u'Transition': {u'Days': 30, u'StorageClass': 'STANDARD_IA'}, u'Expiration': {u'Days': 365}, u'ID': 'test'}]`

After
`[{u'Status': 'Enabled', u'Prefix': '', u'Transition': [{u'Days': 30, u'StorageClass': 'STANDARD_IA'}, {u'Days': 90, u'StorageClass': 'GLACIER'}], u'Expiration': {u'Days': 365}, u'ID': 'test'}]`

### DIFF
--- a/botocore/data/s3/2006-03-01/service-2.json
+++ b/botocore/data/s3/2006-03-01/service-2.json
@@ -4511,7 +4511,7 @@
           "shape":"ExpirationStatus",
           "documentation":"If 'Enabled', the rule is currently being applied. If 'Disabled', the rule is not currently being applied."
         },
-        "Transition":{"shape":"Transition"},
+        "Transition":{"shape":"TransitionList"},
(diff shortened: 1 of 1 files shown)


mtdowling:
`GetBucketLifecycle` (`get_bucket_lifecycle`) is deprecated in favor of `GetBucketLifecycleConfiguration` (`get_bucket_lifecycle_configuration`).

When this feature was initially shipped by S3, it supported only a single rule. They since added the ability to specify multiple rules, which meant that we needed to support the version that provided only a single value as a dict for backwards compatibility, but also support a list of dicts. This meant adding a new operation and marking the old form as deprecated.

PR #734, opened 2015-12-03
Moved to containerized infrastrucutre and added testing with Python 3.5 to the build

thedrow:


thedrow:
Seems that the only tests that fail relate to invoking help() on docstrings. No idea why the fail though.

JordonPhillips:
Thanks for the interest! `sudo: false` is already there, and while we're functional in 3.5, we can't add it to the list until all the tests pass (even if only docs tests fail).

thedrow:
Do you have an idea why the docs tests are failing? I looked into it and it doesn't make sense to me.

JordonPhillips:
At a glance I'm not sure, but it might be partially due to the new `OrderedDict` implementation. Python 3 imposes no particular ordering on dictionaries by default, and so we use `OrderedDict`s where necessary. It could be that there are more dictionaries that need to be converted due to a change in the unspecified ordering from 3.4 to 3.5.

jamesls:
Ended up tracking this down.  This is due to a regression in python3.5.  We'll have to update our code to handle this.  Working on a PR.

jamesls:
Fixed via https://github.com/boto/botocore/pull/742.   Ended up needing to pull in a py3.5 version in the .travis.yml file for #742 to ensure the doc tests pass on py35.

PR #735, opened 2015-12-04
Feature cloudfront sign

rayluo:
Provide the signer class. It will be used in CLI soon.

Also note that there is no (hard) dependency of rsa in this implementation.

### DIFF
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -18,6 +18,7 @@ import base64
 
 import botocore
 import botocore.auth
+from botocore.compat import six
 from botocore.awsrequest import create_request_object, prepare_request_dict
 from botocore.exceptions import UnknownSignatureVersionError
 from botocore.exceptions import UnknownClientMethodError
@@ -266,7 +267,7 @@ class CloudFrontSigner(object):
         :type date_less_than: datetime
         :param date_less_than: The URL will expire after the time has passed
 
-        :type policy: bytes
+        :type policy: str
         :param policy: The custom policy, possibly built by self.build_policy()
 
         :rtype: str
@@ -277,6 +278,8 @@ class CloudFrontSigner(object):
             raise ValueError('Need to provide either date_less_than or policy')
         if policy is None and date_less_than is not None:
             policy = self.build_policy(url, date_less_than)
+        if isinstance(policy, six.text_type):
+            policy = policy.encode('utf8')
         if date_less_than is not None:
             params = ['Expires=%s' % int(datetime2timestamp(date_less_than))]
         else:
@@ -304,8 +307,8 @@ class CloudFrontSigner(object):
         :type ip_address: str
         :param ip_address: Use 'x.x.x.x' for an IP, or 'x.x.x.x/x' for a subnet
 
-        :rtype: bytes
-        :return: The policy in a compact binary string.
+        :rtype: str
+        :return: The policy in a compact string.
         """
         if date_greater_than is None and ip_address is None:
             policy = (
@@ -330,7 +333,7 @@ class CloudFrontSigner(object):
                 sort_keys=True,  # Make it stable
                 separators=(',', ':'))
         LOG.debug("CloudFront policy: %s", policy)
-        return policy.encode('utf8')
+        return policy
 
     def _url_b64encode(self, data):
         return base64.b64encode(
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -315,8 +315,8 @@ class CloudFrontSigner(object):
         """
         # Note:
         # 1. Order in canned policy is significant. Special care has been taken
-        #    to ensure the output will match the order defined by document.
-        #    There is also a test case in this commit to ensure that order.
+        #    to ensure the output will match the order defined by the document.
+        #    There is also a test case to ensure that order.
         #    SEE: http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-canned-policy.html#private-content-canned-policy-creating-policy-statement
         # 2. Albeit the order in custom policy is not required by CloudFront,
         #    we still use OrderedDict internally to ensure the result is stable
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -15,12 +15,6 @@ import weakref
 import logging
 import json
 import base64
-try:
-    maketrans = bytes.maketrans  # For Python 3
-except:
-    from string import maketrans
-
-import rsa
 
 import botocore
 import botocore.auth
@@ -28,7 +22,7 @@ from botocore.awsrequest import create_request_object, prepare_request_dict
 from botocore.exceptions import UnknownSignatureVersionError
 from botocore.exceptions import UnknownClientMethodError
 from botocore.exceptions import UnsupportedSignatureVersionError
-from botocore.utils import fix_s3_host
+from botocore.utils import fix_s3_host, datetime2timestamp
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


botocore/utils.py
```diff
@@ -389,6 +389,23 @@ def parse_to_aware_datetime(value):
     return datetime_obj
 
 
+def datetime2timestamp(dt):
+    """Calculate the timestamp based on the given datetime instance.
+
+    Naive datetime input will be treated as UTC time rather than local time.
```
jamesls: Curious to see what others think about this.  A utility function further down in this module does the same thing, but that was specifically for back-compat to fix a regression.  Not sure if it makes sense to carry that forward.

We'd still have to support the no timezone == UTC in the CLI, but we don't necessarily have to have that throughout the botocore utility methods.  I could go either way.

tests/unit/test_signers.py
```diff
@@ -250,6 +250,63 @@ def test_presigned_url_throws_unsupported_signature_error(self):
...
+        self.assertEqual(signed_url, expected)
+
+    def test_generate_presign_url_with_custom_policy(self):
+        policy = self.signer.build_policy(
+            'foo', datetime.datetime(2016, 1, 1),
+            date_greater_than=datetime.datetime(2015, 12, 1),
+            ip_address='12.34.56.78/9')
+        signed_url = self.signer.generate_presigned_url(
+            'http://test.com/index.html?foo=bar', policy=policy)
+        expected = (
```
jamesls: I would prefer we actually assert the various parts of the URL rather than an exact match, i.e parse the URL and assert the query string dict equals X, the hostname equals X, etc.  You'll get a much better diff on failure.

tests/unit/test_signers.py
```diff
@@ -250,6 +250,63 @@ def test_presigned_url_throws_unsupported_signature_error(self):
             self.signer.generate_presigned_url({})
 
 
+class TestCloudfrontSigner(unittest.TestCase):
+    def setUp(self):
+        self.signer = CloudFrontSigner("MY_KEY_ID", lambda message: b'signed')
+
+    def test_b64encode(self):
+        self.assertEqual(b'dGVzdA__', self.signer._url_b64encode(b'test'))
```
jamesls: Don't test internal functions

tests/unit/test_signers.py
```diff
@@ -250,6 +250,63 @@ def test_presigned_url_throws_unsupported_signature_error(self):
...
+            b'{"Statement":[{"Resource":"foo",'
+            b'"Condition":{"DateLessThan":{"AWS:EpochTime":1451606400}}}]}')
+        self.assertEqual(policy, expected)
+
+    def test_build_custom_policy(self):
+        policy = self.signer.build_policy(
+            'foo', datetime.datetime(2016, 1, 1),
+            date_greater_than=datetime.datetime(2015, 12, 1),
+            ip_address='12.34.56.78/9')
+        expected = (
```
jamesls: Same comment as before, I'd rather just verify the parsed dict.  You'll get a better error diff rather than saying the strings don't match and trying to figure out which key is different.

tests/unit/test_signers.py
rayluo: OK. Added two new helper functions in test cases, and manually ran different combinations of errors to ensure they work.

tests/unit/test_signers.py
rayluo: I like your idea and started working on it.

And, because the order of all keys in the policy dictionary is significant, I tried OrderedDict first. But it turns out the error diff of OrderedDict shows up as a long string, which defeats the purpose.

So I end up using a normal dict (which has better error diff by default), yet I still keep the current string-based diff to ensure the correct order.

botocore/signers.py
```diff
@@ -213,6 +220,126 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+    def __init__(self, key_id, rsa_signer):
+        """Create a CloudFrontSigner.
+
+        :type key_id: str
+        :param key_id: The CloudFront Key Pair ID
+
+        :type rsa_signer: callable
+        :param rsa_signer: An RSA signer.
+               Its only input parameter will be the message to be signed,
+               and its output will be the signed content as a binary string.
```
kyleknap: Also may be good to say that it requires SHA-1

botocore/signers.py
```diff
@@ -213,6 +220,126 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+        """
+        self.key_id = key_id
+        self.rsa_signer = rsa_signer
+
+    def generate_presigned_url(self, url, date_less_than=None, policy=None):
```
kyleknap: Oh I just had a better idea than ``date_less_than``. How does ``expires_at`` or ``expires_on`` sound?

botocore/signers.py
```diff
@@ -213,6 +220,126 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+        """Creates a signed CloudFront URL based on given parameters.
+
+        :type url: str
+        :param url: The URL of the protected object
+
+        :type date_less_than: datetime
+        :param date_less_than: The URL will expire after the time has passed
```
kyleknap: the time --> that date and time

botocore/signers.py
```diff
@@ -213,6 +220,126 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+
+        :type policy: str
+        :param policy: The custom policy, possibly built by self.build_policy()
+
+        :rtype: str
+        :return: The signed URL.
+        """
+        if (date_less_than is not None and policy is not None
+                or date_less_than is None and policy is None):
+            raise ValueError('Need to provide either date_less_than or policy')
```
kyleknap: Would be nice to include in the error message that both cannot be provided as the parameters are mutually exclusive.

botocore/signers.py
```diff
@@ -213,6 +220,126 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+        if policy is None and date_less_than is not None:
+            policy = self.build_policy(url, date_less_than)
+        if isinstance(policy, six.text_type):
+            policy = policy.encode('utf8')
+        if date_less_than is not None:
+            params = ['Expires=%s' % int(datetime2timestamp(date_less_than))]
+        else:
```
kyleknap: Wait why are we building a policy on line 280 if we are going to use ``Expires=`` form if ``date_less_than`` is not ``None``? Do we ever have to build a policy give that date_less_than and policy are mutually exclusive.

botocore/signers.py
```diff
@@ -213,6 +220,126 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+            params = ['Policy=%s' % self._url_b64encode(policy).decode('utf8')]
+        signature = self.rsa_signer(policy)
+        params.extend([
+            'Signature=%s' % self._url_b64encode(signature).decode('utf8'),
+            'Key-Pair-Id=%s' % self.key_id,
+            ])
+        return url + ('&' if '?' in url else '?') + '&'.join(params)
```
kyleknap: A comment here on why you are doing this would be nice. It is a little hard to follow. Breaking it up to its own helper method may be nice as well instead of writing all inline.

botocore/signers.py
```diff
@@ -213,6 +220,126 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+
+    def build_policy(self, resource, date_less_than,
```
kyleknap: I wonder if we need to dangle this off of the cloudfront signer, especially if we do not use it in the generate_presigned_url method? Maybe move it to ``utils.py``?

botocore/signers.py
```diff
@@ -213,6 +220,126 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+
+        :rtype: str
+        :return: The policy in a compact string.
+        """
+        if date_greater_than is None and ip_address is None:
+            policy = (
+                '{"Statement":[{"Resource":"%s",'
+                '"Condition":{"DateLessThan":{"AWS:EpochTime":%s}}}]}') % (
+                resource, int(datetime2timestamp(date_less_than)))
+        else:
```
kyleknap: I am not sure if we need this if statement here. Since resource and date_less_than is always required, can we just always it assume it will be part of the policy that way we will have not the same logic in two places.

botocore/signers.py
```diff
@@ -213,6 +220,126 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+            if date_greater_than:
+                moment = int(datetime2timestamp(date_greater_than))
+                condition["DateGreaterThan"] = {"AWS:EpochTime": moment}
+            custom_policy = {
+                "Statement": [{"Resource": resource, "Condition": condition}]}
+            policy = json.dumps(
+                custom_policy,
+                sort_keys=True,  # Make it stable
+                separators=(',', ':'))
+        LOG.debug("CloudFront policy: %s", policy)
```
kyleknap: Do we need to log this as this is just a helper method and it is not used in flow of generating a signed url? I could see us logging it in the CLI though if we try to build on top of it.

botocore/signers.py
```diff
@@ -213,6 +220,126 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+        return policy
+
+    def _url_b64encode(self, data):
+        return base64.b64encode(
```
kyleknap: It would be good to add a comment on why you are replacing certain characters.

botocore/utils.py
```diff
@@ -389,6 +389,20 @@ def parse_to_aware_datetime(value):
     return datetime_obj
 
 
+def datetime2timestamp(dt, default_timezone=tzutc()):
```
kyleknap: Is tzutc() mutable? If it is, I rather the have default be ``None`` and set the value to ``tzutc()`` in the function if it is ``None``.

botocore/utils.py
```diff
@@ -389,6 +389,20 @@ def parse_to_aware_datetime(value):
...
+    """Calculate the timestamp based on the given datetime instance.
+
+    The optional default_timezone is only used when dt is a naive datetime.
```
kyleknap: It would be good to get complete docstrings for the parameters instead of just that sentence. So something like:
```
:type dt: ...
:param dt: ...
:type default_timezone: ...
:param default_timezone: ...
```

tests/unit/test_utils.py
```diff
@@ -234,6 +235,17 @@ def test_parse_invalid_timestamp(self):
...
 
 
+class TestDatetime2Timestamp(unittest.TestCase):
+    def test_datetime2timestamp_naive(self):
+        self.assertEqual(
+            datetime2timestamp(datetime.datetime(1970, 1, 2)), 86400)
+
+    def test_datetime2timestamp_aware(self):
+        self.assertEqual(
+            datetime2timestamp(datetime.datetime(1970, 1, 2, tzinfo=tzutc())),
```
kyleknap: Let's not use ``tzutc()`` because that is default. Cannot really check that the timezone aware's timezone is respected.

kyleknap:
Looks good. Just had some small comments and questions that would be good to discuss.

botocore/signers.py
rayluo: Currently the `date_less_than` matches its counterpart `date_greater_than` in the build_policy(), and they all comply with the convention `DateLessThan` and `DateGreaterThan` which are used by the underlying policy definition. I believe the person who invented them in the first place, purposely borrows the `>` and `<` concept in maths to avoid the ambiguity of "expires at Monday? do you mean the beginning of Monday or the end of Monday?"

Besides, if we are really going to change one of them, we better come up with a pair of candidates to change them together.

tests/unit/test_signers.py
jamesls: Can you elaborate on which part of the policy is dependent on key order?  Glancing at [the docs](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html#private-content-custom-policy-statement) suggests that's not the case:

> * Use UTF-8 character encoding.
> * Include all punctuation and parameter names exactly as specified. Abbreviations for parameter  names are not accepted.
> * The order of the parameters in the Condition section doesn't matter.

botocore/signers.py
rayluo: Done. See the next line in code.

botocore/signers.py
rayluo: Sorry for the confusion. We discussed and now I added a comment at [line 278](https://github.com/boto/botocore/pull/735/files#diff-763e11dcda6df7b4753ec19afe457d5eR278).

tests/unit/test_signers.py
rayluo: To capture offline discussion: Per API spec, the order is not required by CloudFront as long as the signature matches the order we used. But as an implementation detail, we need a stable order to ensure test cases with hardcoded signature will always pass on different Python versions. The agreed solution is to still use order-aware implementation, but we don't explicitly check the order in test case.

botocore/signers.py
```diff
@@ -213,6 +216,132 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+        :type ip_address: str
+        :param ip_address: Use 'x.x.x.x' for an IP, or 'x.x.x.x/x' for a subnet
+
+        :rtype: str
+        :return: The policy in a compact string.
+        """
+        # Note:
+        # 1. Order in canned policy is significant. Special care has been taken
+        #    to ensure the output will match the order defined by document.
+        #    There is also a test case in this commit to ensure that order.
```
jamesls: Remove reference to "this commit".  That's very easy to get out of date as changes are made.

botocore/signers.py
```diff
@@ -213,6 +216,132 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+
+        :type ip_address: str
+        :param ip_address: Use 'x.x.x.x' for an IP, or 'x.x.x.x/x' for a subnet
+
+        :rtype: str
+        :return: The policy in a compact string.
+        """
+        # Note:
+        # 1. Order in canned policy is significant. Special care has been taken
+        #    to ensure the output will match the order defined by document.
```
jamesls: Clean up the wording here, maybe: "order defined by the document: <link to document>"?

jamesls:
:shipit:

mtdowling:
:ship:

One possibility for a future addition: did you consider adding a helper method to create an rsa_signer for the user? For example, maybe it could be a function that accepts that path to a private key and returns a function that could be given to `rsa_signer` (basically capturing your example in the doc comment).

kyleknap:
Thanks :ship:

botocore/signers.py
```diff
@@ -213,6 +216,132 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
         return request.url
 
 
+class CloudFrontSigner(object):
+    '''A signer to create a signed CloudFront URL.
+
+    First you create a cloudfront signer based on a normalized RSA signer::
+
+        import rsa
```
mekza: You should use [cryptography](https://cryptography.io) instead of rsa.

PR #736, opened 2015-12-08
AWS_DATA_PATH should override customer/builtin data paths

jamesls:
This was a regression introduced in
1225dc7ed2246461166b27822323d61543ae0550.  This was
originally fixed in 0.68.0 of botocore.

At this point, I think a functional test in warranted,
as this did functionality did not surivive a refactoring,
so I've added a basic functional test to ensure we don't
regress on this in the future.

There were also a few tests that need to be updated with this change.
These tests were previously wrong but were not being caught
because of this bug.

### DIFF
--- a/tests/functional/test_loaders.py
+++ b/tests/functional/test_loaders.py
@@ -1,0 +1,40 @@
+# Copyright 2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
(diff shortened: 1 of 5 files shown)


JordonPhillips:
:ship:

mtdowling:
:shipit:

rayluo:
:ship:

tests/functional/test_loaders.py
```diff
@@ -0,0 +1,40 @@
...
+# language governing permissions and limitations under the License.
+import os
+import shutil
+
+from botocore import loaders
+from tests import unittest, temporary_file
+
+
+class TestLoaderAllowsDataPathOverride(unittest.TestCase):
+    def create_file(self, f, contents, name):
```
kyleknap: It would be cool if FileCreator was in botocore.

kyleknap:
Looks good. :ship: Just had a comment about something that would be cool to have in the future if we keep following this pattern of creating a file and cleaning it up in the tests.

PR #737, opened 2015-12-08
Decode JSON pagination tokens.

JordonPhillips:
Fixes aws/aws-cli#1601

The issue was that dynamodb sends back dicts for pagination tokens rather than strings like other services do. When you hit a boundary, the `NextToken` we return would look something like this:

`{'foo': 'bar'}___1234`

When you pass that back in as the `StartingToken` for the next request we take the bits before `___`  and pass them into the request. The issue is that we take that as the string `"{'foo': 'bar'}"` instead  of the dict `{'foo': 'bar'}`. So when we pass that in, parameter validation would fail.

This PR will attempt to parse the prefix into a dict before passing it along. If it fails, it will fall back to previous behavior. 

cc @kyleknap @mtdowling @rayluo @jamesls

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -38,9 +38,10 @@ class PaginatorModel(object):
 class PageIterator(object):
     def __init__(self, method, input_token, output_token, more_results,
                  result_keys, non_aggregate_keys, limit_key, max_items,
-                 starting_token, page_size, op_kwargs):
+                 starting_token, page_size, op_kwargs, op_model):
         self._method = method
         self._op_kwargs = op_kwargs
+        self._op_model = op_model
         self._input_token = input_token
         self._output_token = output_token
         self._more_results = more_results
@@ -324,10 +325,10 @@ class PageIterator(object):
             except ValueError:
                 raise ValueError("Bad starting token: %s" %
                                  self._starting_token)
-        for part in parts:
+        for token_index, part in enumerate(parts):
             if part == 'None':
                 next_token.append(None)
-            else:
+            elif self._is_dict_token(token_index):
                 try:
                     # Some services (such as dynamodb) non-string pagination
                     # tokens, which will fail validation if they aren't
@@ -335,17 +336,24 @@ class PageIterator(object):
                     next_token.append(ast.literal_eval(part))
                 except ValueError:
                     next_token.append(part)
+            else:
+                next_token.append(part)
         return next_token, index
 
+    def _is_dict_token(self, token_index):
+        token_type = self._op_model.input_shape.members.get(
+            self._input_token[token_index]).type_name
+        return token_type == 'map' or token_type == 'structure'
 
 
 class Paginator(object):
 
     PAGE_ITERATOR_CLS = PageIterator
 
-    def __init__(self, method, pagination_config):
+    def __init__(self, method, pagination_config, operation_model):
         self._method = method
         self._pagination_cfg = pagination_config
+        self._operation_model = operation_model
         self._output_token = self._get_output_tokens(self._pagination_cfg)
         self._input_token = self._get_input_tokens(self._pagination_cfg)
         self._more_results = self._get_more_results_token(self._pagination_cfg)
@@ -412,7 +420,7 @@ class Paginator(object):
             page_params['MaxItems'],
             page_params['StartingToken'],
             page_params['PageSize'],
-            kwargs)
+            kwargs, self._operation_model)
 
     def _extract_paging_params(self, kwargs):
         pagination_config = kwargs.pop('PaginationConfig', {})
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 
 import ast
+import logging
 from itertools import tee
 
 from six import string_types
@@ -20,6 +21,9 @@ import jmespath
 from botocore.exceptions import PaginationError
 from botocore.compat import zip
 from botocore.utils import set_value_from_jmespath, merge_dicts
+
+
+logger = logging.getLogger(__name__)
 
 
 class PaginatorModel(object):
@@ -335,6 +339,8 @@ class PageIterator(object):
                     # converted to their proper type.
                     next_token.append(ast.literal_eval(part))
                 except ValueError:
+                    logger.debug(
+                        "Dict pagination token failed to parse as dict.")
                     next_token.append(part)
             else:
                 next_token.append(part)
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 
+import ast
 from itertools import tee
 
 from six import string_types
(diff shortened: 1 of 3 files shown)


mtdowling:
I haven't looked too deeply at how it would be done, but is there any way to use the model to know what the type is before attempting to parse it in some way? For example, the thing you're targeting for DynamoDB is modeled as a map: https://github.com/boto/botocore/blob/develop/botocore/data/dynamodb/2012-08-10/service-2.json#L1223

JordonPhillips:
The issue is that we don't have operation model information at that point. I'll need to find a way to inject it in.

botocore/paginate.py
```diff
@@ -323,22 +325,35 @@ def _parse_starting_token(self):
             except ValueError:
                 raise ValueError("Bad starting token: %s" %
                                  self._starting_token)
-        for part in parts:
+        for token_index, part in enumerate(parts):
```
JordonPhillips: I'm not sure if I can rely on the ordering here. Could one of you confirm/deny?

kyleknap: So I was not sure. After looking at the code, it is defined by the order in which it is defined in the JSON paginator model. So if someone was to add a new input/ouput token that broke the order in the list that will break this logic, but it will also break token parsing logic elsewhere... So it should be fine to rely on ordering I suppose.

botocore/paginate.py
```diff
@@ -323,22 +325,35 @@ def _parse_starting_token(self):
...
             if part == 'None':
                 next_token.append(None)
+            elif self._is_dict_token(token_index):
+                try:
+                    # Some services (such as dynamodb) non-string pagination
+                    # tokens, which will fail validation if they aren't
+                    # converted to their proper type.
+                    next_token.append(ast.literal_eval(part))
+                except ValueError:
+                    next_token.append(part)
```
kyleknap: We should add a logging statement for this case.

tests/unit/test_paginate.py
```diff
@@ -218,7 +224,8 @@ def test_build_full_result_with_single_key(self):
             "result_key": "Users",
             "limit_key": "MaxKeys",
         }
-        self.paginator = Paginator(self.method, self.paginate_config)
+        self.paginator = Paginator(
+            self.method, self.paginate_config, mock.Mock())
```
kyleknap: Wow these are a lot of tests to change. Could we just have a helper method for creating paginators so we do not have to update each test every time for next time?

tests/unit/test_paginate.py
```diff
@@ -334,6 +344,35 @@ def test_s3_list_multipart_uploads(self):
              ])
 
 
+class TestNonStringTokens(unittest.TestCase):
+    def setUp(self):
+        self.method = mock.Mock()
+        # This is something we'd see in s3 pagination.
```
kyleknap: s3 in comment. Looks like residual from copy paste.

tests/unit/test_paginate.py
```diff
@@ -334,6 +344,35 @@ def test_s3_list_multipart_uploads(self):
...
+        self.paginate_config = {
+            "input_token": ["ExclusiveStartKey"],
+            "output_token": ["LastEvaluatedKey"],
+            "result_key": 'Items',
+        }
+        session = botocore.session.get_session()
+        service_model = session.get_service_model('dynamodb')
```
kyleknap: How hard would it be to create your own model? It is nice to be able to avoid relying on service models so tests do not start failing when we update a model.

kyleknap:
Looks fine. Had a few comments.

botocore/paginate.py
```diff
@@ -323,22 +329,37 @@ def _parse_starting_token(self):
...
             if part == 'None':
                 next_token.append(None)
+            elif self._is_dict_token(token_index):
+                try:
+                    # Some services (such as dynamodb) non-string pagination
+                    # tokens, which will fail validation if they aren't
+                    # converted to their proper type.
+                    next_token.append(ast.literal_eval(part))
+                except ValueError:
+                    logger.debug(
```
kyleknap: We should include the ``part`` in the log as well.

kyleknap:
Looks good. Just had a small comment. :ship:

mtdowling:
:shipit:

botocore/paginate.py
```diff
@@ -323,22 +329,37 @@ def _parse_starting_token(self):
             except ValueError:
                 raise ValueError("Bad starting token: %s" %
                                  self._starting_token)
-        for part in parts:
+        for token_index, part in enumerate(parts):
             if part == 'None':
                 next_token.append(None)
+            elif self._is_dict_token(token_index):
+                try:
+                    # Some services (such as dynamodb) non-string pagination
```
rayluo: Missing a word?

    ... dynamodb) USES non-string ...

botocore/paginate.py
```diff
@@ -323,22 +329,37 @@ def _parse_starting_token(self):
...
+                    # tokens, which will fail validation if they aren't
+                    # converted to their proper type.
+                    next_token.append(ast.literal_eval(part))
+                except ValueError:
+                    logger.debug(
+                        "Dict pagination token failed to parse as dict.")
+                    next_token.append(part)
```
rayluo: Out of curiosity, in this case, will the raw `part` still be accepted by the service? If not, shall we use hard failure by `raise ValueError("Dict pagination token failed to parse as dict")`?

botocore/paginate.py
```diff
@@ -323,22 +329,37 @@ def _parse_starting_token(self):
...
+            self._input_token[token_index]).type_name
+        return token_type == 'map' or token_type == 'structure'
 
 
 class Paginator(object):
 
     PAGE_ITERATOR_CLS = PageIterator
 
-    def __init__(self, method, pagination_config):
+    def __init__(self, method, pagination_config, operation_model):
```
rayluo: This reminds me another trick we used [before](https://github.com/boto/botocore/blob/develop/botocore/docs/shape.py#L27). If we define an optional setting like this:

    def __init__(self, method, pagination_config, settings=None):
        self._settings = {} if settings is None else settings

we can then initialize this class in client.py by something like this:

    paginator = documented_paginator_cls(
        ..., ...,
        {"dict_tokens": _find_all_dict_token_names_in_advance()})

Now the `_is_dict_token()` can be implemented as:

    def _is_dict_token(self, token_index):
        return self._input_token[token_index] in self._settings.get('dict_tokens', [])

The benefit is:

* The paginator interface remains backward compatible (although it is not really that important here, because it is not an open API). As a result, we don't need to modify tons of existing test cases.
* The paginator does not couple with the model logic. Should we change model APIs in the future (very unlikely), this paginator will not need to change at all.

jamesls:
Wouldn't it be safer if we just injected a key into the returned hash rather than using the `__` separator?  The intent was to delineate strings, not arbitrary python data.  What happens if a dynamodb key/value contains the `__` string?  I would expect this to break.

JordonPhillips:
@jamesls wouldn't it break any time we ever have a value with that separator in there? I agree that it's an issue, but maybe we need to think about why it's there in the first place.

jamesls:
Why what's there, the separator in the next token? Most of the time, this is an opaque token the service generates that the user has no control over.  I'd have to check with S3 pagination, but other than S3, dynamodb paginators appear the only case where a user can directly influence the token, and therefor have a `__` char.

JordonPhillips:
After some discussion with James, the direction I'm going with this is going to change enough that I'm gonna switch to a new PR.

PR #741, opened 2015-12-15
Support stubbing out operations.

JordonPhillips:
This will allow customers to stub out operations, providing their own responses so that they don't have to hit an endpoint to test their code. The interface for this is the `Stubber` class, which will manage the response queue and and handle validating the provided response stubs.

Usage example:

```python
import botocore.session
import botocore.stub
import datetime


s3_client = botocore.session.get_session().create_client('s3')
stubber = botocore.stub.Stubber(s3_client)
list_objects_response = {
   'IsTruncated': False,
   'EncodingType': 'url',
   'Marker': '',
   'Prefix': '',
   'Name': 'test-bucket',
   'Contents': [
      {
         'Size': 123,
         'Owner': {
            'DisplayName': 'display_name',
            'ID': '123abc'
         },
         'Key': 'file.txt',
         'StorageClass': 'STANDARD',
         'ETag': '"abc123"',
         'LastModified': datetime.datetime(2015, 12, 7, 18, 5, 52)
      },
      {
         'Size': 456,
         'Owner': {
            'DisplayName': 'display_name',
            'ID': '123abc'
         },
         'Key': 'file.py',
         'StorageClass': 'STANDARD',
         'ETag': '"abc123"',
         'LastModified': datetime.datetime(2015, 12, 7, 17, 38, 49)
      }
   ],
   'MaxKeys': 1000
}
stubber.add_response('list_objects', list_objects_response)
stubber.activate()
response = s3_client.list_objects(Bucket='test-bucket')
assert(response == list_objects_response)
```

cc @rayluo @kyleknap @mtdowling @jamesls

### DIFF
--- a/botocore/stub.py
+++ b/botocore/stub.py
@@ -1,0 +1,201 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import copy
+from collections import deque
+from botocore.validate import validate_parameters
+from botocore.exceptions import ParamValidationError, StubResponseError
+from botocore.vendored.requests.models import Response
+
+
+class Stubber(object):
+    """
+    This class will allow you to stub out requests so you don't have to hit
+    an endpoint to write tests. Responses are returned first in, first out.
+    If operations are called out of order, or are called with no remaining
+    queued responses, an error will be raised.
+
+    Example::
+        import datetime
+        import botocore.session
+        from botocore.stub import Stubber
+
+
+        s3 = botocore.session.get_session().create_client('s3')
+        stubber = Stubber(s3)
+
+        response = {
+            'IsTruncated': False,
+            'Name': 'test-bucket',
+            'MaxKeys': 1000, 'Prefix': '',
+            'Contents': [{
+                'Key': 'test.txt',
+                'ETag': '"abc123"',
+                'StorageClass': 'STANDARD',
+                'LastModified': datetime.datetime(2016, 1, 20, 22, 9),
+                'Owner': {'ID': 'abc123', 'DisplayName': 'myname'},
+                'Size': 14814
+            }],
+            'EncodingType': 'url',
+            'ResponseMetadata': {
+                'RequestId': 'abc123',
+                'HTTPStatusCode': 200,
+                'HostId': 'abc123'
+            },
+            'Marker': ''
+        }
+
+        stubber.add_response('list_objects', response)
+        stubber.activate()
+
+        service_response = s3.list_objects(Bucket='test-bucket')
+        assert service_response == response
+    """
+    def __init__(self, client):
+        """
+        :param client: The client to add your stubs to.
+        """
+        self.client = client
+        self._event_id = 'boto_stubber'
+        self._queue = deque()
+
+    def activate(self):
+        """
+        Activates the stubber on the client
+        """
+        self.client.meta.events.register(
+                'before-call.*.*',
+                self._get_response_handler,
+                unique_id=self._event_id)
+
+    def deactivate(self):
+        """
+        Deactivates the stubber on the client
+        """
+        self.client.meta.events.unregister(
+                'before-call.*.*',
+                self._get_response_handler,
+                unique_id=self._event_id)
+
+    def add_response(self, method, service_response):
+        """
+        Adds a service response to the response queue. This will be validated
+        against the service model to ensure correctness. It should be noted,
+        however, that while missing attributes are often considered correct,
+        your code may not function properly if you leave them out. Therefore
+        you should always fill in every value you see in a typical response for
+        your particular request.
+
+        :param method: The name of the client method to stub.
+        :type method: str
+
+        :param service_response: A dict response stub. Provided parameters will
+            be validated against the service model.
+        :type service_response: dict
+        """
+        self._add_response(method, service_response)
+
+    def _add_response(self, method, service_response, http_response=None):
+        if not hasattr(self.client, method):
+            raise ValueError(
+                "Client %s does not have method: %s"
+                % (self.client.meta.service_model.service_name, method))
+
+        if http_response is None:
+            http_response = Response()
+            http_response.status_code = 200
+            http_response.reason = 'OK'
+
+        operation_name = self.client.meta.method_to_api_mapping.get(method)
+        self._validate_response(operation_name, service_response)
+
+        response = (operation_name, (http_response, service_response))
+        self._queue.append(response)
+
+    def add_client_error(self, method, service_error_code='',
+                         service_message='', http_status_code=400):
+        """
+        Adds a ClientError to the response queue.
+
+        :param method: The name of the service method to return the error on.
+        :type method: str
+
+        :param service_error_code: The service error code to return,
+                                   e.g. NoSuchBucket
+        :type service_error_code: str
+
+        :param service_message: The service message to return, e.g.
+                        'The specified bucket does not exist.'
+        :type service_message: str
+
+        :param http_status_code: The HTTP status code to return, e.g. 404, etc
+        :type http_status_code: int
+        """
+        http_response = Response()
+        http_response.status_code = http_status_code
+
+        # We don't look to the model to build this because the caller would need
+        # to know the details of what the HTTP body would need to look like.
+        parsed_response = {
+            'ResponseMetadata': {'HTTPStatusCode': http_status_code},
+            'Error': {
+                'Message': service_message,
+                'Code': service_error_code
+            }
+        }
+
+        operation_name = self.client.meta.method_to_api_mapping.get(method)
+        response = (operation_name, (http_response, parsed_response))
+        self._queue.append(response)
+
+    def assert_no_pending_responses(self):
+        """
+        Asserts that all expected calls were made.
+        """
+        remaining = len(self._queue)
+        if remaining != 0:
+            raise AssertionError("%d responses remaining in queue." % remaining)
+
+    def _get_response_handler(self, model, params, **kwargs):
+        if not self._queue:
+            raise StubResponseError(
+                operation_name=model.name,
+                reason='Unexpected API Call: called with parameters %s' %
+                       params)
+
+        name, response = self._queue.popleft()
+        if name != model.name:
+            raise StubResponseError(
+                operation_name=model.name,
+                reason='Operation mismatch: found response for %s.' % name)
+
+        return response
+
+    def _validate_response(self, operation_name, service_response):
+        service_model = self.client.meta.service_model
+        operation_model = service_model.operation_model(operation_name)
+        output_shape = operation_model.output_shape
+
+        # Remove ResponseMetadata so that the validator doesn't attempt to
+        # perform validation on it.
+        response = service_response
+        if 'ResponseMetadata' in response:
+            response = copy.copy(service_response)
+            del response['ResponseMetadata']
+
+        if output_shape is not None:
+            validate_parameters(response, output_shape)
+        elif response:
+            # If the output shape is None, that means the response should be
+            # empty apart from ResponseMetadata
+            raise ParamValidationError(
+                report="Service response should only contain ResponseMetadata.")
--- a/tests/functional/test_stub.py
+++ b/tests/functional/test_stub.py
@@ -1,0 +1,60 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+from tests import unittest
+
+import botocore
+import botocore.session
+from botocore.stub import Stubber
+from botocore.exceptions import StubResponseError, ClientError
+import botocore.client
+import botocore.retryhandler
+import botocore.translate
+
+
+class TestStubber(unittest.TestCase):
+    def setUp(self):
+        session = botocore.session.get_session()
+        config = botocore.client.Config(signature_version=botocore.UNSIGNED)
+        self.client = session.create_client('s3', config=config)
+
+        self.stubber = Stubber(self.client)
+
+    def test_stubber_returns_response(self):
+        service_response = {'ResponseMetadata': {'foo': 'bar'}}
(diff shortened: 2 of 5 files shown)


garnaat:
Nice!  +1

jamesls:
Awesome.  I'd like to look at this in more depth, but I'm curious if there's a reason using an existing event like `before-call` was insufficient?  Are there additional capabilities you get from hooking deeper down into endpoint vs. `before-call`?  Just want to ensure there's not a pre-existing event we can use before we go down the path of adding another event.  I'm also hesitant to expose the requests's `Response` class to users.

mtdowling:
Looks good. Some high level feedback:

1. With these types of abstractions, it's often nice to be able to get the requests that were sent through the stubber class. Do you think it would be useful to have some way to be able to ask the stubber what requests it stubbed, and then in turn some way to clear the history? In the PHP SDK, we just track the last request and last response -- https://github.com/aws/aws-sdk-php/blob/master/src/MockHandler.php
2. What do you think about adding support for stubbing errors so that when invoked, the class raises a pre-created exception?
3. Do you think there would be any value in allowing different strategies for determining which response to use when stubbing? For example, one could just pop from a queue (what you've implemented), while another strategy might be to just call a function where a user could do various things in a test each time a call is stubbed. Another use case might be a strategy that has some predetermined expectation on what requests will be sent and the response associated with them. We didn't include this type of functionality in the PHP SDK because it is really easy to create anonymous functions that act as the HTTP client.
4. Instead of my third bullet-point (and possibly in addition to...), what about adding functions to the queue as well? When a function is popped of the queue the function is invoked with the request params and the return value is the result of the function (or an exception is raised)? Using functions might be required for raising exceptions if the exception cannot be created in advance (because it needs a reference the request or something).

garnaat:
I did something similar to this is https://github.com/garnaat/placebo but I ended up having to rely on the client's private ``_endpoint`` attribute to mock out the call and get my data in there.  Maybe there is another way but I couldn't find it.  Something like this would make things much easier.

JordonPhillips:
@jamesls There are several events in `Endpoint` that `before-call` would trample if that were used. I didn't want to interfere with anything else that we (or a customer) might be doing. Further, intercepting at that point allows for the possibility of testing out retry logic. As far as `Response` is concerned, how about letting them pass in the status code and reason of the response, and we create it for them? Or maybe even a `ClientError`?

@mtdowling 

1. I would be +1 on something like this. We could potentially have a map of requests to responses and the user would pass in the history size they want on creation of the `Stubber` class (probably defaulting to 1).
2. Right now you can stub errors by passing in an error response. I think letting them pass a status code / reason would probably suffice though.
3. I think these would be cases for subclasses. We could encourage this by making `Stubber` an abstract class and providing some number of implementations.
4. I think this could potentially be useful, but it skirts around validation. Additionally, they would need to provide this for every single call where they want their function to run, and I'm not sure that's a wonderful interface.

@garnaat in the current state, there's not really a better way to go about it since you can't short-circuit `Endpoint`. With this PR you would be able to short-circuit with an event handler.

JordonPhillips:
After much discussion, this is going to be changed and broken up a bit. The first part will be addressed in #743. We will then build on top of that.

JordonPhillips:
The tests on this PR will not pass until #743 is merged and I can safely rebase.

JordonPhillips:
It is now safe

botocore/stub.py
```diff
@@ -0,0 +1,164 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import copy
+from botocore.validate import validate_parameters
+from botocore.exceptions import ParamValidationError, StubResponseError
+from botocore.vendored.requests.models import Response
+
+
+class QueueStubber(object):
```
jamesls: Is the `Queue` part here necessary?

I don't think we'd ever create a `StackStubber`, and I don't think I've seen any kind of mock library where responses weren't returned in the order you added them.

botocore/stub.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    This class will allow you to stub out requests so you don't have to hit
+    an endpoint to write tests. Responses are returned first in, first out.
+    If operations are called out of order, or are called with no remaining
+    queued responses, an error will be raised.
```
jamesls: It would be nice either here or as a module docstring to show a basic example.

botocore/stub.py
```diff
@@ -0,0 +1,164 @@
...
+        :param client: The client to add your stubs to.
+        """
+        self.client = client
+        self._queue = []
+
+    def activate(self):
+        """
+        Activates the stubber on the client
+        """
+        self.client.meta.events.register(
```
jamesls: I would suggest using a `unique_id` here to ensure we don't register this same handler twice.  Otherwise if you call `activate()` twice you'll have identical handlers here.

This *shouldn't* matter because of the short circuit behavior of before-call but it's still something that could cause issues in the future.

botocore/stub.py
```diff
@@ -0,0 +1,164 @@
...
+    def assert_no_pending_responses(self):
+        """
+        Asserts that all expected calls were made.
+        """
+        remaining = len(self._queue)
+        if remaining != 0:
+            raise AssertionError("%d responses remaining in queue." % remaining)
+
+    def _get_response_handler(self, model, **kwargs):
+        if len(self._queue) < 1:
```
jamesls: More idiomatically: `if not self._queue`

botocore/stub.py
```diff
@@ -0,0 +1,164 @@
...
+            raise StubResponseError(
+                operation_name=model.name,
+                reason='No responses remaining.')
```
jamesls: What about updating this error message to also include something like `Unexpected API call: <the API call + args>` to give more context to the user about what went wrong.  A  misconfigured mock can be tedious to track down if the error messages don't give a lot of detail.

botocore/stub.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # Pops from index 0 so it's FIFO
+        name, response = self._queue.pop(0)
```
jamesls: If you're going to be popping from the front of the list, you should generally prefer `collections.deque` over a list.  `collections.deque` is a doubly linked list so you'll avoid the O(n) pops from the front of the list.  Note a huge deal until the number of queued responses are large, but definitely a best practice.

botocore/stub.py
```diff
@@ -0,0 +1,164 @@
...
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import copy
+from botocore.validate import validate_parameters
+from botocore.exceptions import ParamValidationError, StubResponseError
+from botocore.vendored.requests.models import Response
```
jamesls: I'd still like to minimize the use of requests directly within botocore.  Is it feasible to use `awsrequest.AWSRequest`?

botocore/stub.py
```diff
@@ -0,0 +1,164 @@
...
+
+        operation_name = self.client.meta.method_to_api_mapping.get(method)
+        self._validate_response(operation_name, service_response)
+
+        response = (operation_name, (http_response, service_response))
+        self._queue.append(response)
+
+    def add_client_error(self, method, service_error_code='',
+                         service_message='', http_status_code=400,
+                         http_reason=''):
```
jamesls: What's the reason for including the `http_reason`?  This is tied to the http status code and isn't something the user can generally override is it?

If we find this useful to have we can automatically lookup the reason for a given status code (I believe it's somewhere either in requests/urllib3/httplib), but it doesn't seem like the user should be able to provide this.

botocore/stub.py
```diff
@@ -0,0 +1,164 @@
+# Copyright 2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jamesls: 2016

tests/functional/test_stub.py
```diff
@@ -0,0 +1,162 @@
+# Copyright 2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jamesls: 2016

tests/unit/test_stub.py
```diff
@@ -0,0 +1,115 @@
...
+        with self.assertRaises(ValueError):
+            self.stubber.add_response('foo', {})
+
+    def test_validates_service_response(self):
+        self.stubber.add_response('foo', {})
+        self.assertTrue(self.validate_parameters_mock.called)
+
+    def test_validate_ignores_response_metadata(self):
+        service_response = {'ResponseMetadata': {'foo': 'bar'}}
+        service_model = mock.Mock()
```
jamesls: I'd prefer not to mock service model objects.  Given this is an object we own (and is primarily a data object) I'd prefer to just create the models directly.

We've seen these mocks quickly get out of date as the corresponding APIs they're mocking change.

botocore/stub.py
```diff
@@ -0,0 +1,164 @@
...
+            http_response.status_code = 200
+            http_response.reason = 'OK'
+
+        operation_name = self.client.meta.method_to_api_mapping.get(method)
+        self._validate_response(operation_name, service_response)
+
+        response = (operation_name, (http_response, service_response))
+        self._queue.append(response)
+
+    def add_client_error(self, method, service_error_code='',
```
jamesls: It doesn't look like this is tested.

tests/functional/test_stub.py
```diff
@@ -0,0 +1,162 @@
...
+from botocore.stub import QueueStubber
+from botocore.credentials import Credentials
+from botocore.exceptions import StubResponseError
+from botocore import hooks
+import botocore.client
+import botocore.retryhandler
+import botocore.translate
+
+
+class TestStubber(unittest.TestCase):
```
jamesls: One of the things I was hoping to see in the functional tests was trying to stub out an existing real AWS service to ensure that everything works with a real service model.

This is a final test to ensure that all the pieces fit together and won't be subject to false test passes if the underlying APIs being mocked are changed.

jamesls:
Overall I think the API's nice and easy to use.  I like the validation of responses.  I think customers are going to find this really useful.  I have some small comments/question above but overall looks nice.

botocore/stub.py
JordonPhillips: No. `AWSRequest` does not conform to `Response`

botocore/stub.py
kyleknap: I agree on this point.

botocore/stub.py
```diff
@@ -0,0 +1,201 @@
...
+    def add_response(self, method, service_response):
+        """
+        Adds a service response to the response queue. This will be validated
+        against the service model to ensure correctness. It should be noted,
+        however, that while missing attributes are often considered correct,
+        your code may not function properly if you leave them out. Therefore
+        you should always fill in every value you see in a typical response for
+        your particular request.
+
+        :param method: The name of the service method to stub.
```
kyleknap: service --> client. I see the service method as the version that is all camel case

botocore/stub.py
```diff
@@ -0,0 +1,201 @@
...
+            response = copy.copy(service_response)
+            del response['ResponseMetadata']
+
+        if output_shape is not None:
+            validate_parameters(response, output_shape)
+        elif response:
+            # If the output shape is None, that means the response should be
+            # empty apart from ResponseMetadata
+            raise ParamValidationError(
+                report="Service response contains unknown members.")
```
kyleknap: This message is a bit ambiguous. Maybe say something that the response should have at most the ResponseMetadata member.

botocore/stub.py
```diff
@@ -0,0 +1,201 @@
...
+    def deactivate(self):
+        """
+        Deactivates the stubber on the client
+        """
+        self.client.meta.events.unregister(
+                'before-call.*.*',
+                self._get_response_handler,
+                unique_id=self._event_id)
+
+    def add_response(self, method, service_response):
```
kyleknap: This does need to be addressed in this PR, but something that would be cool is that for each add_response, you can add the option to assert the parameters sent for that response. I think that is one thing missing to get close to a complete replacement to mock.

tests/functional/test_stub.py
```diff
@@ -0,0 +1,122 @@
...
+    def setUp(self):
+        credentials = Credentials('access-key', 'secret-key')
+
+        with mock.patch('botocore.client.EndpointCreator') \
+                as endpoint_creator_patch:
+            endpoint_creator = endpoint_creator_patch.return_value
+            self.endpoint = mock.Mock()
```
kyleknap: So for this endpoint mocking, let's just move it to the unit test. These should not be in the functional tests. What I would do for these functional test is remove all instances of asserting the endpoint was hit and compact them into a few unit tests that do all of patching and then ensure that the endpoint is not hit. In the end, I would prefer no patches or mocks being used.

Everything else in the functional test look fine though.

tests/functional/test_stub.py
```diff
@@ -0,0 +1,122 @@
...
+        self.assertTrue(self.endpoint.make_request.called)
+        self.endpoint.make_request.reset_mock()
+
+        self.stubber.add_response('list_objects', {})
+        self.client.list_objects(Bucket='foo')
+        self.assertTrue(self.endpoint.make_request.called)
+
+    def test_deactivated_stubber_does_nothing(self):
+        self.stubber.add_response('list_objects', {})
+        self.stubber.activate()
```
kyleknap: Now that I think about it. I would even suggest moving a lot of these tests to the unit directory if the purpose of the test is to see if the endpoint was hit. Then, in the functional test, it could be a some smoke cases that ensure the proper responses and errors are returned and the external interface is as you would expect.

botocore/stub.py
JordonPhillips: The catch there is accounting for the substantial number of parameters we inject.

tests/functional/test_stub.py
JordonPhillips: In that case, these should just go away entirely. There's already a unit test in `test_client` that verifies the endpoint isn't hit.

botocore/stub.py
kyleknap: Yeah it may be tricky but it would be very useful because we currently do not have a way to do mock's ``assert_called_with()`` method.

tests/functional/test_stub.py
```diff
@@ -0,0 +1,68 @@
...
+
+class TestStubber(unittest.TestCase):
+    def setUp(self):
+        session = botocore.session.get_session()
+        config = botocore.client.Config(signature_version=botocore.UNSIGNED)
+        self.client = session.create_client('s3', config=config)
+
+        self.stubber = Stubber(self.client)
+
+        self.parser_patch = mock.patch(
```
kyleknap: This patch is no longer used.

JordonPhillips:
:+1:

kyleknap:
Awesome. Had that one suggestion that we could add later, but otherwise looks good. :ship:

botocore/stub.py
JordonPhillips: done

mtdowling:
Looks great. :shipit: 

One thing that might be useful to add at a later point is updating the Stubber to track the params and operations that were intercepted so that the user of the stubber could perform assertions on the requests that were sent (e.g., ensure that a handler added a custom parameter).

kyleknap:
@mtdowling I actually have an implementation for that in some of the s3 work that I have been doing. Will probably put together and send a PR once this gets merged.

mtdowling:
@kyleknap :+1:

kapilt:
really looking forward to this, one comment on api was i often just pass a session due to multiple service/client interactions, it looks like that would still need a mock on the session with this api for each client interaction, vs the session patching that placebo does.

jamesls:
:shipit:  Thanks for incorporating the feedback.

PR #742, opened 2015-12-17
Fix doc generation on python35

jamesls:
There's a regression on python35 where the final parts of the
document aren't parsed unless you call close.

Looks like there's a fix for this here:
https://bugs.python.org/issue23144 but this won't be available
until at least 3.5.1.

Either way, calling close() should work going forward even once
the original behavior is fixed.

I did verify that it's ok to re-use the parser after you call
close().

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
   - "2.7"
   - "3.3"
   - "3.4"
+  - "3.5"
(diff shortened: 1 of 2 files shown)


kyleknap:
:ship:

JordonPhillips:
:boat:

PR #743, opened 2015-12-17
Use before-call event to allow short-circuiting the endpoint call.

JordonPhillips:


kyleknap:
The code change looks fine. The only issue that I can forsee is that if we ever realize that we want to exercise more of stack (the endpoint/retry logic) before we short-circuit the sending of the request to the http session we may muddy the code more with event-based if statement trees. So @jamesls @mtdowling @rayluo are we sure using the ``before-call`` event is the direction we want to go to short-circuit the sending of the request over the wire?

jamesls:
I don't think it's necessarily bad if in the future we end up wanting to allow customers to drop down to the http layer and expose some event in endpoints.  Right now, using this event is the lowest level we have before dropping down to endpoints/http and it doesn't require adding any new events.  It also allows libraries such as @garnaat's placebo to use entirely public APIs to achieve their desired functionality.

The only thing this doesn't allow customers to do is to specify HTTP responses.  I don't really see that being all that common of a use case for customers (it's arguable more interesting for us botocore devs), I'd prefer to be very cautious about not exposing internals of either the Endpoint class or the Request/Response classes in emitted events in such a way that we can't refactor the internals in the future because of backwards compatibility concerns.  This is why I'd suggest holding off on adding events to endpoints.py for now until we feel more confident about the endpoint and http interfaces.

So for me, I'm inclined to :shipit: as is.

garnaat:
FWIW, this looks pretty much ideal for what I was looking for.  As long as I can inject the response data and an HTTP response code (which I think I can do with this) that would solve all of my problems.

kyleknap:
I agree that we should not expose any of the internals of the endpoint. So I guess the way I was seeing it was that the arguments emitted in this event would not expose any of the internals of the endpoint, but I have not really thought too much about how that would look.

I do feel that we will need to support the endpoint event at some point for whatever reason, but I guess for now I am fine with making the ``before-call`` event a switch as I can see it being it intuitive that if you return something from the ``before-call`` event, botocore would use that instead of making the request.

kyleknap:
:ship:

PR #750, opened 2016-01-07
Add 429 as a status code to retry

jamesls:
Encountered this with API gateway:
http://docs.aws.amazon.com/apigateway/api-reference/handling-errors/

Given 429 Too Many Requests is a generic HTTP status code, I added
to the default set of status codes we retry instead of just for API
gateway.

cc @mtdowling @kyleknap @rayluo @JordonPhillips

### DIFF
--- a/botocore/data/_retry.json
+++ b/botocore/data/_retry.json
@@ -16,6 +16,13 @@
         }
       }
     },
+    "too_many_requests": {
+      "applies_when": {
+        "response": {
+          "http_status_code": 429
+        }
+      }
+    },
     "general_socket_errors": {
       "applies_when": {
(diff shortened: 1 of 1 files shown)


mtdowling:
:ship:

Do you know if they include a [Retry-After](https://tools.ietf.org/html/rfc6585#section-4) header? If they do, that might be an interesting thing to look into utilizing in the future.

rayluo:
:ship: Agree with @mtdowling

jamesls:
Just checked.  They do not send a retry-after header.  Might be worth sending them a feature request.

Going to go ahead and merge now.

mtdowling:
:cake:

PR #752, opened 2016-01-09
Fix decode_list_object handler to urldecode all values that can contain S3 key strings

fsangiovanni:
Hi,

I've noticed that botocore.handlers.decode_list_object() only addresses url-decoding of the 'Contents' key of the response dict.

This leaves undecoded values of response keys that can contain S3 object names:
CommonPrefixes
Marker
NextMarker
Prefix
Delimiter

I've been able to track back the problem until version 1.3.9 (last working version).
This PR should address the issue.
Please let me know if it's ok.

Thanks!

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -531,11 +531,22 @@ def decode_list_object(parsed, context, **kwargs):
     # This is needed because we are passing url as the encoding type. Since the
     # paginator is based on the key, we need to handle it before it can be
     # round tripped.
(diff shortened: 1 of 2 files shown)


fsangiovanni:
Hi,
is there something wrong with this PR?
Anything I can do to help?

mtdowling:
Thanks! This has been merged in https://github.com/boto/botocore/issues/765

PR #757, opened 2016-01-13
Special case versionId query param in urlencoding

jamesls:
A while back there was a bug fix that did not url encode
the query string.  This was so customers could provide
a ?versionId to indicate they wanted a specific version
copied (https://github.com/boto/botocore/issues/509).

However, it's possible that a customer's key name ends
with something that looks like a query string, e.g.
"foo?a=b&c=d".

This modifies this code to only skip the uri encoding
if there's a query string with versionId.

There's still one edge case.  It's ambiguous for a key
such as 'foo?versionId=bar' whether you meant the entire
string to be the key name or whether you meant
key=foo, versionId=bar.  In this case, we assume you meant
the latter, so it's not possible to copy a key with
'?versionId=...'.

In order to fix this edge case we'd need to support some
explicit way of the customer specifying the bucket, key,
and versionId.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -33,6 +33,7 @@ from botocore.signers import add_generate_presigned_url
 from botocore.signers import add_generate_presigned_post
 from botocore.exceptions import ParamValidationError
 from botocore.exceptions import UnsupportedTLSVersionWarning
+from botocore.utils import percent_encode, SAFE_CHARS
 
 from botocore import retryhandler
 from botocore import utils
@@ -51,6 +52,7 @@ REGISTER_LAST = object()
 # combination of uppercase letters, lowercase letters, numbers, periods
 # (.), hyphens (-), and underscores (_).
 VALID_BUCKET = re.compile('^[a-zA-Z0-9.\-_]{1,255}$')
+VERSION_ID_SUFFIX = re.compile(r'\?versionId=[^\s]+$')
 
 
 def check_for_200_error(response, **kwargs):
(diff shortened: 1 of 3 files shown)


botocore/handlers.py
```diff
@@ -51,6 +52,7 @@
 # combination of uppercase letters, lowercase letters, numbers, periods
 # (.), hyphens (-), and underscores (_).
 VALID_BUCKET = re.compile('^[a-zA-Z0-9.\-_]{1,255}$')
+VERSION_ID_SUFFIX = re.compile(r'\?versionId=[^\s]+$')
```
mtdowling: Should this allow for `?versionId=...` and `&versionId=...` (e.g., the query string contains values before `versionId`)?

mtdowling:
> In order to fix this edge case we'd need to support some explicit way of the customer specifying the bucket, key, and versionId.

Given the edge cases that are coming up, something we might want to consider adding is a custom parameter for this operation like `RawCopySource`. With this parameter, we would either do no encoding at all, or possibly we just encode characters in the URI fragment that are specifically not allowed while ensuring that we do not double encode values (something akin to [this URI code](https://github.com/guzzle/psr7/blob/master/src/Uri.php#L577-L600)).

jamesls:
I chatted with a few people here and I like the idea of doing what ruby does where you can either provide a string or a dict of bucket, key, version_id.  The preferred interface would then be the dict so the user doesn't have to worry about these edge cases.  They also wouldn't need to know the `{bucket}/{key}?versionId={version_id}` serialization format either.

botocore/handlers.py
jamesls: Not opposed.  The main reason I didn't do this is because S3 will return validation errors if you have anything other than versionId in your query string.  It's possible this can change in the future on S3's side, but with the proposed change of supporting both string and dict, hopefully this won't matter as we can steer people towards using the dict format.

kyleknap:
Yeah I am in favor of the ruby implementation as well. It also ensures that we do not have to add an extra parameter to the signature because CopySource will now except a string or dictionary with the components.

JordonPhillips:
+1 to ruby style

mtdowling:
The ruby solution sounds good too. One small suggestion is to use `versionId` in the dict to be consistent with the rest of the camelCased args.

jamesls:
Alternate form for CopySource added in #758.

botocore/handlers.py
```diff
@@ -269,13 +271,16 @@ def add_expect_header(model, params, **kwargs):
...
-        # We only want to quote the path.  If the user specified
-        # extra parts, say '?versionId=myversionid' then that part
-        # should not be quoted.
-        quoted = quote(p[2].encode('utf-8'), '/~')
-        final_source = urlunsplit((p[0], p[1], quoted, p[3], p[4]))
-        params['headers']['x-amz-copy-source'] = final_source
+        result = VERSION_ID_SUFFIX.search(value)
+        if result is None:
+            params['headers']['x-amz-copy-source'] = percent_encode(
+                value, safe=SAFE_CHARS + '/')
```
kyleknap: It is looking like ``SAFE_CHAR`` is picking up some extra characters other than ``~``. Is this intentional?

kyleknap:
Had one comment that would be good to be confirmed. Otherwise looks fine. :ship:

jamesls:
Closing, superseded by https://github.com/boto/botocore/pull/758

PR #758, opened 2016-01-14
Add support for alternate form of CopySource

jamesls:
Follow up from https://github.com/boto/botocore/pull/757, this adds support for a dict value for `CopySource`.  This ensures we automatically handle all cases correctly for the customer because they're explicitly telling us the bucket, key, and optional version id.

```
s3.copy_object(CopySource={'Bucket': 'sourcebucket', 'Key': 'sourcekey'})
```

I've also handed a handler to update the documentation for this new format.

I'll add a handwritten example to boto3 shortly as well.

NOTE: I've branched this off of #757 because I had to modify the existing handler for the string value of `CopySource` so you'll see the diff of both in this PR.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -33,6 +33,7 @@ from botocore.signers import add_generate_presigned_url
 from botocore.signers import add_generate_presigned_post
 from botocore.exceptions import ParamValidationError
 from botocore.exceptions import UnsupportedTLSVersionWarning
+from botocore.utils import percent_encode, SAFE_CHARS
 
 from botocore import retryhandler
 from botocore import utils
@@ -51,6 +52,7 @@ REGISTER_LAST = object()
 # combination of uppercase letters, lowercase letters, numbers, periods
 # (.), hyphens (-), and underscores (_).
 VALID_BUCKET = re.compile('^[a-zA-Z0-9.\-_]{1,255}$')
+VERSION_ID_SUFFIX = re.compile(r'\?versionId=[^\s]+$')
 
 
 def check_for_200_error(response, **kwargs):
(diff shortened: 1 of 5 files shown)


botocore/handlers.py
```diff
@@ -266,16 +268,67 @@ def add_expect_header(model, params, **kwargs):
...
+        )
+
+
+def support_dict_copy_source(params, context, **kwargs):
+    source = params.get('CopySource')
+    if source is None or not isinstance(source, dict):
+        return
+    source_dict = params['CopySource']
+    try:
+        bucket = source_dict['Bucket']
```
kyleknap: Does the bucket need to be url encoded as well? It looks like it gets urlencoded for the string branch of this logic.

botocore/handlers.py
```diff
@@ -266,16 +268,67 @@ def add_expect_header(model, params, **kwargs):
...
+    except KeyError as e:
+        raise ParamValidationError(
+            report='Missing required parameter: %s' % str(e))
+    params['CopySource'] = '%s/%s' % (bucket, key)
+    if version_id is not None:
+        params['CopySource'] += '?versionId=%s' % version_id
+    context['CopySourceEncoded'] = True
+
+
+def quote_source_header(params, context, **kwargs):
```
kyleknap: I wonder if this would be better to combine these two handlers into one so that we can avoid this whole context logic? If your reason for keeping quote_source_header is for backwards compatibility that is already broken because context is now required in the function signature. I also not sure if there is a difference in doing the encoding at before-parameter-build instead of before-call event?

kyleknap:
Looks fine. I had a couple questions on the logic that I would like answered first before we merge it.

botocore/handlers.py
jamesls: It doesn't need to be encoded.  I think it was done that way initially because it doesn't matter.  The characters allowed for a bucket name don't change when urlencoded.

botocore/handlers.py
jamesls: Good point.  I tracked through the history and doesn't seem like it needs to be in before-call.

The only backwards compatibility issue isn't really with calling it directly (no one should be doing that), it's if a customer was unregistering the handler directly.  For example, if they wanted to work around this bug they could url encode the value themselves and unregister this handler.  Do you think that's worth preserving?

kyleknap: I do not think it is worth preserving. Unregistering default handlers is not something we suggest to do to users and I do not think they should as it may have unintended consequences unless we say it is fine to unregister. Also in most issues involving CopySource that it has came up, that solution has never been suggested: https://github.com/boto/botocore/issues/509.

jamesls: Ok I'l update.  I'll close out #757 in favor of this because the changes are sufficiently coupled to each other that tracking them through 2 separate PRs doesn't make sense.

jamesls:
@kyleknap Feedback incorporated.  Ready for another look.

kyleknap:
Thanks. Looks much cleaner. :ship:

PR #763, opened 2016-01-20
Refactored paginators to use a dict rather than a custom seperator.

JordonPhillips:
Previously, our custom pagination token came in the form: `serviceToken___5000` where serviceToken represented the service's pagination token(s) and the number at the end represented where in the array we should slice the response. The problem is that a token for dynamodb or s3 can include a triple underscore. So instead of seeing one token and an index, we see two tokens and an index. This causes the paginator to fall on its face.

The solution is to use a list as a token instead. So our token will now look like `"['serviceToken', '5000']"`. We then use ast to evaluate that back into a list that we can easily use. The service token won't cause any issues because python will see that it's a string and leave it at that. Additionally, if the service token is not a string (such as in the case of dynamodb), we will get the original object back. This fixes issues we had previously where we were trying to pass the dynamodb `ExclusiveStartKey` back as a string when it's actually a dict.

I would also like to discuss another addition: base64 encoding this new token. This would obfuscate it much like the service tokens. Since this isn't something the customer is supposed to be generating on their own, encoding it would discourage such behavior.

Fixes aws/aws-cli#1601

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -16,6 +16,8 @@ from itertools import tee
 from six import string_types
 
 import jmespath
+import json
+import base64
 from botocore.exceptions import PaginationError
 from botocore.compat import zip
 from botocore.utils import set_value_from_jmespath, merge_dicts
@@ -64,8 +66,20 @@ class PageIterator(object):
 
     @resume_token.setter
     def resume_token(self, value):
-        if isinstance(value, list):
-            self._resume_token = '___'.join([str(v) for v in value])
+        if not isinstance(value, dict):
+            raise ValueError("Bad starting token: %s" % value)
+
+        if 'truncate_amount' in value:
+            token_keys = sorted(self._input_token + ['truncate_amount'])
+        else:
+            token_keys = sorted(self._input_token)
+        dict_keys = sorted(value.keys())
+
+        if token_keys == dict_keys:
+            self._resume_token = base64.b64encode(
+                    json.dumps(value).encode('utf-8'))
+        else:
+            raise ValueError("Bad starting token: %s" % value)
 
     @property
     def non_aggregate_part(self):
@@ -74,7 +88,7 @@ class PageIterator(object):
     def __iter__(self):
         current_kwargs = self._op_kwargs
         previous_next_token = None
-        next_token = [None for _ in range(len(self._input_token))]
+        next_token = dict((key, None) for key in self._input_token)
         # The number of items from result_key we've seen so far.
         total_items = 0
         first_request = True
@@ -111,7 +125,7 @@ class PageIterator(object):
                 yield response
                 total_items += num_current_response
                 next_token = self._get_next_token(parsed)
-                if all(t is None for t in next_token):
+                if all(t is None for t in next_token.values()):
                     break
                 if self._max_items is not None and \
                         total_items == self._max_items:
@@ -182,7 +196,7 @@ class PageIterator(object):
             op_kwargs[self._limit_key] = self._page_size
 
     def _inject_token_into_kwargs(self, op_kwargs, next_token):
-        for name, token in zip(self._input_token, next_token):
+        for name, token in next_token.items():
             if token is None or token == 'None':
                 continue
             op_kwargs[name] = token
@@ -243,22 +257,22 @@ class PageIterator(object):
         # However, even though we only kept 1, this is post
         # left truncation so the next starting index should be 2, not 1
         # (left_truncation + amount_to_keep).
-        next_token.append(str(amount_to_keep + starting_truncation))
+        next_token['truncate_amount'] = amount_to_keep + starting_truncation
         self.resume_token = next_token
 
     def _get_next_token(self, parsed):
         if self._more_results is not None:
             if not self._more_results.search(parsed):
-                return [None]
-        next_tokens = []
-        for token in self._output_token:
-            next_token = token.search(parsed)
+                return {}
+        next_tokens = {}
+        for output_token, input_key in zip(self._output_token, self._input_token):
+            next_token = output_token.search(parsed)
             # We do not want to include any empty strings as actual tokens.
             # Treat them as None.
             if next_token:
-                next_tokens.append(next_token)
+                next_tokens[input_key] = next_token
             else:
-                next_tokens.append(None)
+                next_tokens[input_key] = None
         return next_tokens
 
     def result_key_iters(self):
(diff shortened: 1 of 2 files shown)


JordonPhillips:
cc @kyleknap @mtdowling @jamesls @rayluo

kyleknap:
Have not looked into implementation quite yet, but I would be +1 for obsfucating the token. I am good with base64 encoding it especially this would not be hard to reconstruct. I feel like the token before with the triple underscores did a decent job, but turning that into a list makes it easier to construct, which is something that users should not be doing because it is easy to mess up. How would something like this sound: ``token = 'token-' + base64.b64encode(orig_val)``  to get a sample token like this: ``'token-YQ=='``. The reason why I am suggesting a prefix is because I have seen base64 encoded tokens come back from services and I do not want them using the service token over our generated token. What are everyone's thoughts?

mtdowling:
Let's keep obfuscating tokens separate from this PR. I think this PR could be merged and obfuscation could be added in a subsequent PR.

kyleknap:
That's fine with me. Implementation-wise it looks fine. :ship:

jamesls:
I think this is a pretty clever solution.  Regarding base64 encoding, the only downside to not doing that now is that this will almost always require the user to properly quote the value:

```
$ aws s3api list-objects --bucket bucket --max-items 2 --region us-west-2 --output text
...
NEXTTOKEN	[None, '2']

$ aws s3api list-objects --bucket bucket --max-items 2 --region us-west-2 --output text \
        --starting-token "[None, '2']"
                         ^^^^^^^^^^^^^
```

Users should be comfortable with quoting things, but the base64 encoding means there will be one less way to enter an incorrect command.

If we don't address it in this PR, are you guys fine with adding this immediately afterwards in a follow up PR?

jamesls:
Also should mention, looks good to me as is, but I would like to see the base64 added relatively soon.

:shipit:

kyleknap:
@jamesls I am +1 on that idea.

mtdowling:
I think I might be missing something on this PR. I'm trying to understand why this needs to be represented as a string that we parse into an AST, and why we don't just use the actual value type that will be passed into the operation. Is it because we need to be able to represent it as a string when parsing CLI arguments? If it's just to be able to serialize/deserialize, then couldn't we just make the AST parsing code part of the CLI?

JordonPhillips:
@mtdowling The ast parsing allows us to handle services whose pagination tokens aren't strings. Currently we are broken on dynamodb pagination because `LastEvaluatedKey` is a structure. Doing it in this way will also allow for the pagination token to be of any basic python type.

JordonPhillips:
We would need to do that anyway to interpret the list, or otherwise use some ugly regex

mtdowling:
The thing I'm wondering about is why we don't just accept the pagination token as any type and perform no parsing at the Botocore level. The pagination model define the input_token and output_token. For example, from DynamoDB: https://github.com/boto/botocore/blob/develop/botocore/data/dynamodb/2012-08-10/paginators-1.json#L10-L11. When a pagination token comes back from a service, our response parsers parse it into the appropriate Python data types because we JSON decode the response and walk it with the model. At that point, we have a valid value to pass directly into the input_token of a paginator without needing any form of AST or regex parsing because it's already been parsed out of the JSON response.

It seems like being able to accept the pagination token as a string value that is parsed into a Python type is a requirement of the CLI because it's passed in as a string argument. I'm wondering why we need to have any special handling of pagination tokens at the Botocore level.

JordonPhillips:
@mtdowling Ah, I see. Yeah, the string requirement is only a CLI issue since the token should be opaque in botocore. It's more complicated to add in the cli, but that shouldn't be a blocker. @jamesls, @kyleknap what do you think?

jamesls:
Conceptually I agree it doesn't need to be a string in botocore.  The only concern I have is backwards compatibility in changing the return type.  For example, if a user writes the token to disk and later reads it back, they'll be broken by this change (or stores it in a sqlite database, dbm, or any other kind of persistence).  My guess is it's unlikely but I don't know for sure.  For example:

```
if should_pause_s3_transfer():
  open('state-file', 'w').write(paginator.resume_token)  # Now broken, can't write a list() type
```

mtdowling:
We chatted about this in the office the other day, we all agreed that it's fine to stay a string, and obfuscating the string would be fine as well. The most apparent use case of providing an input token to a paginator is to resume pagination from a previously created paginator. The token is, and should be, opaque; however, customers could still provide the other API parameters that could influence pagination.

Some of the things we talked about that might be a good idea are to use JSON instead of relying on Python's AST parser, and update the string to always be a hash of values instead of a list (this would match how Ruby and PHP are implemented).

The backwards compatibility concern is a good point. I think though that even if we continue to use a string, the level of change of the format in this PR would still break users who are using a token that was previously persisted. Like you said though, I highly doubt anyone has done this.

JordonPhillips:
Exactly. Right now I'm working on changing from a list to a dict, which is a much more involved change.

JordonPhillips:
@mtdowling As discussed, the internal representation of the token is now a dictionary where the key is the input argument. I'm assigning the keys in the same way that we used to be passing the arguments in to the next call. For example, a dynamo key (unencoded) would look like this:

```
{
    "ExclusiveStartKey": {"hello": "world"},
    "truncate_amount": 30000
}
```

It is then json dumped and base64 encoded. Since usage of this token is not the common case in botocore/boto3, this encoding will have little to no performance impact there.

botocore/paginate.py
```diff
@@ -64,8 +66,20 @@ def resume_token(self):
...
+        if not isinstance(value, dict):
+            raise ValueError("Bad starting token: %s" % value)
+
+        if 'truncate_amount' in value:
+            token_keys = sorted(self._input_token + ['truncate_amount'])
+        else:
+            token_keys = sorted(self._input_token)
+        dict_keys = sorted(value.keys())
+
+        if len(dict_keys) == len(token_keys) and token_keys == dict_keys:
```
kyleknap: Is the length check needed? The equality check should be able to handle this already right?

kyleknap:
Still looks fine. :ship: had just one comment. I think using ``truncate_amount`` should be fine in the dictionary as well if we are concerned about naming collisions because there is no AWS services today that uses snake cases as their parameter names.

botocore/paginate.py
JordonPhillips: Yep, removed

mtdowling:
Looks good! :shipit:, but I want to mention that I'm a bit worried about `truncate_amount`. While we don't have services right now that use snake_case, I think it's possible in the mid-far future as we adopt a broader set of APIs. I know it's only a tiny bit better and kind of superficial, but perhaps `__truncate_amount` would help to mitigate a collision a little bit more.

JordonPhillips:
@mtdowling I don't know that we need to worry that much. Adding an underscore to the front is pretty ugly, and I'd rather not do it unless we actually encounter a collision.

jamesls:
Can we have something fail is this happens then?

I don't mind making the change now, but if we're not, we should have something complain very loudly if we encounter a name collision.

Or we could just rename the variable ;)

JordonPhillips:
@jamesls We could also call it `boto_trancate_amount`. I cannot imagine a scenario where a service would use that.

JordonPhillips:
Renamed

jamesls:
:shipit:

PR #764, opened 2016-01-20
Retrieve refreshable credentials as a set to avoid auth failures when refresh happens in the middle of auth signing process.

DenverJ:
There is currently a problem that can be triggered during refresh time using refreshable credentials. 

Because multiple properties (access key/secret key/token) of the credentials are accessed independently for the signing process and each access can trigger a refresh for that property we can end up signing our request with inconsistent values. IE an access key from the previous set and a secret and/or token from the refreshed set. 

This addresses that issue by retrieving the credentials as a set (and triggering a refresh if necessary) and using that consistent set during the signing process.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -289,6 +289,10 @@ class RefreshableCredentials(Credentials):
         logger.debug("Retrieved credentials will expire at: %s", self._expiry_time)
         self._normalize()
 
+    def get_credential_set(self):
+        self._refresh()
+        return Credentials(self._access_key, self._secret_key, token=self._token)
+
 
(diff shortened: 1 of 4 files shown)


jamesls:
Taking a look.  I think the approach looks good.  The one change I'd like to make before merging is to come up with a consistent approach for both temporary vs. permanent credentials so we can always use the same code path.

jamesls:
Thanks for the pull request.  I made some modifications to this PR.  It was still possible to trigger a (much smaller) race condition with this current approach.  I've built on your original commit here: https://github.com/boto/botocore/pull/790

Closing in favor of #790.

Issue #762, opened 2016-01-19
Paginator breaks with NextMarker for "aws s3 ls" (non-recursive)


boivie:
Hi,
There is an issue with the `aws s3 ls` command when key names contains non-alphanumeric
characters. The issue only occurs when the number of results exceeds the page size value. 
Here's an example:
```
aws s3 ls s3://some-bucket/test/yearmonth=201601/
2016-01-19 12:10:42          0 file1
2016-01-19 12:10:46          0 file2
2016-01-19 12:10:49          0 file3
```
If we set the --page-size value to be less than the number of files then the following error occurs:
```
aws s3 --page-size=2 ls s3://some-bucket/test/yearmonth=201601/
2016-01-19 12:10:42          0 file1
2016-01-19 12:10:46          0 file2
2016-01-19 12:10:42          0 file1
2016-01-19 12:10:46          0 file2
Error during pagination: The same next token was received twice: ['test/yearmonth%3D201601/file2']
```
Note, listing files with the --recursive flag works fine.
```
aws s3 ls --page-size 2 --recursive s3://some-bucket/test/yearmonth=201601/
2016-01-19 12:10:42          0 test/yearmonth=201601/file1
2016-01-19 12:10:46          0 test/yearmonth=201601/file2
2016-01-19 12:10:49          0 test/yearmonth=201601/file3
```

The problem seems to be that the results for "aws s3 ls" (without recursive) gives the marker in "NextMarker" for the first response, and it's returned urlencoded (`=` becomes `%3D`) by the s3 service. The paginator will then specify that value (still urlencoded) as "Marker" for the following request. The S3 service expects it to be decoded (unquoted) (`%3D` becomes `=`).

One way to solve this - which may not be correct at all, is to modify `data/s3/2006-03-01/service-2.json`:
```
diff --git a/data/s3/2006-03-01/service-2.json b/data/s3/2006-03-01/service-2.json
index c3227f9..22972c5 100644
--- a/data/s3/2006-03-01/service-2.json
+++ b/data/s3/2006-03-01/service-2.json
@@ -3100,6 +3100,7 @@
         "Marker":{"shape":"Marker"},
         "NextMarker":{
           "shape":"NextMarker",
+          "location":"querystring",
[...]

PR #765, opened 2016-01-21
Adding missing url decoding from S3 responses.

mtdowling:
The previous commit to add automatic endcoding and decoding of S3 responses missed a few keys: https://github.com/boto/botocore/pull/726

Specifically, we were not decoding Marker, Delimiter, NextMarker, and CommonPrefixes[*].Prefix. The decode_list_object handler has been updated to decode each of these members in the parsed response if they are present.

This change will fix a CLI issue: https://github.com/boto/botocore/issues/762

For more information, see http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html

(supersedes and builds on https://github.com/boto/botocore/pull/752)

@jamesls @kyleknap @JordonPhillips @rayluo

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -600,22 +600,24 @@ def decode_list_object(parsed, context, **kwargs):
     # This is needed because we are passing url as the encoding type. Since the
     # paginator is based on the key, we need to handle it before it can be
     # round tripped.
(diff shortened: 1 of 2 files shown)


JordonPhillips:
:boat:

kyleknap:
Looks good. :ship:

Issue #339, opened 2014-09-10
What regions a service is available in?


garnaat:
Not really an issue, just a question.  We have reworked all of this code to use an algorithmic approach to computing the actual service endpoint for a region rather than looking that up in a JSON file.  That's good but is there anyway now to know what regions a service is available in?  For example, if I ask for a Redshift endpoint in ``us-west-1`` the ``Service`` object is happy to give me a nicely formed one but there is no Redshift service in that region.

Am I missing something?

Issue #594, opened 2015-06-26
Don't allow session with bad services or regions.


AMeng:
Boto should not allow you to create a session with a service and region combination that is not available.

For example, if I create a redshift session in us-west-1, boto allows me to create this session and then takes quite a while to fail my requests. I'm seeing response times in excess of 10 seconds for boto to raise an  `EndpointConnectionError`. I see the same thing when using the AWS CLI:

```
$ time aws --region=us-west-1 redshift describe-clusters

Could not connect to the endpoint URL: "https://redshift.us-west-1.amazonaws.com/"

real	0m11.699s
user	0m0.272s
sys	0m0.036s
```

It looks like a list of available endpoints is kept in the tests [here](https://github.com/boto/botocore/blob/80b0161316dbbe355149f619da4ecf1565123e93/tests/unit/test_regions.py#L27).

Ideally boto would raise an error when a user attempts to even create the session. But if that does not sound reasonable, it would be nice to at least have formal access to that `KNOWN_REGIONS` list so that I can perform the validation myself.

This is not an issue with EC2 because it provides the `describe_regions` method. Most services don't seem to provide that.

AMeng:
Oops. Looks like this is a duplicate of https://github.com/boto/botocore/issues/339

PR #769, opened 2016-01-25
Updating to use a new endpoint system

mtdowling:
This commit updates Botocore to utilize a new system for region and endpoint resolution, allowing customers to know the endpoints available for a service and endpoints available per/partition.

## Motivation

We currently allow customers to resolve endpoints based on a set of rules in _endpoints.json. While the current _endpoints.json system can resolve endpoints, it is not able to tell customers which regions and endpoints are available for a specific service, which services are available in a specific region, and which regions are available in a partition. The new proposed system for resolving endpoints can provide all of this missing information and provides a more declarative mechanism for describing endpoint data.

A secondary motivation of this PR is to help consolidate the various region and endpoints formats in the various AWS SDKs (similar PRs will be submitted to our other SDKs). By utilizing the same format as other SDKs, we can better share the maintenance burden of keeping endpoints up to date and make it more straightforward for third-parties implementing SDKs to understand and implement endpoint resolution.

Another important reason for this PR is to add a feature back into Botocore/Boto3 that previously existed in Boto2. Customers that were relying on similar region and endpoints metadata behavior in Boto2 would now be able to update to Boto3.

Other motivation:

- Resolves: #339, https://github.com/boto/boto3/issues/188
- Related: https://github.com/boto/botocore/issues/594, https://github.com/boto/boto3/issues/431
- This has also been a requested feature for the CLI, but the only solution currently is to issue an EC2 command that doesn't actually tell you which services are in which regions. For example: http://askubuntu.com/questions/94612/how-do-i-get-a-list-of-amazon-regions-via-the-command-line

## Resolving endpoints

In order to resolve an endpoint, an SDK or tool must be provided a service and region. We will use these values to determine if the provided combination is valid for a partition, and if so, what endpoint metadata to use.

### 1. Check if the service is in the partition

First check to see if the provided service name is present in the partition. This can be done by checking if the `services` definition contains a key matching the provided service value.

If the service cannot be found, then go to step 3.

### 2. Check if the provided region is valid for the service

Check to see if the provided region is present in the list of `endpoints` of the service. If the region is present in the list, then that specific endpoint will be utilized and we proceed to step 4 (endpoint resolution).

### 3. Determine the appropriate endpoint

If the region provided by the user does not match an exact endpoint name for the service, then check to see if the region provided matches one of the following constraints:

1. The region exactly matches one of the regions specified in the `regions` attribute of the partition.
2. The region matches the `regionRegex` regular expression attribute of the partition.

If none of the above cases are true, then we cannot construct an endpoint for the user, and must stop further endpoint resolution.

If the service being requested has a `parititonEndpoint` and `isRegionalized` is explicitly set to `false`, then use the endpoint named in the `partitionEndpoint` attribute of the service and go to step 4. This would occur for example if a user requested the `us-west-200` endpoint for IAM. At the time of writing this document, the `us-west-200` endpoint does not exist in the `aws` partition, IAM is not regionalized  `isRegionalized=false`, and IAM is has a `partitionEndpoint`, so we'll send the user to the partition endpoint.

If the service does not have a `partitionEndpoint`, `isRegionalized` is not explicitly set to `false`, or the exact service is not found in the partition, then we need to construct an endpoint for the user based on the default `hostname` URI template of the partition. For example, this could occur if (prior to the release of `us-east-200`) a user requested an S3 `us-east-200` endpoint. We're confident we can guess the appropriate endpoint based on the partition metadata, so we attempt it resolve the endpoint for the customer. Forcing users to update their SDK just to resolve a new endpoint for a standard region is a bad customer experience, so we offer regular expression based endpoint resolution. Allowing endpoint resolution based on a regular expression allows users to resolve endpoints for newer regions that may have not yet been pulled into the customer's copy of the SDK.

### 4. Assimilate endpoint data

Next we must create the actual resolved endpoint result dict. This result will be called `result`.

1. Take the `endpoint` definition for the endpoint of the service. The endpoint definition might be an empty object or an object with various key value pairs. If the endpoint is not present for the service or the service is not present in the partition, then start with an empty `{}`.
2. If `defaults` is set for the service, shallow merge each value present in the `defaults` hash onto the `result` is the value is not already present. If `result` already contains one of the values provided in `defaults`, then do not modify the existing value on the `result`.
3. If `defaults` is set for the partition, shallow merge each value present in the `defaults` hash onto the `result` if the value is not already present.
4. Expand the `hostname` URI template of the `result`, replacing `{service}` with the given service, `{region}` with the resolved endpoint name, and `dnsSuffix` with the `dnsSuffix` of the partition.

   Note that in Botocore, we will use the `sslCommonName` attribute if present
   when connecting to an endpoint. `sslCommonName` is also a URI template.

## Enumerating endpoints

A user can enumerate the available partitions and endpoints of a service using a session object:

```python
import botocore.session
session.get_available_partitions()
# ['aws', 'aws-cn', 'aws-us-gov']
session.get_available_regions('s3')
# [u'us-east-1', u'us-west-1', u'us-west-2', u'ap-northeast-1', u'ap-northeast-2', u'ap-southeast-1', u'ap-southeast-2', u'sa-east-1', u'eu-west-1', u'eu-central-1']
session.get_available_regions(service_name='s3', allow_non_regional=True)
# [u'us-east-1', u's3-external-1', u'us-west-1', u'us-west-2', u'ap-northeast-1', u'ap-northeast-2', u'ap-southeast-1', u'ap-southeast-2', u'sa-east-1', u'eu-west-1', u'eu-central-1']
session.get_available_regions(service_name='s3', partition_name='aws-us-gov')
# [u'us-gov-west-1']
```


botocore/endpoint.py
```diff
@@ -234,41 +248,39 @@ def create_endpoint(self, service_model, region_name=None, is_secure=True,
...
             final_endpoint_url = endpoint_url
         else:
-            final_endpoint_url = endpoint['uri']
+            # Construct a URI from the scheme and hostname. This will
+            # only resolve and endpoint. It won't override signature
+            # versions, signing names, etc. That logic is all in
+            # ClientCreator. Note that ClientCreator will inject an
+            # endpoint_url into create_endpoint(), so this branch is
+            # typically not invoked unless used directly.
+            final_endpoint_url = self._get_endpoint_from_resolver(
```
mtdowling: I kept this here for backwards compatibility, but I was wondering if we should refactor this slightly to require an endpoint_url and region, given that endpoint resolution occurs in ClientCreator (eliminating this conditional branch and the need for the EndpointCreator to utilize an endpoint resolver). Is this class deemed part of the public API (meaning we should not change this aspect of it), or would it be ok to make this change considering endpoint_url is always injected from ClientCreator?

mtdowling:
I just pushed a commit to this pull request that adds the ability to specify if you want to use an OrderedDict or not when loading data. This has a dramatic effect on the speed of loading partition data.

```
>>> timeit.timeit(
...     "loader.load_file('/Users/dowling/projects/botocore/botocore/data/partitions/aws')",
...     setup='import botocore.loaders; loader = botocore.loaders.JSONFileLoader()',
...     number=1000)
7.847162961959839
>>> timeit.timeit(
...     "loader.load_file('/Users/dowling/projects/botocore/botocore/data/partitions/aws', False)",
...     setup='import botocore.loaders; loader = botocore.loaders.JSONFileLoader()',
...     number=1000)
0.8665060997009277
```

We could also look at utilizing this for other JSON files that don't actually need to use OrderedDict.

kyleknap:
Cool. Since you are working on making the partitions into one file, I am going to try to review everything but the lazy file-loader logic.

botocore/data/partitions/aws-us-gov.json
```diff
@@ -0,0 +1,179 @@
...
+            "https"
+          ]
+        }
+      }
+    },
+    "sqs": {
+      "endpoints": {
+        "us-gov-west-1": {
+          "sslCommonName": "{region}.queue.{dnsSuffix}",
+          "signatureVersions": [
```
kyleknap: sqs uses "s3" and "s3v4" in only US Gov cloud? Seems fishy. I feel like it would not have a signature version specified.

botocore/endpoint.py
kyleknap: Anything in the ``endpoint`` module in my mind is deemed as internal and is free to change. I would be +1 with moving this logic out of endpoint. Have always wanted to move the endpoint resolution out of this class.

botocore/client.py
```diff
@@ -199,37 +176,59 @@ def _resolve_verify_value(self, provided_verify_value, scoped_config):
...
+        # Determine what region the user provided either via the
+        # region_name argument or the client_config. Note that
+        # endpoint resolvers are responsible for enforcing whether
+        # or not a region is required.
+        if region_name is None:
+            if client_config and client_config.region_name is not None:
+                region_name = client_config.region_name
+        return region_name
+
+    def _resolve_endpoint(self, region_name, endpoint_config, endpoint_url,
```
kyleknap: Question, one thing that have never liked about the client class related code is that it has a bunch of internal methods that are used to resolve endpoints. What would your thoughts be of having an abstraction on top of the endpoint resolver that resolves all of the defaults for these pieces of data required to correctly create the client. So this is how I would see it (with a better class name):
```py
class ClientDefaultResolver(object):
    def __init__(self, endpoint_resolver):
       self._endpoint_resolver = endpoint_resolver
   def get_default_values(self, service_model, region, is_secure):
       endpoint_config = self._endpoint_resolver.construct_endpoint(service_model.endpoint_prefix, region_name)
       # Move logic from and do it:
       # Client._resolve_endpoint
       # Endpoint.create_endpoint_url
       # Client._get_preferred_signature_version
       return {endpoint_url, region_name, signing_name, signature_version}
```
From there I would use this helper class to get some default values that can then be overriden by values passed in by the client's constructor within the client class.

The main reason I am suggesting this is because this class is getting really bloated and hard to follow. It would be nice to separate some of this logic that does not need to live in the client class, just needs to be used by it. Thoughts?

botocore/endpoint.py
```diff
@@ -38,6 +38,20 @@
 filter_ssl_warnings()
 
 
+def create_endpoint_url(endpoint_config, is_secure):
```
kyleknap: I would be in favor of not adding any of this logic to ``endpoint`` module. Anything to simplify/remove logic from this module is ideal for me. Maybe make it a part of the helper default value determiner class I was talking about?

botocore/loaders.py
```diff
@@ -140,7 +140,7 @@ def exists(self, file_path):
         """
         return os.path.isfile(file_path + '.json')
 
-    def load_file(self, file_path):
+    def load_file(self, file_path, use_ordered_dict=True):
```
kyleknap: Let's add the ``use_ordered_dict`` to the docstring.

botocore/endpoint.py
```diff
@@ -234,41 +248,39 @@ def create_endpoint(self, service_model, region_name=None, is_secure=True,
                         response_parser_factory=None, timeout=DEFAULT_TIMEOUT):
```
kyleknap: Might be better to address later, but it would be really cool if we could get rid of any arguments such as ``service_model``, ``region_name``, ``is_secure`` as a result of moving endpoint resolving logic out of the endpoint.

botocore/loaders.py
```diff
@@ -353,7 +389,7 @@ def load_service_model(self, service_name, type_name, api_version=None):
         return self.load_data(full_path)
 
     @instance_cache
-    def load_data(self, name):
+    def load_data(self, name, use_ordered_dict=True):
```
kyleknap: Hmm I wondering if there is a better way to expose ``use_ordered_dict`` value as this will only be applicable to JSON loading. So if we want to add a plain text loader as the ``FILE_LOADER_CLASS`` this would not make much sense. I will try to think of something.

botocore/data/partitions/aws-us-gov.json
mtdowling: Oh, good catch. That should not have been there.

botocore/endpoint.py
mtdowling: Yeah, I'm in favor of refactoring EndpointCreator a bit by removing endpoint and other unnecessary arguments. I take it you are too, so I'll make these changes.

kyleknap: Definitely infavor

botocore/endpoint.py
mtdowling: I'll move this method, and remove endpoint related code from the endpoint module.

botocore/client.py
mtdowling: I'll push the other changes, and then play with this idea. I'm not sure if it will help yet because it will need a lot of the collaborators present in the ClientCreator, but worth a look.

botocore/regions.py
```diff
@@ -10,89 +10,230 @@
...
+
+    def construct_endpoint(self, service_name, region_name):
+        for resolver in self.endpoint_resolvers:
+            endpoint = resolver.construct_endpoint(service_name, region_name)
+            if endpoint:
+                return endpoint
+        return None
+
+
+class LazyPartitionEndpointResolver(object):
```
kyleknap: Yeah making the partitions into one file is going to simplify a lot of this logic

botocore/client.py
kyleknap: Sounds good. I feel like it was worth mentioning as the logic in this class seems to keep growing and growing and at some point I feel like it will be due for a refactoring before it ends up like endpoint where too much unneeded logic lives directly in the class.

mtdowling:
I've updated the PR to use a single file for endpoint resolution. I'm still working on removing endpoint stuff from `endpoint`.

kyleknap:
Cool I will keep looking at it then.

botocore/regions.py
```diff
@@ -10,89 +10,194 @@
...
+        :type allow_non_regional: bool
+        :param allow_non_regional: Set to True to include endpoints that are
+             not regional endpoints (e.g., s3-external-1,
+             fips-us-gov-west-1, etc).
+        :return: Returns a list of endpoint names (e.g., ["us-east-1"]).
+        """
+        raise NotImplementedError
+
+
+class S3CompatResolver(BaseEndpointResolver):
```
kyleknap: I wonder if it would be better to change this around a bit. It seems sort of weird that there is a S3CompatResolver that wraps the EndpointResolver (which follows the specification to a T) because if there are ever further backwards compatibilities that we need to add the pattern seems to be that we will keep wrapping the original EndpointResolver. So like:
```py
        def create_default_resolver():
            loader = self.get_component('data_loader')
            endpoints = loader.load_data('endpoints', use_ordered_dict=False)
            return OtherCompatResolver(S3CompatResolver(EndpointResolver(endpoints)))
```
Especially since there are already compatibility logic in both the endpoint and the client layers, I feel that some compatibility layer that conjoins all of the various compatibility logic in each class would be very convenient as it will clean up the codebase by removing the class and make it easier to find and wrap your head around all of the compatibility logic.

botocore/data/endpoints.json
```diff
@@ -0,0 +1,1200 @@
+{
```
kyleknap: Oh can we name it ``_endpoints.json``? We have some scripts where we would have to update them to ``endpoints.json``.

botocore/loaders.py
kyleknap: Three ideas that have been juggling here to make it work for the use case I pointed out:
1) Add ``OBJECT_PAIRS_HOOK`` class variable in ``JSONLoader`` to store the class to be used for ``object_pairs_hook``. Downside is that we would have to create a new JSONLoader just for regions and endpoint resolution to avoid affecting loading of service models.
2) Add an extra set of args called ``load_file_params`` that is a dictionary that will be translated to keyword args passed to the underlying ``load_file`` method.
3) We save the use of non-ordered dict to another pull request when we have a better idea of how we want to account for these changes since endpoints are a single file now and load time is not affected as much.

I am not sure if I really like any of these ideas. My preference in order from highest to lowest would be 2, 3, 1. Thoughts?

botocore/client.py
```diff
@@ -106,22 +108,25 @@ def _register_retries(self, service_model):
...
-        # request signer.
-        resolver = self._endpoint_resolver
-        scheme = 'https' if is_secure else 'http'
-        endpoint_config = resolver.construct_endpoint(
-            service_model.endpoint_prefix, region_name, scheme=scheme)
-
-        # Signature version override from endpoint.
+    def _get_preferred_signature_version(self, versions):
+        # Use signers in the following preference order.
+        preference_order = ['s3', 's3v4', 'v4', 'v2']
```
jamesls: Could you explain how this works? I see that the endpoints file have a `signatureVersion` key in their config, for example, S3 has `signatureVersion: ['s3', 's3v4']`.  I don't quite follow why the client creator needs to know about signature order preferences.

kyleknap:
Looks good. I really like the single file format. Here are the three things that would be cool to get addressed:
1) Figure out how to combine all of the different compatibility logic.
2) Figure out what we want to do about the option for turning ordered dict on and off
3) Renaming endpoint.json to _endpoints.json

But overall is looking good to me.

botocore/endpoint.py
mtdowling: I just pushed a commit that removes this and other unrelated arguments from EndpointCreator. endpoint_url and region_name are now injected from ClientCreator.

botocore/loaders.py
mtdowling: Added

botocore/data/endpoints.json
mtdowling: I think this is actually an argument in favor of the rename -- it forces us (and anyone else relying on `_endpoints.json`) to update code to account for the change.

tests/unit/test_session.py
```diff
@@ -45,7 +46,7 @@ def setUp(self):
         self.environ_patch = mock.patch('os.environ', self.environ)
         self.environ_patch.start()
         self.environ['FOO_PROFILE'] = 'foo'
-        self.environ['FOO_REGION'] = 'moon-west-1'
```
jamesls: Interesting.  I assume this is because the region didn't match any of the regexs for the partition right?

To make sure I understand, this would mean a customer would have to upgrade if a new region prefix is added in a partition right?  I'm guessing there's not much we can do to avoid this is there?

botocore/client.py
mtdowling: The endpoint resolver will provide all of the metadata about an endpoint, including available signature versions. For example, `aws-cn` regions only support `s3v4` while most `aws` regions support `s3` and `s3v4`. Higher level code that consumes this endpoint metadata needs to make a choice as to how to use that metadata. In Boto3 and the CLI, we currently always pick `s3` over `s3v4`.

I'm open to suggestions on where we pick which signature version to use.

mtdowling:
@kyleknap Can you elaborate on "Figure out how to combine all of the different compatibility logic."?

kyleknap:
So ideally if we are starting a brand new sdk, we would pick our signing name, signature version, and endpoint url straight from the information provided by ``EndpointResolver``. But that is not the case since botocore is GA, so there are compatibility logic spread around botocore to account for this such as in client creation, the S3CompatResolver, and that create_endpoint method you moved out of ``endpoint`` to ``regions``. I am saying it would be nice to move all of the logic that we can into one class or location.

mtdowling:
Ah I see. I can take a crack at that, and possibly add the ability to fallback to just assuming that the user wants the default partition of no endpoint can be matched. This matches the current behavior, and we could possibly make it configurable as to which partition is assumed.

botocore/loaders.py
jamesls: 3

My reasons being:

1. This will apply to not just endpoints loading.  In fact, the only thing we actually use ordered dict for (that I can remember) is for consistent ordering in the docs for member names.  I'm not even sure botocore needs ordered dicts internally.  We might be able to default to dict() and have doc generation figure out some way to configure ordered dicts for html generation (not sure about inline docstrings in a repl)

2. We haven't really figured out how we're going to do text/rst file loading, and how that's implemented could change how (2) would be implemented.

botocore/data/endpoints.json
```diff
@@ -0,0 +1,1200 @@
...
+          }
+        },
+        "importexport": {
+          "partitionEndpoint": "aws-global",
+          "isRegionalized": false,
+          "endpoints": {
+            "aws-global": {
+              "hostname": "importexport.amazonaws.com",
+              "signatureVersions": [
+                "v2"
```
jamesls: Isn't this covered by the service-2.json for importexport having the signature version as v2?

botocore/client.py
jamesls: Do you have an example where the v2/v4 logic is needed?  I now follow the `s3->s3v4` preferences but I couldn't find an example of v4/v2.  v4 doesn't seem to be used in endpoints.json and `v2` is used for importexport but that syncs up with the service model.

botocore/data/endpoints.json
mtdowling: It is, however, I think we should no longer use the `signatureVersion` value from model files. With aws-cn and newer region deployments, it's clear that the signature version used with a service is not a dictation of the service but of an individual service endpoint.

botocore/loaders.py
mtdowling: Sure, I can remove it from the PR and do a follow up that adds it later.

tests/unit/test_session.py
mtdowling: I'm working right now on making it so that we can guess the endpoint if we're unable to resolve an endpoint. This would allow us to keep the existing behavior of guessing as a last resort.

mtdowling:
I've pushed an update that consolidates all of the bridging between Botocore and endpoint metadata into a single class, `ClientEndpointBridge`. The bridge class takes the endpoint values, extracts the data needed by the client creator, and returns the resolved hash. This change allows moves all endpoint resolution logic into a single class, moving it out of ClientCreator and removing `S3CompatResolver`. This change also made it significantly easier to test various edge cases.

I've also removed the `use_ordered_dict` change.

tests/unit/test_session.py
mtdowling: This has now been updated to guess an endpoint as a last resort.

kyleknap:
Big fan of the refactor. Looks great. Thanks for all the work! :ship:

botocore/client.py
```diff
@@ -239,16 +162,16 @@ def _get_client_args(self, service_model, region_name, is_secure,
             if client_config.user_agent_extra is not None:
                 user_agent += ' %s' % client_config.user_agent_extra
 
-        signer = RequestSigner(service_model.service_name, region_name,
-                               service_model.signing_name,
-                               signature_version, credentials,
-                               event_emitter)
+        signer = RequestSigner(
+            service_name, endpoint_config['signing_region'], service_name,
```
jamesls: Looks like signing_name isn't being used anymore.  How do we account for this now?

botocore/data/endpoints.json
jamesls: Interesting, must have missed that in the spec.  Didn't realize we were phasing out signature version from the model.

If that's the case, I'd prefer to move the default value of "v4" from botocore code into the endpoints json file somehow.  Seems like you should be able to derive what signature version to use entirely from the endpoints file.

botocore/client.py
```diff
@@ -325,6 +246,144 @@ def _api_call(self, *args, **kwargs):
...
+
+    This class handles taking out the relevant arguments from the endpoint
+    resolver and determining which values to use, taking into account any
+    client configuration options and scope configuration options.
+
+    This class also handles determining what, if any, region to use if no
+    explicit region setting is provided. For example, Amazon S3 client will
+    utilize "us-east-1" by default if no region can be resolved."""
+
+    SIGNATURE_PRECEDENCE = ['s3', 's3v4', 'v4', 'v2']
```
jamesls: Just want to make sure my previous question didn't get lost in the shuffle with the recent refactoring.

I get the s3->s3v4 preference.

What's an example of where we need the `v4/v2` preference?

botocore/client.py
```diff
@@ -325,6 +246,144 @@ def _api_call(self, *args, **kwargs):
...
+
+    def _check_default_region(self, service_name, region_name):
+        if region_name is not None:
+            return region_name
+        # Use the client_config region if no explicit region was provided.
+        if self.client_config and self.client_config.region_name is not None:
+            return self.client_config.region_name
+        # Use us-east-1 as the default region for S3.
+        if service_name == 's3':
+            return 'us-east-1'
```
jamesls: Is there anyway we can avoid hard coded logic of service names/region names in code?  I get this is special behavior we need, but previously this was captured in the old endpoints file.  Is there anyway this information can be moved into endpoints.json?

botocore/client.py
mtdowling: The preference order is a way for us to check for signature versions that we explicitly have signers for. If a service came out with a signer that we didn't understand, then none of our preferred signers would match and we'd use sigv4. The preferred signers order is checked in a specific order to capture how Botocore favors signature versions. We would use v2 for services that only have v2 support, but only if explicitly stated and only as a last resort. The preference order is SDK specific because other SDKs use the same data source but might prefer v4 over s3 (e.g., PHP and Ruby).

botocore/client.py
mtdowling: I don't see how we could capture a default region in the spec I'm presenting. Many of our other SDKs require a region and have no default, so a default region for S3 (currently the only default region service) is SDK specific. I don't see a better way to do this, but I'm open to suggestions.

botocore/client.py
mtdowling: Good catch. I can add this to the bridge class as a return value.

botocore/data/endpoints.json
mtdowling: We can add a default of "v4" to all of the partitions that are shipped in the default `endpoints.json`.

Do you think we should fail if we're given an endpoints file that does not contain a signature version or does not contain any signature versions we understand?

botocore/client.py
mtdowling: This has now been added and tested.

botocore/data/endpoints.json
jamesls: I believe the current behavior is to fail if it sees a signature version it doesn't understand.  I think that's also the case for no signature version.  I'm ok with keeping that behavior for now.

The other alternative we might have to change in the future is that no signature version present means the requests aren't signed but I don't think we'd need to change that now.

botocore/client.py
jamesls: Are there any specific services that use the "prefer v4 over v2" logic now?  I'm wondering if the s3->s3v4 is a special case because we can't migrate to s3v4 in a backwards compatible way just yet.  The prefer v4 over v2 case would only be triggered if a service endpoint has `['v4', 'v2']` in its `signatureVersion` list right?  Given that there's no entries right now that have that, and future services/endpoints will likely use v4, is this necessary?

Just trying to understand if this is addressing an existing issue we have, is more of a future proof thing, etc.

botocore/data/endpoints.json
mtdowling: If we ever did need unsigned requests, I think we could use something explicit like "anonymous" or something similar.

I can update the PR to fail in the case of being able to find no known signers.

botocore/client.py
jamesls: I'm hesitant to keep it in code, but I think if no one else feels that strongly about it, let's keep it.

As an aside, the only reason botocore's data files are called `_endpoints` and `_retry` are because back in the day, botocore didn't track service models by api version.  They were all in a single, top level directory by name, e.g: `['ec2.json', 's3.json', 'elb.json', ...]`.  Therefore, putting an `endpoints.json` in this same directory would make botocore think there's an "endpoints" service.  Even if the botocore loaders were hard coded to ignore `endpoints.json` from the `listdir()` call, we'd still be in trouble if an actual service called "endpoints" ever came out.  To workaround this, I just prefixed all non service model files with `_` and had the loaders ignore anything starting with `_`, giving us `_endpoints.json` and `_retry.json`.  This is no longer necessary now that each service is in its own directory, but we never went back and removed the leading `_` because there wasn't a real pressing reason to do so.  I never meant it as any kind of statement of internal/implementation detail.

botocore/client.py
mtdowling: It's both a future proof thing and an easy way to define which signers we support in Botocore. If we find one of the preferred signers in the list of signature versions, then we know that we can send requests to a service.

Checking in a predefined order is a future proof thing. The reason to check for v4 and prefer it over v2 is to correctly handle the case of one of the few remaining v2 services adding support for v4. At that point, it is unlikely the service would stop supporting v2, so we'd need to account for the case of both v2 and v4 as available signature versions.

botocore/data/endpoints.json
mtdowling: I just pushed the change so that we will fail if no known signature version can be found in signatureVersions.

mtdowling:
The only remaining concern was how signature versions are checked in a predetermined order: `['s3', 's3v4', 'v4', 'v2']`. The outstanding question was: should we move the two s3 signers out into an if statement so that s3 and s3v4 are only checked if the service is "s3".

I also want to capture a few followup items from this PR:

- Add corresponding endpoint methods to Boto3's Session
- Add a way to disable ordered_dict when loading JSON
- Add commands to the CLI that allow customers to list endpoints

botocore/client.py
jamesls: > It's both a future proof thing and an easy way to define which signers we support in Botocore. If we find one of the preferred signers in the list of signature versions, then we know that we can send requests to a service.

That's handled from the auth type maps though isn't it? https://github.com/boto/botocore/blob/develop/botocore/auth.py#L736

>The reason to check for v4 and prefer it over v2 is to correctly handle the case of one of the few remaining v2 services adding support for v4.  At that point, it is unlikely the service would stop supporting v2, so we'd need to account for the case of both v2 and v4 as available signature versions.

I'm comparing how we've rolled out v4 to services so far.  With the exception of S3 (which has its own special signature version anyways) the `v2->v4` upgrade has been completely transparent.  We just update the signatureVersion to say `v4` in the service model.  I picked a handful of services and flipped them back to v2, and they all still work (EC2, ELB, IAM, etc.).  I suspect every service that once supported v2 still supports v2.  Yet we don't call them out here in this endpoints file.  And I don't think we need to.  If the last remaining v2 service, importexport, flips to v4, I imagine we'd just update this file to say that we should start using `v4` for importexport wouldn't we?  There's no need to specify both signature versions in this file.

I'm just thinking if maybe this is just yet-another special case required for S3 only.

botocore/client.py
```diff
@@ -239,16 +163,18 @@ def _get_client_args(self, service_model, region_name, is_secure,
             if client_config.user_agent_extra is not None:
                 user_agent += ' %s' % client_config.user_agent_extra
 
-        signer = RequestSigner(service_model.service_name, region_name,
-                               service_model.signing_name,
-                               signature_version, credentials,
-                               event_emitter)
+        signer = RequestSigner(
+            service_name, endpoint_config['signing_region'],
+            endpoint_config['signing_name'],
```
jamesls: Looks like this isn't quite working.  Is it just because the signing names haven't been ported over from the models file?

```
$ aws iot-data get-thing-shadow --thing-name foo /tmp/a --region us-east-1

A client error (ForbiddenException) occurred when calling the GetThingShadow operation: Credential should be scoped to correct service.

$ aws dynamodbstreams list-streams --region us-east-1

A client error (InvalidSignatureException) occurred when calling the ListStreams operation: Credential should be scoped to correct service: 'dynamodb'.
```

If so, it's probably worth auditing all existing services with a signing name and making sure they're accounted for in the endpoints.json file.

botocore/client.py
```diff
@@ -325,6 +249,155 @@ def _api_call(self, *args, **kwargs):
...
+                        "Switching signature version for service %s "
+                        "to version %s based on config file override.",
+                        service_name, version)
+                    return version
+        # Pick a signature version from the endpoint metadata if present.
+        if 'signatureVersions' in resolved:
+            for preference in self.SIGNATURE_PRECEDENCE:
+                if preference in resolved['signatureVersions']:
+                    return preference
+        raise UnknownSignatureVersionError(
```
jamesls: This should probably not fail here unless we actually don't know the signature version.  The AUTH_TYPE_MAP map in auth.py has the definitive source of which signature versions we support.  There's a couple of other signature versions in with AUTH_TYPE_MAP that we support that will now fail, e.g `v3`.  While nothing uses that signature version anymore, it still seems like we should only claim an `UnknownSignatureVersionError` if it's truly an unknown signature version.

jamesls:
Just had some small feedback, otherwise looking good.

botocore/client.py
mtdowling: Yeah, I think it makes sense to not fail here but instead rely on `AUTH_TYPE_MAP` (as it's a separate concern) to see if the resolver signature version is supported.

Also related to https://github.com/boto/botocore/pull/769#discussion_r51489723:

The problem we have is that the list of `signatureVersions` provided from resolving an endpoint can contain one or more signature version names. We need to somehow look in the list and choose the most appropriate signature version for Botocore. We know the following:

1. If the service is s3

   a.  if "s3" is available, then use "s3"
   b. use "s3v4" if available.

2. If "v4" is available use it.
3. ?  -- at this point how do we choose a signer? Just take element `[0]` of the list? Or should we iterate over the list of `signatureVersions` until we find a signer that is supported in `AUTH_TYPE_MAP`?

botocore/client.py
mtdowling: Good catch. I just pushed an update that adds the missing credentialScope information for all of the services that are currently called out in the model, and I also added support for taking the `signingName` from models if `credentialScope.service` is not present. I was able to connect and successfully call each of these services with this change, and we will now continue to honor `signingName`.

botocore/client.py
mtdowling: I just pushed a commit that incorporates the above approach, but step 3 now iterates over the `signatureVersions` list in order and uses the first found value that matches one of the known supported signature versions present in `AUTH_TYPE_MAP`.

botocore/client.py
mtdowling: I agree in that S3 is a special case here. I've just pushed a commit to address this and I utilzied `AUTH_TYPE_MAP`.

mtdowling:
Feedback incorporated

botocore/client.py
jamesls: Is `SIGNATURE_PRECEDENCE` needed anymore?

mtdowling: No. My mistake for not removing it. It's removed now.

botocore/client.py
```diff
@@ -325,6 +251,173 @@ def _api_call(self, *args, **kwargs):
...
+        if region_name is not None:
+            return region_name
+        # Use the client_config region if no explicit region was provided.
+        if self.client_config and self.client_config.region_name is not None:
+            return self.client_config.region_name
+        # Use us-east-1 as the default region for S3.
+        if service_name == 's3':
+            return 'us-east-1'
+        # We require a region for everything except S3, so we must fail.
+        raise NoRegionError()
```
jamesls: S3 isn't the only service that doesn't require a region.  Anything in the current `_endpoints` file with a catchall rule will work, which typically were the "global" endpoints: `iam`, `cloudfront`, `sts`, etc.

So on the develop branch now with no region configured the above services will work, but now fail with a NoRegionError.  I'm thinking we probably still need to preserve this behavior for now.

mtdowling: That's a good point. I've pushed a change that will now not require a region for any service that has a partitionWide endpoint defined, but instead we'll assume the partitionWide endpoint. This makes importexport, cloudfront, waf, route53, s3, sts, and iam all work the same way they work now in the rules based approach, and it removes the hardcoded s3 default region of us-east-1. This approach also allows customers to provide completely custom partitions with partitionWide endpoints to get the same behavior without requiring a code change.

Does this approach sound ok?

botocore/session.py
```diff
@@ -796,6 +799,37 @@ def create_client(self, service_name, region_name=None, api_version=None,
...
+    def get_available_partitions(self):
+        """Lists the available partitions found on disk
+
+        :rtype: list
+        :return: Returns a list of partition names (e.g., ["aws", "aws-cn"])
+        """
+        resolver = self.get_component('endpoint_resolver')
+        return resolver.get_available_partitions()
+
+    def get_available_regions(self, service_name, partition_name='aws',
```
jamesls: I'm using this in a script I'm working on and I noticed that the `service_name` is actually the endpoint prefix.  What do you think about getting this updated to use botocore's service name?

The code I had was:

```
for service_name in session.get_available_services():
    for region_name in session.get_available_regions(service_name):
        # ...
```

Given other session methods that take service name use botocore's service names, it seems like `get_available_regions` should probably also use the same names:

```
>>> s.create_client('elb')
<botocore.client.ElasticLoadBalancing object at 0x10f82df10>
>>> s.get_service_model('elb')
<botocore.model.ServiceModel object at 0x10f780e10>
>>> s.get_available_regions('elb')
[]
>>> s.get_available_regions('elasticloadbalancing')
[u'us-east-1', u'us-west-1', u'us-west-2', ...]
```

botocore/client.py
jamesls: Sounds good to me.

botocore/session.py
mtdowling: Yeah, that's true. I just pushed a commit that allows you to use the short form of the name or the endpoint_prefix -- e.g., `elb` or `elasticloadbalancing`. We first try to resolve endpoints using the given name. If that result is empty, we then try to load the service data associated with the given service_name to grab it's `endpointPrefix`. If the service cannot be found (e.g., `foo`), then we return an empty list. Otherwise, we return the result of resolving endpoints using the service's `endpointPrefix`.

jamesls: Sounds good, it'd be worth updating the docstring to say that endpointprefix and service name are both supported.

jamesls:
:shipit:  Looks good.  Thanks for incorporating all the feedback.

mtdowling:
Great. I've pushed an update with the updated docblock. Let's merge this in tomorrow morning pending some internal tasks. After merging, we have the following remaining tasks:

[ ] Implement the ability to turn OrderedDict JSON loading on/off
[ ] Expose the new botocore session methods in Boto3
[ ] Add a command to the CLI to list regions and endpoints

PR #770, opened 2016-01-26
Add paginators for ssm

kyleknap:
We were missing paginators for ssm. Here are the API docs to double check: http://docs.aws.amazon.com/ssm/latest/APIReference/API_ListDocuments.html


jamesls:
:shipit:  Assuming travis runs cleanly.

WhileLoop:
Any reason there is no paginator for describe_parameters?

PR #779, opened 2016-01-30
Removing "max" trait validation.

mtdowling:
This commit updates Botocore to no longer validate the "max" constraint.
This allows customers to use older versions of Botocore/CLI even if the
service relaxes the "max" constraint in a subsequent release.

@jamesls @kyleknap @rayluo @JordonPhillips


kyleknap:
Looks good :ship:

PR #780, opened 2016-01-30
Wrap string bodies as StringIO

JordonPhillips:
There was an issue where young buckets would fail to PUT  when virtual host addressing was used and when the body was a string of a few MB. This was because urllib3 will send the body as a separate request on file-like objects, but not on strings. S3 sends back a redirect after the initial request. When we're using two requests, we see the redirect before sending the body. When using small strings, S3 will redirect it for us. But when using long strings, we will continue sending data until S3 closes the connection on us, resulting in an exception.

This commit fixes the issue by wrapping string bodies in a StringIO object, which is a file-like object.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -23,6 +23,7 @@ import xml.etree.cElementTree
 import copy
 import re
 import warnings
+import io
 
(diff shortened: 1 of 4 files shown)


JordonPhillips:
Closing this as I discovered some large issues with the current setup.

PR #783, opened 2016-02-02
SAML forms-based authentication

rayluo:
Now customers can configure a profile to authenticate against a SAML Identity Provider (IdP). Only forms-based authentication is supported so far. The profile will look like this.

Tested with Okta.

```
[default]
saml_endpoint = https://dev-123456.oktapreview.com/home/amazon_aws/hash/123
saml_authentication_type = form
saml_username = your_username@your_idp.com
role_arn = arn:aws:iam::123456789012:role/okta_websso
```

Tested with ADFS.

```
[profile jamesls]
saml_endpoint = https://your.adfs.example.com/adfs/ls/IdpInitiatedSignOn.aspx?loginToRp=urn:amazon:webservices
saml_authentication_type = form
saml_provider = adfs
role_arn = arn:aws:iam::123456789012:role/okta_websso
```

In both cases,
- `saml_username` can be provided by callback function, so it is optional in the configuration.
- ~~you don't have to specify the `role_arn`, CLI will prompt and ask for your input~~ You STILL NEED to specify the `role_arn` in config.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1106,7 +1106,7 @@ class SamlGenericFormsBasedAuthenticator(SamlAuthenticator):
         return config.get('saml_authentication_type') == 'form'
 
     def authenticate(self, config, username_prompter, password_prompter):
-        verify = config.get('saml_verify_ssl') != 'false'
+        verify = True
         if not config.get('saml_username'):
             config['saml_username'] = username_prompter("Username: ")
         endpoint = config['saml_endpoint']
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -35,7 +35,7 @@ from botocore.exceptions import PartialCredentialsError
 from botocore.exceptions import ConfigNotFound
 from botocore.exceptions import InvalidConfigError
 from botocore.exceptions import RefreshWithMFAUnsupportedError
-from botocore.exceptions import RefreshUnsupportedError, SamlError
+from botocore.exceptions import RefreshUnsupportedError, SAMLError
 from botocore.utils import InstanceMetadataFetcher, parse_key_val_file
 import botocore.vendored.requests as requests
 from botocore.client import Config
@@ -1052,12 +1052,12 @@ class AssumeRoleWithSAMLProvider(AssumeRoleProvider):
             raise ValueError("Unsupported saml_authentication_type: %s"
                              % config.get('saml_authentication_type'))
         if not assertion:
-            raise SamlError(
+            raise SAMLError(
                 detail='Failed to login at %s' % config['saml_endpoint'])
         idp_roles = self._parse_roles(assertion)
         role = self.role_selector(config.get('role_arn'), idp_roles)
         if not role:
-            raise SamlError(detail='Unable to choose role "%s" from %s' % (
+            raise SAMLError(detail='Unable to choose role "%s" from %s' % (
                 config.get('role_arn'), [r['RoleArn'] for r in idp_roles]))
         role['SAMLAssertion'] = assertion
         return role
@@ -1108,7 +1108,7 @@ class SAMLGenericFormsBasedAuthenticator(SAMLAuthenticator):
         endpoint = config['saml_endpoint']
         login_form = self._get_form(requests.get(endpoint, verify=verify).text)
         if login_form is None:
-            raise SamlError(detail='Login form is not found in %s' % endpoint)
+            raise SAMLError(detail='Login form is not found in %s' % endpoint)
         form_action = urljoin(endpoint, login_form.attrib.get('action', ''))
         remind = ''
         if not form_action.lower().startswith('https://'):
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -966,7 +966,20 @@ class AssumeRoleWithSamlProvider(AssumeRoleProvider):
                  password_prompter=getpass.getpass,
                  username_prompter=raw_input,
                  authenticators=None):
-        """
+        """To initiate an AssumeRoleWithSamlProvider.
+
+        :type role_selector: callable
+        :param role_selector: A function to choose a role based on both
+            user configuration and the roles allowed by IdP.
+
+        :type password_prompter: callable
+        :param password_prompter: A callable that receives a prompt for user,
+            and then returns a password provided by the user, such as getpass()
+
+        :type username_prompter: callable
+        :param username_prompter: A callable that receives a prompt for user,
+            and then returns the username provided by the user, such as input()
+
         :type authenticators: list
         :param authenticators: A list of authenticators, which are instances
             with an is_suitable() method and an authenticate() method.
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -38,6 +38,7 @@ from botocore.exceptions import RefreshWithMFAUnsupportedError
 from botocore.exceptions import RefreshUnsupportedError, SamlError
 from botocore.utils import InstanceMetadataFetcher, parse_key_val_file
 import botocore.vendored.requests as requests
+from botocore.client import Config
 
 
 logger = logging.getLogger(__name__)
@@ -1029,10 +1030,8 @@ class AssumeRoleWithSAMLProvider(AssumeRoleProvider):
             refresh_using=refresh_func)
 
     def _create_client(self):
-        # sts.assume_role_with_saml() requires no access keys,
-        # but we still need a pair of dummy values here to break recursion.
         return self._client_creator(
-            'sts', aws_access_key_id='dummy', aws_secret_access_key='dummy')
+            'sts', config=Config(signature_version='unsigned'))
 
     def _retrieve_temp_credentials(self):
         logger.debug("Retrieving credentials via AssumeRoleWithSaml.")
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -972,7 +972,9 @@ class AssumeRoleWithSamlProvider(AssumeRoleProvider):
         self.role_selector = role_selector
 
     def _create_cache_key(self):
-        return self._profile_name.replace(':', '_')
+        role_arn = self._get_role_config_values().get('role_arn')
+        cache_key = "%s--%s" % (self._profile_name, role_arn)
+        return cache_key.replace(':', '_').replace('/', '-')
 
     def _get_role_config_values(self):
         return self._loaded_config.get('profiles', {}).get(
@@ -1013,7 +1015,7 @@ class AssumeRoleWithSamlProvider(AssumeRoleProvider):
         return creds, response
 
     def _assume_role_base_kwargs(self, config):
-        if config.get('saml_authentication_type')=='form':
+        if config.get('saml_authentication_type') == 'form':
             if not config.get('saml_username'):
                 config['saml_username'] = self.username_prompter("Username: ")
             mapping = {
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -18,6 +18,10 @@ import os
 import getpass
 import threading
 from collections import namedtuple
+import re
+import xml.etree.ElementTree as ET
+import base64
+import json
 
 from dateutil.parser import parse
 from dateutil.tz import tzlocal
@@ -25,12 +29,16 @@ from dateutil.tz import tzlocal
 import botocore.config
 import botocore.compat
 from botocore.compat import total_seconds
+from botocore.compat import raw_input
+from botocore.compat import urljoin
 from botocore.exceptions import UnknownCredentialError
 from botocore.exceptions import PartialCredentialsError
 from botocore.exceptions import ConfigNotFound
 from botocore.exceptions import InvalidConfigError
 from botocore.exceptions import RefreshWithMFAUnsupportedError
+from botocore.exceptions import RefreshUnsupportedError
 from botocore.utils import InstanceMetadataFetcher, parse_key_val_file
+import botocore.vendored.requests as requests
 
 
 logger = logging.getLogger(__name__)
@@ -56,6 +64,12 @@ def create_credential_resolver(session):
     providers = [
         env_provider,
         AssumeRoleProvider(
+            load_config=lambda: session.full_config,
+            client_creator=session.create_client,
+            cache={},
+            profile_name=profile_name,
+        ),
+        AssumeRoleWithSamlProvider(
             load_config=lambda: session.full_config,
             client_creator=session.create_client,
             cache={},
@@ -147,6 +161,21 @@ def create_mfa_serial_refresher():
         # when the temp creds expire.
         raise RefreshWithMFAUnsupportedError()
     return _refresher
+
+
+def credential_normalizer(credentials):
+    # We need to normalize the credential names returned from assume_role()
+    # or assume_role_with_saml(), to the values expected by the refresh creds.
+    return {
+        'access_key': credentials['AccessKeyId'],
+        'secret_key': credentials['SecretAccessKey'],
+        'token': credentials['SessionToken'],
+        'expiry_time': _serialize_if_needed(credentials['Expiration']),
+    }
+
+
+def exception_raiser(exception):
+    raise exception()
 
 
 class Credentials(object):
@@ -825,7 +854,7 @@ class AssumeRoleProvider(CredentialProvider):
         role_config = self._get_role_config_values()
         # On windows, ':' is not allowed in filenames, so we'll
         # replace them with '_' instead.
-        role_arn = role_config['role_arn'].replace(':', '_')
+        role_arn = role_config[self.ROLE_CONFIG_VAR].replace(':', '_')
         role_session_name=role_config.get('role_session_name')
         if role_session_name:
             cache_key = '%s--%s--%s' % (self._profile_name, role_arn, role_session_name)
@@ -919,6 +948,127 @@ class AssumeRoleProvider(CredentialProvider):
         if config['role_session_name'] is not None:
             assume_role_kwargs['RoleSessionName'] = config['role_session_name']
         return assume_role_kwargs
+
+
+class AssumeRoleWithSamlProvider(AssumeRoleProvider):
+    METHOD = 'assume-role-with-saml'
+    ROLE_CONFIG_VAR = 'saml_endpoint'
+
+    def _create_cache_key(self):
+        return self._profile_name.replace(':', '_')
+
+    def _get_role_config_values(self):
+        settings = self._loaded_config.get('profiles', {}).get(
+            self._profile_name, {})
+        if 'saml_params' in settings and '_params_list' not in settings:
+            settings['_params_list'] = json.loads(settings['saml_params'])
+        return settings
+
+    def _create_creds_from_response(self, response):
+        config = self._get_role_config_values()
+        need_human_input = any(
+            p for p in config.get('_params_list', []) if 'value' not in p)
+        if need_human_input:
+            # If some parameter(s) would require prompting the user for input,
+            # we use a different refresh_func.
+            # We can explore an option in the future to support reprompting for
+            # input, but for now we just error out when the temp creds expire.
+            refresh_func = lambda: exception_raiser(RefreshUnsupportedError)
+        else:
+            refresh_func = lambda: credential_normalizer(
+                self._create_client().assume_role_with_saml(
+                    **self._assume_role_base_kwargs(config))['Credentials'])
+        return RefreshableCredentials(
+            access_key=response['Credentials']['AccessKeyId'],
+            secret_key=response['Credentials']['SecretAccessKey'],
+            token=response['Credentials']['SessionToken'],
+            method=self.METHOD,
+            expiry_time=_parse_if_needed(
+                response['Credentials']['Expiration']),
+            refresh_using=refresh_func)
+
+    def _create_client(self):
+        return self._client_creator(
+            'sts', aws_access_key_id='dummy', aws_secret_access_key='dummy')
+
+    def _retrieve_temp_credentials(self):
+        logger.debug("Retrieving credentials via AssumeRoleWithSaml.")
+        response = self._create_client().assume_role_with_saml(
+            **self._assume_role_base_kwargs(self._get_role_config_values()))
+        creds = self._create_creds_from_response(response)
+        return creds, response
+
+    def _assume_role_base_kwargs(self, config):
+        assertion = self._get_saml_assertion(
+            config[self.ROLE_CONFIG_VAR],
+            self._input_params(config.get('_params_list', [])))
+        if not assertion:
+            raise ValueError('Login failed')
+        roles = self._get_roles(assertion)
+        if not roles:
+            raise ValueError('Identity provider provides no role.')
+        if len(roles) > 1:
+            logging.debug('TODO: Let user choose one among many roles.')
+        else:
+            role = roles[0]
+        role['SAMLAssertion'] = assertion
+        return role
+
+    def _get_form(self, html, form_xpath='(//form)[0]'):
+        # TODO: Support form_xpath
+        form_snippet = re.search('(<form.+</form>)', html, flags=re.DOTALL)
+        if form_snippet:
+            return ET.fromstring(form_snippet.group(0))
+
+    def _get_saml_assertion(self, login_url, params):
+        login_form = self._get_form(requests.get(login_url).text)
+        if login_form is None:
+            raise ValueError('Login form is not found in %s' % login_url)
+        payload = dict((tag.attrib['name'], tag.attrib.get('value', ''))
+                        for tag in login_form.findall(".//input"))
+        for param in params:
+            if param['field'] in payload:
+                payload[param['field']] = param.get('value', '')
+        aws_form = self._get_form(requests.post(
+            urljoin(login_url, login_form.attrib['action']),
+            data=payload).text)
+        if aws_form is not None:
+            responses = aws_form.findall(".//input[@name='SAMLResponse']")
+            if responses:
+                return responses[0].attrib['value']
+        # Login failed, typically caused by incorrect username and/or password
+
+    def _input_params(self, params):
+        # This implementation collects input from terminal. Suitable for CLI.
+        for param in params:
+            if 'value' in param:
+                continue
+                # So we keep the predefined value as-is.
+                # But then it means user could use this to store password.
+                # We'll probably remove this feature in finalized design.
+            if param.get('type')=='password':
+                input_method = getpass.getpass
+            else:
+                input_method = raw_input
+            prompt = param.get('label', param['field'])  # A good UI needs it
+            param['value'] = input_method('%s: ' % prompt)
+        return params
+
+    def _get_roles(self, assertion):
+        attribute = '{urn:oasis:names:tc:SAML:2.0:assertion}Attribute'
+        attr_value = '{urn:oasis:names:tc:SAML:2.0:assertion}AttributeValue'
+        awsroles = []
+        for attr in ET.fromstring(base64.b64decode(assertion)).iter(attribute):
+            if attr.get('Name')=='https://aws.amazon.com/SAML/Attributes/Role':
+                for value in attr.iter(attr_value):
+                    parts = value.text.split(',')
+                    # Deals with "role_arn,pricipal_arn" or its reversed order
+                    if 'saml-provider' in parts[0]:
+                        role = {'PrincipalArn': parts[0], 'RoleArn': parts[1]}
+                    else:
+                        role = {'PrincipalArn': parts[1], 'RoleArn': parts[0]}
+                    awsroles.append(role)
+        return awsroles
 
 
 class CredentialResolver(object):
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1111,20 +1111,16 @@ class SAMLGenericFormsBasedAuthenticator(SAMLAuthenticator):
         if login_form is None:
             raise SAMLError(detail='Login form is not found in %s' % endpoint)
         form_action = urljoin(endpoint, login_form.attrib.get('action', ''))
-        remind = ''
         if not form_action.lower().startswith('https://'):
-            remind = ("Your IdP is not using secure HTTPS connection. "
-                      "Abort if you don't want to proceed.\n")
-            logger.warn(remind)
+            raise SAMLError(detail='Your SAML IdP must use HTTPS connection')
         if not config.get('saml_username'):
-            config['saml_username'] = username_prompter(remind + "Username: ")
+            config['saml_username'] = username_prompter("Username: ")
         payload = dict((tag.attrib['name'], tag.attrib.get('value', ''))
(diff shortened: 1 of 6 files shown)


jamesls:
High level feedback:

1. Could you squash your commits into logical commits?
2. Could you expand the PR description?   Assume people have no idea what the feature is (or new people later on may want to look at this PR).

Looking at this now.

botocore/credentials.py
```diff
@@ -16,6 +16,10 @@
 import logging
 import os
 import getpass
+import re
+import xml.etree.ElementTree as ET
+import base64
+import json
```
jamesls: import from compat module.

rayluo: Turns out my current implementation does not need to use json module at all. I'll remove that in next commit.

botocore/credentials.py
```diff
@@ -791,6 +821,157 @@ def _assume_role_base_kwargs(self, config):
...
+                 password_prompter=getpass.getpass,
+                 username_prompter=raw_input):
+        super(AssumeRoleWithSamlProvider, self).__init__(
+            load_config, client_creator, cache, profile_name)
+        self.username_prompter = username_prompter
+        self.password_prompter = password_prompter
+        self.role_selector = role_selector
+
+    def _create_cache_key(self):
+        return self._profile_name.replace(':', '_')
```
jamesls: I don't think this is unique enough.  We need to account for the principal/role arns don't we?

botocore/credentials.py
```diff
@@ -791,6 +821,157 @@ def _assume_role_base_kwargs(self, config):
...
+
+    def _retrieve_temp_credentials(self):
+        logger.debug("Retrieving credentials via AssumeRoleWithSaml.")
+        response = self._create_client().assume_role_with_saml(
+            **self._assume_role_base_kwargs(self._get_role_config_values()))
+        creds = self._create_creds_from_response(response)
+        return creds, response
+
+    def _assume_role_base_kwargs(self, config):
+        if config.get('saml_authentication_type')=='form':
```
jamesls: spaces around the `==`

botocore/credentials.py
```diff
@@ -791,6 +821,157 @@ def _assume_role_base_kwargs(self, config):
...
+            if field in payload:
+                payload[field] = value
+        response_form = self._get_form(requests.post(
+            urljoin(endpoint, login_form.attrib.get('action', '')),
+            data=payload, verify=verify).text)
+        if response_form is not None:
+            return self._get_value_of_first_tag(
+                response_form, 'input', 'name', 'SAMLResponse')
+        # Login failed, typically caused by incorrect username and/or password
+        return None
```
jamesls: I think it would help users if we distinguish between couldn't find a SAML assertion vs. login failed (one is something we need to fix, the other is something the user can fix).  Maybe just checking the status code would suffice?

botocore/credentials.py
```diff
@@ -791,6 +821,157 @@ def _assume_role_base_kwargs(self, config):
...
+
+    def _get_value_of_first_tag(self, root, tag, attr, trait):
+        ## This is backported from the following Python 2.7+ implementation:
+        # found = root.findall(".//tag[@attr='trait']")
+        # return found[0].attrib.get('value') if found else None
+        for element in root.findall(tag):
+            if element.attrib.get(attr)==trait:
+                return element.attrib.get('value')
+
+    def _get_form(self, html):
```
jamesls: What's this doing?  I read the comment but I still don't understand why this method is necessary.

botocore/credentials.py
```diff
@@ -791,6 +821,157 @@ def _assume_role_base_kwargs(self, config):
...
+                response_form, 'input', 'name', 'SAMLResponse')
+        # Login failed, typically caused by incorrect username and/or password
+        return None
+
+    def _get_value_of_first_tag(self, root, tag, attr, trait):
+        ## This is backported from the following Python 2.7+ implementation:
+        # found = root.findall(".//tag[@attr='trait']")
+        # return found[0].attrib.get('value') if found else None
+        for element in root.findall(tag):
+            if element.attrib.get(attr)==trait:
```
jamesls: spaces around `==`

botocore/credentials.py
```diff
@@ -791,6 +821,157 @@ def _assume_role_base_kwargs(self, config):
...
+        return creds, response
+
+    def _assume_role_base_kwargs(self, config):
+        if config.get('saml_authentication_type')=='form':
+            if not config.get('saml_username'):
+                config['saml_username'] = self.username_prompter("Username: ")
+            mapping = {
+                "adfs": {
+                    "username": "ctl00$ContentPlaceHolder1$UsernameTextBox",
+                    "password": "ctl00$ContentPlaceHolder1$PasswordTextBox"},
```
jamesls: what's the significance of the placeholder here?

botocore/credentials.py
```diff
@@ -791,6 +821,157 @@ def _assume_role_base_kwargs(self, config):
...
+                    if 'saml-provider' in parts[0]:
+                        role = {'PrincipalArn': parts[0], 'RoleArn': parts[1]}
+                    else:
+                        role = {'PrincipalArn': parts[1], 'RoleArn': parts[0]}
+                    awsroles.append(role)
+        return awsroles
+
+
+class SamlFormsBasedAuthenticator(object):
+    def authenticate(self, endpoint, params, verify=True):
```
jamesls: Docstring.  I'm following this call from the assume role provider and don't quite understand what `params` is suppose to be.

botocore/credentials.py
```diff
@@ -791,6 +821,157 @@ def _assume_role_base_kwargs(self, config):
...
+            **self._assume_role_base_kwargs(self._get_role_config_values()))
+        creds = self._create_creds_from_response(response)
+        return creds, response
+
+    def _assume_role_base_kwargs(self, config):
+        if config.get('saml_authentication_type')=='form':
+            if not config.get('saml_username'):
+                config['saml_username'] = self.username_prompter("Username: ")
+            mapping = {
+                "adfs": {
```
jamesls: What's this mapping for?  (might also be nice to name it something more clear).

botocore/credentials.py
rayluo: It is not a placeholder. It is the field names used by ADFS forms-based login page.

botocore/credentials.py
rayluo: It is used to grab the "<form>...</form>" part out of the login page, AND we do it again on the response page. So this logic is encapsulated here.

botocore/credentials.py
```diff
@@ -791,6 +821,157 @@ def _assume_role_base_kwargs(self, config):
...
+                 mapping["password"]: self.password_prompter("Password: ")},
+                verify=config.get('saml_verify_ssl') != 'false')
+        else:
+            raise ValueError("Unsupported saml_authentication_type: %s"
+                             % config.get('saml_authentication_type'))
+        if not assertion:
+            raise ValueError('Login failed: SAML assertion not found')
+        idp_roles = self._parse_roles(assertion)
+        if not idp_roles:
+            raise ValueError('Identity provider provides no role.')
```
jamesls: Is this a `ValueError`?  If the IdP doesn't provide any roles it's not necessarily because of a bad value error.

rayluo: Good point. What exception would you like to suggest then? TypeError? Typically, "IdP doesn't provide any roles" is the administrator's fault.

jamesls:
Quick question:  How can a user extend the form based login?

Let's say someone wanted to use this with form based auth, but they need to customize how they retrieve the SAML assertion.

So that we're not in the critical path here, can we slightly refactor this to make the SAML retriever pluggable?  Seems like it shouldn't be too bad.

botocore/credentials.py
jamesls: And okta uses the default fields of "username"/"password"?

Maybe renaming the variable would make this more clear.  It might also be worth pulling out of this method so users can easily modify this mapping.

botocore/credentials.py
rayluo: As far as I know, SAML response always ends up in an HTTP 200 OK status code, so we can not rely on status code. Typically, an unsuccessful login will result in a plain page saying something like "login fails", possibly with a (re)login form, but there won't be a form with `<input name="SAMLResponse" value="..."/>`. So there is not really a difference here.

botocore/credentials.py
rayluo: Done.

botocore/credentials.py
rayluo: Do you mean something like this?

    mapping = {
                "adfs": {
                    "username": "ctl00$ContentPlaceHolder1$UsernameTextBox",
                    "password": "ctl00$ContentPlaceHolder1$PasswordTextBox"},
                "okta": {
                    "username": "username",
                    "password": "password"},
                }.get(config.get('saml_provider'),
                      {"username": "username", "password": "password"})

But I kind of purposely not explicitly give "okta" a name, simply because its field settings are so generic. The point is I don't want in future, when we find another 3rd IdP named "FooBar" who happens to also use the same generic field names "username" and "password", and then we need to configure it as "saml_provider=okta".

We can pull this part of logic out to become a separated function, though.

rayluo:
Since previous commit, the SAML authenticators are organized in a pluggable style.

botocore/credentials.py
```diff
@@ -791,6 +820,193 @@ def _assume_role_base_kwargs(self, config):
...
+    METHOD = 'assume-role-with-saml'
+    DISTINCTION_VAR = 'saml_endpoint'
+
+    def __init__(self, load_config, client_creator, cache, profile_name,
+                 role_selector=_role_selector,
+                 password_prompter=getpass.getpass,
+                 username_prompter=raw_input,
+                 authenticators=None):
+        """
+        :type authenticators: list
```
kyleknap: Docstrings for the params would be nice here. Not completely obvious what all of the params represent.

botocore/credentials.py
```diff
@@ -791,6 +820,193 @@ def _assume_role_base_kwargs(self, config):
...
+            method=self.METHOD,
+            expiry_time=_parse_if_needed(
+                response['Credentials']['Expiration']),
+            refresh_using=refresh_func)
+
+    def _create_client(self):
+        # sts.assume_role_with_saml() requires no access keys,
+        # but we still need a pair of dummy values here to break recursion.
+        return self._client_creator(
+            'sts', aws_access_key_id='dummy', aws_secret_access_key='dummy')
```
kyleknap: Are dummy credenitals still needed even though this is anonymous? https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L681

botocore/credentials.py
```diff
@@ -791,6 +820,193 @@ def _assume_role_base_kwargs(self, config):
...
+
+class SamlGenericFormsBasedAuthenticator(SamlAuthenticator):
+    username_field = 'username'
+    password_field = 'password'
+
+    def is_suitable(self, config):
+        return config.get('saml_authentication_type') == 'form'
+
+    def authenticate(self, config, username_prompter, password_prompter):
+        verify = config.get('saml_verify_ssl') != 'false'
```
kyleknap: Why would we give them the option to turn ssl off when they may be sending a password over the wire? I do not see a reason of why they would want to. We should just always have it on to help protect the user from unknowingly setting this.

botocore/credentials.py
```diff
@@ -791,6 +820,193 @@ def _assume_role_base_kwargs(self, config):
...
+                        role = {'PrincipalArn': parts[0], 'RoleArn': parts[1]}
+                    else:
+                        role = {'PrincipalArn': parts[1], 'RoleArn': parts[0]}
+                    awsroles.append(role)
+        return awsroles
+
+
+class SamlAuthenticator(object):
+    def is_suitable(self, config):
+        """Return True if this instance intends to perform authentication"""
```
kyleknap: Docstrings of parameters would be good here as well.

kyleknap: Especially for the structure of the callable for the password prompter. Was not obvious that the string ``"Password: "`` was used to call it.

botocore/credentials.py
rayluo: Yes, they are still needed. Otherwise the underlying create_client() will trigger a load_credentials(), which ends up in a infinite loop.

kyleknap: That's unfortunate. We should look into fixing that eventually, but not needed now.

botocore/credentials.py
rayluo: On a second thought, perhaps I should just rename all those `config` to `profile`. That should be self-documented. Isn't it?

botocore/credentials.py
jamesls: I meant not keeping the mapping in the method.  This seems like it should be a class constant or something so people can extend it if they need to.

botocore/credentials.py
kyleknap: Interesting. It would be cool if we at least debug the reason why it failed or pass some information in the raised SamlError to help the user out a bit. As opposed to just raising an opaque "oh sorry this failed" message for the first if statement of checking the existence of ``assertion``.

botocore/credentials.py
jamesls: :+1: Strongly agree.  Turning off SSL verification should not be possible here.

botocore/credentials.py
kyleknap: Ehh how about ``profile_config``? I would still like a small snippet of what it represents.

botocore/credentials.py
```diff
@@ -791,6 +820,206 @@ def _assume_role_base_kwargs(self, config):
...
+        # return found[0].attrib.get('value') if found else None
+        for element in root.findall(tag):
+            if element.attrib.get(attr) == trait:
+                return element.attrib.get('value')
+
+    def _get_form(self, html):
+        # Scrape a form from html page, and return it as an elementtree element
+        form_snippet = re.search('(<form.+</form>)', html, flags=re.DOTALL)
+        if form_snippet:
+            # To handle &nbsp;, on Python 2 we can use an undocumented parser:
```
jamesls: I think my previous comment wasn't clear.  Why do we need *this portion* of the code.  I'm not understanding:

1) what it means to handle &nbsp.  In what way were we not properly handling spaces?  How was it breaking?
2) how injecting some html snippet with the entity decls helps.

botocore/credentials.py
```diff
@@ -791,6 +820,206 @@ def _assume_role_base_kwargs(self, config):
...
+        ## This is backported from the following Python 2.7+ implementation:
+        # found = root.findall(".//tag[@attr='trait']")
+        # return found[0].attrib.get('value') if found else None
+        for element in root.findall(tag):
+            if element.attrib.get(attr) == trait:
+                return element.attrib.get('value')
+
+    def _get_form(self, html):
+        # Scrape a form from html page, and return it as an elementtree element
+        form_snippet = re.search('(<form.+</form>)', html, flags=re.DOTALL)
```
jamesls: Also, why are we using a regex to parse through html?  There's a number of edge cases this doesn't handle.  If we're going to return a DOM element anyways, why not use the dom api/xpath/etc to find the form element in a more robust manner?

botocore/credentials.py
```diff
@@ -791,6 +820,206 @@ def _assume_role_base_kwargs(self, config):
...
+            # but it won't work on Python 3.
+            # So we use a pure XML way to handle it, for now.
+            return ET.fromstring(
+                '''<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
+                "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd" [
+                <!ENTITY nbsp ' '>
+                ]>''' + form_snippet.group(0))
+
+
+class SamlAdfsFormsBasedAuthenticator(SamlGenericFormsBasedAuthenticator):
```
jamesls: pep8, all your class names should have SAML, not Saml.

botocore/credentials.py
```diff
@@ -791,6 +820,206 @@ def _assume_role_base_kwargs(self, config):
...
+        role['SAMLAssertion'] = assertion
+        return role
+
+    def _parse_roles(self, assertion):
+        attribute = '{urn:oasis:names:tc:SAML:2.0:assertion}Attribute'
+        attr_value = '{urn:oasis:names:tc:SAML:2.0:assertion}AttributeValue'
+        awsroles = []
+        root = ET.fromstring(base64.b64decode(assertion))
+        for attr in root.getiterator(attribute):
+            if attr.get('Name')=='https://aws.amazon.com/SAML/Attributes/Role':
```
jamesls: spaces around the `==`.  I'd suggest just running flake8/pep8 through this PR code and fixing up any styling issues that get flagged.

tests/unit/test_credentials.py
```diff
@@ -972,5 +973,137 @@ def test_source_profile_does_not_exist(self):
...
+    @mock.patch(GET, return_value=mock.Mock(text=adfs_login_form))
+    def test_unmatch_role(self, _get):
+        profile = {
+            'saml_endpoint': 'http://example.com/login.asp',
+            'saml_authentication_type': 'form',
+            'saml_provider': 'adfs',
+            'saml_username': 'joe',
+            'role_arn': 'arn:aws:iam::123456789012:role/does_not_match',
+        }
+        with self.assertRaises(botocore.exceptions.SamlError):
```
kyleknap: Maybe use ``assertRaisesRegexp`` to make sure it was the iam role that caused the failure as ``SamlError`` can be thrown for a few reasons?

botocore/credentials.py
```diff
@@ -791,6 +820,206 @@ def _assume_role_base_kwargs(self, config):
         return assume_role_kwargs
 
 
+def _role_selector(role_arn, roles):
```
jamesls: pep8, functions at the top, not inbetween classes.

also docstring conventions (pep257) summary on one line, blank line, remaining docs.

botocore/credentials.py
```diff
@@ -791,6 +820,206 @@ def _assume_role_base_kwargs(self, config):
...
+        :param authenticators: A list of authenticators, which are instances
+            with an is_suitable() method and an authenticate() method.
+            You can use it to add your own implementation for 3rd party IdP.
+        """
+        super(AssumeRoleWithSamlProvider, self).__init__(
+            load_config, client_creator, cache, profile_name)
+        self.username_prompter = username_prompter
+        self.password_prompter = password_prompter
+        self.role_selector = role_selector
+        self.authenticators = authenticators or [
```
jamesls: anytime you have a default value of `None`, you should be using `if authenticators is not None`.  A user may want an empty list of authenticators to start with (maybe they'll inject them later).

botocore/credentials.py
```diff
@@ -791,6 +820,206 @@ def _assume_role_base_kwargs(self, config):
...
+    username_field = 'username'
+    password_field = 'password'
+
+    def is_suitable(self, config):
+        return config.get('saml_authentication_type') == 'form'
+
+    def authenticate(self, config, username_prompter, password_prompter):
+        verify = config.get('saml_verify_ssl') != 'false'
+        if not config.get('saml_username'):
+            config['saml_username'] = username_prompter("Username: ")
```
kyleknap: @jamesls I am not sure about this but does directly changing a loaded config in its dictionary form, can it affect other components of the session? I forget.  I want to say no, but it may be safer to treat the config as read only.

botocore/credentials.py
```diff
@@ -791,6 +820,206 @@ def _assume_role_base_kwargs(self, config):
...
+                    config=config, username_prompter=self.username_prompter,
+                    password_prompter=self.password_prompter)
+                break
+        else:
+            raise ValueError("Unsupported saml_authentication_type: %s"
+                             % config.get('saml_authentication_type'))
+        if not assertion:
+            raise SamlError(detail='Login failed: SAML assertion not found')
+        idp_roles = self._parse_roles(assertion)
+        if not idp_roles:
```
kyleknap: I would be nice to get a test case for when identity provider provides no roles. Or you could look into consolidating it into the SamlError below it.

botocore/credentials.py
jamesls: This doesn't seem that hard to fix does it?  I'd prefer not to keep workarounds like this in code if a reasonable fix isn't too much work.

Rather that solve the recursion issue, what about  if you allow creating a client with a config that has a signature version of unsigned?  Then we don't try to load credentials and instead use dummy creds.  Seems like a much cleaner way to handle this, has the added benefit of allowing creation of unsigned clients (useful for anon s3 access), and is relatively low effort to implement.

botocore/credentials.py
rayluo: I don't have preference here. Just want to share the background here, so that we make an informed decision.

This (undocumented) option is necessary, only when authenticating against some testing SAML IdP setup. This would probably happen again, when customer administrator is setting up their internal SAML IdP from scratch, without a SSL certificate yet, and wanting to test its function really quick. Besides, it will be the CLI customer who setup that most important "saml_endpoint" setting in their `~/.aws/credentials`, so it seems logically true that they should know what IdP they are sending their login credentials to. If they don't trust an IdP, they don't use it in the first place. An "saml_verify_ssl = false" just provides options when needed.

Let me know your preference. I am all ears. Meanwhile I'm finding some security review too.

tests/unit/test_credentials.py
```diff
@@ -972,5 +973,137 @@ def test_source_profile_does_not_exist(self):
             provider.load()
 
 
+class TestSamlGenericFormsBasedAuthenticator(unittest.TestCase):
+    authenticator = credentials.SamlGenericFormsBasedAuthenticator()
```
jamesls: these should be added to setUp().  That's the standard way of test fixture setup.

botocore/credentials.py
kyleknap: I say we always verify ssl until we get requests to have the option. It would be an additive change if we add the option.

tests/unit/test_credentials.py
```diff
@@ -972,5 +973,137 @@ def test_source_profile_does_not_exist(self):
...
+    @mock.patch(GET, return_value=mock.Mock(text='<html>wrong way</html>'))
+    def test_login_form_not_exist(self, _get):
+        with self.assertRaises(botocore.exceptions.SamlError):
+            if self.authenticator.is_suitable(self.profile):
+                self.authenticator.authenticate(
+                    self.profile, lambda prompt: 'foo', lambda prompt: 'bar')
+
+    @mock.patch(POST, return_value=mock.Mock(text='<html>login failed</html>'))
+    @mock.patch(GET, return_value=mock.Mock(text=login_form))
+    def test_no_saml_assertion(self, _get, _post):
```
jamesls: these don't need underscores, they're only accessible in the local scope anyways.

tests/unit/test_credentials.py
kyleknap: Agreed

botocore/credentials.py
jamesls: >  Besides, it will be the CLI customer who setup that most important "saml_endpoint" setting in their ~/.aws/credentials, so it seems logically true that they should know what IdP they are sending their login credentials to. 

No, that's exactly why this should not be an option.  You *cannot* know you're talking to the correct IdP when you disable SSL validation.

This options really needs to be removed.

botocore/credentials.py
rayluo: Good catch. Also, after some minor change, I ran pep8/pyflake/flake8 just now, they report no warning (except those preexist ones).

botocore/credentials.py
rayluo: Does it also include changing the class `AssumeRoleWithSamlProvider` to `AssumeRoleWithSAMLProvider`? Such `ContainingCAPITALWordInTheMiddle` style just look a little bit weird to me.

botocore/credentials.py
rayluo: OK then. The fix is easy: `verify = True`.

tests/unit/test_credentials.py
rayluo: It is true that their proper name may be "patched_get", "patched_post". The fact is, the code here doesn't even need to use those patched methods in most cases. In this case, do you mind I just use a single underline character (`_`) as that parameter name? I think using `_` to represent "this is a variable which I don't care" is a convention, but I haven't checked that practice in PEP8 or something. I'm just curious to know.

botocore/credentials.py
kyleknap: That would be cool

botocore/credentials.py
rayluo: Documenting the off-line conversation here.

* Feature-wise, the current implementation is capable to handle the ADFS and Okta IdP; and the plugin architecture is flexible to support new IdPs.
* We are still open to explore a different implementation approach based on HTML Parser. We will schedule this part as a lower priority for now.

tests/unit/test_credentials.py
rayluo: OK. By default I'll move this instance to setUp(). For the other 3, they are just constant definition, so I'll leave them here. Let me know if you think otherwise.

botocore/credentials.py
rayluo: It would be hard to debug the reason and present it to user, because the login fail page is totally undefined in SAML standard. That being said, for a forms-based authentication, user can always open the endpoint url in browser and manually login and see what goes wrong. I'll change that "assertion not found" exception message to include the endpoint url, so that the user will be encouraged to try it out by themselves.

botocore/credentials.py
rayluo: Please refer to [this conversation](https://github.com/boto/botocore/pull/783#discussion_r51782767)

botocore/credentials.py
jamesls: I see this has been already updated, just wanted to follow up on your question to make sure we're on the same page.  It's not about which one looks weird or not (I personally don't think either one is objectively better), it's about following our standard coding convention, pep8, so we don't have to get into discussions of "preferences" of individual developers.  Relevant part from pep8 (emphasis mine):

>CapitalizedWords (or CapWords, or CamelCase -- so named because of the bumpy look of its letters [3] ). This is also sometimes known as StudlyCaps.

> Note: **When using abbreviations in CapWords, capitalize all the letters of the abbreviation**. Thus HTTPServerError is better than HttpServerError.

botocore/credentials.py
jamesls: That's a good catch.  I did a quick search through botocore, and I didn't find anything that was modifying the config dif.  I agree with you, that has the potential to be buggy, and I'd suggest treating config as read only and not modifying the config dict.

tests/unit/test_credentials.py
jamesls: Can't you just use a context manager to handle patching if that's your concern?

tests/unit/test_credentials.py
jamesls: If you don't need the value provided in a mock.patch you can add the mock value as an argument to patch().  Then the decorated function won't be passed in the arguments.

botocore/session.py
```diff
@@ -779,7 +779,9 @@ def create_client(self, service_name, region_name=None, api_version=None,
         event_emitter = self.get_component('event_emitter')
         response_parser_factory = self.get_component(
             'response_parser_factory')
-        if aws_secret_access_key is not None:
+        if config and config.signature_version == 'unsigned':
```
kyleknap: Let's use the constant for ``UNSIGNED`` in botocore: https://github.com/boto/botocore/blob/develop/botocore/__init__.py#L54

botocore/credentials.py
```diff
@@ -791,6 +829,212 @@ def _assume_role_base_kwargs(self, config):
...
+        endpoint = config['saml_endpoint']
+        login_form = self._get_form(requests.get(endpoint, verify=verify).text)
+        if login_form is None:
+            raise SAMLError(detail='Login form is not found in %s' % endpoint)
+        form_action = urljoin(endpoint, login_form.attrib.get('action', ''))
+        remind = ''
+        if not form_action.lower().startswith('https://'):
+            remind = ("Your IdP is not using secure HTTPS connection. "
+                      "Abort if you don't want to proceed.\n")
+            logger.warn(remind)
```
kyleknap: Will the user have enough time to abort if this message is displayed to them? I wonder if prompting or straight up erroring out is better.

botocore/session.py
kyleknap: Can we also add a functional or unit test to make sure ``UNSIGNED`` is used for the signature if it is specified in the client config as we are now making it a full fledged feature?

botocore/credentials.py
```diff
@@ -791,6 +829,212 @@ def _assume_role_base_kwargs(self, config):
...
+        if not config.get('saml_username'):
+            config['saml_username'] = username_prompter(remind + "Username: ")
```
kyleknap: I think I mentioned this before but can we not modify the config to be safe from unintended consequences by making it read-only?

kyleknap:
Looks good. Just had a few comments.

botocore/credentials.py
rayluo: The user will have enough time to response, because the flow will stop waiting user to input username and/or password anyway. Besides, I also put this remind in the username and password remind.

I'm open to other approach. If we choose to error out, then do we also provide a way for user to confirm "yes I know I'm not protected by HTTPS but I still want to do that"? If yes, how? Another config option?

BTW, the purpose of this change is literally: `Before sending the username & password to the SAML Provider, verify if its using HTTPS, if not raise a warning and make the customer agree to send their credentials through a non-encrypted channel.`

botocore/credentials.py
rayluo: Well, actually I knew it was a side effect when I wrote it like this at the first place. I thought it was harmless (or even arguably desirable), because we already asked user to input it once, so we could save it here just in case we would need it later. That being said, I don't mind to remove it in my next commit.

kyleknap:
Thanks! Looks good. :ship:

botocore/session.py
```diff
@@ -779,7 +779,9 @@ def create_client(self, service_name, region_name=None, api_version=None,
         event_emitter = self.get_component('event_emitter')
         response_parser_factory = self.get_component(
             'response_parser_factory')
-        if aws_secret_access_key is not None:
+        if config and config.signature_version == UNSIGNED:
```
jamesls: Use `is` for object identity.  If `config.signature_version is UNSIGNED`.

jamesls:
Changes so far look good!

Looks like the only pending feedback is the research into using HTMLParser vs. the regex for form parsing.

botocore/credentials.py
rayluo: [This commit](https://github.com/rayluo/botocore/commit/9543c2fbdb65ceb3419907ac06689e99800b193a) implements an HTML Parser. Because this approach is also not perfect (see [the previous conversation](https://github.com/boto/botocore/pull/783#discussion_r51782767)), I tentatively keep the previous regex approach for now, so that we can switch between them more easily. Let me know if you want to get rid of regex. New commit is ready for review.

jamesls:
Superseded via https://github.com/boto/botocore/pull/1316

PR #784, opened 2016-02-02
Assert ``expected_params`` stubber

kyleknap:
Allows you to specify what parameters that you are to expect for a response and if they do not match, throw an error.

I have been writing a fair amount of tests using this stubber. I have found this sort of response/expected_params matching as being pretty useful and was using it an awful lot.

The only behavior I am not sure about and would be up for opinions is what happens to the stubber's corresponding response when an ``expected_params`` does not match. Should it use up the response in the queue? Or should it keep the original response in the queue so that it can be called later? I decided to go with the later because in the current stubber logic if there is a parameter validation error with the client, we will not pop the response off of the queue. So I thought ``expected_params`` would have a similar expectation. 

### DIFF
--- a/botocore/stub.py
+++ b/botocore/stub.py
@@ -54,7 +54,9 @@ class Stubber(object):
             'Marker': ''
         }
 
-        stubber.add_response('list_objects', response)
+        expected_params = {'Bucket': 'test-bucket'}
+
+        stubber.add_response('list_objects', response, expected_params)
         stubber.activate()
 
         service_response = s3.list_objects(Bucket='test-bucket')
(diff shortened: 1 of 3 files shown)


mtdowling:
Nice! :ship:

> I decided to go with the later because in the current stubber logic if there is a parameter validation error with the client, we will not pop the response off of the queue.

Seems reasonable. It probably won't matter much in practice -- once a mismatch occurs, an exception is thrown and the test will likely halt at that point.

botocore/stub.py
```diff
@@ -154,7 +176,13 @@ def add_client_error(self, method, service_error_code='',
         }
 
         operation_name = self.client.meta.method_to_api_mapping.get(method)
-        response = (operation_name, (http_response, parsed_response))
+        # Note that we do not allow for expected_params in the
+        # adding errors into the queue yet.
```
JordonPhillips: I would replace 'in the' with 'while'.

JordonPhillips:
:boat: Nice!

jamesls:
:shipit:

PR #787, opened 2016-02-08
Remove caching in request signer

jamesls:
This simplifies the code and fixes potential edge cases from
caching.

This was a legacy artifact from having to support both the
service/operation objects and the client interfaces at the
same time.  This is now no longer needed.

The only concern would be if caching actually helps performance.
To verify this, I ran a test consisting of:

* Mocking out the _send_request in endpoints.py to return a precanned
  response.
* Making a client call 1000 times (I chose ec2 DescribeInstances
  arbitrarily).  This is essentially timing the time it takes from
  a client method until it hits the http layer in botocore.

I ran the tests against botocore's develop branch and this PR.  There's
negligible difference::

```
    (simple-auth %) $ repeat 10 python scripts/signerperf.py
    Total time: 0.7235
    Total time: 0.7022
    Total time: 0.7190
    Total time: 0.7032
    Total time: 0.7112
    Total time: 0.7059
    Total time: 0.7099
    Total time: 0.6895
    Total time: 0.7079
    Total time: 0.7254
    (simple-auth %) $ git checkout develop
    Switched to branch 'develop'
    (develop %) $ repeat 10 python scripts/signerperf.py
    Total time: 0.7085
    Total time: 0.6812
    Total time: 0.7016
    Total time: 0.6995
    Total time: 0.7055
    Total time: 0.7314
    Total time: 0.6950
    Total time: 0.7105
    Total time: 0.7140
    Total time: 0.7129
```

cc @kyleknap @rayluo @JordonPhillips

PR #790, opened 2016-02-11
Improve thread safety of temp cred refresh

jamesls:
This builds on the pull request from https://github.com/boto/botocore/pull/764 and introduces a few additional enhancements for multithreaded credential refresh.  Includes:

* The original fix from #764.  Credentials must be accessed as a fixed set to ensure you don't get access_key from time t_1 and secret_key from time t_2.
* Introduce locking to ensure only one thread starts the refresh process.  The previous behavior was that every thread, once determining a refresh was needed, would all call the refresh function.  Additionally, double checked locking is used to avoid lock acquisition in the common case where a refresh isn't needed.
* Don't hard fail if refreshing initially fails.

To expand on the last point, there's now two time periods for refresh, the advisory period and the mandatory period.

##### Advisory period

* Only the thread that acquires the lock will refresh.  If a thread can't acquire the refresh lock during this period it uses the existing temporary creds.
* If the refresh function fails, a warning is logged, and the existing set of temporary creds are used.

##### Mandatory period

* All threads block on the refresh lock.
* Any exceptions raised during the refresh function will propagate back to the user.

There's still a few improvements that can be added (backoff delay during advisory period), but I think this is a good set of improvements.

Note: this builds on https://github.com/boto/botocore/pull/787 so the diff will show up here until that PR gets merged.

cc @kyleknap @rayluo @JordonPhillips

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -203,7 +203,12 @@ class RefreshableCredentials(Credentials):
     :ivar session: The ``Session`` the credentials were created for. Useful for
         subclasses.
     """
-    refresh_timeout = 15 * 60
+    # The time at which we'll attempt to refresh, but not
+    # block if someone else is refreshing.
+    _advisory_refresh_timeout = 15 * 60
+    # The time at which all threads will block waiting for
+    # refreshed credentials.
+    _mandatory_refresh_timeout = 10 * 60
 
     def __init__(self, access_key, secret_key, token,
                  expiry_time, refresh_using, method,
@@ -214,8 +219,10 @@ class RefreshableCredentials(Credentials):
         self._token = token
         self._expiry_time = expiry_time
         self._time_fetcher = time_fetcher
-        self._refresh_lock = threading.RLock()
+        self._refresh_lock = threading.Lock()
         self.method = method
+        self._frozen_credentials = ReadOnlyCredentials(
+            access_key, secret_key, token)
         self._normalize()
 
     def _normalize(self):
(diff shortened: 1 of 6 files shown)


botocore/credentials.py
```diff
@@ -255,27 +270,99 @@ def _seconds_remaining(self):
...
+                self._refresh_lock.release()
+        elif self.refresh_needed(self._mandatory_refresh_timeout):
+            # If we're within the mandatory refresh window,
+            # we must block until we get refreshed credentials.
+            with self._refresh_lock:
+                if not self.refresh_needed(self._mandatory_refresh_timeout):
+                    return
+                self._protected_refresh(mandatory=True)
+
+    def _protected_refresh(self, mandatory):
```
kyleknap: Ah so ``mandatory`` is suggesting if refresh is mandatory. Maybe call it ``is_mandatory``?

botocore/credentials.py
```diff
@@ -289,6 +376,43 @@ def _set_from_data(self, data):
...
 
+    def get_frozen_credentials(self):
+        """Return immutable credentials.
+
+        The ``access_key``, ``secret_key``, and ``token`` properties
+        on this class will always check and refresh credentials if
+        needed before returning the particular credentials.
+
+        This has an edge case where you can get inconsistent
+        credentials.  Image this:
```
kyleknap: Image --> Imagine

tests/functional/test_credentials.py
```diff
@@ -0,0 +1,76 @@
...
+        threads = []
+        for _ in range(20):
+            threads.append(threading.Thread(target=func, args=(collected,)))
+        start = time.time()
+        for thread in threads:
+            thread.start()
+        for thread in threads:
+            thread.join()
+        for triple in collected:
+            self.assertTrue(triple[0] == triple[1] == triple[2], triple)
```
kyleknap: Might be good to explain what is in the triple and why you are checking that they are all equal. It is because we keep track of how many times refresh is called and the number of refresh times all need to be the same in order to ensure that the credentials set all came from the same refresh. Right?

tests/__init__.py
```diff
@@ -242,3 +247,76 @@ def cmd(self, *cmd):
...
+            self._current_datetime() +
+            datetime.timedelta(seconds=creds_last_for))
+        if refresh_function is None:
+            refresh_function = self._do_refresh
+        super(IntegerRefresher, self).__init__(
+            '1', '1', '1', expires_in,
+            refresh_function, 'INTREFRESH')
+        self.creds_last_for = creds_last_for
+        self.refresh_counter = 0
+        if advisory_refresh is not None:
```
kyleknap: What does ``None`` do for you here. You already have a default set. Not sure if the ``if`` check is needed.

jamesls:

tests/functional/test_credentials.py
jamesls: I added a comment here that explains what's going on, but yes, that's what we're verifying.

tests/__init__.py
```diff
@@ -242,3 +247,74 @@ def cmd(self, *cmd):
...
+                 advisory_refresh=_advisory_refresh_timeout,
+                 mandatory_refresh=_mandatory_refresh_timeout,
+                 refresh_function=None):
+        expires_in = (
+            self._current_datetime() +
+            datetime.timedelta(seconds=creds_last_for))
+        if refresh_function is None:
+            refresh_function = self._do_refresh
+        super(IntegerRefresher, self).__init__(
+            '1', '1', '1', expires_in,
```
kyleknap: Is their a particular reason you picked ``1`` as the starting point? It is a little confusing in the unit test that you assert that the value is ``2`` which I would see as the credentials being successfully refreshed twice when you are making sure that they get immediately refreshed. Maybe start at 0? Or have a helper function for asserting how many times credentials were refreshed?

tests/functional/test_credentials.py
```diff
@@ -0,0 +1,97 @@
...
+        # we should see a maximum number of refreshes being (6 / 2.0) + 1 = 4
+        max_calls_allowed = math.ceil((end - start) / 2.0) + 1
+        self.assertTrue(creds.refresh_counter <= max_calls_allowed,
+                        "Too many cred refreshes, max: %s, actual: %s, "
+                        "time_delta: %.4f" % (max_calls_allowed,
+                                              creds.refresh_counter,
+                                              (end - start)))
+
+    def test_no_race_for_immediate_expiration(self):
+        creds = IntegerRefresher(
```
kyleknap: How does this setup for immediate expiration? If the creds are set to last for 1 sec and the advisory period is 1 sec while the mandatory refresh is at 0 sec, should the credentials still be in the advisory period? Or is there a clock tick somewhere I am missing in my logic?

kyleknap:
I only had a couple questions about the tests that be cool to get addressed. Otherwise looks good, :ship:

tests/__init__.py
jamesls: I think starting at 0 makes sense.  That way the value of akid/sak/token is the number of refreshes that have occurred.

tests/functional/test_credentials.py
jamesls: I can rename this, but it was immediate "advisory" expiration, that is, the credentials need to be refreshed immediately after they're refreshed.

kyleknap:
Thanks. :ship:

PR #792, opened 2016-02-12
Re-added and updated  documentation on supported Lambda runtimes

Bitbored:
Documentation on supported runtimes got removed in 9b14a65, however I think having this information is pretty useful.

Python2.7 has been added as a supported runtime.


JordonPhillips:
We don't generally accept PR's to the models because we source those from the service teams. In this case, valid values are documented in the Request Syntax section:

thing

PR #793, opened 2016-02-13
Fix nested lists in service doc strings.

JordonPhillips:
Fixes boto/boto3#310

There were three issues at play:

1. We were completely disabling newlines when going through lists
2. The doc strings contain a vestigial space after each `<li>` and we also add a space after a list item opening, resulting in two spaces before content starts.
3. Nested lists need to be indented, but the top level list should not be.


JordonPhillips:
I had some tests to fix, and I updated, but I'm gonna take a look at the output to make sure everything looks good.

JordonPhillips:
Closing this because the approach changed wildly

PR #795, opened 2016-02-16
NextToken should be a string, not bytes.

JordonPhillips:
On python 3, the output of base64 encoding is a `bytes` object, which does not play well with the JSON encoder / decoder.

cc @jamesls @kyleknap @rayluo


jamesls:
:shipit:

PR #797, opened 2016-02-17
Fix create_serializer() when invoked without validation

carl0FF:
If ``create_serializer()`` is called with ``include_validation=False`` an error will ocurr. This pull request fixes it by recovering the original intent of the function: decorating the corresponding serializer depending on the given value of the parameter ``include_validation``.


JordonPhillips:
Could you add some tests to make sure it still serializes properly without validation?

carl0FF:
Tests added :)

jamesls:
Merged in 1a915f199cede0dcbb3ce9fdbc4e12869b06eb15, thanks for the pull request.

PR #798, opened 2016-02-17
Fix issue where AWS_CA_BUNDLE was not being used

kyleknap:
Was not using the session's configuration and instead relied on the scoped config.

@jamesls @rayluo @JordonPhillips


tests/unit/test_client.py
```diff
@@ -1094,21 +1094,6 @@ def test_client_s3_addressing_style_path_registers_correctly(self):
             client.meta.events.register.call_args_list
         )
 
-    def test_create_with_ca_bundle_from_scoped_config(self):
```
jamesls: Are these tests no longer relevant or were they wrong before?

jamesls:
:shipit:

tests/unit/test_client.py
kyleknap: We talked about it, but the logic is now handled in the session's get_config_variable() method

PR #731, opened 2015-11-28
fixed parsers.BaseJSONParser error.

tell-k:
botocore authors.

First, thanks for the awesome tool.
I encountered errors such as the following.
This error seems to occur when  status  504 is returned from CloudSearch.

```
    parsed = self._do_error_parse(response, shape)
  File "/Users/tell_k/.virtualenvs/testproj-sns/lib/python3.4/site-packages/botocore/parsers.py", line 639, in _do_error_parse
    error = super(RestJSONParser, self)._do_error_parse(response, shape)
  File "/Users/tell_k/.virtualenvs/testproj-sns/lib/python3.4/site-packages/botocore/parsers.py", line 495, in _do_error_parse
    body = json.loads(response['body'].decode(self.DEFAULT_ENCODING))
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/json/__init__.py", line 318, in loads
    return _default_decoder.decode(s)  
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/json/decoder.py", line 343, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/json/decoder.py", line 361, in raw_decode
    raise ValueError(errmsg("Expecting value", s, err.value)) from None
ValueError: Expecting value: line 1 column 1 (char 0)
```

dump response object.

```
{'body': b'', 'status_code': 504, 'headers': {'connection': 'keep-alive', 'content-length': '0'}}
```

The cause of this error is because parsers.BaseJSONParser cannot handle empty response body well.
I have fixed so that botocore.exceptions.ClientError is raised. Please feel free to merge it.

see also http://docs.aws.amazon.com/cloudsearch/latest/developerguide/uploading-data.html#bulk-uploads

thx.

PR #572, opened 2015-06-17
Always populate a 'Message' key in the errors dict

jamesls:
Fixes #532.

The error message will now look like:

```
ClientError: An error occurred (InvalidInput) when calling the ListResourceRecordSets operation:
```

cc @kyleknap @mtdowling

PR #799, opened 2016-02-18
Handle 504 error from cloudsearch

jamesls:
This pulls in https://github.com/boto/botocore/pull/731 with a few changes:

* Move up common functionality into base class.  The logic for parsing a JSON body, handling encoding, and defaulting to an empty dict when no body is present is now consolidated to a single base class method.
* Group all error tests into a single class.  The eventual goal is to pull these into the shared protocol tests,.  Grouping them into a single test class makes this easier.

It's hard to see the new test because I've grouped all the error tests together, but the test case for this bug fix is here: https://github.com/boto/botocore/pull/799/files#diff-16d168f91b157ec88dcddaf854150629R503

Closes #731.

cc @kyleknap @rayluo @JordonPhillips

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -507,8 +507,11 @@ class BaseJSONParser(ResponseParser):
         return self._timestamp_parser(value)
 
     def _do_error_parse(self, response, shape):
-        body = json.loads(response['body'].decode(self.DEFAULT_ENCODING))
-        error = {"Error": {}, "ResponseMetadata": {}}
+        if response['body']:
+            body = json.loads(response['body'].decode(self.DEFAULT_ENCODING))
+        else:
+            body = {}
+        error = {"Error": {"Message": None, "Code": None}, "ResponseMetadata": {}}
         # Error responses can have slightly different structures for json.
         # The basic structure is:
         #
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -507,21 +507,18 @@ class BaseJSONParser(ResponseParser):
         return self._timestamp_parser(value)
 
     def _do_error_parse(self, response, shape):
-        if response['body']:
-            body = json.loads(response['body'].decode(self.DEFAULT_ENCODING))
-        else:
-            body = {}
-        error = {"Error": {"Message": None, "Code": None}, "ResponseMetadata": {}}
+        body = self._parse_body_as_json(response['body'])
+        error = {"Error": {"Message": '', "Code": ''}, "ResponseMetadata": {}}
         # Error responses can have slightly different structures for json.
         # The basic structure is:
         #
         # {"__type":"ConnectClientException",
         #  "message":"The error message."}
-        # Code, Type
 
         # The error message can either come in the 'message' or 'Message' key
         # so we need to check for both.
-        error['Error']['Message'] = body.get('message', body.get('Message'))
+        error['Error']['Message'] = body.get('message',
+                                             body.get('Message', ''))
         code = body.get('__type')
         if code is not None:
             # code has a couple forms as well:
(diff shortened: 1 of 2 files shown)


botocore/parsers.py
```diff
@@ -507,18 +507,18 @@ def _handle_timestamp(self, shape, value):
         return self._timestamp_parser(value)
 
     def _do_error_parse(self, response, shape):
-        body = json.loads(response['body'].decode(self.DEFAULT_ENCODING))
-        error = {"Error": {}, "ResponseMetadata": {}}
+        body = self._parse_body_as_json(response['body'])
+        error = {"Error": {"Message": '', "Code": ''}, "ResponseMetadata": {}}
```
kyleknap: I feel like you have pushed back on this before where if the Message or Code cannot be found we set a default empty value because it hides up a possible problem in our JSON error parsing. Am I remembering correctly or not?

I am personally fine with having the empty default value because having a runtime AttributeNotFound/KeyNotFound error is a pretty bad experience. But just checking that you are fine with it.

jamesls: I can dig into this.  Do you happen to remember any PR/issues?  I see a similar change a while back that added defaults in the parser: https://github.com/boto/botocore/pull/572/files

The only thing that I can remember offhand was that I didn't want *upstream* sources to silently handle these, that is, having the Exception classes handle missing error keys.  Doing that would mean that every layer up would have to gracefully handle the absence of these keys.

botocore/parsers.py
```diff
@@ -535,6 +535,13 @@ def _inject_response_metadata(self, parsed, headers):
             parsed.setdefault('ResponseMetadata', {})['RequestId'] = (
                 headers['x-amzn-requestid'])
 
+    def _parse_body_as_json(self, body_contents):
+        if not body_contents:
+            return {}
```
kyleknap: This is just me thinking out loud but should there a difference between a body with an empty string and a body with an empty JSON structure? It is really weird in a first place an empty string is the body of a JSON protocol service. I think that there is not a great way to represent it other than an empty dictionary. Have a value of ``None`` is not a good option because there could be no body on the response.

botocore/parsers.py
kyleknap: I think the other one I thought of was this: https://github.com/boto/botocore/pull/677. In this case, we catch the exception and log why it failed for future reference.

It looks like there is precedence on this so I'd be fine with having defaults.

kyleknap:
Looks fine. Had a couple comments but I do not think any actions need to be taken on them. :ship:

botocore/parsers.py
jamesls: It would be nice to use something different, but I don't really have any other ideas for what we could use instead of `{}`.  I agree that `None` isn't a good option.

botocore/parsers.py
jamesls: Ah yes I remember that PR.  I think the common theme is that I'd prefer to try to fix failures at their root cause rather than fixing the "effects" of those failure.  In that case, #677 was fixing the the symptom of a bug rather than the "real" root cause, which ended up being https://github.com/boto/botocore/pull/690.

With the `Code/Message` issue, I remember that rather than just have the Exception class handle missing `code/messages`, I wanted to update our parsers/serializers to ensure we always populate code/message, or in this specific case to use the missing `__type` key we weren't looking at.

Would it be fair to say that this isn't the same issue here?

kyleknap: Yeah I think so. I am just saying if there is an issue in the future in the way we parse for a message and code, it may get covered up by having defaults such as with this line:
```py
+        error['Error']['Message'] = body.get('message',
+                                             body.get('Message', ''))
```
But yet again, I am fine with defaults.

PR #801, opened 2016-02-19
Automatically wrap S3 string upload bodies with BytesIO.

JordonPhillips:
This is to avoid sending large initial requests to S3, which can result in S3 terminating the connection and a broken pipe on our end.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -25,7 +25,7 @@ import re
 import warnings
 
 from botocore.compat import urlsplit, urlunsplit, unquote, \
-    json, quote, six, unquote_str, ensure_unicode
+    json, quote, six, unquote_str, ensure_bytes
 from botocore.docs.utils import AutoPopulatedParam
 from botocore.docs.utils import HideParamFromOperations
 from botocore.docs.utils import AppendParamDocumentation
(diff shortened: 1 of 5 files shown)


botocore/handlers.py
```diff
@@ -619,11 +620,22 @@ def decode_list_object(parsed, context, **kwargs):
                 for member in parsed[top_key]:
                     member[child_key] = unquote_str(member[child_key])
 
+
+def convert_string_body_to_file_like_object(params, **kwargs):
+    if 'Body' in params and isinstance(params['Body'], six.string_types):
+
+        body = ensure_unicode(params['Body']).encode('utf-8')
```
jamesls: What if we just add an "ensure_bytes" function in compat?  The problem with `ensure_unicode` is that it requires the contents to be able to be decoded from utf-8, but s3 objects can have arbitrary binary data.  So instead, we'll have `ensure_bytes` be a noop if it's a bytes type, otherwise it will encode to utf-8 or whatever encoding you've provided.  That should handle the case of arbitrary binary data.

Side note:  if we don't have a test for this, it'd be worth adding.

jamesls:
Approach looks good, had some small feedback.  We should also add a unit test for this as well, something in `test_handlers.py` that invokes the handler function directly.

tests/integration/test_s3.py
```diff
@@ -252,6 +252,12 @@ def test_can_get_and_put_object(self):
             Bucket=self.bucket_name, Key='foobarbaz')
         self.assertEqual(data['Body'].read().decode('utf-8'), 'body contents')
 
+    def test_can_put_large_string_body_on_new_bucket(self):
+        body = '*' * (5 * (1024 ** 2))
+        response = self.client.put_object(
+            Bucket=self.bucket_name, Key='foo', Body=body)
+        self.assertEqual(response['ResponseMetadata']['HTTPStatusCode'], 200)
```
jamesls: Can you also use `assert_can_put_object` here as well?  I noticed that it's used in one test but not the other.

jamesls:
:shipit: Looks good.

tests/integration/test_s3.py
JordonPhillips: The tests are in different classes, this one doesn't have that method.

jamesls: They seem identical except for the body content correct?

If that's the case, can we just pull up the test helper into a base class?

JordonPhillips: :+1:

jamesls:
:shipit: Looks good to me.

PR #802, opened 2016-02-19
Change one StubResponseError to StubAssertionError

rayluo:
The newly introduced StubAssertionError derives from AssertionError,
so that it will be (properly) classified as test failure rather than
test error when your test cases fail on that.


kyleknap:
Looks good. :ship:

JordonPhillips:
:boat:

jamesls:
:shipit:

PR #804, opened 2016-02-23
Don't use MD5 with s3v4

JordonPhillips:
We have a function which avoids md5 when in sigv4, but forgot to include s3v4 even though we attach the function to several s3 commands.

cc @jamesls @kyleknap @rayluo

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -150,7 +150,7 @@ def _calculate_md5_from_file(fileobj):
 def conditionally_calculate_md5(params, **kwargs):
     """Only add a Content-MD5 when not using sigv4"""
     signer = kwargs['request_signer']
-    if signer.signature_version != 'v4':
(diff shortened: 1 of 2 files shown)


botocore/handlers.py
```diff
@@ -150,7 +150,7 @@ def _calculate_md5_from_file(fileobj):
 def conditionally_calculate_md5(params, **kwargs):
     """Only add a Content-MD5 when not using sigv4"""
     signer = kwargs['request_signer']
-    if signer.signature_version != 'v4':
+    if signer.signature_version != 'v4' and signer.signature_version != 's3v4':
```
jamesls: What about `if signer.signature_version not in ['v4, 's3v4']`?

jamesls:
:shipit:

rayluo:
:ship:

PR #806, opened 2016-02-24
Nested docstring lists

JordonPhillips:
Fixes boto/boto3#310

There were three issues at play:

1. We were completely disabling newlines when going through lists
2. Nested lists need to be indented, but the top level list should not be.
3. The doc strings contain, in some cases, several spaces between the start of an `li` and the first instance of text. This causes all sorts of problems in rst.

To make things easier, I separated this PR into three commits, one for each of the items above. Item 3 is the trickiest customer, and required a significant change.

Basically, we need to have full context of the lists before we write them so that we can remove vestigial spaces. What I have done here is change the DocStringParser so that it writes to a tree, and then flushes it all out at close. This allows us to do inspections before anything is written.

We could, alternatively, pull in beautifulsoup to generate this tree for us. The advantage would be that it is far more robust, but at the cost of a large dependency that may take more time than our specialized need requires.

Another potential option is to modify `ReSTStyle` such that it generates a string that is then output to rst, allowing us to apply transformations after generation but before writing. Since this affects more than just doc strings, however, it could result in massive strings being buffered into memory. While we can't be sure of the size of doc strings, they aren't usually more than a few paragraphs.

cc @jamesls @rayluo @kyleknap

### DIFF
--- a/botocore/docs/bcdoc/docstringparser.py
+++ b/botocore/docs/bcdoc/docstringparser.py
@@ -11,10 +11,9 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 from botocore.compat import six
-from six.moves import html_parser
 
 
-class DocStringParser(html_parser.HTMLParser):
+class DocStringParser(six.moves.html_parser.HTMLParser):
     """
     A simple HTML parser.  Focused on converting the subset of HTML
     that appears in the documentation strings of the JSON models into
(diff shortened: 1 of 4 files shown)


botocore/docs/bcdoc/docstringparser.py
```diff
@@ -22,30 +21,175 @@ class DocStringParser(html_parser.HTMLParser):
...
     def handle_data(self, data):
-        if data.isspace():
-            data = ' '
+        self.tree.add_data(data)
+
+    def close(self):
+        # HTMLParser is an old style class, so the super() method will not work.
+        six.moves.html_parser.HTMLParser.close(self)
+        self.tree.write()
+        self.tree = None
```
kyleknap: What happens if I tried to feed data into the parse after closing it? Would setting it to ``None`` cause attribute errors? I wonder if you should add a reset() method or just create a new HTMLTree after closing it.

botocore/docs/bcdoc/docstringparser.py
```diff
@@ -22,30 +21,175 @@ class DocStringParser(html_parser.HTMLParser):
...
+
+class LineItemNode(TagNode):
+    def __init__(self, attrs=None, parent=None):
+        super(LineItemNode, self).__init__('li', attrs, parent)
+
+    def write(self, doc):
+        self._lstrip()
+        super(LineItemNode, self).write(doc)
+
+    def _lstrip(self):
```
kyleknap: I would be fine with removing this method. To simplify the code a bit. It was a little confusing with the two different methods with the one proxying to the other. It would make more sense to have it if _lstrip was not recursive.

botocore/docs/bcdoc/docstringparser.py
```diff
@@ -22,30 +21,175 @@ class DocStringParser(html_parser.HTMLParser):
...
 
     def handle_data(self, data):
-        if data.isspace():
-            data = ' '
+        self.tree.add_data(data)
+
+    def close(self):
+        # HTMLParser is an old style class, so the super() method will not work.
+        six.moves.html_parser.HTMLParser.close(self)
+        self.tree.write()
```
kyleknap: It's a little unfortunate that we have to buffer the entire tree and write it out on the ``close()`` as opposed to before we actually just wrote out with the style as we were feeding the html. I understand the need for an abstraction on top of the style but would it be possible to write out the data to the doc as you add the children? I think it may be possible, but I do not think it is a dealbreaker if we are unable to it because we only invoke the html parser feed on docstings for single shapes and members.

kyleknap:
It looks good. I like this approach more than modifying the Style classes as the changes are only isolated to how we parse the html docstrings that we receive, which ensures we do not mess up any of the hand generated documentation. I did have a few comments and questions.

botocore/docs/bcdoc/docstringparser.py
JordonPhillips: I can imagine a few ways to write as we go, but I think that would make the code much more complicated. We have no forward context if we do that, and no way to go back on previous writes. We would need to either give every node `li` specific state, or do a reverse tree traversal to find out if it is a child of an `li` and then back down to see if it is the first instance of text. I agree it is unfortunate that we have to buffer the whole thing, but I think for the very specific case of doc strings, it is fine. We know that the doc strings will never be more than a few paragraphs, whereas if we buffered ALL the docs (even for a single service) it could be a massive memory footprint.

JordonPhillips:
Changes:
* Moved `write` into `feed`
* Reset the tree instead of dropping it on close
* Removed unnecessary `_lstrip` function.

kyleknap:
Thanks! :ship:

PR #807, opened 2016-02-24
Verify MD5 is available

JordonPhillips:
On some systems, such as a FIPS enabled system, MD5 hashing is unavailable. This commit will detect when that is the case, avoiding it where possible and raising a better error otherwise.

cc @jamesls @rayluo @kyleknap

### DIFF
--- a/botocore/compat.py
+++ b/botocore/compat.py
@@ -237,3 +237,17 @@ def total_seconds(delta):
     day_in_seconds = delta.days * 24 * 3600.0
     micro_in_seconds = delta.microseconds / 10.0**6
     return day_in_seconds + delta.seconds + micro_in_seconds
+
+
+def md5_available():
+    """
+    Checks to see if md5 is available on this system. A given system might not
+    have access to it for various reasons, such as FIPS mode being enabled.
+    :return: True if md5 is available. False if not.
+    """
+    try:
+        import hashlib
+        hashlib.md5()
+        return True
+    except:
+        return False
--- a/botocore/compat.py
+++ b/botocore/compat.py
@@ -253,7 +253,7 @@ except ValueError:
     MD5_AVAILABLE = False
 
 
-def get_md5(raise_error_if_unavailable=True, *args, **kwargs):
+def get_md5(*args, **kwargs):
     """
     Attempts to get an md5 hashing object.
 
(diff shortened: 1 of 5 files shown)


botocore/compat.py
```diff
@@ -237,3 +237,17 @@ def total_seconds(delta):
...
     return day_in_seconds + delta.seconds + micro_in_seconds
+
+
+def md5_available():
+    """
+    Checks to see if md5 is available on this system. A given system might not
+    have access to it for various reasons, such as FIPS mode being enabled.
+    :return: True if md5 is available. False if not.
+    """
+    try:
```
jamesls: I don't think we need to check this each time.  It's either available on the system or not.  I'd prefer to just move this to a constant in this module.

botocore/compat.py
```diff
@@ -237,3 +237,17 @@ def total_seconds(delta):
...
+        import hashlib
+        hashlib.md5()
+        return True
+    except:
```
jamesls: I'd prefer to check the specific Exception here (`ValueError` I believe?).

botocore/handlers.py
```diff
@@ -188,6 +191,9 @@ def copy_source_sse_md5(params, **kwargs):
 def _sse_md5(params, sse_member_prefix='SSECustomer'):
     if not _needs_s3_sse_customization(params, sse_member_prefix):
         return
+    if not md5_available():
```
jamesls: Rather than require each call site to check if md5 is available before creating checksums, can we just have a compat function that either calculate the md5 if possible or raises an exception if not available?

I'd rather have this logic once rather than throughout the code base.  It'll be easier to maintain.

botocore/compat.py
```diff
@@ -237,3 +242,33 @@ def total_seconds(delta):
...
+# Checks to see if md5 is available on this system. A given system might not
+# have access to it for various reasons, such as FIPS mode being enabled.
+try:
+    hashlib.md5()
+    MD5_AVAILABLE = True
+except ValueError:
+    MD5_AVAILABLE = False
+
+
+def get_md5(raise_error_if_unavailable=True, *args, **kwargs):
```
jamesls: After discussing, we can remove the `raise_error_if_unavailable` argument implemented for the CLI and update the CLI code to not compute MD5's if it's not possible.

jamesls:
Looks good, just had that small comment about removing `raise_error_if_unavailable` for now.

jamesls:
:shipit:

Issue #796, opened 2016-02-17
What would be the implications of disabling validation in serialization?


carl0FF:
In an application that uses boto to massively insert to DynamoDB we've been able to observe how a significant amount of CPU time is dedicated to serialization and a part of it is dedicated parameter validation. 

Disabling validation (through forcing ``include_validation=False`` in ``create_serializer()`` at serialize.py) leads to a ~23% performance increase. 

Now, the question is: **how (un)safe would be to disable validation in a production system?**

python 2.7 
Amazon Linux AMI 2015.09
botocore 1.3.15

JordonPhillips:
That depends on how confident you are that your inputs are correct, and what your upgrade process looks like. If the data you send is invalid, the service will reject it, producing a `ClientError` on your side *after* you've already sent all the data. Many services have per-request pricing, so disabling local validation could potentially cost more than the saved cpu time. Dynamo gives free data in, though I don't know how invalid requests count against write capacity, if at all. I would guess that it won't count against you. Additionally, if there is a minor service update that changes parameter validation for Dynamo and you upgrade botocore without testing, you may start to see requests failing since you're using the updated model. This is not likely to happen, however.

In summary, here are the implications:

- You won't know about invalid data until after you've spent bandwidth sending it
- There will be a different class of errors when invalid data is found
- Depending on pricing model specifics, you could still be charged for invalid requests
- Updating botocore will require that you ensure your data still passes validation

carl0FF:
Thank you very much for your reply, Jordon. That's pretty much what we thought, but it always is reassuring to hear it from someone with deeper insights, thanks! :)

PR #809, opened 2016-02-25
Ability to deactivate parameter validation in serializer through config

carl0FF:
The natural follow-up to #797 - exposing parameter validation as an option at serialization time when configuring the client!! :)


botocore/client.py
```diff
@@ -617,6 +618,10 @@ class Config(object):
         thrown when attempting to read from a connection. The default is
         60 seconds.
 
+    :type parameter_validation: str
+    :param parameter_validation: Whether parameter validation should occur
+        when serializing requests. The default is True.
+
```
carl0FF: My only concern for this PR is whether the caveats so kindly laid out by @JordonPhillips in #796 should be somehow mentioned here or not.

jamesls:
Feature seems reasonable to me.  Any concerns, @kyleknap @JordonPhillips?

kyleknap:
Yeah I am fine with it.

JordonPhillips:
I'm good with it, though I do think the caveats should be mentioned.

jamesls:
Ok, I'll rebase the commits and merge.

jamesls:
Merged.  Thanks for the pull request.

PR #812, opened 2016-02-29
Updated endpoints pr

kyleknap:
Keeps this PR up to date: https://github.com/boto/botocore/pull/769

I also removed the behavior where you can specify the endpoint prefix of a service instead of the service name and still be able to get the regions.

cc @jamesls @JordonPhillips


jamesls:
:shipit:

PR #751, opened 2016-01-08
Generate RoleSessionName before creating assume_role refresher

jamesboehmer:
When using a session profile that assumes a role, and when the profile config has no "role_session_name" attribute, the RoleSessionName argument to assume_role is generated the first time, but then never provided to the refresher function.  As a result, the refresher fails to assume the role and raises an exception.  It's not possible to recover from.

I moved the RoleSessionName creation into _assume_role_base_kwargs so that's it's available before the refresher is created, as well as to anything else that may need to generate assume_role args from profile configs.

PR #814, opened 2016-02-29
Add testcase for #751 (assume role provided bug)

jamesls:
This pulls in #751 and adds a testcase that demonstrates the issue.

Essentially, the `RoleSessionName` argument was only being provided in the first `assume_role` call, and not in subsequent `assume_role` refresh calls.

cc @kyleknap @JordonPhillips 

Closes #751.


kyleknap:
:ship:

PR #818, opened 2016-03-01
Add autoscaling waiters

kyleknap:
Fixes https://github.com/boto/boto3/issues/318
cc @jamesls @JordonPhillips

### DIFF
--- a/botocore/data/autoscaling/2011-01-01/waiters-2.json
+++ b/botocore/data/autoscaling/2011-01-01/waiters-2.json
@@ -1,0 +1,62 @@
+{
+    "version": 2,
+    "waiters": {
+        "GroupExists": {
+            "acceptors": [
+                {
+                    "argument": "length(AutoScalingGroups) > `0`",
+                    "expected": true,
+                    "matcher": "path",
+                    "state": "success"
+                },
+                {
+                    "argument": "length(AutoScalingGroups) > `0`",
+                    "expected": false,
+                    "matcher": "path",
+                    "state": "retry"
+                }
+            ],
+            "delay": 5,
+            "maxAttempts": 10,
+            "operation": "DescribeAutoScalingGroups"
+        },
+        "GroupInService": {
+            "acceptors": [
+                {
+                    "argument": "contains(AutoScalingGroups[].[length(Instances[?LifecycleState=='InService']) >= MinSize][], `false`)",
+                    "expected": false,
+                    "matcher": "path",
+                    "state": "success"
+                },
+                {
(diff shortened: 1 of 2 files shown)


botocore/data/autoscaling/2011-01-01/waiters-2.json
```diff
@@ -0,0 +1,62 @@
+{
+    "version": 2,
+    "waiters": {
+        "GroupExists": {
+            "acceptors": [
+                {
+                    "argument": "length(AutoScalingGroups) > `0`",
```
jamesls: We should probably discuss this.  This is misleading if you specify multiple groups.  Autoscaling will just omit that group from its returned list.

kyleknap: Bringing in relevant parties: @awood45 @jeskew

@jamesls do you mean that it will omit it if it does not exist thus possibly returning a partial result? If that is the case, would it be impossible to wait for multiple groups because we cannot inject input parameters to the acceptors using our current waiters spec since we would need to know how many groups the user specified?

jamesls: Yes.  This is a "wait until at least one group exists out of the N you might have specified."

kyleknap: That is really unfortunate.

kyleknap:
Just want to capture some of the offline discussion here:

So when we were pulling in the waiters, we realized that the implementation does not handle multiple autoscaling groups. It might be a while till we can pull this into botocore for two reasons:

1) We could try to rename the waiter by prefixing the names with "Single" to indicate that you cannot wait for multiple waiters. However, the ruby sdk already shipped this waiter and a resource model that uses the waiter and we like to make sure that we match all of the different models that we ship across SDKs for consistency sake. So if we branch off, it would be difficult to pick up updates from other SDKs to the waiter and the resource model, which is not ideal.
2) We need to update our waiter specification to take into account inputs. The only way to support waiting for multiple groups is to make sure that the length returned from DescribeAutoscalingGroups is equal to the number of group ids provided to the waiter.

If you really need this waiter, you can pull down the waiter and add it to ``~/.aws/models/autoscaling/2011-01-01/waiters-2.json`` to use it in the CLI and boto3, but I do not think at this time we can merge it into botocore until we figure out the story of handling multiple autoscaling groups.

jamesls:
@kyleknap given we closed out the boto3 PR for waiter resources, should we also close this one for the time being?

kyleknap:
Closing due to need for discussion

PR #820, opened 2016-03-02
Move Config into its own file

JordonPhillips:
To help with import speed, and so we can alias it in boto3

cc @kyleknap @jamesls


botocore/config.py
```diff
@@ -11,231 +10,147 @@
...
-
-    If there are no profiles in the provided parsed INI contents, then
-    an empty dict will be the value associated with the ``profiles`` key.
-
-    .. note::
-
-        This will not mutate the passed in parsed_ini_config.  Instead it will
-        make a deepcopy and return that value.
-
+from collections import OrderedDict
```
jamesls: I believe this should be `from botocore.compat` for py26 compat.

jamesls:
Small comment about py26 compat (should get a test failure), but otherwise looks good.  :shipit:

PR #821, opened 2016-03-02
Add retry for kinesis DescribeStream

kyleknap:
It is here in the docs: http://docs.aws.amazon.com/kinesis/latest/APIReference/API_DescribeStream.html

Also here is an error that may be returned:
```
DEBUG:botocore.parsers:Response body:
{"__type":"LimitExceededException","message":"Rate exceeded for stream request_counters_temp under account XXXXXXXX."}
DEBUG:botocore.hooks:Event needs-retry.kinesis.DescribeStream: calling handler <botocore.retryhandler.RetryHandler object at 0x7f824d7bc4d0> 
```

cc @jamesls @JordonPhillips

### DIFF
--- a/botocore/data/_retry.json
+++ b/botocore/data/_retry.json
@@ -123,6 +123,20 @@
         }
       }
     },
+    "kinesis": {
+      "DescribeStream": {
+        "policies": {
+          "request_limit_exceeded": {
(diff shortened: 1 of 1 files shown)


jamesls:
:shipit:

PR #826, opened 2016-03-02
Charade -> Chardet in NOTICE

JordonPhillips:
When we updated requests, we forgot to update our NOTICE

cc @jamesls @kyleknap


kyleknap:
:ship:

PR #829, opened 2016-03-03
Don't use hardcoded service-2 for type name

jamesls:
This makes it general enough to use for boto3 resources.

In order to make this backwards compatible, the UnknownServiceError
subclasses from DataNotFoundError.

cc @kyleknap @JordonPhillips


JordonPhillips:
:boat:

kyleknap:
:ship:

PR #830, opened 2016-03-04
Use 'protocols' key to influence scheme

### LINKED ISSUE #819: Dynamodb local uses HTTPS instead of HTTP
In our endpoints.json we have:

```
            "local": {
              "hostname": "localhost:8000",
              "protocols": [
                "http"
              ],
              "credentialScope": {
                "region": "us-east-1"
              }
            }

```

However, creating a client with this region will use HTTPS:

```
>>> import botocore.session
>>> botocore.session.get_session().create_client('dynamodb', 'local').list_tables()
...
botocore.vendored.requests.exceptions.SSLError: [SSL: UNKNOWN_PROTOCOL] unknown protocol (_ssl.c:581)
```

My best guess is that we're not honoring the `protocol` key from endpoints.json when creating clients.

jamesls:
This fixes a bug where if you try to use
DynamoDB Local via `--region local`,
we'll try to connect over https:// despite
the fact that the `protocols` key for that
region indicates that only http is supported.

Also verified I could connect to dynamodb local
after these changes.

Fixes #819.

cc @kyleknap @JordonPhillips

### DIFF
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -291,7 +291,8 @@ class ClientEndpointBridge(object):
         if endpoint_url is None:
             # Use the sslCommonName over the hostname for Python 2.6 compat.
             hostname = resolved.get('sslCommonName', resolved.get('hostname'))
-            endpoint_url = self._make_url(hostname, is_secure)
+            endpoint_url = self._make_url(hostname, is_secure,
+                                          resolved.get('protocols', []))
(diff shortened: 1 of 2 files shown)


kyleknap:
:ship:

JordonPhillips:
:boat:

PR #831, opened 2016-03-04
Pass service_signing_name to ClientEndpointBridge

jamesls:
This was being passed as the 4th positional argument
to ClientEndpointBridge which is the ``default_endpoint``,
not the ``service_signing_name``.  Passing this value
as a kwarg to be explicit.

I also fixed a case where the signingName from the service
was not being used in the case where we assume an endpoint
if we can't resolve anything from the endpoint resolver.

Fixes boto/boto3#538

cc @kyleknap @JordonPhillips


kyleknap:
:ship: I can confirm it works as well

JordonPhillips:
:boat:

PR #832, opened 2016-03-04
Hide PasswordLastUsed from CreateUser response

JordonPhillips:
CreateUser returns a User shape, which contains PasswordLastUsed. This attribute will never show up in a CreateUser response, however, so we hide it in the documentation.

Fixes boto/boto3#487

cc @kyleknap @jamesls

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -25,9 +25,9 @@ import warnings
 
 from botocore.compat import urlsplit, urlunsplit, unquote, \
     json, quote, six, unquote_str, ensure_bytes, get_md5, MD5_AVAILABLE
-from botocore.docs.utils import AutoPopulatedParam
-from botocore.docs.utils import HideParamFromOperations
-from botocore.docs.utils import AppendParamDocumentation
+from botocore.docs.utils import (
+    AutoPopulatedParam, HideParamFromOperations,
+    HideUnusedShapeMember, AppendParamDocumentation)
 from botocore.signers import add_generate_presigned_url
 from botocore.signers import add_generate_presigned_post
 from botocore.exceptions import ParamValidationError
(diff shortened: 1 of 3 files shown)


jamesls:
Tests?

botocore/handlers.py
```diff
@@ -749,5 +749,7 @@ def convert_body_to_file_like_object(params, **kwargs):
           'PutBucketLifecycle', 'PutBucketLogging', 'PutBucketNotification',
           'PutBucketPolicy', 'PutBucketReplication', 'PutBucketRequestPayment',
           'PutBucketTagging', 'PutBucketVersioning', 'PutBucketWebsite',
-          'PutObjectAcl']).hide_param)
+          'PutObjectAcl']).hide_param),
+    ('docs.*.iam.CreateUser.complete-section',
```
kyleknap: I think this needs to get undocumented in a bit more locations like ``get_group``, ``create_virtual_mfa_device`` and ``list_virtual_mfa_devices`` and possibly some other locations. I wonder if the list should be exclusive as opposed to inclusive such that all instances where the User shape and the operation is not GetUser or ListUser then apply the handler? It may be easier then auditing everything, but may lead to unintended consequences such as if User is used say in the input to an operation or something.

tests/unit/docs/test_utils.py
```diff
@@ -205,6 +206,86 @@ def test_hides_param_from_example(self):
...
+            shape_signature, '}'
+        ])
+        self.assert_not_contains_line(member_signature)
+
+    def test_hides_member_from_example_and_removes_trailing_comma(self):
+        shape_section = self.doc_structure\
+            .add_new_section('structure-value')\
+            .add_new_section(self.shape)
+        shape_signature = ':%s: {' % self.shape
+        shape_section.write(shape_signature)
```
kyleknap: This is pretty verbose. Pretty hard to imagine what the contents look like. Could we reuse one of the documenters or combine this into one helper method with comments.

PR #833, opened 2016-03-04
Document streaming payload inputs

kyleknap:
Fixes https://github.com/boto/boto3/issues/477

Here are samples of how it looks:
screen shot 2016-03-04 at 1 45 42 pm

cc @jamesls @JordonPhillips

### DIFF
--- a/botocore/docs/example.py
+++ b/botocore/docs/example.py
@@ -47,7 +47,10 @@ class BaseExampleDocumenter(ShapeDocumenter):
 
     def document_shape_default(self, section, shape, history, include=None,
                                exclude=None, **kwargs):
-        py_type = py_default(shape.type_name)
+        py_type = self._get_special_py_default(shape)
+        if py_type is None:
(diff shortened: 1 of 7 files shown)


jamesls:
Do we define SeekableStream anywhere?

Is it worth just saying it's a `file` type?

kyleknap:
I just made up a class name. We can call it ``file`` if we want. I just want to make the point that we accept file like objects that must be seekable. Thoughts?

JordonPhillips:
It's unfortunate that there's no IO base class that is defined by seek support. Maybe call it `file` and inject something in the doc string about it requiring seek?

jamesls:
Yeah something like that would work.  My main concern is that someone sees `SeekableStream`, thinks that's some base class/interface they need to implement, and can't find any information about that.

kyleknap:
Alright how about this one?

jamesls:
:shipit:  Looks good.

Issue #772, opened 2016-01-27
Add ability to alias arguments


kyleknap:
We need the ability to rename and alias arguments, like we do in the CLI. Here are some examples of where this would be nice or is actually needed:
https://github.com/boto/boto3/issues/463
https://github.com/boto/boto3/issues/456

Furthermore, it would be cool if these aliases can be become aliases for parameters in the CLI.

jamesls:
Another request for this, this time from a CLI customer.

How would we actually do this?  Just have the shape object from the model be marked with some new deprecated flag?  That way the doc gen has something to determine if it should be documented or not?

kyleknap:
That is what I am thinking. I am not sure if we could do something clever and manipulate the runtime model such that we add the new member for ``filters`` that is a copy of the ``filter`` and add a flag that means the ``filter`` member is deprecated so docs handle that appropriately as well.

jamesls:
What would `.members` return from the struct shape?  Should it include both `filter` and `filters` in the return dict?

Thinking through that, this would require all downstream consumers to change their code to filter out deprecated members, i.e the AWS CLI, AWS Shell, etc.

kyleknap:
I am just throwing out ideas. Have not thought about it too much. I think ``.members`` would still be a structure shape. The undocumented part would just be somehow added as a member trait. And the members structure would include both members.

Ideally, I would want to just rename the ``filter`` member in the model at runtime and add a handler to make ``Filter`` an alias for ``Filters`` but CLI cannot do that because it generates its arguments from the model and I do not think we can easily alias this in the CLI other than taking advantage of the argparse "feature" of auto-completing if you type an argument that does not exist and is a subset one that does exist, it will fill it out for you. But that is a bit hacky...

jamesls:
Another idea (also thinking out loud):

Have `.members` return a dict that includes *only* non-deprecated members.  This represents the "ideal" case that customers, tools should use going forward.  For example, the AWS Shell wouldn't autocomplete both `--filter` and `--filters` and would work without any changes.

Add a new property `.all_members` or some other better name that included all members, including aliased, deprecated members.

Update our internal code (param validator, request serializer) to use the new `.all_members`.

That way most downstream tooling wouldn't have to change, though the CLI would probably need to use `.all_members` so it could tell argparse about the deprecated feature.

kyleknap:
[...]

PR #839, opened 2016-03-08
Add ability to alias parameters

kyleknap:
Fixes parameters that shadow reserved keywords in python and inconsistent
parameter names in ec2.

Fixes https://github.com/boto/botocore/issues/772

As to implementation, I followed a similar format (just a tad different) as the CLI so if the alias needed to be in both the CLI and boto3, we could import the dictionary in botocore and modify the ``HIDDEN_ALIAS`` dictionary in the CLI or just copy and paste entities in that dictionary from botocore to the CLI. I went with this path because it required less work and less of a wider impact in the code base then the ``all_member`` route we discussed . Also it would not be difficult for the CLI to pick up these aliases as it already has this functionality.

cc @jamesls @JordonPhillips

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -31,6 +31,7 @@ from botocore.docs.utils import AppendParamDocumentation
 from botocore.signers import add_generate_presigned_url
 from botocore.signers import add_generate_presigned_post
 from botocore.exceptions import ParamValidationError
+from botocore.exceptions import AliasConflictParameterError
 from botocore.exceptions import UnsupportedTLSVersionWarning
 from botocore.utils import percent_encode, SAFE_CHARS
 
@@ -671,6 +672,12 @@ class ParameterAlias(object):
             # input shape.
             if self._original_name in model.input_shape.members:
                 if self._alias_name in params:
+                    if self._original_name in params:
+                        raise AliasConflictParameterError(
+                            original=self._original_name,
+                            alias=self._alias_name,
+                            operation=model.name
+                        )
                     # Remove the alias parameter value and use the old name
                     # instead.
                     params[self._original_name] = params.pop(self._alias_name)
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -628,6 +628,79 @@ def convert_body_to_file_like_object(params, **kwargs):
             params['Body'] = six.BytesIO(params['Body'])
 
 
+def _add_parameter_aliases(handler_list):
+    # Mapping of original parameter to parameter alias.
+    # The key is <service>.<operation>.parameter
+    # The first part of the key is used for event registration.
+    # The last part is the original parameter name and the value is the
+    # alias to expose in documentation.
+    aliases = {
+        'ec2.*.Filter': 'Filters',
+        'logs.CreateExportTask.from': 'fromTime',
+        'cloudsearchdomain.Search.return': 'returnFields'
+    }
+
+    for original, new_name in aliases.items():
+        event_portion, original_name = original.rsplit('.', 1)
+        parameter_alias = ParameterAlias(original_name, new_name)
+
+        # Add the handlers to the list of handlers.
+        # One handler is to handle when users provide the alias.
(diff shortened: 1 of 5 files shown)


tests/unit/test_handlers.py
```diff
@@ -804,3 +807,77 @@ def test_add_md5_with_bytes_object(self):
         self.assertEqual(
             request_dict['headers']['Content-MD5'],
             'OFj2IjCsPJFfMAxmQxLGPw==')
+
+
+class TestParameterAlias(unittest.TestCase):
```
jamesls: We should have a test for when a both the alias and the original are entered.

tests/unit/test_handlers.py
```diff
@@ -291,10 +295,9 @@ def test_run_instances_userdata_blob(self):
         # user data.
         user_data = b'\xc7\xa9This is a test'
         b64_user_data = base64.b64encode(user_data).decode('utf-8')
-        event = 'before-parameter-build.ec2.RunInstances'
         params = dict(ImageId='img-12345678',
                       MinCount=1, MaxCount=5, UserData=user_data)
-        self.session.emit(event, params=params)
+        handlers.base64_encode_user_data(params=params)
```
jamesls: Why did this need to be changed?

botocore/handlers.py
```diff
@@ -628,6 +628,79 @@ def convert_body_to_file_like_object(params, **kwargs):
...
+
+    for original, new_name in aliases.items():
+        event_portion, original_name = original.rsplit('.', 1)
+        parameter_alias = ParameterAlias(original_name, new_name)
+
+        # Add the handlers to the list of handlers.
+        # One handler is to handle when users provide the alias.
+        # The other handler is to update the documentation to show only
+        # the alias.
+        handler_for_parameter_build = (
```
jamesls: Can we rename this?  This isn't actually the handler, this is a tuple of event name to handler.

botocore/handlers.py
```diff
@@ -628,6 +628,79 @@ def convert_body_to_file_like_object(params, **kwargs):
...
+            'before-parameter-build.' + event_portion,
+            parameter_alias.alias_parameter_in_call)
```
jamesls: I think these probably make sense as `REGISTER_FIRST` don't they?  Seems like we want to map these as soon as we possibly can.

jamesls:
Looks good, just had a few questions/comments.

tests/unit/test_handlers.py
kyleknap: By emitting that event, it will call every handler and that fake emission of the event does not include an operation model which is required by the alias handler causing a runtime error. By removing the emission of an event, I feel like that it has become more of a unit test as well as it is decoupled from all of the other handlers that may be registered to this event. Thoughts?

tests/unit/test_handlers.py
kyleknap: What should be the desired behavior? We remove the alias but use the original's value or remove the alias and use the alias's value. Currently because of how its written the alias will clobber the original. Do you think an error should be thrown instead for this scenario saying the parameters provided is ambiguous?

botocore/handlers.py
kyleknap: True. Will do.

botocore/handlers.py
kyleknap: Yeah we can do that to be safe.

tests/unit/test_handlers.py
JordonPhillips: Agree. Running less unrelated code is better.

tests/unit/test_handlers.py
JordonPhillips: I would say the expected result would be failure on the old without any special ambiguity.

JordonPhillips:
Adding the test @jamesls mentioned would be nice, but otherwise :boat:

kyleknap:
Just added the error case. Should be good to look at.

jamesls:
:shipit:

PR #840, opened 2016-03-11
Add api_versions to session variable

kyleknap:
Also allows use to specify ``api_versions`` in the config file so the CLI can start picking these up as well. It would looks like this in the config file:
```
api_versions =
    cloudsearch = 2011-02-01
    ec2 = 2014-10-01
```
If you want to set the api versions on the session, you would have to use the session's ``get_config_variable()`` and ``set_config_variable()`` methods.

cc @jamesls @JordonPhillips


jamesls:
:shipit:

PR #845, opened 2016-03-19
Document automatic base64 encoding of lambda ZipFile

JordonPhillips:
Lambda already updated their docs for `ZipFile` in `create_function` so we don't need that here.

screen shot 2016-03-18 at 5 25 38 pm

Resolves boto/boto3#201

cc @jamesls @kyleknap


jamesls:
:shipit:

PR #849, opened 2016-03-23
Fall back to deprecated starting token parser

JordonPhillips:
A recent change to paginators broke customers who were manually crafting pagination tokens in services where they did not make them opaque. This adds support for the old starting token format while still returning next tokens in the new format. If we are unable to parse the starting token in the new format, we'll attempt to parse it with the old format.

This addresses the root cause of aws/aws-cli#1835, but the docs should be updated to reflect the new tokens.

cc @kyleknap @jamesls

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -353,7 +353,8 @@ class PageIterator(object):
         This handles parsing of old style starting tokens, and attempts to
         coerce them into the new style.
         """
-        log.debug("Attempting to fall back to old starting token parser.")
+        log.debug("Attempting to fall back to old starting token parser. For "
+                  "token: %s" % self._starting_token)
         if self._starting_token is None:
             return None
 
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -18,9 +18,13 @@ from six import string_types
 import jmespath
 import json
 import base64
+import logging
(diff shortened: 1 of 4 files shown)


botocore/paginate.py
```diff
@@ -341,9 +345,43 @@ def _parse_starting_token(self):
...
-            raise ValueError("Bad starting token: %s" % self._starting_token)
+            next_token, index = self._parse_starting_token_deprecated()
         return next_token, index
 
+    def _parse_starting_token_deprecated(self):
+        """
+        This handles parsing of old style starting tokens, and attempts to
+        coerce them into the new style.
+        """
+        log.debug("Attempting to fall back to old starting token parser.")
```
kyleknap: Might be good to say what the token was that was received that started this attempt.

kyleknap:
Looks good. You pretty much copied code from the old implementation :ship:. Did you test your implementation with some of the services where we saw regressions? It may not hurt to add an integration or functional test for one of those services like rds or route53.

jamesls:
:shipit:  You can also use s3 as well for an integration test.  There'a already some existing tests for pagination so it might be easier to build on those tests.

JordonPhillips:
tests added

kyleknap:
Thanks! :ship:

PR #850, opened 2016-03-23
Handle the case where we receive a generic html error response

jamesls:
Fixes boto/boto3#339.

cc @kyleknap @JordonPhillips

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -237,7 +237,9 @@ class ResponseParser(object):
         LOG.debug("Received a non protocol specific error response from the "
                   "service, unable to populate error code and message.")
         return {
-            'Error': {'Code': '', 'Message': ''},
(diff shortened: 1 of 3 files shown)


JordonPhillips:
What do you think about populating those values with the HTTP error codes? Otherwise CLI users are not going to have any idea what just happened.

jamesls:
@JordonPhillips feedback incorporated.

JordonPhillips:
:boat:

kyleknap:
Looks good. :ship: Do you think that this fix also partially resolves this issue: https://github.com/aws/aws-cli/issues/1679 for when the proxy returns straight html?

jamesls:
@kyleknap It should.  Looking at their output it looks like this should fix that case because we're getting an HTML error page back from the proxy.  Excellent.

Issue #843, opened 2016-03-16
Amazon Inspector API change


joewing:
The Amazon Inspector API has been updated and it appears that botocore (as of 1.4.4) still provides the old API, which no longer works.  Are there plans to update this?  The documentation is here:
http://docs.aws.amazon.com/inspector/latest/APIReference/Welcome.html

JordonPhillips:
Thanks for letting us know, I'll let the service team know the old api version isn't being routed correctly and get the updates pushed out ASAP.

mjschultz:
This is affecting us too, any tips on how to get the boto API calls to match the new version of the Amazon API?

Raybuster:
I found this forum post on how to adjust the CLI to the new API: https://forums.aws.amazon.com/thread.jspa?threadID=227591

I think there may be a way to update boto using the model file linked from that post

rayluo:
@Raybuster The workaround mentioned from your forum link is very likely also effective for botocore. Did you try that?

JordonPhillips:
We have updated the Inspector api to the latest version, sorry for the delay.

PR #854, opened 2016-03-28
Update to 2016-02-16 Amazon Inspector API

mjschultz:
I know botocore generally generates these from your "[internal knowledge of the AWS services.](https://github.com/boto/botocore/blob/develop/CONTRIBUTING.rst)" but barring that, this should at least make the [Amazon Inspector API](https://github.com/boto/botocore/issues/843) usable.

I think the missing part is the documentation parameters, I'm not sure if those are auto-generated or not.

(This is based on the code generated by running `aws configure add-model --service-model https://s3-us-west-2.amazonaws.com/inspector-service-model/inspector-2016-02-16.normal.json --service-name inspector`.)


JordonPhillips:
As of today's release, we have updated to the latest API version (along with docs). Thanks for your interest, and sorry it took so long to get out.

PR #859, opened 2016-03-30
Consistently parse error responses with no body

jamesls:
The generic error parsing code can now additionally be triggered by a 5xx response with an empty body in addition to an html response body.

This also required updating one test which was previously leaving the code/message blank in this scenario.

Fixes boto/boto3#574.

cc @kyleknap @JordonPhillips


kyleknap:
:ship:

freimer:
Thank you!

PR #861, opened 2016-03-31
Validate S3 metadata only contains ascii chars

jamesls:
Fixes boto/boto3#478.

cc @kyleknap @JordonPhillips

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -450,6 +450,38 @@ def document_base64_encoding(param):
     return append.append_documentation
 
 
+def validate_ascii_metadata(params, **kwargs):
+    """Verify S3 Metadata only contains ascii characters.
+
+    From: http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
+
+    "Amazon S3 stores user-defined metadata in lowercase. Each name, value pair
+    must conform to US-ASCII when using REST and UTF-8 when using SOAP or
+    browser-based uploads via POST."
+
+    """
+    metadata = params.get('Metadata')
+    if not metadata or not isinstance(metadata, dict):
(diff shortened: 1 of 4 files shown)


jamesls:
I fixed the failing test by defaulting to unicode literals in exceptions.py.  That has the possibility for unintended side effects, but all the botocore/cli tests pass.

The difference being that `type('foo %s' % u'bar') -> unicode`, but `type('foo {}'.format(u'bar')) -> str`.

JordonPhillips:
:boat:

botocore/handlers.py
```diff
@@ -450,6 +450,38 @@ def document_base64_encoding(param):
...
+
+    From: http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
+
+    "Amazon S3 stores user-defined metadata in lowercase. Each name, value pair
+    must conform to US-ASCII when using REST and UTF-8 when using SOAP or
+    browser-based uploads via POST."
+
+    """
+    metadata = params.get('Metadata')
+    if not metadata or not isinstance(metadata, dict):
```
kyleknap: Yeah that is an interesting problem in that we cannot validate with the validator that the parameter passed by the user is correct prior to the handler. It would be interesting to expose a new event down the line passed the validation or a helper method to run a quick validation in the future if we run into more of these scenarios.

kyleknap:
Looks good. Just had a comment to leave out there for future reference. :ship:

botocore/handlers.py
jamesls: I looked through several `before-parameter-build`event handlers, and there's several that are susceptible to this problem.  I'm thinking it's probably worth having this event in the near future.

PR #866, opened 2016-04-04
Allow non-specified input tokens in old starting token format

### LINKED ISSUE #862: Paginator fix still doesn't handle all cases from before
https://github.com/boto/botocore/pull/849 requires that the starting token matches the length of the input tokens, which means all input tokens need to be specified.  While this was the case for any token that came back from a botocore response, it's not the case for customers that are handcrafting starting tokens that have semantic value.  For example, here's differing behaviors on two CLI versions:

v1.9.0

```
$ aws route53 list-resource-record-sets --hosted-zone-id $ZONE_ID \
    --max-items 1 --starting-token 'my.domain.name.' --debug
{
    "NextToken": "None___None___None___1",
    "ResourceRecordSets": [
        {
            "ResourceRecords": [

                {
                    "Value": "..."
                },
                {
                    "Value": "..."
                }
            ],
            "Type": "NS",
            "Name": "my.domain.name.",
            "TTL": 172800
        }
    ]
}
```

v1.10.17

```
$ aws route53 list-resource-record-sets --hosted-zone-id $ZONE_ID \
    --max-items 1 --starting-token 'my.domain.name.' --debug

Bad starting token: my.domain.name.
```

In this example, the use case is that the user wants to get information about a specific resource record set (in this case for the domain name of `my.domain.name.`).  This maps to the `StartRecordName` param name in the API.

Note: even though this is a CLI specific example, I'm filing it here because the fix for this issue will need to come from botocore.

JordonPhillips:
This fixes a regression that wasn't handled with the recent fallback fix. Closes #862 

cc @jamesls @kyleknap

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -379,8 +379,15 @@ class PageIterator(object):
         This attempts to convert a deprecated starting token into the new
         style.
         """
-        if len(deprecated_token) != len(self._input_token):
+        len_deprecated_token = len(deprecated_token)
+        len_input_token = len(self._input_token)
+        if len_deprecated_token > len_input_token:
             raise ValueError("Bad starting token: %s" % self._starting_token)
+        elif len_deprecated_token < len_input_token:
+            log.debug("Old format starting token does not contain all input "
+                      "tokens. Setting the rest, in order, as None.")
+            for i in range(len_input_token - len_deprecated_token):
+                deprecated_token.append(None)
         return dict(zip(self._input_token, deprecated_token))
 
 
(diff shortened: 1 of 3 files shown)


kyleknap:
Code looks fine :ship: but with like the last PR can you add an integration or functional test for this?

I want to make sure that we never see this issue or issues related to this regression again and I think we could have avoided this one with an additional functional/integration test.

Looks like we already have pagination tests for route53: https://github.com/boto/botocore/blob/develop/tests/integration/test_route53.py. So maybe look there to start?

JordonPhillips:
:+1:

kyleknap:
Thanks! :ship:

PR #870, opened 2016-04-07
Use compat unittest

JordonPhillips:
A few tests were using base unittest rather than our compat version.

cc @jamesls @kyleknap


jamesls:
:shipit:

PR #871, opened 2016-04-07
Add missing % in datetime format string

### LINKED ISSUE #869: RefreshableCredentials serializing without timezone info
Hey boto folks -

I'm currently implementing my own CredentialsProvider that provides RefreshableCredentials. I'm having issues with code that has been working, but recently broke. I have to wonder if either one of my dependencies got bumped (although I can't prove it) or because we just entered daylight savings.

Basically, I'm always hitting this condition: 

> Credentials were refreshed, but the refreshed credentials are still expired.

To create my meta-data, I'm using the same(?) code used in botocore:
```
from datetime import datetime, timedelta
from dateutil.tz import tzlocal
expiration = datetime.now(tzlocal()) + timedelta(minutes=5)
expiration.strftime('%Y-%m-%dT%H:%M:%SZ')
```
> '2016-04-07T10:26:04Z'

This doesn't encode my local time zone information, as %Z is needed.

I can demonstrate this inconsistency using some of the internal boto funcs
```
from botocore.credentials import _local_now, _serialize_if_needed, _parse_if_needed
now = _local_now()
print now
serialized = _serialize_if_needed(now)
print serialized
deserialized = _parse_if_needed(serialized)
print deserialized
print now - deserialized
```
> 2016-04-07 10:24:12.336556-04:00
> 2016-04-07T10:24:12Z
> 2016-04-07 10:24:12+00:00
> 4:00:00.336556

I can fix this issue on my end by encoding the timezone information in my format string `'%Y-%m-%dT%H:%M:%S%Z'`, but I'm convinced botocore should do the same in its default time serialization code

TL;DR
It seems like your `time_fetcher` is using a timezone aware (tzlocal) datetime, but you're losing the timezone information upon serialization, defaulting to UTC. So in my case, it's comparing EST to UTC when determining expiration time.

JordonPhillips:
Fixes #869 

cc @jamesls @kyleknap

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -121,7 +121,7 @@ def _parse_if_needed(value):
 
 def _serialize_if_needed(value):
     if isinstance(value, datetime.datetime):
-        return value.strftime('%Y-%m-%dT%H:%M:%SZ')
+        return value.strftime('%Y-%m-%dT%H:%M:%S%Z')
     return value
 
 
--- a/tests/unit/test_credentials.py
+++ b/tests/unit/test_credentials.py
@@ -15,10 +15,10 @@ from datetime import datetime, timedelta
 import mock
 import os
 
-from dateutil.tz import tzlocal
+from dateutil.tz import tzlocal, tzutc
 
 from botocore import credentials
-from botocore.credentials import EnvProvider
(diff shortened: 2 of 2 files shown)


jamesls:
:shipit:

kyleknap:
:ship:

PR #873, opened 2016-04-08
Don't overwrite the Accept header if it is already set

JordonPhillips:
APIGateway has at least one function, `get_export`, which relies on the use of the `Accept` header. We were over-writing that with `application/json`, so this checks to see if the header is already set before we put that in.

cc @kyleknap @jamesls

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -519,8 +519,9 @@ def add_glacier_version(model, params, **kwargs):
 
 
 def add_accept_header(model, params, **kwargs):
-    request_dict = params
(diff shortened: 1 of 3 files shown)


jamesls:
Can you confirm this doesn't break on response parsing?  If we're sending application/yaml for the accept header, wouldn't we get get yaml back?  I would expect this to break at the JSON response parser unless this was modeled as a streaming payload.

JordonPhillips:
This function is indeed modeled as a streaming payload.

jamesls:
:shipit:  Ok cool.

PR #874, opened 2016-04-08
Remove dead code in response.py

JordonPhillips:
This little bit of code is called by nothing in botocore, boto3, or the CLI. Looks like it was renamed and moved into StreamingBody, but the original location didn't get removed.

cc @jamesls @kyleknap


kyleknap:
:ship:

PR #878, opened 2016-04-13
fix when use custom endpoint_url

jingtaozf:
If I use a custom endpoint_url which is not s3.amazonaws.com, boto3 will fail to work.


jamesls:
Looks like @JordonPhillips has a PR for this as well.  @JordonPhillips can we consolidate PRs?

PR #883, opened 2016-04-14
Add new-change script to botocore

jamesls:
Same as the CLI version, with the default values being switched to `boto/botocore`.

cc @kyleknap @JordonPhillips


kyleknap:
:ship:

PR #886, opened 2016-04-18
Respect custom endpoints in fix_s3_host

### LINKED ISSUE #875: fix_s3_hosts always assumes you're using central aws instead of a local S3.
I've noticed when attempting to generate presigned urls that the logic behind fix_s3_hosts and the virtual styling always assumes you're using AWS S3. [fix_s3_host](https://github.com/boto/botocore/blob/6065566237de957507fc63ae4d3a034c87552adf/botocore/utils.py#L649)

However, we have a local s3 api on top of ceph, as such it throws away our local s3 endpoint and substitutes s3.amazonaws.com, although the presigned urls work when you switch out the virtual path for our local s3, it's a pain.

This is because the default_endpoint_url passed to [switch_to_virtual_host_style](https://github.com/boto/botocore/blob/6065566237de957507fc63ae4d3a034c87552adf/botocore/utils.py#L656) is always set to 's3.amazonaws.com'

The request object is available in the fix_s3_host function so it should be a fairly simple fix.

JordonPhillips:
Currently we inject the hard-coded value `s3.amazonaws.com` as the endpoint
url for S3 requests when switching from path style to virtual host addressing.
This has the effect of clobbering custom endpoint urls. This PR will try to
determine if the endpoint url is an S3 endpoint, and only inject the global S3
url if so.

There are two discussion items I would like to bring up with this PR:

* Should we pass along the user's netloc (current behavior) or short-circuit
  fix_x3_host? The former will be consistent with how we handle S3, and the
  latter may respect the user's intentions better.
* Is there a better, reasonable way to determine if the URL is an S3 url? We
  have the endpoint resolver available, but since we don't have access to the
  session inside fix_s3_host we would have to load the endpoints file
  separately. I'm not a fan of using regex to solve problems like this, but I'm
  also not a fan of duplicated reads.

Closes #875
Closes #878

P.S. In addition, I also did some clean-up in the files I was working in.

cc @jamesls @kyleknap


botocore/utils.py
```diff
@@ -644,24 +646,32 @@ def fix_s3_host(request, signature_version, region_name, **kwargs):
...
     elif not _allowed_region(region_name):
         return
+
+    # We need to respect custom endpoint urls, but unfortunately we have no
+    # way at this point to know how the url got set. We unfortunately don't have
+    # access to the endpoint resolver here, so we can't do a data-driven test.
+    # A regular expression is therefore used to determine if the request url is
+    # an S3 url.
+    endpoint_url = None
+    match = S3_ENDPOINT_RE.match(request.url)
```
kyleknap: I wonder if there are better option here? Seems could be flakey. @jamesls do you have any ideas. I can't think of any ideas.

botocore/utils.py
```diff
@@ -644,24 +646,32 @@ def fix_s3_host(request, signature_version, region_name, **kwargs):
...
-        switch_to_virtual_host_style(
-            request, signature_version, 's3.amazonaws.com')
+        switch_to_virtual_host_style(request, endpoint_url)
     except InvalidDNSNameError as e:
         bucket_name = e.kwargs['bucket_name']
         logger.debug('Not changing URI, bucket is not DNS compatible: %s',
                      bucket_name)
 
 
-def switch_to_virtual_host_style(request, signature_version,
```
kyleknap: This is arguably a breaking change by removing the variable. I think that I am fine with it because I do not see users trying to use this method since we already have configuration options. @jamesls what are your thoughts.

tests/functional/test_s3.py
```diff
@@ -170,14 +183,14 @@ def test_default_endpoint_for_virtual_addressing(self):
     def test_provided_endpoint_url_for_virtual_addressing(self):
         s3 = self.session.create_client(
             's3', config=Config(s3={'addressing_style': 'virtual'}),
-            endpoint_url='https://foo.amazonaws.com')
```
kyleknap: Why does the amazonaws part need to be removed?

kyleknap:
The approach looks fine to me. Had a couple questions. Leaning toward a ship depending on the answers. Also would be good if @jamesls took a look at it.

botocore/utils.py
JordonPhillips: We could spin up an endpoint resolver, but that would require a load from disk...

tests/functional/test_s3.py
JordonPhillips: We aren't likely to be passed an amazonaws address as a custom endpoint url. And in the case of this regex change, it could fail.

botocore/utils.py
jamesls: What's the ideal behavior?  If a user provides an endpoint url we don't mess with the hostname?

kyleknap: That is what I would say is ideal. If an endpoint_url is provided, then we always use that. The only issue is that this handler is way too far down the stack to know if user provided the url or if we came up with the endpoint ourselves. We would need something to flag us that the user provided an endpoint_url or move this handler up the stack.

jamesls: Could we just unregister the handler in the client creation if `endpoint_url` is provided, similar to how we deal with `s3_addressing_style`?

The only wrinkle is we still want virtual hosted addressing for s3-accelerate endpoints given, but it seems easier to handle that case specifically.

jamesls:
Want to make sure the comments aren't lost here.  Can we deregister the handler if a user provides an endpoint url (assuming we take care of the S3 accelerate endpoint logic)?

JordonPhillips:
Switched to a deregister strategy. I like this muuuuuch better.

botocore/client.py
```diff
@@ -202,6 +207,7 @@ def _get_client_args(self, service_model, region_name, is_secure,
         # Add any additional s3 configuration for client
         self._inject_s3_configuration(
             config_kwargs, scoped_config, client_config)
+        self._conditionally_unregister_fix_s3_host(endpoint_url, event_emitter)
```
kyleknap: I was concerned that this may be unregistering the non-copied version of the event emitter, but you are fine. Ideally this would be moved to the BaseClient class where we handle the other s3 registering and unregistering, but we never have access to the customer provided endpoint url so that is not doable without a big refactoring. So I think I am fine with this here.

kyleknap:
I like the approach more. Based on the behavior, it looks like if a custom endpoint url is provided it will use the endpoint url and default to path style. This is fine with me, but it would be good to have a functional test to capture this behavior. It should not be hard to add since we have similar functional tests already in place.

JordonPhillips:
Added functional test

kyleknap:
Thanks. :ship:

Issue #882, opened 2016-04-14
boto2.num_retries equivalent in retryhandler.py or _retry.json 


taraslayshchuk:
Hi!

I have a problem with using logic to retry failed requests.

```bash
14:00:52 Traceback (most recent call last):
14:00:52   File "deploy/build_scripts/gen-security-groups.py", line 638, in <module>
14:00:52     main()
14:00:52   File "deploy/build_scripts/gen-security-groups.py", line 631, in main
14:00:52     gsga_results = Security_Groups_get_IP_Addresses()
14:00:52   File "deploy/build_scripts/gen-security-groups.py", line 420, in Security_Groups_get_IP_Addresses
14:00:52     security_group_addresses = Security_Groups_get_Addresses_for_Stacks(stacks_security_groups)
14:00:52   File "deploy/build_scripts/gen-security-groups.py", line 407, in Security_Groups_get_Addresses_for_Stacks
14:00:52     security_group_addresses.update(Security_Groups_get_Addresses_for_Groups(client, sec_groups))
14:00:52   File "deploy/build_scripts/gen-security-groups.py", line 387, in Security_Groups_get_Addresses_for_Groups
14:00:52     groups = client.desc_security_groups(GroupIds=sec_groups)
14:00:52   File "/usr/lib/python2.7/site-packages/botocore/client.py", line 228, in _api_call
14:00:52     return self._make_api_call(operation_name, kwargs)
14:00:52   File "/usr/lib/python2.7/site-packages/botocore/client.py", line 492, in _make_api_call
14:00:52     raise ClientError(parsed_response, operation_name)
14:00:52 botocore.exceptions.ClientError: An error occurred (RequestLimitExceeded) when calling the DescribeSecurityGroups operation: Request limit exceeded.
````
I had used script for creating bulk of security group and I met RequestLimitExceeded Exception.
I don't see any probably workaround for result this issue, because I don't have any built in function to change retry config. (Yes, I can change json, or in some way do something like that:
````python
client = boto3.client('ec2', region_name='us-west-2', config=boto3_config)
client.meta.events._unique_id_handlers['retry-config-ec2']['handler']._checker.__dict__['_max_attempts'] = 20
````
, or maybe I can create retryhandler with correct retry_config and transfer this object in some constructor, but I did not find this solution). So, is it true that botocore does not provide any supported and built-in methods for change retries count variable or delay base value?

kyleknap:
[...]

PR #891, opened 2016-04-22
Add possibility of modifying retry_config

taraslayshchuk:
Hi, according to issue #882 with num_retries there is some improvements to change retry_config over Api which include: max_attempts, delay.base, delay.growth_factor.

Usage example:

``` python
retry_config = {
    "max_attempts": 10,
    "base": 15,
    "growth_factor": 60
}
boto3_config = botocore.client.Config(retry_config=retry_config)
```

What do you think about it?


jamesls:
Not opposed to the feature in general.

I wonder how much of the retry logic we should expose though.  Once it's exposed we can't change/remove it.  I'm slightly concerned that base/growth_factor are too low level, and might make it hard to support other types of backoff calculations in the future besides exponential backoff.

What do others think?

taraslayshchuk:
In general if we made some changes in `~/.aws/models/_retry.json` it will cause changes in all scripts, which used on same PC with same user. So if you want made some changes for one script and did not want make for another you have to write some wrapper which will put this json-file before script run and delete it after run (and add some logic to delete json-file if script fails). It is not simple solution and anyway we need have some possibility to make changes in retry logic from code and use it per script.

copumpkin:
I'd love this, since I've had some cases where the default number of retries wasn't enough.

rafaelcaricio:
Please just expose the `max_attempts` to make this feature move on.

lapointexavier:
@rafaelcaricio @taraslayshchuk I was wondering, is this still relevant? I was looking for a solution to this and came across this PR

rafaelcaricio:
@lapointexavier  Yes, would be nice to have this options available.

lapointexavier:
@rafaelcaricio What's missing to get this merged, better patch coverage?

thenoid:
Can we get this merged please?

spstarr:
Would like this merged also:

Merely just trying to use a filter will trigger API limits when I have way too many volumes, snapshots etc:

` sched_volumes = list(ec2.volumes.filter(Filters=vol_filter))`
...
botocore.exceptions.ClientError: An error occurred (RequestLimitExceeded) when calling the DescribeVolumes operation (reached max retries: 4): Request limit exceeded.

Anything that can help me override this retry limit would help vs me having to use python @retry decorator fun.

thehesiod:
time to get this or something related to this puppy landed

kyleknap:
@taraslayshchuk We recently merged a PR: https://github.com/boto/botocore/pull/1260 that allows you to set the maximum retry attempts. It provides a similar functionality to yours except it does not expose the base or growth factor. We were not quite comfortable exposing those two parameters quite yet and wanted to start small with just maximum retry attempts before including some of the more granular/implementation related configurations. For this reason, I am closing out this PR, but thanks again for the contribution!

PR #896, opened 2016-04-25
Patch out warnings.warn call in iot-data tests

jamesls:
Warnings are not emitted after the first time, so this
test was dependent on the order in which the tests are run.
It relied on the fact that the test was triggering the warning
for the first time in the test suite.

On a machine that actually has an older version of TLS that would
normally trigger this warning, any previous call to ``.create_client()``
would trigger this warning.  As a result, this test would fail, assuming
it was run after a test that had created an ``iot-data`` client.

cc @kyleknap @JordonPhillips


kyleknap:
:ship:

PR #897, opened 2016-04-26
Retry 502 and 504 status codes

jamesls:
There was a bug with the 502 retries where it was added to
the definitions, but not to the default list of retry policies.

I've verified that 502 and 504 are in fact retried now.

cc @kyleknap @JordonPhillips


JordonPhillips:
⛵

PR #898, opened 2016-04-27
Don't sign content-type

JordonPhillips:
Fixes boto/boto3#610

cc @jamesls @kyleknap


jamesls:
I don't think this actually solves the customer's issue if I'm reading the linked issue properly.  The issue is that *S3* (not us) is including it in its canonicalization and we're not.  Moreso, we don't even have a way in botocore for someone to specify that they'd like us to sign that value.

As an aside, this change would also prevent content type from not being signed even for operations such as `PutObject` where it's a modeled param.  We should *always* sign parameters the customer provides us.  These are params the customer wants to send to a service, and it's our responsibility to ensure that those values make it to S3 without anything messing with those values on the way.  That's what the signature gives us.

PR #904, opened 2016-04-29
Update SIX

JordonPhillips:
My primary motivation for updating this package is that I want the updated print function that includes the `flush` argument in 2.6+.

cc @kyleknap @jamesls


jamesls:
Is there a reason we can't use use `.flush()` on stderr/stdout?

JordonPhillips:
We can, but this would allow us to take advantage of the built in automatic flush in python 3. It also saves us a line of code, not that that's a huge deal.

JordonPhillips:
Closing as the code ended up being more understandable with the explicit calls.

PR #905, opened 2016-04-30
Add session var to disable parameter validation

jamesls:
Corresponds to the same value in the client Config object.
This can now be specified in ``~/.aws/config``.

Alternative to https://github.com/aws/aws-cli/pull/1943

cc @kyleknap @JordonPhillips


jamesls:
Just realized, the docs for the config option will be in boto3 (http://boto3.readthedocs.io/en/latest/guide/configuration.html#configuration).

kyleknap:
Approach looks good to me. :ship:

JordonPhillips:
⛵

Issue #906, opened 2016-05-03
Ec2 InstanceExists waiter throws an exception immediately if the instance does not exist


nicholasbishop:
Code that reproduces the issue:

```python
import boto3
import botocore

print('boto3 version:', boto3.__version__)
print('botocore version:', botocore.__version__)

# This is an instance that does not exist
invalid_instance_id = 'i-12345678'

ec2_client = boto3.client('ec2')
waiter = ec2_client.get_waiter('instance_exists')
waiter.wait(InstanceIds=[invalid_instance_id])
```

(Note that although this is an artificial example with an obviously-fake instance ID, the issue also reproduces occasionally with real instances if the waiter is called immediately after creating the instance.)

Output:
```
$ python3 repro.py 
boto3 version: 1.3.1
botocore version: 1.4.15
Traceback (most recent call last):
  File "repro.py", line 12, in <module>
    waiter.wait(InstanceIds=[invalid_instance_id])
  File "/usr/lib/python3.4/site-packages/botocore/waiter.py", line 53, in wait
    Waiter.wait(self, **kwargs)
  File "/usr/lib/python3.4/site-packages/botocore/waiter.py", line 292, in wait
    if acceptor.matcher_func(response):
  File "/usr/lib/python3.4/site-packages/botocore/waiter.py", line 194, in acceptor_matches
    return expression.search(response) == expected
  File "/usr/lib/python3.4/site-packages/jmespath/parser.py", line 524, in search
    result = interpreter.visit(self.parsed, value)
  File "/usr/lib/python3.4/site-packages/jmespath/visitor.py", line 66, in visit
    return method(node, *args, **kwargs)
  File "/usr/lib/python3.4/site-packages/jmespath/visitor.py", line 114, in visit_comparator
    self.visit(node['children'][0], value),
  File "/usr/lib/python3.4/site-packages/jmespath/visitor.py", line 66, in visit
    return method(node, *args, **kwargs)
  File "/usr/lib/python3.4/site-packages/jmespath/visitor.py", line 129, in visit_function_expression
    return self._functions.call_function(node['value'], resolved_args)
  File "/usr/lib/python3.4/site-packages/jmespath/functions.py", line 89, in call_function
    self._validate_arguments(resolved_args, signature, function_name)
  File "/usr/lib/python3.4/site-packages/jmespath/functions.py", line 100, in _validate_arguments
    return self._type_check(args, signature, function_name)
[...]

PR #907, opened 2016-05-03
Fix for EC2.Waiter.InstanceExists raising JMESPathTypeError

nicholasbishop:
Reverted the "success" acceptor for that waiter to an earlier version
that checks only the response code, not the response body.

Added a unit test that reproduces the issue and passes with the
change to the JSON file.

Bug: https://github.com/boto/botocore/issues/906

### DIFF
--- a/botocore/data/ec2/2015-10-01/waiters-2.json
+++ b/botocore/data/ec2/2015-10-01/waiters-2.json
@@ -7,9 +7,8 @@
       "operation": "DescribeInstances",
       "acceptors": [
         {
-          "matcher": "path",
-          "expected": true,
-          "argument": "length(Reservations[]) > `0`",
+          "matcher": "status",
+          "expected": 200,
           "state": "success"
         },
         {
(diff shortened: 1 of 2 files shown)


botocore/data/ec2/2015-10-01/waiters-2.json
```diff
@@ -7,9 +7,8 @@
       "operation": "DescribeInstances",
       "acceptors": [
         {
-          "matcher": "path",
-          "expected": true,
-          "argument": "length(Reservations[]) > `0`",
```
jamesls: I believe this was added because it's possible to get an empty reservations list even if you specify an instance id in some cases.  I think we'll just need to modify the waiter code to handle this case appropriately.

jamesls:
I pulled in your PR and updated the code to only search non error responses.  Closing in favor of https://github.com/boto/botocore/pull/917

Thanks for the pull request.

PR #912, opened 2016-05-06
Bump delay time for VpnConnection waiter

kyleknap:
Users were seeing waiter timeouts with this.  So bumped up the delay time as it can take roughly 30 minutes for it to be available.

cc @jamesls @JordonPhillips


JordonPhillips:
⛵

PR #914, opened 2016-05-06
Provide better error when page size is not supported

### LINKED ISSUE #876: Cloudformation Paginator PageSize does not work
We're trying to use page size with the CloudFormation paginator.  We're having the following issue.

Example code:
client = boto3.client('cloudformation')
paginator = client.get_paginator('list_stacks')
for page in paginator.paginate(StackStatusFilter=['CREATE_COMPLETE'], PaginationConfig={'PageSize': 1}):
  ......

Will throw error:
.....
botocore/paginate.py", line 175, in _make_request
    return self._method(**current_kwargs)
TypeError: _api_call() keywords must be strings

We've traced it down to _limit_key being None, which is used in a kargs dictionary.  So you get {None: 1} (assuming 1 was the PageSize arg sent earlier).

Guessing what _limit_key string should be, we always get unexpected parameter (below trying _limit_key="PageSize'):
Unknown parameter in input: "PageSize", must be one of: NextToken, StackStatusFilter

Using
boto3==1.3.0
botocore==1.4.11

Are you able to reproduce?
Thanks,
Mike

JordonPhillips:
Some paginate-able operations do not support page size. Before, we would
just blindly try to toss it in anyway, leading to a very confusing error.
This raises an error on a call to paginator.paginate if page size is
provided, but not supported.

Fixes #876

cc @jamesls @kyleknap

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -45,7 +45,6 @@ class PageIterator(object):
                  result_keys, non_aggregate_keys, limit_key, max_items,
                  starting_token, page_size, op_kwargs):
         self._method = method
-        self._op_kwargs = op_kwargs
         self._input_token = input_token
         self._output_token = output_token
         self._more_results = more_results
@@ -472,6 +471,10 @@ class Paginator(object):
             max_items = int(max_items)
         page_size = pagination_config.get('PageSize', None)
         if page_size is not None:
+            if self._pagination_cfg.get('limit_key', None) is None:
+                raise PaginationError(
+                    message="PageSize parameter is not supported for the "
+                            "pagination interface for this operation.")
             page_size = int(page_size)
         return {
             'MaxItems': max_items,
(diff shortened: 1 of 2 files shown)


botocore/paginate.py
```diff
@@ -472,6 +471,10 @@ def _extract_paging_params(self, kwargs):
             max_items = int(max_items)
         page_size = pagination_config.get('PageSize', None)
         if page_size is not None:
+            if self._pagination_cfg.get('limit_key', None) is None:
+                raise PaginationError(
+                    message="Page size unsupported by the service for this "
```
kyleknap: Can we make the error message a little more clear like (mainly the spelling of Page size):
``PageSize parameter is not supported for the pagination interface for this operation``

kyleknap:
Looks good just had a small comment on the error message.

JordonPhillips:
Updated error message

kyleknap:
Thanks! :ship:

PR #915, opened 2016-05-10
Implement context mangager functions for Stubber

JordonPhillips:
It can be useful to use a stubber as a context manager to make code a
bit more readable and to ensure you aren't leaking events.

I was responding to a question on StackOverflow about stubbing responses
in which they were attempting to use mock. While they didn't complain
about having to use `activate`, I realized my sample code didn't look
as clean as their (unsuccessful) attempt at mocking. Since it is such
a tiny change, I figured it would be useful if I just knocked it out.

## Example
```python
import datetime
import botocore.session
from botocore.stub import Stubber


s3 = botocore.session.get_session().create_client('s3')

response = {
    "Owner": {
        "ID": "foo",
        "DisplayName": "bar"
    },
    "Buckets": [{
        "CreationDate": datetime.datetime(2016, 1, 20, 22, 9),
        "Name": "baz"
    }]
}


with Stubber(s3) as stubber:
    stubber.add_response('list_buckets', response, {})
    service_response = s3.list_buckets()

assert service_response == response
```

cc @kyleknap @jamesls

### DIFF
--- a/botocore/stub.py
+++ b/botocore/stub.py
@@ -63,6 +63,37 @@ class Stubber(object):
 
         service_response = s3.list_objects(Bucket='test-bucket')
         assert service_response == response
+
+
+    This class can also be called as a context manager, which will handle
+    activation / deactivation for you.
+
+    **Example:**
+    ::
+        import datetime
+        import botocore.session
+        from botocore.stub import Stubber
+
+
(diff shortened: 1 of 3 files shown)


kyleknap:
Nice. :ship:

PR #916, opened 2016-05-12
Ssm waiter

aereinha:
Added waiter for a successfully run command from SSM.


kyleknap:
Logic looks good. One thing though could we rename the waiter to ``"CommandStatusSuccess"``? Usually we like to the name the waiter for whatever specific value that we are waiting for. In this case, we are waiting for ``Success`` not really the status of ``Ok``.

radykal-com:
Hello,

I'd like to have this waiter merged. Is there any thing I can help with?

PR #920, opened 2016-05-13
Use MD5 to sign S3 SigV4 bodies

JordonPhillips:
S3 now supports optional SHA256 body signing, so this switches from SHA256 to MD5 to boost speed while reducing CPU and Memory usage when using sigv4 with S3. SHA256 will still be used on systems which do not support using MD5.

cc @kyleknap @jamesls


botocore/auth.py
```diff
@@ -367,7 +368,15 @@ def _modify_request_before_signing(self, request):
         super(S3SigV4Auth, self)._modify_request_before_signing(request)
         if 'X-Amz-Content-SHA256' in request.headers:
             del request.headers['X-Amz-Content-SHA256']
-        request.headers['X-Amz-Content-SHA256'] = self.payload(request)
+
+        # S3 allows optional body signing, so to minimize the performance
+        # impact, we opt to not SHA256 sign the body. Instead, we MD5 it.
+        # We will fall back to SHA256 when MD5 is not available, such as with
+        # FIPS systems.
+        if MD5_AVAILABLE:
```
jamesls: Can't we just check if there's a Content-MD5 value in the request headers?  Seems more definitive.

jamesls:
I think we need to have a config option to enable/disable this.  Especially given that users can reject requests without unsigned payloads via IAM policies (http://docs.aws.amazon.com/AmazonS3/latest/API/bucket-policy-s3-sigv4-conditions.html), we need to give them a way to enable this behavior if needed.

I think I still prefer only doing this for the streaming s3 requests (put_object/upload_part) but I'm interested to hear other's thoughts.

JordonPhillips:
I updated to check for the header and added a config flag. Still need to look into making this exclusive to UploadPart and PutObject

JordonPhillips:
Personally, I think we should go for unsigned payloads regardless. It keeps things simple on our side and provides the performance boost consistently. What do you think @kyleknap

kyleknap:
I talked to you about this, but wanted to capture the conversation. Lets only do unsigned payload for ``PutObject`` and ``UploadPart`` for the following reasons:
1) The behavior change will only be scoped to ``PutObject`` and ``UploadPart`` leaving a lot of the control plane operations of S3 untouched.
2) We have knowledge upfront if the payload is streaming by looking at the operation model when we create the context so we won't be hard coding ``PutObject`` and ``UploadPart`` into auth signer if we pass that knowledge to the context and we would probably would want to disable body signing for any future S3 operation that has a streaming body for its input.

jamesls:
I'm going to switch this to incorporating-feedback.  Seems like this is going to change based on @kyleknap 's feedback so I'll defer reviewing this until those changes are made.

JordonPhillips:
Rebased on top of #936 to take advantage of context plumbing.

JordonPhillips:
Squashed the commits. Only the top commit needs to be reviewed.

botocore/handlers.py
```diff
@@ -148,11 +148,20 @@ def _calculate_md5_from_file(fileobj):
...
+def conditionally_calculate_md5(params, context, request_signer, **kwargs):
+    """Only add a Content-MD5 if the system supports it."""
+    if not MD5_AVAILABLE:
+        return
+
+    s3_config = context.get('client_config').s3
+    if s3_config is None:
+        s3_config = {}
+    use_sha256 = s3_config.get('sha256_sign_s3v4_payload', False)
+    if request_signer.signature_version == 's3v4' and \
```
kyleknap: Questions for everyone: if the users opt-in to sha256_sign_s3v4_payload, should we also do MD5 calculation? Based on the docstring you have it is not clear that you turn off MD5 signing.

tests/functional/test_s3.py
```diff
@@ -148,6 +149,34 @@ def test_500_error_with_non_xml_body(self):
...
+
+    def get_sent_headers(self):
+        return self.http_session_send_mock.mock_calls[0][1][0].headers
+
+    def test_content_md5_set(self):
+        self.client.put_object(Bucket='foo', Key='bar', Body='baz')
+        self.assertIn('content-md5', self.get_sent_headers())
+
+    def test_content_sha256_set_if_md5_is_unavailable(self):
+        with mock.patch('botocore.auth.MD5_AVAILABLE', False):
```
kyleknap: This can't be patched from ``botocore.compat``?

tests/functional/test_s3.py
```diff
@@ -148,6 +149,34 @@ def test_500_error_with_non_xml_body(self):
...
+            with mock.patch('botocore.handlers.MD5_AVAILABLE', False):
+                self.client.put_object(Bucket='foo', Key='bar', Body='baz')
+        sent_headers = self.get_sent_headers()
+        unsigned = 'UNSIGNED-PAYLOAD'
+        self.assertNotEqual(sent_headers['x-amz-content-sha256'], unsigned)
+        self.assertNotIn('content-md5', sent_headers)
+
```
kyleknap: Should we add a test for when the config value of opting-in to sigv4 signing is turned on?

tests/functional/test_s3.py
JordonPhillips: No

botocore/handlers.py
JordonPhillips: You're right, we shouldn't disable MD5. Rather, we should enforce SHA256 if that config isn't explicitly set and MD5 isn't present.

JordonPhillips:
Moved the config value checking into auth so that MD5 is not impacted.

kyleknap:
Changes look good here. I would like to see some testing with the CLI done before we merge this one. Mainly switching everything to sigv4 and running the s3 integration tests. If those pass, I would feel pretty confident about merging it.

botocore/auth.py
```diff
@@ -362,12 +363,52 @@ def _set_necessary_date_headers(self, request):
...
+        # provided that we're on https.
+        client_config = request.context.get('client_config')
+        s3_config = getattr(client_config, 's3', None)
+
+        # The config could be None if it isn't set, or if the customer sets it
+        # to None.
+        if s3_config is None:
+            s3_config = {}
+
+        use_sha256 = s3_config.get('sha256_sign_s3v4_payload')
```
jamesls: I would prefer this is named something like `sign_payload` or `enable_signed_payloads`.

The sha256 seems implementation specific and the `s3` is implied by being in the `s3` config.

After thinking about it, I don't think it's ambiguous with the ContentMD5 because ContentMD5 is used as a checksum and the x-amz-sha256 is used to sign the payload.

@kyleknap thoughts?

jamesls:
Main concern was about the name otherwise looks good.

botocore/auth.py
kyleknap: Makes sense. I think I would vote for ``sign_payloads``.

JordonPhillips: I disagree, I think `sign_payloads` is far too broad. That would enable/disable ALL kinds of payload signing, including MD5. You would have to take all or nothing. If you don't, it's not at all obvious to the customer what that variable does.

kyleknap: Hmm. So I think we can take the ``s3`` part out at least since it is in the s3 config already. Honestly, I do not have a strong opinion on what the name should be. People will have to read the docs anyway to understand what it does so as long as we are clear there I am fine.

As to ``sign_payloads``, if there is a new auth, that comes along would that mean we have to not sign the payload there as well if the config option is turned off? It may be good to keep the sigv4 part in the name for backwards compatibility reasons.

kyleknap: Yet again now that I think of it. If we do want the option to turn off payload signing for a newer auth mechanism, we would not be able to reuse the variable.

jamesls: @JordonPhillips ContentMD5 is not being used as part of the signature.  ContentMD5 and X-Amz-Sha256 are both *checksums* of the body, but only one is being used in the context of signing.

I think sign_payload or enable_signed_payloads clearly communicate this.  

@kyleknap  I would be ok with something slightly more specific like sign_v4_payloads, but I think that can be confusing.  I would expect customers not to care what "signature version 4" is.  They'd only care whether or not we're computing a sha256 of the body as part of signing a request.

JordonPhillips: It's just so broad. What happens when I say `sign_payloads = True`? What am I signing my payload with? Everything? Maybe `disable_payload_signing` is a bit more explicit. It says the same thing, but when it's False all you're promising is that payload signing isn't disabled. You're still forced to make sure every form of payload signing is disabled, but the the opposite case is more clearly defined.

jamesls: A given auth mechanism will define precisely how to sign a payload.  With signature version 4, there are two, and only two options:

* sha256 of the body
* the string UNSIGNED-PAYLOAD.

Again, I think you're confusing ContentMD5 as a mechanism for signing a payload.  It's not.

Another more explicit option could be "use_unsigned_payloads".

JordonPhillips: I just talked with some of the Java guys. What they're using now, which has been released, is `payload_signing_enabled`. That basically the same as what I was talking about, and it would be good to be consistent. What do you two think?

jamesls: That's also pretty similar to what I suggested in `use_unsigned_payloads`.  Not opposed.

The one thing we should keep in mind is that it would be nice for us to standardize on the naming of boolean values for client config.

JordonPhillips:
Rebased against the latest redirector and tentatively changed the name to `payload_signing_enabled`

botocore/auth.py
```diff
@@ -362,12 +363,52 @@ def _set_necessary_date_headers(self, request):
...
+        client_config = request.context.get('client_config')
+        s3_config = getattr(client_config, 's3', None)
+
+        # The config could be None if it isn't set, or if the customer sets it
+        # to None.
+        if s3_config is None:
+            s3_config = {}
+
+        sign_payload = s3_config.get('payload_signing_enabled')
+        if sign_payload in [True, 'True', 'true']:
```
jamesls: -1 to this.  We shouldn't let the lack of typed parsing in the parser/config objects leak into dependent objects.  I would prefer we try to normalize this as high up the stack as we can.  Likely in the client creator for now.

JordonPhillips:
Updated

jamesls:
:shipit:  Looks good to me.  Thanks for incorporating the feedback.

jamesls:
da6f25fbf541885a243ad90a3ecee3604aefa444 looks good to me.

jamesls:
@JordonPhillips failing build?

kyleknap:
:ship: Thanks for all the work you put into it.

JordonPhillips:
@jamesls it's codecov. What is our official target percent?

PR #924, opened 2016-05-17
Add paginator for ListObjectsV2

### LINKED ISSUE #922: Add paginator for s3 list_objects_v2
I started off using `client.get_paginator('list_objects')` but then [read that their is a new "v2" version of list objects][1] that should be used. 

But, when I tried to use it `client.get_paginator('list_objects_v2')` I get:

```
    File "/home/rsyring/.virtualenvs/filebrowser/lib/python3.5/site-packages/botocore/client.py", line 602, in get_paginator
    raise OperationNotPageableError(operation_name=operation_name)
    botocore.exceptions.OperationNotPageableError: Operation cannot be paginated: list_objects_v2
```

I see there are other requests for paginators (#919), so I'd like to request one for this method as well.

Thanks.

[1]: http://boto3.readthedocs.io/en/latest/reference/services/s3.html?highlight=list_objects#S3.Client.list_objects_v2

JordonPhillips:
Fixes #922

cc @kyleknap @jamesls

### DIFF
--- a/.changes/next-release/feature-s3.json
+++ b/.changes/next-release/feature-s3.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "description": "Add paginator for ListObjectsV2",
+  "category": "s3"
+}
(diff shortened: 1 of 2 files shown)


kyleknap:
:ship:

PR #928, opened 2016-05-25
Add additional keys to error response

JordonPhillips:
Some services return additional error keys that may be relevant
in debugging, this adds the ability to add arbitrary keys to the
stubbed error response.

cc @kyleknap @jamesls


jamesls:
:shipit:

PR #933, opened 2016-05-28
Remove docutil dependency from setup.py

Fale:
Docutil is used only in the following places:
- botocore/docs/bcdoc/textwriter.py

And is also referenced in:
- setup.py
- setup.cfg
- requirements.txt

Since during the setup.py execution the documentation is not generated,
I think we should not require docutil to be installed. We could create a
secondary target to create the documentation, but this is a different
problem/opportunity/situation.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,6 +5,5 @@ universal = 1
 requires-dist =
 	python-dateutil>=2.1,<3.0.0
 	jmespath>=0.7.1,<1.0.0
-	docutils>=0.10
 	ordereddict==1.1; python_version=="2.6"
 	simplejson==3.3.0; python_version=="2.6"
--- a/setup.py
+++ b/setup.py
@@ -6,8 +6,7 @@ from setuptools import setup, find_packages
 
 
 requires = ['jmespath>=0.7.1,<1.0.0',
-            'python-dateutil>=2.1,<3.0.0',
-            'docutils>=0.10']
+            'python-dateutil>=2.1,<3.0.0']
 
 
 if sys.version_info[:2] == (2, 6):


jamesls:
Just wanted to follow up on this.  I don't think we can remove the dependency while we're still importing docutils in botocore's codebase.

I think we'll need to first move textwriter.py up to the CLI (where it's used) before we can merge this.

amw:
Can I help with this issue? I would like to remove the `docutils` dependency before I start using `boto3` in my project.

The issue is significant, because `docutils` uses non-trivial license. Parts of it are even covered by GPL3.

http://docutils.sourceforge.net/COPYING.html

amw:
I am still interested in helping getting this through. I don't understand this dependency though. `docutils` is used in `botocore/docs/bcdoc/textwriter.py`, but searching for the file (`git grep textwriter`) and the class it implements (`git grep TextWriter`) returns no results.

Fale:
@amw do you think we can just drop that class?

amw:
@Fale I don't see where it is used, but @jamesls said it is used in "CLI" – but I don't know what he means. Maybe it's in some other repo, but `boto3` doesn't reference it either.

Fale:
@amw cli is https://github.com/aws/aws-cli

amw:
Oh, yeah. It's used there. I'll see what we can do.

stephen-goveia:
Any movement on this? Would really like to see botocore GPL free, i.e. remove docutils. Thanks team!

borlafu:
I'd also love to see this change happening. Any update? Thanks!

nateprewitt:
This issue should now be resolved in Botocore 1.18.0.

PR #934, opened 2016-05-28
Remove has_key that is deprecated

Fale:


kyleknap:
Can confirm it is the two are the same as well: https://bitbucket.org/mirror/docutils/src/39b7b6b6fa0370d983a66931e5aa4c8ba19a74d1/docutils/docutils/nodes.py?at=default&fileviewer=file-view-default#nodes.py-560,647,650. Merging.

PR #936, opened 2016-05-31
Automatically redirect S3 sigv4 requests sent to the wrong region

JordonPhillips:
S3 generally provides enough information for us to redirect
requests, so this attempts to do so. To prevent a massive
number of additional requests, bucket regions are cached.

When a redirect occurs, a warning is printed that tells the
customer they should use a client configured to the proper
region to avoid additional requests.


botocore/auth.py
```diff
@@ -154,7 +154,8 @@ def __init__(self, credentials, service_name, region_name):
         # We initialize these value here so the unit tests can have
         # valid values.  But these will get overriden in ``add_auth``
         # later for real requests.
-        self._region_name = region_name
+        self._signing_region = region_name
```
kyleknap: Could we just keep these changes to the S3Sigv4 signer? I would prefer that this change did not affect all AWS services that use sigv4 signing.

botocore/auth.py
```diff
@@ -303,14 +304,17 @@ def signature(self, string_to_sign, request):
...
-        k_region = self._sign(k_date, self._region_name)
+        k_region = self._sign(k_date, self._signing_region)
         k_service = self._sign(k_region, self._service_name)
         k_signing = self._sign(k_service, 'aws4_request')
         return self._sign(k_signing, string_to_sign, hex=True)
 
     def add_auth(self, request):
         if self.credentials is None:
             raise NoCredentialsError
+        signing_context = request.context.get('signing', {})
```
kyleknap: Just want to make sure here. Each request gets one auth instance and individual requests can never share auth instances? If that is so lets make a note here as if we ever do consider sharing them that could be a hard concurrency issue to figure out.

botocore/client.py
```diff
@@ -534,7 +539,7 @@ def _make_api_call(self, operation_name, api_params):
             http, parsed_response = event_response
         else:
             http, parsed_response = self._endpoint.make_request(
-                operation_model, request_dict)
+                operation_model, request_dict, request_context)
```
kyleknap: I am not a fan of adding the request_context to the ``make_request`` signature because we would have to modify all of the other methods of the ``Endpoint`` class. What if we did the following:

1) Keep the signature the same for all of the ``endpoint`` module
2) Introduce the idea of passing a context to ``prepare_request_dict``: https://github.com/boto/botocore/blob/develop/botocore/awsrequest.py#L263
3) Add the context to AWSRequest object from the request dict when we create it: https://github.com/boto/botocore/blob/develop/botocore/awsrequest.py#L294.

Another thing that I am not sure if I am liking is the fact we take a context that we expose in the all of the events related to parameter building and strap it into a context living on ``AWSRequest`` that we have never exposed. It seems that there is a chance for really hard to determine bugs. I would prefer that the user could never affect the context of ``AWSRequest`` because that could affect behavior that we rely on.

botocore/utils.py
```diff
@@ -794,5 +799,102 @@ def _switch_hosts(request, new_endpoint, use_new_scheme=True):
...
+            context = {}
+
+        signing_context = {
+            'region': new_region,
+            'bucket': bucket,
+            'endpoint': endpoint
+        }
+        context['signing'] = signing_context
+
+        self._cache[bucket] = signing_context
```
kyleknap: Do you think we should strap a lock around modifying and grabbing from the cache? If there are multiple threads sharing a client is there a chance that we hit an issue from the fact that the threads are sharing state of the cache?

botocore/utils.py
```diff
@@ -794,5 +799,102 @@ def _switch_hosts(request, new_endpoint, use_new_scheme=True):
...
+            return http_response.headers['x-amz-bucket-region']
+
+        # Next, check the error body
+        region = service_response.get('Error', {}).get('Region')
+        if region:
+            return region
+
+        # Finally, HEAD the bucket. No other choice sadly.
+        endpoint = 'https://' + response[1]['Error']['Endpoint']
+        head_response = requests.head(endpoint)
```
kyleknap: Hmmm. Is this how other SDKs are doing this? Or are they using the client interface? I feel like we may be missing some of the knobs that are tuned in ``Endpoint`` like verify and proxies.

Also this request does not need to be signed then? Is there any edge cases we have to worry about or is a head bucket guaranteed to work (i.e. no permission or signing requirements)?

tests/unit/test_utils.py
```diff
@@ -940,5 +950,135 @@ def test_uses_original_endpoint_scheme(self):
...
+        self.redirector.set_request_url(params, context)
+        self.assertEqual(
+            params['url'], 'https://us-west-2.amazonaws.com/foo')
+
+    def test_sets_signing_context_from_cache(self):
+        signing_context = {'endpoint': 'bar'}
+        cache = {'foo': signing_context}
+        params = {'Bucket': 'foo'}
+        context = {}
+        self.redirector._cache = cache
```
kyleknap: We should not be relying on modifying internal properties. Have you considered being able to inject a cache or make the property public?

botocore/auth.py
JordonPhillips: Possibly

botocore/auth.py
JordonPhillips: Each client gets one auth instance, each request gets one context object that is discarded after the request is prepared.

kyleknap:
Looks like a good start. My main comments were related to a different approach to injecting context and some possible edge cases.

I think one cool thing to try if you have not already done so is run the CLI integration test with an incorrect region and sigv4 enabled. That should help improve our confidence that we handled all of the edge cases.

botocore/client.py
JordonPhillips: That context gets discarded when the request is prepared, and this signing context is added in addition to whatever is there. An end user has no way to modify the existing contents of that request context besides changing the signing context we inject.

The request dict thing could work.

botocore/utils.py
JordonPhillips: That's a good point, I hadn't considered that.

botocore/utils.py
JordonPhillips: This is at least how Ruby is doing it. You don't need to be signed in to head a bucket. In fact, adding auth is far more brittle because bad auth will cause it to fail. A raw HEAD will always succeed since bucket names are not secret.

tests/unit/test_utils.py
JordonPhillips: I think injecting it would be best.

botocore/endpoint.py
```diff
@@ -199,24 +199,24 @@ def _get_response(self, request, operation_model, attempts):
...
-                     caught_exception=None):
+    def _needs_retry(self, attempts, operation_model, request_dict,
+                     response=None, caught_exception=None):
         event_name = 'needs-retry.%s.%s' % (self._endpoint_prefix,
                                             operation_model.name)
         responses = self._event_emitter.emit(
             event_name, response=response, endpoint=self,
             operation=operation_model, attempts=attempts,
-            caught_exception=caught_exception)
+            caught_exception=caught_exception, request_dict=request_dict)
```
kyleknap: If we are just going for the context here, we do not need the ``request_dict``. I believe you should be able to get it from ``response.request.original.context`` and we will not need to add an extra keyword arg to the emitted event or ``_needs_retry`` method.

botocore/client.py
```diff
@@ -208,6 +209,10 @@ def _get_client_args(self, service_model, region_name, is_secure,
         self._inject_s3_configuration(
             config_kwargs, scoped_config, client_config)
         self._conditionally_unregister_fix_s3_host(endpoint_url, event_emitter)
+        if service_name == 's3':
+            # We want to make double sure that we don't create this object
```
kyleknap: Why do we need to check this only for s3? The fact that you have all events scoped to s3 should be fine enough right?

botocore/utils.py
kyleknap: Do they take into account ca certs or proxies (Pretty much any parameter we set in the ``http_session.send()`` method in endpoint)? I think this could be an issue for those who may rely on them. Could we store them in the context possibly and use them for the head request?

botocore/client.py
JordonPhillips: We only need this for S3, and at this point we have a trivially easy way to limit it to S3. Otherwise we have that object laying around taking up memory and hanging on to the endpoint bridge for services that don't need it.

kyleknap: We already limit it to S3 by how we register the events. I also do not think it would be a significant memory imprint. I just find it ugly when reading the code to have a big if statement for a specific service in the middle of generic logic. @jamesls want to be the tie breaker?

kyleknap:
It is looking good. Just had a couple of comments. Would be good to get @jamesls to look at it as well.

tests/unit/test_utils.py
```diff
@@ -102,12 +104,18 @@ def test_already_snake_cased(self):
...
-        self.assertEqual(xform_name('DescribeCachediSCSIVolumes', '-'), 'describe-cached-iscsi-volumes')
-        self.assertEqual(xform_name('DescribeStorediSCSIVolumes', '-'), 'describe-stored-iscsi-volumes')
-        self.assertEqual(xform_name('CreateStorediSCSIVolume', '-'), 'create-stored-iscsi-volume')
+        self.assertEqual(xform_name('SwapEnvironmentCNAMEs'),
+                         'swap_environment_cnames')
+        self.assertEqual(xform_name('SwapEnvironmentCNAMEs', '-'),
+                         'swap-environment-cnames')
+        self.assertEqual(xform_name('CreateCachediSCSIVolume', '-'),
+                         'create-cached-iscsi-volume')
+        self.assertEqual(xform_name('DescribeCachediSCSIVolumes', '-'),
```
jamesls: In the future, please at least have separate commits for functional changes vs. code cleanup.

It doesn't have to be an entirely separate PR (though that's helpful), but it clutters PRs when reviewers have to sift through what's relevant to your changes vs. what's not.  I know I do this sometimes to so feel free to remind me if you see me doing it.

botocore/client.py
jamesls: I agree that it's more readable if we don't have the conditional.  I would actually prefer to remove as much service specific logic as possible in the client creator (I already think it's too much).

botocore/awsrequest.py
```diff
@@ -305,9 +305,11 @@ def create_request_object(request_dict):
 
     """
     r = request_dict
-    return AWSRequest(method=r['method'], url=r['url'],
-                      data=r['body'],
-                      headers=r['headers'])
+    request_object = AWSRequest(
+        method=r['method'], url=r['url'], data=r['body'], headers=r['headers'])
+    if 'context' in r:
```
jamesls: Is there ever a reason not to have context in the request dict?  I'd prefer to have less of these conditionals if possible.  I can't think of a reason why this should be optional.

botocore/client.py
```diff
@@ -577,6 +582,7 @@ def _convert_to_request_dict(self, api_params, operation_model,
             api_params, operation_model)
         prepare_request_dict(request_dict, endpoint_url=self._endpoint.host,
                              user_agent=self._client_config.user_agent)
+        request_dict['context'] = context
```
jamesls: Is there a reason not to fold this into `prepare_request_dict`?

botocore/utils.py
jamesls: I think we should be using clients here.  I don't see any reason to use `requests` directly.  That should be thought of as an implementation detail.

We can already  make unauth'd requests in clients, so if that's the main driving reason, I think we should still be using clients here.

botocore/utils.py
```diff
@@ -794,5 +800,117 @@ def _switch_hosts(request, new_endpoint, use_new_scheme=True):
...
+
+    def get_region_from_response(self, response):
+        """
+        There are multiple potential sources for the new region to redirect to,
+        but they aren't all universally available for use. This will try to
+        find region from response elements, but will fall back to calling
+        HEAD on the bucket if all else fails.
+        """
+        # First try to source the region from the headers.
+        http_response = response[0]
```
jamesls: I'd prefer not to rely on `http_response`.  Can we add this into the service response somehow?  What about some way of injecting this into the request metadata?

botocore/utils.py
```diff
@@ -794,5 +800,117 @@ def _switch_hosts(request, new_endpoint, use_new_scheme=True):
...
+        if response is None:
+            # This could be none if there was a ConnectionError or other
+            # transport error.
+            return
+
+        error = response[1].get('Error', {})
+        if error.get('Code') != 'PermanentRedirect':
+            return
+
+        if 'context' not in request_dict:
```
jamesls: Same thing as above.  Unless there's a strong reason to have context be optional, it simplifies code if make make guarantees about the contents of the request dict.

botocore/utils.py
```diff
@@ -794,5 +800,117 @@ def _switch_hosts(request, new_endpoint, use_new_scheme=True):
...
+        }
+        request_dict['context']['signing'] = signing_context
+
+        self._add_bucket_to_cache(bucket, signing_context)
+        self.set_request_url(request_dict, request_dict['context'])
+
+        # Return 0 so it doesn't wait to retry
+        return 0
+
+    def get_region_from_response(self, response):
```
jamesls: This name doesn't match with the description.  This isn't just getting the region from the response.  It will potentially make network calls if necessary.

tests/integration/test_s3.py
```diff
@@ -1033,5 +1034,54 @@ def setUp(self):
...
+        self.bucket_region = 'eu-central-1'
+        self.client_region = 'us-west-2'
+
+        self.client = self.session.create_client(
+            's3', region_name=self.client_region,
+            config=Config(signature_version='s3v4'))
+
+    def create_bucket(self, region_name, bucket_name=None):
+        saved_client = self.client
+        if region_name != self.client.meta.region_name:
```
jamesls: Seems like it's cleaner if you just allow create_bucket to be passed a client to use.

tests/integration/test_s3.py
```diff
@@ -1033,5 +1034,54 @@ def setUp(self):
...
+        self.client = saved_client
+        return bucket
+
+    def test_region_redirects(self):
+        self.bucket = self.create_bucket(self.bucket_region)
+        try:
+            response = self.client.list_objects(Bucket=self.bucket)
+            self.assertEqual(
+                response['ResponseMetadata']['HTTPStatusCode'], 200)
+        except ClientError:
```
jamesls: Should this check be more precise?  Seems like we should ensure we don't get the bucket is in wrong region code and allow any other errors to propagate.

botocore/utils.py
JordonPhillips: We can't use the client though. Our auth will cause the request to fail. We can't correct the auth because we don't know where the bucket is. We would need to construct an entirely new client using the endpoint they give us.

botocore/endpoint.py
JordonPhillips: But we need to modify the request dict so that the next request created has the correct signing context. The old request object is not used when constructing the new one.

botocore/client.py
JordonPhillips: It's already S3 specific, the name of the class includes 'S3' as do the names of the functions above. I don't see how this is any more ugly. If I had to go to any amount of effort to get the name of the service I would agree, but it's readily available. It seems irresponsible to create this object and hold it in memory for the life of the client when it isn't being used for any other service given how easy it is to avoid that.

botocore/utils.py
JordonPhillips: If we're going to do something like that, I would rather have it added to the service model. Why isn't `http_response` reliable?

JordonPhillips:
Updated in response to some feedback, minus that which has outstanding comments.

botocore/client.py
kyleknap: Probably is a bit extreme for this change. But what if we created specific subclasses for the ``ClientCreator`` and/or ``BaseClient`` like ``S3ClientCreator`` and/or ``S3Client``? That would remove the issue of having specific if statements, registering unused handlers, and instantiating unused objects.

botocore/endpoint.py
kyleknap: Instead of updating it [here](https://github.com/boto/botocore/pull/936/commits/f32d6241da8c137008e93a9e3647127a2a2a4c59#diff-cb00fc815f8f80ae15e7d0cecc74b395R311) you could just assign the reference to it and that should allow you to modify it. I do not see why it even needs to be an update since you start with a brand new dictionary each time. I honestly think you should even be able to inject it into the ``AWSRequest`` object. I am just really trying to avoid adding a new parameter to an event if we do not need to.

botocore/utils.py
kyleknap: I think being able to use a client here would be ideal because it handles all of the configuration for us, but I am not sure how to plumb it through and get it to use the correct endpoint as well. That is why I am saying we should at least match tuning parameters sent to http_session.send(). @jamesls any ideas on how would you do it?

botocore/client.py
JordonPhillips: I think this is something we should do, though probably out of scope for this pr.

botocore/endpoint.py
JordonPhillips: That could be done, but it would again require we inspect the http response. @jamesls, what do you think?

JordonPhillips:
Updated again. I'm now passing the HTTP headers in through the response metadata, the bucket cache is now on the session, and I'm using the client to make the head bucket request.

tests/unit/__init__.py
```diff
@@ -11,3 +11,30 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import copy
+from tests import unittest
+
+
+class BaseResponseTest(unittest.TestCase):
+    def assert_response(self, actual_response, expected_response):
```
kyleknap: Can we rename this to something more explicit it. ``assert_response`` seemed vague to me when I read it. I think my main issue was that I did not realize you do a subset check to assert the response. I would have thought it would be an equality check. Maybe something like ``assert_response_with_subset_metadata``?

botocore/utils.py
```diff
@@ -794,5 +800,138 @@ def _switch_hosts(request, new_endpoint, use_new_scheme=True):
...
+            return response_headers['x-amz-bucket-region']
+
+        # Next, check the error body
+        region = service_response.get('Error', {}).get('Region')
+        if region:
+            return region
+
+        # Finally, HEAD the bucket. No other choice sadly.
+        bucket = service_response.get('Error', {}).get('Bucket')
+        self.unregister()
```
kyleknap: I would be very wary of doing unregister and registering mid-request for concurrencyu. If a client is being shared what happens if you unregistered the request mid another thread using the client to send a request that has the incorrect bucket region? It seems that the redirection functionality may have a chance of slipping through the cracks here if we are possibly registering and unregistering these events.

botocore/utils.py
```diff
@@ -794,5 +800,138 @@ def _switch_hosts(request, new_endpoint, use_new_scheme=True):
...
+        This handler retrieves a given bucket's signing context from the cache
+        and adds it into the request context.
+        """
+        bucket = params.get('Bucket')
+        signing_context = self._cache.get(bucket)
+        if signing_context is not None:
+            context['signing'] = signing_context
+
+
+class S3BucketRegionCache(object):
```
kyleknap: Might be good to document this since it is a publicly exposed object on the session through its ``get_component``.  Like it is not clear what the signing context schema is.

botocore/client.py
```diff
@@ -159,6 +163,18 @@ def _conditionally_unregister_fix_s3_host(self, endpoint_url, emitter):
         if endpoint_url is not None:
             emitter.unregister('before-sign.s3', fix_s3_host)
 
+    def _create_s3_redirector(self, client, scoped_config):
+        if client.meta.service_model.service_name != 's3':
+            return
+
+        service_signing_name = client.meta.service_model.signing_name
+        endpoint_bridge = ClientEndpointBridge(
```
kyleknap: It is unfortunate that we have to recreate this. Is there any other way we could retrieve the original endpoint bridge so we do not have to make a new one?

kyleknap:
Its looking better. I had some more comments where some were small some were larger.

botocore/client.py
JordonPhillips: We can create it before the get args call and pass it in, should be trivial

botocore/utils.py
JordonPhillips: I actually realized that I don't need to do that anyway.

botocore/session.py
```diff
@@ -164,6 +165,7 @@ def _register_components(self):
         self._register_endpoint_resolver()
         self._register_event_emitter()
         self._register_response_parser_factory()
+        self._register_bucket_region_cache()
```
jamesls: I know we chatted about this, but I really don't like the session have service specific concepts in it.  The session should remain general purpose.  Otherwise if we allow service specific stuff in here, it can quickly deteriorate into a hard to maintain class.

Is there a way we can do this without have S3 concepts fixed in the session/clientcreator?  I know we want to share bucket caches across clients but it seems like there's a less hard coded way to do it.

botocore/signers.py
```diff
@@ -492,7 +492,7 @@ def generate_presigned_url(self, ClientMethod, Params=None, ExpiresIn=3600,
 
     # Prepare the request dict by including the client's endpoint url.
     prepare_request_dict(
-        request_dict, endpoint_url=self.meta.endpoint_url)
+        request_dict, endpoint_url=self.meta.endpoint_url, context={})
```
jamesls: This is kind of a weird idiom to pass an empty dict through to the prepare function if we don't need to prepare/reference it before hand.  We should just let the `prepare_request_dict` own creation of the context dict.

botocore/utils.py
jamesls: Why is this so specific to S3?  This is basically a restricted dict interface.  I'd expect it to just be a MutableMapping or some other similar class.

botocore/session.py
JordonPhillips: Is there? I guess we could make it a singleton, but...

botocore/client.py
kyleknap: I would be more in favor of that.

jamesls:
An unfortunate wrinkle:  this doesn't seem to be working with the CLI:

```
$ aws s3 cp 00/19/c583bbfdd5c26af85616862a10b8d52cd9e4 s3://bucket/ --profile default --region us-west-1
upload failed: 00/19/c583bbfdd5c26af85616862a10b8d52cd9e4 to s3://bucket/c583bbfdd5c26af85616862a10b8d52cd9e4 A client error (301) occurred when calling the HeadBucket operation: Moved Permanently
```

I haven't really looked into this but I do see that we are making the HeadBucket request:

```
2016-06-08 15:26:49,661 - Thread-5 - botocore.endpoint - DEBUG - Making request for OperationModel(name=PutObject)
2016-06-08 15:26:50,247 - Thread-5 - botocore.endpoint - DEBUG - Making request for OperationModel(name=HeadBucket)
```

Anyone else seeing this?  Any ideas?

JordonPhillips:
For some reason cp is seeing the error that I catch and propagating it anyway. Not sure why.

botocore/session.py
jamesls: Thinking out loud:

a) If we think a general purpose cache is useful that can be shared amongst various clients then we can generalize the cache and not make it S3 specific.  Might be dangerous to let any arbitrary objects write to it.
b) We need to design a way to allow for service specific objects to be integrated into client creation in a more general purpose way.  It doesn't seem like S3 will be the last service that might need service specific objects.

kyleknap:
Doesn't the CLI hook into the error parsing with a handler: https://github.com/aws/aws-cli/blob/develop/awscli/errorhandler.py? It may be short-circuiting your handlers if it is registered first...

JordonPhillips:
That's a really odd way to handle exceptions, why not just catch them as normal?

kyleknap:
I do not remember what is needed for. It might be legacy code from when botocore did not have clients and we did not fully move off of it.

jamesls:
Yeah that's from the pre-client era.  However, be very careful ripping that code out (not opposed though).  Lots and lots of code in the CLI depends on that so you'll need to update various customizations and CLI internals.

JordonPhillips:
I could have that ClientError inherit from botocore. That should be relatively straightforward.

jamesls:
Not sure about that.  They have different attrs.

JordonPhillips:
Oh it can be done. It isn't pretty, but it can be done.

JordonPhillips:
The following would be workable, if awful. I'll need to look into how python resolves `super`.

```python
class CompatClientError(ClientError, botocore.exceptions.ClientError):
        code = error_response['Error'].get('Code', 'Unknown')
        message = error_response['Error'].get('Message', 'Unknown')
        msg = self.MSG_TEMPLATE.format(
            error_code=code, error_message=message,
            error_type='client', operation_name=operation_name)
        Exception.__init__(self, msg)
        self.error_code = code
        self.error_message = message
        self.error_type = 'client'
        self.operation_name = operation_name
        self.http_status_code = error_response['ResponseMetadata']['HTTPStatusCode']
        self.response = error_response
```

kyleknap:
I feel like this not the right approach to go down for this feature. Can't you just unregister the error handler for this? We already do this for a few customizations. Or make sure we register your redirect handler first?

I rather see if it is possible to completely remove the error handler from the CLI and clean up the CLI codebase accordingly, which I realize is a really large task and is probably why it has not happened yet. Making these complicated error classes seems that it would only make the technical debt worse.

jamesls:
Yeah, one of us can probably separately tackle removing the legacy error handling in the CLI.  This PR can still continue, it'll just be blocked on being merged until the CLI work is finished.

JordonPhillips:
It's in a completely different handler, and even if I could unregister it there's the concurrency issues to worry about. The only way to fix it would be to do that refactoring, and we would still need something like this for any existing plugins.

JordonPhillips:
I think that's a discussion to have separately. I don't think we can fix it in this PR, so lets get this to a ready-to-merge state and then worry about the CLI.

botocore/session.py
JordonPhillips: Well for b we could add an event to client creation. a would also work, though I agree it could be dangerous. We could also just make it session based later.

JordonPhillips:
I removed the bucket cache from the session.

botocore/endpoint.py
kyleknap: Ping @jamesls. I think you will have an opinion on this.

botocore/utils.py
```diff
@@ -794,5 +802,152 @@ def _switch_hosts(request, new_endpoint, use_new_scheme=True):
...
+        This handler retrieves a given bucket's signing context from the cache
+        and adds it into the request context.
+        """
+        bucket = params.get('Bucket')
+        signing_context = self._cache.get(bucket)
+        if signing_context is not None:
+            context['signing'] = signing_context
+
+
+class ThreadSafeCache(MutableMapping):
```
kyleknap: Can we add some unit tests for get, set, delete, len?

kyleknap:
Alright, I think you are really close for getting a ship it from me but I need two things addressed:

1) Closure on whether we add the ``request_dict`` to the ``need-retry`` event. (I am not inclined if there is a way around it)
2) Get tests for the ``ThreadSafeCache``

JordonPhillips:
Tests have been added, though I may need some guidance if we want to add tests that verify thread safety.

botocore/endpoint.py
jamesls: Looking...looks like I missed this comment initially.

kyleknap:
The tests part looks good. I do not think we need to add threaded test cases. We usually do not go to that extent for unit tests. I am sure if there is a problem, we will figure it out for the CLI s3 command tests if we see how many of them pass when we switch to an incorrect region.

botocore/endpoint.py
jamesls: Initially I was inclined to say that let's avoid adding a kwarg if possible.  However, I'm not opposed to adding new kwargs if it exposes new general purpose functionality.

The way I see it, we need the general ability to be able to not only indicate that a request should be retried, but allow handlers to alter the parameters of subsequent requests.

I think Jordon's approach here does that.  It opens up the possibility for handlers to change how requests are retried by modifying the `request_dict`.  I think that's pretty reasonable case for adding a new kwarg.

botocore/client.py
```diff
@@ -522,7 +530,7 @@ def _service_model(self):
         return self.meta.service_model
 
     def _make_api_call(self, operation_name, api_params):
-        request_context = {}
+        request_context = {'client_region': self.meta.region_name}
```
jamesls: It would be nice to have consistent casing for keys in this dict.  Looks like camel case is being used for the encoding type key.

botocore/utils.py
jamesls: Remind me again why this just can't be a plain old dict?  The operations you're exposing here are already thread safe for the built in dicts.  This class seems unnecessary. 

Where you seeing test failures?   Or can you reproduce any scenarios in which the code does not behave as expected without explicit locking?

JordonPhillips: @kyleknap had concerns about potential race conditions if we didn't lock down the the cache.

botocore/client.py
JordonPhillips: I would rather change that to be snake case than to make everything else camel case

botocore/endpoint.py
kyleknap: Alright. I'm fine with exposing request_dict then.

botocore/utils.py
kyleknap: Looking through the code again, I do not see any race conditions. We should be fine with using a plain dict. I think I forgot that since dict is a built-in it is thread safe in nature or there might have been something that you were doing before that was not safe, but right now it would be fine to remove the lock.

JordonPhillips:
Squashed commits and moved back to using a plain dict cache

botocore/awsrequest.py
```diff
@@ -289,6 +290,7 @@ def prepare_request_dict(request_dict, endpoint_url, user_agent=None):
         else:
             url += '&%s' % encoded_query_string
     r['url'] = url
+    r['context'] = context or {}
```
jamesls: Whenever you have a default argument of None, you should always use `if context is not None`.  Especially for dicts and lists which have meaningful false values.

botocore/client.py
jamesls: I don't have a strong preference.  I suggested camel case because I prefer not to change unrelated code, but it should be a safe change to make either way (I think).

botocore/utils.py
```diff
@@ -794,5 +802,122 @@ def _switch_hosts(request, new_endpoint, use_new_scheme=True):
...
-    request.url = final_endpoint
+    logger.debug('Updating URI from %s to %s' % (
+        original_endpoint, final_endpoint))
+    return final_endpoint
+
+
+class S3RegionRedirector(object):
+    def __init__(self, endpoint_bridge, client, cache=None):
+        self._endpoint_resolver = endpoint_bridge
+        self._cache = cache or {}
```
jamesls: `cache is not None`

kyleknap:
Code looks good. Once we can confirm the CLI s3 integration tests pass with it, I say we can merge it.

tests/unit/__init__.py
```diff
@@ -11,3 +11,31 @@
...
+
+        self.assertEqual(actual, expected)
+        self.assert_dict_is_proper_subset(actual_metadata, expected_metadata)
+
+    def assert_dict_is_proper_subset(self, superset, subset):
+        """
+        Asserts that a dictionary is a proper subset of another.
+        """
+        self.assertTrue(all((k in superset and superset[k] == v)
+                            for k, v in subset.items()))
```
jamesls: Missing newline?

jamesls:
Looks pretty good, just had some small feedback.

botocore/client.py
JordonPhillips: I'll just do that in a separate PR

JordonPhillips:
Updated

jamesls:
I think it looks good.  @JordonPhillips @kyleknap what's next?  Get a clean run of the CLI integ tests before merging?

I think it would be good to get this branch and the unsigned payload branch merged together, then run the CLI integ tests, then merge to develop.  Does that seem reasonable?

kyleknap:
Yep that was what I was thinking.

botocore/client.py
```diff
@@ -319,6 +319,12 @@ def resolve(self, service_name, region_name=None, endpoint_url=None,
             return self._assume_endpoint(service_name, region_name,
                                          endpoint_url, is_secure)
 
+    def resolve_default(self, service_name, is_secure=True):
```
kyleknap: Does it matter that you make a default here of ``True`` for ``is_secure`` when you actually do not use it in the ``redirect_create_bucket()``?

botocore/utils.py
```diff
@@ -925,12 +925,33 @@ def set_request_url(self, params, context, **kwargs):
...
+        bucket = params.get('Bucket')
+        self.redirect_from_cache(bucket, context)
+        if model.name == 'CreateBucket':
+            self.redirect_create_bucket(bucket, params, context)
+
+    def redirect_create_bucket(self, bucket, params, context):
+        configuration = params.get('CreateBucketConfiguration', {})
+        location = configuration.get('LocationConstraint', None)
+
+        if location is not None:
```
kyleknap: So do we know for certain that only ``us-east-1`` is the only region that allows for an empty ``LocationConstraint``? With the way you have it now, the default endpoint for ``aws`` will always be used. And so if you are in a different partition that had a default endpoint, it would not use that.

tests/integration/test_s3.py
```diff
@@ -1099,3 +1099,36 @@ def test_redirects_head_object(self):
...
+
+        waiter = self.client.get_waiter('bucket_exists')
+        waiter.wait(Bucket=bucket_name)
+
+        response = self.client.head_bucket(Bucket=bucket_name)
+        headers = response['ResponseMetadata']['HTTPHeaders']
+        region = headers.get('x-amz-bucket-region', None)
+        self.assertEqual(region, 'us-east-1')
+
+    def test_create_bucket_with_location_does_not_redirect(self):
```
kyleknap: Based on my reading of this, this test implies that you used to redirect even though you had the client region correct? How did that happen in the first place?

tests/unit/test_client.py
```diff
@@ -1378,6 +1378,34 @@ def test_uses_us_east_1_by_default_for_s3(self):
...
+            'endpointName': 'us-east-1', 'signatureVersions': ['s3', 's3v4'],
+            'protocols': ['https']}
+        bridge = ClientEndpointBridge(resolver)
+        resolved = bridge.resolve_default('s3')
+        self.assertEqual('us-east-1', resolved['region_name'])
+        self.assertEqual('us-east-1', resolved['signing_region'])
+        self.assertEqual('https://s3.amazonaws.com',
+                         resolved['endpoint_url'])
+
+    def test_resolve_defualt_unaffected_by_client_config(self):
```
kyleknap: defualt --> default

tests/integration/test_s3.py
JordonPhillips: I'm not entirely sure what you're saying. We redirect if location constraint isn't set. We do not redirect if it is set. In this test I just happen to use `self.client_region` because it is readily available.

kyleknap: Nevermind. I understand now, you were testing both possible cases.

botocore/client.py
JordonPhillips: Didn't seem like it needed to be taken out. I'm 0 on keeping it though.

botocore/utils.py
JordonPhillips: That I'm actually not sure about. I'm fairly certain it's only good for us-east-1.

JordonPhillips:
Updated to make sure us-east-1 is resolvable

JordonPhillips:
create bucket redirection may not have been a thing

JordonPhillips:
I'm going to go ahead and close this in favor of merging it all at once via #920

PR #937, opened 2016-06-03
Allow stubber to be used with GetBucketLocation

Jc2k:
Right now you can't use the `Stubber` object with s3's `get_bucket_location` because it has an override that assumes a body of xml is available. This patch allows the override to be skipped where something (in this case the stubber) has already 'parsed' the response.


JordonPhillips:
Looks good, but could you put in a quick unit test? Also, could you please use `scripts/new-change` to generate a changelog entry?

Jc2k:
@JordonPhillips done :-)

JordonPhillips:
⛵

PR #938, opened 2016-06-03
Redshift EnhancedVpcRouting response data

infame-io:
The PR enables to determine which redshift clusters are and which are not using enhanced VPC routing. Currently there is no way to check whether enhanced VPC routing is enabled when using ```describe_cluster```

I'll be adding the unittests soon


JordonPhillips:
We don't accept PRs to service JSON files, as we generate those upstream. (see [contributing guide](EnhancedVpcRouting))

PR #939, opened 2016-06-03
Allow paramters to be ignored in Stubber validation

### LINKED ISSUE #931: Feature request:  Add ability to ignore params in stubber
It would be nice if the stubber had a solution for being able to ignore randomly generated values in the input.  For example:

* idempotency tokens
* statement ids

And really anything that typically is a random value.  

There's a few ways I can think of to accomplish this.  In `mock`, you can use `mock.ANY` to indicate that you don't care about the value.   In the CLI we have an `ignore_params` for `assert_params_for_cmd`, though this only works for top level params (which typically is sufficient).

JordonPhillips:
This allows users to specify stub.ANY for expected parameters if
they don't care about a specific parameter's value.

Resolves #931

cc @kyleknap  @jamesls

### DIFF
--- a/botocore/stub.py
+++ b/botocore/stub.py
@@ -20,6 +20,9 @@ from botocore.exceptions import ParamValidationError, \
 from botocore.vendored.requests.models import Response
 
 
+ANY = object()
+
+
 class Stubber(object):
     """
     This class will allow you to stub out requests so you don't have to hit
@@ -96,6 +99,50 @@ class Stubber(object):
             service_response = s3.list_buckets()
 
         assert service_response == response
+
+
+    If you have an input paramter that is a randomly generated value, or you
+    otherwise don't care about its value, you can use stub.ANY to ignore it in
+    validation.
+
+    **Example:**
+    ::
+        import datetime
+        import botocore.session
+        from botocore.stub import Stubber, ANY
+
+
+        s3 = botocore.session.get_session().create_client('s3')
+        stubber = Stubber(s3)
+
+        response = {
+            'IsTruncated': False,
+            'Name': 'test-bucket',
+            'MaxKeys': 1000, 'Prefix': '',
+            'Contents': [{
+                'Key': 'test.txt',
+                'ETag': '"abc123"',
+                'StorageClass': 'STANDARD',
+                'LastModified': datetime.datetime(2016, 1, 20, 22, 9),
+                'Owner': {'ID': 'abc123', 'DisplayName': 'myname'},
+                'Size': 14814
+            }],
+            'EncodingType': 'url',
+            'ResponseMetadata': {
+                'RequestId': 'abc123',
+                'HTTPStatusCode': 200,
+                'HostId': 'abc123'
+            },
+            'Marker': ''
+        }
+
+        expected_params = {'Bucket': ANY}
+        stubber.add_response('list_objects', response, expected_params)
+
+        with stubber:
+            service_response = s3.list_objects(Bucket='test-bucket')
+
+        assert service_response == response
     """
     def __init__(self, client):
         """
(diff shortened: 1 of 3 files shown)


botocore/stub.py
```diff
@@ -265,7 +314,21 @@ def _get_response_handler(self, model, params, **kwargs):
...
         expected_params = self._queue[0]['expected_params']
-        if expected_params is not None and params != expected_params:
+        if expected_params is None:
+            return
+
+        # Validate the parameters are equal
+        for param, value in expected_params.items():
+            if value is ANY:
+                continue
+            elif expected_params[param] != params[param]:
```
kyleknap: I feel like there may be key errors here if the expected param is missing from the actual params.

botocore/stub.py
```diff
@@ -265,7 +314,21 @@ def _get_response_handler(self, model, params, **kwargs):
...
+                raise StubAssertionError(
+                    operation_name=model.name,
+                    reason='Expected parameters:\n%s,\nbut received:\n%s' % (
```
kyleknap: I wonder if we can make the error message much cleaner in telling what parameters are differing and how instead of having to compare the dictionaries because I actually find that a bit annoying.

kyleknap:
I like the idea. Allows me to remove mock completely from ``s3transfer``. But had a couple comments.

botocore/stub.py
```diff
@@ -20,6 +20,9 @@
 from botocore.vendored.requests.models import Response
 
 
+ANY = object
```
jamesls: `s/object/object()`.

botocore/stub.py
jamesls: I agree but I think that should be a separate PR right?

tests/functional/test_stub.py
```diff
@@ -148,3 +150,19 @@ def test_still_relies_on_param_validation_with_expected_params(self):
         # Throw an error for invalid parameters
         with self.assertRaises(ParamValidationError):
             self.client.list_objects(Buck='bar')
+
+    def test_any_ignores_param_for_validation(self):
```
jamesls: Can we add a test with multiple params?  One stub.ANY and one with an expected value and verify you're allowed to mix the two types in the same response call?

jamesls:
Just had some small feedback.

botocore/stub.py
kyleknap: Yeah probably more appropriate in separate PR.

JordonPhillips:
Updated

botocore/stub.py
```diff
@@ -265,7 +314,21 @@ def _get_response_handler(self, model, params, **kwargs):
...
         expected_params = self._queue[0]['expected_params']
-        if expected_params is not None and params != expected_params:
+        if expected_params is None:
+            return
+
+        # Validate the parameters are equal
+        for param, value in expected_params.items():
+            if value is ANY:
+                continue
+            elif expected_params[param] != params.get(param, None):
```
jamesls: I think it's useful to distinguish between explicitly set values versus missing values.  I'd expect something like this to trigger an assertion error:

`{"foo": None} == {}`

kyleknap: I had the same comment.

JordonPhillips:
updated again

kyleknap:
Thanks. :ship:

jamesls:
Where's the test that covers the last condition @kyleknap and myself asked you to address?  I'm not seeing it.

If you haven't added one, then as a reminder, you should be adding tests for all the various cases you're handling.

JordonPhillips:
Seemed at the time to be testing an implementation detail, but I realize now it really isn't. Updated.

jamesls:
:shipit:

PR #940, opened 2016-06-03
Convert boolean config values to boolean types

JordonPhillips:
All values read from the config file are imported as strings,
even if they look like other python data types. This will convert
strings that look like booleans to boolean types to avoid subtle
bugs since strings are truthy.

cc @kyleknap @jamesls

### DIFF
--- a/botocore/config.py
+++ b/botocore/config.py
@@ -93,12 +93,32 @@ class Config(object):
         config_vars = copy.copy(self.OPTION_DEFAULTS)
         config_vars.update(self._user_provided_options)
 
+        # All args come in as strings, so it is useful to convert them to other
+        # data types if necessary. i.e. to 'True' to True
+        config_vars = self._preprocess_config(config_vars)
+
         # Set the attributes based on the config_vars
         for key, value in config_vars.items():
             setattr(self, key, value)
 
         # Validate the s3 options
         self._validate_s3_configuration(self.s3)
+
+    def _preprocess_config(self, config_vars):
+        for key, value in config_vars.items():
+            if isinstance(value, str):
+                config_vars[key] = self._string_to_boolean(value)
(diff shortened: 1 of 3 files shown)


jamesls:
The type of a config var is based on the definition of that config var.  I don't think we should be making the assumption that string values of `true` should be converted to bools.  I'd rather see the types/schema be moved into the definition of the config vars (possible in session.py).

It's probably a safe assumption now that that `'true' -> True` is safe, but I'd rather have a more explicit solution.

I'd also try to enforce that anyone using `Config` classes directly should be passing the correct config var types when possible.

kyleknap:
Agreed. We should only be making the conversion if it is intended and the way to intend it is through a schema.

I also have logic for this [here](https://github.com/boto/botocore/blob/develop/botocore/client.py#L135) so it would be good to remove that as when you get the chance.

JordonPhillips:
The section in `Session.py` doesn't appear to be able to address nested variables. This is also not for session variables, but client variables. We could use similar logic in `configloader`, which seems like the most appropriate spot.

jamesls:
What is the specific bug you're trying to address?

JordonPhillips:
If we have a true/false config value, it comes in as a string. That could lead to bugs if it is assumed to be a boolean. It isn't obvious that it will come in as a string until you hit the issue.

jamesls:
I mean the *specific* bug you are trying to solve.  Is there a config option right now that is not being handled properly?  Do you have a specific config example where this is not being handled appropriately?

I'm thinking that unless there is a specific bug being addressed, we should defer this until later.  It's going to require a slightly more involved redesign anyways.

JordonPhillips:
For #920, I will have to try to figure out if its a true string or false string either way. I put up this PR because I would rather have that sort of thing generalized.

PR #851, opened 2016-03-23
Assume role via instance metadata credentials

pgcp:
Allows a client to assume a role by using the credentials from the EC2 instance profile. If a profile in the configuration has a `role_arn` but not a `source_profile` the credentials used for assuming the role will come from the instance metadata.

This pull request is to address [AWS CLI issue 1390](https://github.com/aws/aws-cli/issues/1390). The issue was linked to the AWS CLI repo since that's where the AssumeRoleProvider code was at the time the issue was created. The AssumeRoleProvider was moved into botocore after that, so I've updated the relevant botocore code.

PR #946, opened 2016-06-10
Add recursive role assumption.

wt:
This change will allow a profile to use a profile that uses a role as a
source profile. This will allow hierarchies of profiles where a user is
used to assume a role which is then used to assume another role. This is
useful for when another account grants cross account access and the role
that can assume the other role is assumable by users. Here's an example.

``` ~/.aws/credentials:
[default]
aws_access_key_id = <blah_id>
aws_secreet_access_key = <blah_secret>

~/.aws/config
[profile customer_accessor_role]
role_arn=arn:aws:iam:<my_acct_number>:role/customer_accessor_role
source_profile=default

[profile customer_role]
role_arn=arn:aws:iam:<customer_acct_number>:role/customer_role
source_profile=customer_accessor_role
external_id=<whatever>
```

Now I can do things like the following:

```
aws --profile customer_role s3 ls
```

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -727,8 +727,8 @@ class AssumeRoleProvider(CredentialProvider):
     # EXPIRY_WINDOW.
     EXPIRY_WINDOW_SECONDS = 60 * 15
 
-    def __init__(self, load_config, client_creator, cache, profile_name,
-                 prompter=getpass.getpass):
+    def __init__(self, load_config, client_creator, cache,
+                 profile_name, prompter=getpass.getpass):
         """
 
         :type load_config: callable
@@ -778,11 +778,31 @@ class AssumeRoleProvider(CredentialProvider):
     def load(self):
         self._loaded_config = self._load_config()
         if self._has_assume_role_config_vars():
-            return self._load_creds_via_assume_role()
+            if self._source_profile_has_keys():
+                return self._load_creds_via_assume_role()
+            else:
+                return self._load_creds_via_assume_role_from_indirect()
 
     def _has_assume_role_config_vars(self):
         profiles = self._loaded_config.get('profiles', {})
-        return self.ROLE_CONFIG_VAR in profiles.get(self._profile_name, {})
+        profile_config = profiles.get(self._profile_name, {})
+        if self.ROLE_CONFIG_VAR not in profile_config:
+            return False
+
+        if 'source_profile' not in profile_config:
+            msg = ('Profile (%s) has no source_profile defined.' %
+                   (self._profile_name,))
+            raise PartialCredentialsError(provider=self.METHOD,
+                                          cred_var=msg)
+
+        return True
+
+    def _source_profile_has_keys(self):
+        source_profile = (self._loaded_config['profiles']
+                          .get(self._profile_name, {})['source_profile'])
+        source_profile_config = (
+            self._loaded_config['profiles'].get(source_profile, {}))
+        return 'aws_access_key_id' in source_profile_config
 
     def _load_creds_via_assume_role(self):
         # We can get creds in one of two ways:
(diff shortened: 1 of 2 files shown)


wt:
Hi,

First of all, this isn't quite ready for pulling, but I have a questions regarding the implementation. How can I reliably pass the cache that is being used in the AssumeRoleProvider into the new resolver that I create with get_credentials? With that, the mfa that I input doesn't get cached to disk when using the aws cli tool since the aws cli tool has to override the cache to make it persist to disk.

Also, I am working on adding tests for this. Do I need to cover every test case that is currently covered for roles that use direct profiles (that is ones that have a access_key, a secret_key, and possibly a token)?

Thanks,
wt

wt:
Hi,

I am working on adding tests for this. Do I need to cover every test case that is currently covered for roles that use direct profiles (that is ones that have a access_key, a secret_key, and possibly a token)?

Thanks,
wt

wt:
At this point, this change passes all the old tests.

wt:
Okay, I've now added tests that cover all new code paths. I found that there was an unnecessary exception being raised since I needed the same check in an ealier location in the code, so I just removed it. The old tests are completely unchanged. I copied and modified some old tests to test the functionality specific to the indirect role resolution.

I think this change is not ready for merging if you are comfortable with it.

jamesls:
Thanks for the pull request.  The one thing that I want to think about is whether or not there's some way to unify the idea of pulling credentials from somewhere other than hard coded creds from a `source_profile`.  There's another related PR (https://github.com/boto/botocore/pull/851) that has something similar.  They want to be able to pull creds from the instance metadata service instead of from assuming a role.

I think there's probably a way to generalize the concept of pulling creds for the initial assume role call to STS.

Do you have any thoughts on that?

wt:
I actually do have some thoughts on that, but my thoughts would require changing the interface for get_credentials and create_credential_resolver. I think it'd be backward compatible. My basic idea is to add a named arg for a dict that contains all the args used with all the providers. Then, pass that dict and the session (as named args) to all providers during initialization. Then, I would be able to call get_credentials with the session and that dict to pass their values into a new credential resolver for each level of resolution. It would probably be good to have some way to prevent loops in the resolution by keeping a set of already resolved names and passing those in as well. The create_credential_resolver and get_credentials function would end up with this definition:

```
def create_credential_resolver(
        session, provider_config=None,  **kwargs):
    if provider_config is None:
        provider_config = set()
    if 'profile_name' in kwargs:
       profile_name = session.get_config_variable('profile') or 'default'
    ...and so on

def get_credential(
        session, provider_config=None, **kwargs):
    ...
```

Then I could pass the values into the get_credentials for any providers that needed to do a recursive resolution instead of doing something like constructing a new AssumeRoleProvider explicitly like I do in  my implementation of the recursive resolution now.

Having said all this, I feel like this change is a step between here and there. I think it's going to take a lot of fiddling to get the new interface right considering that create_credential_resolver current resolves a lot of magic variables. I think that logic may need to be refactored. I'd like to see this change go in as it demonstrably improves the status quo by allowing recursive role assumption when the leaf identity is an AWS user. I think this is a clear improvement that doesn't impede a more general solution like I described above, and I'd hate to gate this functionality on that functionality.

wt:
Penny for your thoughts?

wt:
*ping*

This functionality is sorely needed so that using roles is easier. Please consider not blocking this solution since it isn't perfect. It really is helping me to use the boto3 lib and aws-cli when using assume role functionality.

wt:
Do you have any feedback for changes that would make this change acceptable? It would be really helpful to get some clear feedback if there is something specific lacking in this change. I'd be happy to address it if I knew what is lacking.

PR #947, opened 2016-06-10
Fix regression in stubber

kyleknap:
Sometimes the list returned from dict.keys() will not be in the same order
which will cause the equality of keys check to fail:
```
    >>> ['key1', 'key2'] == ['key2', 'key1']
    False
```
To fix it, the keys need to be sorted.

Noticed it because of failing tests here: https://github.com/boto/s3transfer/pull/15

cc @jamesls @JordonPhillips


JordonPhillips:
⛵

PR #953, opened 2016-06-16
Replace chars we can't decode with placeholder char

jamesls:
It's possible that the console output from EC2 contains
non string chars we can't utf-8 decode.  When that happens
we were previously just leaving the input untouched, i.e
base64 encoded.

This made for an awkward interface for customers.  We couldn't guarantee
whether or not the content was base64 decoded because it would depend
on the actual ec2 console output content.

This change attempts to make things more consistent by
replacing any chars we can't decode with the unicode
`\ufffd` char (the unicode replace char, i.e the question mark).

The one downside with this approach is that it is possible
to sometimes not get the unaltered contents of the console
output because chars have been replaced with `\ufffd`.
We'll need to weigh whether or not we think that's worth
the benefit.  Given this content is usually text content,
this seems like a reasonable compromise.

cc @kyleknap @JordonPhillips

### DIFF
--- a/.changes/next-release/bugfix-EC2-81604.json
+++ b/.changes/next-release/bugfix-EC2-81604.json
@@ -1,0 +1,5 @@
+{
+  "category": "EC2", 
+  "type": "bugfix", 
+  "description": "Replace chars in the EC2 console output we can't decode with replacement chars.  We were previously returning either the decoded content or the original base64 encoded content.  We now will consistently return decoded output, any any chars we can't decode are substituted with a replacement char. (`#953 <https://github.com/boto/botocore/issues/953>`__)"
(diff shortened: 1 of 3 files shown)


kyleknap:
:ship: The only backwards incompatibility I see is if a user consistently got base64 encoded data back they would have base64 decoded it, but that seems very unlikely since it is a console output. It would also be pretty difficult and a weird workflow to accurately differentiate between base64 encoded and utf-8 encoded.

jamesls:
I agree.  I'll add a little more info in the changelog description for this change to explain this.  I do think that not having a consistent return type is reasonably in the "bug fix" bucket.

PR #958, opened 2016-06-21
Use SigV4 by default for S3

JordonPhillips:
All S3 requests will now use SigV4 by default so that we can take advantage of stronger security algorithms and improved access key isolation.

cc @kyleknap @jamesls


jamesls:
Flip in incorporate feedback pending failing test being fixed.

botocore/client.py
```diff
@@ -460,6 +499,27 @@ def _resolve_signature_version(self, service_name, resolved):
             signature_version=resolved.get('signatureVersions'))
 
 
+def _get_configured_signature_version(service_name, client_config,
```
jamesls: Why is this an internal module level function?  Doesn't seem to be used outside the client creator class.  I'd suggest moving it into the class unless it needs to be module level.

botocore/client.py
```diff
@@ -178,6 +178,15 @@ def _create_s3_redirector(self, client, endpoint_bridge):
             return
         S3RegionRedirector(endpoint_bridge, client).register()
 
+    def _get_default_s3_region(self, service_name, endpoint_bridge):
+        # If a user is providing a custom URL, the endpoint resolver will
+        # refuse to infer a signing region. If we want to default to s3v4,
```
jamesls: Why is this?  Is this intentional?  What's the point of not inferring a signing name if you provide an endpoint_url, but inferring a region if you don't provide an endpoint_url?

botocore/client.py
JordonPhillips: It is used in the client creator and the endpoint bridge.

botocore/client.py
JordonPhillips: It is intentional: https://github.com/boto/botocore/blob/develop/botocore/client.py#L409

jamesls: Ah right, I completely forgot about that sts edge case.

botocore/client.py
```diff
@@ -206,8 +215,14 @@ def _get_client_args(self, service_model, region_name, is_secure,
             if client_config.user_agent_extra is not None:
                 user_agent += ' %s' % client_config.user_agent_extra
 
-        signer = RequestSigner(
-            service_name, endpoint_config['signing_region'],
+        signing_region = endpoint_config['signing_region']
+        endpoint_region_name = endpoint_config['region_name']
+        if signing_region is None and endpoint_region_name is None:
+            signing_region, endpoint_region_name = \
+                self._get_default_s3_region(service_name, endpoint_bridge)
```
jamesls: `_get_default_s3_region` can sometimes return None, but this function is always assuming a 2 element tuple is being returned.

botocore/client.py
```diff
@@ -248,6 +267,37 @@ def _get_client_args(self, service_model, region_name, is_secure,
...
+        return signer_class(service_name, region_name, signing_name,
+                            signature_version, credentials, event_emitter)
+
+    def _set_s3_presign_signature_verion(
+            self, service_name, endpoint_resolver, client_config,
+            scoped_config, request_signer, region_name, endpoint_metadata):
+        if service_name != 's3':
+            return
+
+        provided_signature_version = _get_configured_signature_version(
```
kyleknap: Can we get comments for each of these blocks.

botocore/client.py
```diff
@@ -460,6 +495,27 @@ def _resolve_signature_version(self, service_name, resolved):
...
+def _get_configured_signature_version(service_name, client_config,
+                                      scoped_config):
+    # Client config overrides everything.
+    client = client_config
+    if client and client.signature_version is not None:
+        return client.signature_version
+    # Scoped config overrides picking from the endpoint metadata.
+    scoped = scoped_config
+    if scoped is not None:
+        service_config = scoped.get(service_name)
```
kyleknap: Comments would be nice here.

JordonPhillips:
Updated

tests/functional/test_s3.py
```diff
@@ -423,3 +423,21 @@ def test_region_redirect_cache(self):
                      '?encoding-type=url')
         self.assertEqual(calls[1].url, fixed_url)
         self.assertEqual(calls[2].url, fixed_url)
+
+
+class TestPresign(BaseS3OperationTest):
```
kyleknap: We should also check these cases for presigned posts as they are a completely different auth identifier.

botocore/client.py
```diff
@@ -206,8 +215,14 @@ def _get_client_args(self, service_model, region_name, is_secure,
             if client_config.user_agent_extra is not None:
                 user_agent += ' %s' % client_config.user_agent_extra
 
-        signer = RequestSigner(
-            service_name, endpoint_config['signing_region'],
+        signing_region = endpoint_config['signing_region']
+        endpoint_region_name = endpoint_config['region_name']
+        if signing_region is None and endpoint_region_name is None:
```
kyleknap: May be good to make a comment here as well. It is really weird that we are checking for no signing_region and no endpoint_region_name but then making an explicit call to ``_get_default_s3_region``

JordonPhillips:
To capture some offline discussion: we decided to use an event strategy to set the signer. To simplify things here, I opened a new PR to address that: #964 

I'll rebase on top of that once it gets through.

JordonPhillips:
rebased on top of #964 to get the ball rolling

jamesls:
Flipping to incorporating-feedback pending travis build investigation.

JordonPhillips:
I'm unable to reproduce that error locally...

JordonPhillips:
I got it. See #969 for more details.

kyleknap:
This looks good to me. :ship:. Let's merge it once we have a run of the CLI integration tests and some stats on how much of an affect this will affect the ``s3`` commands.

botocore/client.py
```diff
@@ -248,6 +268,37 @@ def _get_client_args(self, service_model, region_name, is_secure,
...
+        # version, we always want to use what they set.
+        provided_signature_version = _get_configured_signature_version(
+            service_name, client_config, scoped_config)
+        if provided_signature_version is not None:
+            return
+
+        # Check to see if the region is a region that we know about. If we
+        # don't know about a region, then we can safely assume it's a new
+        # region that is sigv4 only, since all new S3 regions only allow sigv4.
+        regions = endpoint_resolver.get_available_endpoints(
```
jamesls: We should take a follow up action item to remove this.  This should live in the `endpoints.json` file.  I think it would be a reasonable change to have the default values be s3v4 in the endpoints file and call out all the existing regions that support `s3` explicitly.

botocore/utils.py
```diff
@@ -632,6 +632,15 @@ def check_dns_name(bucket_name):
     return True
 
 
+def switch_to_s3_sigv2_presigner(signature_version, signing_name, **kwargs):
```
jamesls: docstring please.

jamesls:
Seems reasonable to me.  Pending @kyleknap 's items, :shipit:

botocore/utils.py
jamesls: By docstring, I meant a full docstring, params, return values etc.

Also, "switching" implies some sort of mutation/side effect.  Here all you're doing is figuring out what the signature version name should be for presigning a request and returning that value.  Might be better to have the name reflect that.

JordonPhillips:
For performance, the difference is negligible for large requests and large numbers of requests. For an individual typical request, the difference is ~4ms and for the worst case it's ~15ms.

botocore/utils.py
JordonPhillips: That's not strictly true either. I only return if I think it should be switched.

JordonPhillips:
Updated

kyleknap:
Thanks for the stats. Let's go ahead and merge it as most use cases do not see a significant dip in performance.

PR #961, opened 2016-06-22
Make output JSON serializable

bashtoni:
92568700ad1bd560d3f92213b52a6b11a1647d13 breaks assumed roles in AWS CLI because it adds HTTP headers in a `botocore.vendored.requests.structures.CaseInsensitiveDict`.  AWS CLI caches details to disk by serializing to JSON.

Resolve this by coercing into a normal dict.


JordonPhillips:
Could you please add a test for this?

jamesls:
I'll go ahead and take this and add a test.  I'd like to get this fixed asap so I'll add a test.

jamesls:
Pulled in this PR and added a test.  Let's track the remaining work in https://github.com/boto/botocore/pull/962

PR #962, opened 2016-06-22
Fix regression in assume role caching

jamesls:
This pulls in https://github.com/boto/botocore/pull/961 and adds tests for the change.

Right now you get an error:
```
$ aws ec2 describe-instances --profile some-role
Value cannot be cached, must be JSON serializable: {u'AssumedRole...
```

This was from a recent addition to botocore that added the HTTP headers into the response metadata returned from client calls.

I agree with the original change from @bashtoni to fix this in the parser.  We've had an implicit contract that you can JSON dump response metadata so while we could fix this in the assume role cacher, it's possible customer code could be relying on this property, in which case we would not be able to fix.

cc @kyleknap @JordonPhillips

### DIFF
--- a/.changes/next-release/bugfix-AssumeRole-9717.json
+++ b/.changes/next-release/bugfix-AssumeRole-9717.json
@@ -1,0 +1,5 @@
+{
+  "category": "AssumeRole", 
+  "type": "bugfix", 
+  "description": "Fix regression introduced in `#920 <https://github.com/boto/botocore/issues/920>`__ where assume role responses error out when attempting to cache a response. (`#961 <https://github.com/boto/botocore/issues/961>`__)"
+}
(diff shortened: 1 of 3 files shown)


JordonPhillips:
```
--    .-""-.
   ) (     )
  (   )   (
     /     )
    (_    _)                     0_,-.__
      (_  )_                     |_.-._/
       (    )                    |_--..\
        (__)                     |__--_/
     |''   ``\                   |
     |        \                  |      /b.
     |         \  ,,,---===?A`\  |  ,==y'
   ___,,,,,---==""\        |M] \ | ;|\ |>
           _   _   \   ___,|H,,---==""""bno,
    o  O  (_) (_)   \ /          _     AWAW/
                     /         _(+)_  dMM/
      \@_,,,,,,---=="   \      \\|//  MW/
--''''"                         ===  d/
                                    //
                                    ,'__________________________
   \    \    \     \               ,/~~~~~~~~~~~~~~~~~~~~~~~~~~~
                         _____    ,'  ~~~   .-""-.~~~~~~  .-""-.
      .-""-.           ///==---   /`-._ ..-'      -.__..-'
            `-.__..-' =====\\\\\\ V/  .---\.
                      ~~~~~~~~~~~~, _',--/_.\  .-""-.
                            .-""-.___` --  \|         -.__..-
```

kyleknap:
:ship:

PR #964, opened 2016-06-23
Add choose-signer event to presigners

JordonPhillips:
This adds the choose-signer event to the presigers, allowing
for hot swapping the signature version when presigning.
Additionally, adds support for using botocore.UNSIGNED with the
presigners.

cc @kyleknap @jamesls

### DIFF
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -89,7 +89,8 @@ class RequestSigner(object):
         # Don't call this method directly.
         return self.sign(operation_name, request)
 
-    def sign(self, operation_name, request):
+    def sign(self, operation_name, request, region_name=None,
+             signing_type='standard', expires_in=None):
         """Sign a request before it goes out over the wire.
 
         :type operation_name: string
@@ -97,18 +98,26 @@ class RequestSigner(object):
                                ``ListBuckets``.
         :type request: AWSRequest
         :param request: The request object to be sent over the wire.
-        """
-        signature_version = self._signature_version
 
-        # Allow overriding signature version. A response of a blank
-        # string means no signing is performed. A response of ``None``
-        # means that the default signing method is used.
-        handler, response = self._event_emitter.emit_until_response(
-            'choose-signer.{0}.{1}'.format(self._service_name, operation_name),
-            signing_name=self._signing_name, region_name=self._region_name,
-            signature_version=signature_version)
-        if response is not None:
-            signature_version = response
+        :type region_name: str
+        :param region_name: The region to sign the request for.
+
+        :type signing_type: str
+        :param signing_type: The type of signing to perform. This can be one of
+            three possible values:
+
+            * 'standard'     - This should be used for most requests.
+            * 'presign-url'  - This should be used when pre-signing a request.
+            * 'presign-post' - This should be used when pre-signing an S3 post.
+
+        :type expires_in: int
+        :param expires_in: The number of seconds the presigned url is valid
+            for. This parameter is only valid for signing type 'presign-url'.
+        """
+        if region_name is None:
+            region_name = self._region_name
+
+        signature_version = self._choose_signer(operation_name, signing_type)
 
         # Allow mutating request before signing
         self._event_emitter.emit(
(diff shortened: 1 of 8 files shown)


botocore/signers.py
```diff
@@ -188,31 +195,49 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
-            signature_version += query_prefix
+        if signature_version is not botocore.UNSIGNED and not \
+                signature_version.endswith(query_suffix):
+            signature_version += query_suffix
 
-        kwargs = {'signing_name': self._signing_name,
-                  'region_name': region_name,
-                  'signature_version': signature_version,
-                  'expires': expires_in}
+        handler, response = self._event_emitter.emit_until_response(
```
kyleknap: I think we should consolidate the event emission into one helper function to make sure that if we update the event at all we account for all places.

botocore/signers.py
```diff
@@ -188,31 +195,49 @@ def generate_presigned_url(self, request_dict, expires_in=3600,
...
+            signature_type = signature_version.split('-', 1)[0]
 
         request = create_request_object(request_dict)
 
         # Fix s3 host for s3 sigv2 bucket names
         fix_s3_host(request, signature_type, region_name)
 
-        auth.add_auth(request)
-        request.prepare()
+        if signature_type is not botocore.UNSIGNED:
```
kyleknap: What are your thoughts on consolidating ``signature_type`` and ``signature_version`` and just use ``signature_version``? Looks like ``signature_type`` was only used for error messaging. This would help reduce all of the if else statements throughout the code.

botocore/signers.py
```diff
@@ -411,29 +436,48 @@ def generate_presigned_post(self, request_dict, fields=None,
...
-        signature_type = signature_version.split('-', 1)[0]
+        if response is not None:
+            signature_version = response
 
-        try:
-            auth = self._request_signer.get_auth_instance(**kwargs)
-        except UnknownSignatureVersionError:
-            raise UnsupportedSignatureVersionError(
-                signature_version=signature_type)
+        if signature_version is botocore.UNSIGNED:
```
kyleknap: I feel like this a lot of just copying and pasting from generate_presigned_url. Can we consolidate this as well into one helper method?

tests/unit/test_signers.py
```diff
@@ -401,6 +469,38 @@ def test_generate_presigned_post(self):
...
+                             {'s3v4-presign-post': self.auth}):
+            self.signer.generate_presigned_post(self.request_dict)
+        self.emitter.emit_until_response.assert_called_with(
+            'choose-signer.s3.PutObject',
+            signing_name='signing_name', region_name='region_name',
+            signature_version='s3v4-presign-post')
+
+    def test_generate_presigned_post_choose_signer_override(self):
+        auth = mock.Mock()
+        self.emitter.emit_until_response.return_value = (None, 'custom')
```
kyleknap: Should we add test cases for when you return a ``s3v4-presign-post`` or ``s3-query``?

kyleknap:
Looks good. Had some refactoring comments and comment about adding more tests. It would also be good to add a functional for unauthenticated posts and urls. I know we have integration tests for those, but we do not need those for the unauthenticated case.

botocore/signers.py
JordonPhillips: It is used in `fix_s3_host`, but we can alter the logic there account for it.

botocore/signers.py
```diff
@@ -127,6 +117,36 @@ def sign(self, operation_name, request):
                                             signature_version)
             signer.add_auth(request=request)
 
+    def choose_signer(self, operation_name, suffix=''):
```
kyleknap: I was first hesitant about exposing this as a public method. But it looks like the ``generate_presigned_post`` uses it, and I think it is a reasonable workflow for people want to use the signer to implement their own auth to call:

1) call choose_signer
2) call get_auth_instance with the retrieved signature version
3) sign the request with the auth_instance

So I am fine with adding this.

botocore/utils.py
```diff
@@ -645,7 +646,8 @@ def fix_s3_host(request, signature_version, region_name, **kwargs):
     """
     # By default we do not use virtual hosted style addressing when
     # signed with signature version 4.
-    if signature_version in ['s3v4', 'v4']:
+    if signature_version is not botocore.UNSIGNED and \
+            's3v4' in signature_version:
```
kyleknap: What about ``'v4'``? How is that handled now?

JordonPhillips: `v4

kyleknap: Ah the handlers protect it: https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L820. Yeah that's fine then.

tests/unit/test_signers.py
```diff
@@ -401,6 +469,49 @@ def test_generate_presigned_post(self):
...
+            's3v4-presign-post': self.auth,
+            'custom-presign-post': auth
+        }
+        with mock.patch.dict(botocore.auth.AUTH_TYPE_MAPS, auth_types):
+            self.signer.generate_presigned_post(self.request_dict)
+        auth.assert_called_with(
+            credentials=self.fixed_credentials, region_name='region_name',
+            service_name='signing_name')
+
+    def test_generate_presigned_post_bad_signer_raises_error(self):
```
kyleknap: You covered the not happy case. I was thinking what if the user returned s3v4-presign-post for the presign_post, I would expect it would properly do the presign post without any issues. You covered the case of mismatching for presign posts and urls.

kyleknap:
I think it looks good. Just one more test I think we should add.

botocore/signers.py
```diff
@@ -123,6 +113,36 @@ def sign(self, operation_name, request):
...
+        """
+        Allow overriding the signature version via the choose-signer event.
+        A value of `botocore.UNSIGNED` means no signing will be performed.
+
+        :param operation_name: The operation to sign.
+        :param suffix: A suffix that needs to be on all valid signers. This
```
jamesls: This arg doesn't seem to make sense as a public interface.  First, suffix is completely implementation specific on how we decided to name signature versions.  Secondly, seems like it should be the callers responsibility to deal with interpreting signature versions if they care.

botocore/signers.py
```diff
@@ -123,6 +113,36 @@ def sign(self, operation_name, request):
                                             signature_version)
             signer.add_auth(request=request)
 
+    def choose_signer(self, operation_name, suffix=''):
+        """
+        Allow overriding the signature version via the choose-signer event.
+        A value of `botocore.UNSIGNED` means no signing will be performed.
```
jamesls: This docstring doesn't seem right.  We're not really "overriding" signature versions.  We're simply giving it an operation, and asking what signature version the request signer would use to sign the request.

botocore/signers.py
```diff
@@ -410,33 +429,29 @@ def generate_presigned_post(self, request_dict, fields=None,
...
+        signature_version = self._request_signer.choose_signer(
+            'PutObject', '-presign-post')
 
         # Store the policy and the fields in the request for signing
         request = create_request_object(request_dict)
         request.context['s3-presign-post-fields'] = fields
         request.context['s3-presign-post-policy'] = policy
 
-        auth.add_auth(request)
+        if signature_version is not botocore.UNSIGNED:
```
jamesls: This code seems like copy/paste from lines 220.  Why not fold this logic into the request signer itself, rather than exposing choose signer and forcing callers to all have similar logic?

botocore/signers.py
JordonPhillips: That brings us right back to the problem of having copied code all over the place.

jamesls: Public in this context as being called as an API within our botocore code (non `_internal`).  I think there's an increasingly higher bar for sensible APIs going from `_internal` methods used within a class, non-internal methods used amongst botocore classes, and methods exposed for customers to call.

I made another comment below about an alternate approach you could use.

JordonPhillips: We can't roll everything together without exposing suffix in some way or having duplicate code. The suffix needs to be added before and after the call to `choose-signer`.

JordonPhillips: Now that I think about it, we could make it a private module level function.

JordonPhillips:
Updated. One thing I know you're going to ask about is 'where is fix_s3_host'. The answer is that since we're going through `sign`, `before-sign` will be called, and thus `fix_s3_host` will be called from the standard handler.

jamesls:
:shipit:  Nice, I think this looks much better.

PR #968, opened 2016-06-23
Save operation_name on ClientError exception object

Bazze:
The operation name could be useful when working with exceptions raised by boto, currently you'll have to match on the error message which is far from optimal.


jamesls:
Seems pretty reasonable to me.  We just need to add a quick test for this and we can merge.

Bazze:
I added two tests. Seems OK? :)

kyleknap:
:ship: Looks good to me.

cc @jamesls @JordonPhillips for additional review

jamesls:
:shipit:

PR #969, opened 2016-06-24
Use weakref for client in S3 redirector

JordonPhillips:
The S3RegionRedirector holds a reference to the client, which was
causing memory leaks on python 2.6. We didn't notice this previously
because our tests give us a certain tolerance and we were still
falling under that tolerance. If more events are added to the
event emitter by default, it can cause the amount of memory leaked
to go past our tolerance.

This fixes the issue by making the S3 redirector hold a weakref to
the client instead of a strong reference. We don't have to worry
about it being GC'd because customers will be holding a strong ref
to the client for us.

cc @jamesls @kyleknap

### DIFF
--- a/.changes/next-release/bugfix-S3-4368.json
+++ b/.changes/next-release/bugfix-S3-4368.json
@@ -1,0 +1,5 @@
+{
+  "category": "S3",
+  "description": "Fixed a bug where the S3 region redirector was potentially causing a memory leak on python 2.6.",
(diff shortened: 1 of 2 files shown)


jamesls:
:shipit:

I'm thinking maybe we should start gradually lowering the max threshold to ensure that these issues get caught with the associated change.

JordonPhillips:
@jamesls agreed

PR #970, opened 2016-06-28
Populate location constraint in integration tests

JordonPhillips:
When switching to sigv4 by default, I missed this S3 test in the client
integration tests. I did a search of the other integration tests to
look for create bucket, and only found one in test_elastictranscoder
which was configured for us-east-1 and thus does not need that value
to be set.

cc @jamesls @phjordon

### DIFF
--- a/tests/integration/test_client.py
+++ b/tests/integration/test_client.py
@@ -41,7 +41,12 @@ class TestBucketWithVersions(unittest.TestCase):
         # 1. Create a bucket
         # 2. Enable versioning
         # 3. Put an Object
-        self.client.create_bucket(Bucket=self.bucket_name)
+        self.client.create_bucket(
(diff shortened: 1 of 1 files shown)


kyleknap:
:ship:

jamesls:
Can you explain why the existing integration test doesn't work with these changes?

If we have to change integ tests wouldn't a customer have to change their code as well?

I'm trying to understand why this isn't considered a back-compat change.

JordonPhillips:
Because you must specify the location constraint when you aren't sending a request to `s3.amazonaws.com`. We discussed this before, and the conclusion was that it was undesired behavior to create a bucket in region that you aren't configured for.

jamesls:
Ok, sounds good to me, :shipit:

PR #971, opened 2016-06-28
Force virtual host style for thread safe auth test

JordonPhillips:
The thread safe auth test depends on `auth_path`, which is only
set when using virtual host style addressing. Since sigv4 does
not use virtual host style by default, the test broke. This
configures the client to use virutal host style for that test.

cc @kyleknap @jamesls


jamesls:
:shipit:

PR #975, opened 2016-07-06
allow bytearray in calculating md5 from bytes

foobarna:
Passing `bytearray` to S3 multipart upload fails when calculating md5. Having the below code, it fails w/o the modification.
https://gist.github.com/foobarna/913a84136a4357bbdc4112c924c0c41e

https://docs.python.org/2/whatsnew/2.6.html#pep-3112-byte-literals


JordonPhillips:
Could you please add a test to verify this change?

foobarna:
test added

foobarna:
@JordonPhillips bump

jamesls:
:shipit:  Looks good to me.

PR #980, opened 2016-07-20
Bug fix for _get_response receives 'bytes' instead of  text

gu1234:
Fixing a bug that when requests module returns a byte response instead of text _get_response crashes
requests module can grantee a text response and handles the encoding when using the text method


gu1234:
Fixes this issue:
https://github.com/boto/boto3/issues/718

jamesls:
Fixed via https://github.com/boto/botocore/pull/982

PR #982, opened 2016-07-21
ECS: use 'response.decode' to convert bytes -> string on python3

aclowes:
Use 'response.text' for more flexible decoding on error; this uses errors='replace'
and tries to auto-detect encoding. Makes sense since an error may not be
generated by AWS and might not always be UTF-8.

This is a bug in how the ECS ContainerMetadataFetcher gets the credentials when running inside a task.

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -980,13 +980,13 @@ class ContainerMetadataFetcher(object):
             if response.status_code != 200:
                 raise MetadataRetrievalError(
                     error_msg="Received non 200 response (%s) from ECS metadata: %s"
-                    % (response.status_code, response.content))
+                    % (response.status_code, response.text))
             try:
-                return json.loads(response.content)
+                return json.loads(response.content.decode('utf-8'))
             except ValueError:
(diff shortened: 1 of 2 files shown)


botocore/utils.py
```diff
@@ -980,13 +980,13 @@ def _get_response(self, full_url, headers, timeout):
             if response.status_code != 200:
                 raise MetadataRetrievalError(
                     error_msg="Received non 200 response (%s) from ECS metadata: %s"
-                    % (response.status_code, response.content))
+                    % (response.status_code, response.text))
             try:
-                return json.loads(response.content)
+                return json.loads(response.content.decode('utf-8'))
```
jamesls: Given your rationale for using `response.text` in the PR description, seems like using `response.text` here also would make sense.

aclowes: Yes, either would be correct, or go straight to `response.json()`. I didn't because:
- we know that AWS is using utf-8, so we don't need to guess
- other parts of the code use `response.content.decode` instead of `response.text`

Thanks for merging.

PR #977, opened 2016-07-09
add CodeDeploy waiters for deployments

quiver:
This PR implements waiters for CodeDeploy' s most basic deployment flows.
 

Deployment state as a whole can take Created/Queued/InProgress/Succeeded/Failed/Stopped.
http://docs.aws.amazon.com/codedeploy/latest/APIReference/API_DeploymentInfo.html#CodeDeploy-Type-DeploymentInfo-deploymentOverview

This waiter waits until deployments Succeeded/Failed/Stopped..

codedeploy-waiter-flow

Closes boto/boto3#708

## Usage
```
$ aws deploy create-deployment \
  --application-name foo \
  --deployment-config-name CodeDeployDefault.OneAtATime \
  --deployment-group-name bar \
  --description "baz" \
  --s3-location bucket=YOUR-BUCKET,bundleType=zip,key=YOUR/KEY.zip
{
    "deploymentId": "d-123456789"
}

$ aws deploy wait deployment-successful --deployment-id d-123456789
```

## API names

OpsWorks has wait/deployment-successful API, so I followed this naming conventions.
http://docs.aws.amazon.com/cli/latest/reference/opsworks/wait/deployment-successful.html

PR #983, opened 2016-07-26
Add changelog entry for DeploymentSuccessful watier

JordonPhillips:
This is a changelog entry for #977

cc @kyleknap @jamesls @chrisradek


chrisradek:
:shipit:

PR #984, opened 2016-07-27
Fixes #976 by keeping old netloc field in new endpoint

### LINKED ISSUE #976: Region redirect does not honor scheme
Bucket in us-west-2
```
$ aws s3api get-bucket-location --bucket jamesls-test-sync
{
    "LocationConstraint": "us-west-2"
}
```

Specifying an http endpoint to us standard instead of us-west-2:

```
aws s3 cp /tmp/foo.txt s3://jamesls-test-sync/ --endpoint-url http://s3.amazonaws.com/ --debug
```

I can see we have the redirect logic now, but I can see it getting redirected from http to https:

```
2016-07-08 08:57:31,621 - Updating URI from http://s3.amazonaws.com/jamesls-test-sync to https://s3-us-west-2.amazonaws.com/jamesls-test-sync
```

I would expect that if we're going to change the URl we should still honor the scheme they've requested right?

cc @JordonPhillips

srri:
Fixes https://github.com/boto/botocore/issues/976
The old `netloc` field was not being used when it should be:
```
>>> urlparse.urlsplit('http://s3.amazonaws.com/jamesls-test-sync')
SplitResult(scheme='http', netloc='s3.amazonaws.com', path='/jamesls-test-sync', query='', fragment='')
```
This will allow the scheme to be upgraded but won't change the endpoint location
@jamesls @kyleknap

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -799,7 +799,7 @@ def _get_new_endpoint(original_endpoint, new_endpoint, use_new_scheme=True):
         scheme = new_endpoint_components.scheme
     final_endpoint_components = (
         scheme,
-        new_endpoint_components.netloc,
+        original_endpoint_components.netloc,
         original_endpoint_components.path,
         original_endpoint_components.query,
         ''


JordonPhillips:
We *want* the new netloc, the issue is with the scheme.

PR #985, opened 2016-07-27
Honor old scheme when redirecting S3 requests

### LINKED ISSUE #976: Region redirect does not honor scheme

JordonPhillips:
If the old URL was using http over https, we will no continue to
use http.

Fixes #976
Closes #984

cc @kyleknap @jamesls

### DIFF
--- a/.changes/next-release/bugfix-s3-36114.json
+++ b/.changes/next-release/bugfix-s3-36114.json
@@ -1,0 +1,5 @@
+{
+  "description": "S3 region redirector will now honor the orginial url scheme.",
+  "category": "s3",
+  "type": "bugfix"
+}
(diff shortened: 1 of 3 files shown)


kyleknap:
:ship:

PR #988, opened 2016-07-28
Make s3 integ test more resilient

jamesls:
Use bucket_client instead of self.client.
Also, use a waiter to ensure the bucket exists.  There are times
when this test can fail with a bucket does not exist.

cc @kyleknap @JordonPhillips


JordonPhillips:
⛵

PR #987, opened 2016-07-28
Check if access_key_id is None in session.create_client

srri:
Fixes https://github.com/boto/boto3/issues/686

PR #989, opened 2016-07-28
Raise error on partial creds provided

jamesls:
This builds on #987 to make this condition an error case.
A user providing just an access key or secret key is never
valid and should result in an error.

This is also consistent with other cred providers that raise
PartialCredentialsError for partial creds.

https://github.com/boto/boto3/issues/686

cc @kyleknap @JordonPhillips


kyleknap:
:ship: assuming tests pass.

srri:
:shipit:

PR #990, opened 2016-07-29
Properly handle boolean query string arguments

srri:
Fixes https://github.com/aws/aws-cli/issues/2030
Read there for background. Also let me know if you think we need a test for this
cc: @JordonPhillips

### DIFF
--- a/.changes/next-release/bugfix-Serialization-71811.json
+++ b/.changes/next-release/bugfix-Serialization-71811.json
@@ -1,0 +1,5 @@
+{
+  "category": "Serialization", 
+  "type": "bugfix", 
+  "description": "Account for boolean in query string serialization"
+}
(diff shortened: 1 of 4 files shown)


JordonPhillips:
Always need new tests to prevent regressions

srri:
Working on one, thanks

srri:
Added a couple tests also fixed a test that was being executed but not checked. Can we re-run codecov?

srri:
Tests added, good to go

jamesls:
I think these would make the most sense as shared protocol tests (https://github.com/boto/botocore/blob/develop/tests/unit/protocols/input/rest-xml.json) so other SDKs can pick them up.

srri:
Updated test to live in `procotols/input/rest-xml.json`

jamesls:
:shipit:  Looks good to me.

JordonPhillips:
⛵

PR #991, opened 2016-07-29
Add test case for configloader

srri:
Saw that test coverage was on the lower side for [this file](https://codecov.io/gh/boto/botocore/commit/cc40fafc886dab916d29baf5f8cc89bdf619a2ee/graphs/tree.svg), although it is small. So I decided to add a simple test for one of the functions.

Tests the functionality of _merge_list_of_dicts.
Also removed some unused variables.


kyleknap:
Hmm you are testing an internal method: ``_merge_list_of_dicts``. Would it be possible to test this functionality from a public interface (i.e. one that does not begin with an underscore)?

srri:
It has no public interface, it's used from a single spot.

PR #996, opened 2016-08-03
Add support for aws_chunked with s3v4.

### LINKED ISSUE #995: Add support for streaming (transfer-encoding: chunked) v4 uploads
Currently, botocore (and boto3) when using v4 signatures will upload an object to S3 with a SHA256 of the entire content in the signature. This is a problem for streaming uploads, as the entire stream must be buffered first, the SHA256 computed, and then submitted as a PUT to S3.

AWS S3 describes a method for v4 signer streaming uploads: http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html. The premise is to set `transfer-encoding: chunked` and upload each chunk of a set size (except for the last one) where only a chunk's content is signed (as opposed to the entire stream). The one limitation is that the content length must be known ahead of time and set as the `x-amz-decoded-content-length`. Still, supporting such uploads would be greatly beneficial when retrieving a stream of known length from one location and putting it into S3 without having to buffer on disk or memory and seeking through it to compute the signature.

When submitting such requests, the content hash is initially set to `STREAMING-AWS4-HMAC-SHA256-PAYLOAD` and each individual chunk is submitted with its hash.

timuralp:
When uploading to S3 from a stream, it would be useful to opt into S3
aws_chunked uploads with v4 signatures. The mechanism is documented
here:
http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html

The feature is implemented through a new auth class --
S3SigV4ChunkedAuth. The class wraps the body of the request with a
ChunkedUploadWrapper instance. It takes care of producing fixed size
upload chunks and computing the required signatures for each one.

To opt into this scheme, the caller is required to set the aws_chunked
option for client configuration and it will only be used with the
PutObject operations.

Fixes #995

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -300,14 +300,16 @@ class SigV4Auth(BaseSigner):
         sts.append(sha256(canonical_request.encode('utf-8')).hexdigest())
         return '\n'.join(sts)
 
-    def signature(self, string_to_sign, request):
+    def signing_key(self, request):
         key = self.credentials.secret_key
         k_date = self._sign(('AWS4' + key).encode('utf-8'),
                             request.context['timestamp'][0:8])
         k_region = self._sign(k_date, self._region_name)
         k_service = self._sign(k_region, self._service_name)
-        k_signing = self._sign(k_service, 'aws4_request')
-        return self._sign(k_signing, string_to_sign, hex=True)
+        return self._sign(k_service, 'aws4_request')
+
+    def signature(self, signing_key, string_to_sign, request):
+        return self._sign(signing_key, string_to_sign, hex=True)
 
     def add_auth(self, request):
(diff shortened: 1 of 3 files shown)


timuralp:
I wasn't sure if this is the right approach to this problem and the PR doesn't have any unit tests at the moment. If there is agreement that this make sense or guidance on how to rework it, I'm happy to work further on the PR and add the tests.

jamesls:
Linking issues, boto3 tracking issue: https://github.com/boto/boto3/issues/751

Pending discussion over on the issue.

timuralp:
I don't know if there is anything I could've done to draw more attention to this or generate discussion. In light of lack of desire to discuss this feature, it probably makes sense to close it.

Issue #860, opened 2016-03-31
Implement CloudFormation ChangeSet waiter


webratz:
The newly available CloudFormation ChangeSets often require you to wait until a certain state is reached. Especially when calling the clients describe_change_set function pretty much always a wait is required.
It would be great to have waiters similar to those already existing for CloudFormation Stacks.

jamesls:
Marking as feature request.  I agree this would be helpful.

Nath-P:
+1   Yeah this would be nice, currently a fair bit of Handling legwork needs to be done to use change sets i with boto,  Even nicer would be  a resource object for changesets thats a sub resource of the Stack object

PR #999, opened 2016-08-05
#860 Implement CloudFormation ChangeSet waiter

grahamar:


jamesls:
The latest `develop` branch has some additional validation for waiters that are failing with this change.

Looks like the matchers need to be `path` instead of `pathAll/pathAny`:

```
AssertionError: Attempted to use 'pathAll' matcher in waiter 'ChangeSetCreateComplete' with non list result in JMESPath expression: Status
AssertionError: Attempted to use 'pathAll' matcher in waiter 'ChangeSetDeleteComplete' with non list result in JMESPath expression: Status
```

PR #1003, opened 2016-08-11
add paginator for ECR ListImages

memory:
This API is paginated, but botocore lacks the paginator definition.

### DIFF
--- a/botocore/data/ecr/2015-09-21/paginators-1.json
+++ b/botocore/data/ecr/2015-09-21/paginators-1.json
@@ -1,9 +1,9 @@
 {
   "pagination": {
     "ListImages": {
-      "input_token": "NextToken",
-      "output_token": "NextToken",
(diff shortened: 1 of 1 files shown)


jamesls:
:shipit:  Looks good to me.

Shape: https://github.com/boto/botocore/blob/develop/botocore/data/ecr/2015-09-21/service-2.json#L992


cc @kyleknap @JordonPhillips

kyleknap:
@jamesls does the casing matter? I noticed in the service model the names for the members are: ``nextToken`` and ``maxResults`` not ``NextToken`` and ``MaxResults``

jamesls:
Ah good point, I missed that.  I thought our pagination tests would have caught that.

memory:
Likewise surprised the tests didn't catch that!  Updated. :)

kyleknap:
Thanks :ship:. Merging.

PR #1005, opened 2016-08-14
ClientError subclass factory

kislyuk:
See also: https://github.com/boto/boto3/issues/167, https://github.com/boto/boto3/issues/361

I'm using this PR to start a discussion about subclassing ClientError. Is there a plan for how to implement this? (I couldn't find any relevant notes in https://github.com/boto/boto3/issues/167.) The implementation here allows more useful error processing patterns in ClientError consumers, like:

```
import boto3
from botocore.errorfactory import NoSuchEntity

iam = boto3.resource("iam")

try:
    iam.Role("fakeRole").load()
except NoSuchEntity as e:
    print(e)
```

There are a few caveats, including the fact that the error factory has no ability to sanity check error names. I can add docs if I receive feedback that I'm on the right track.


kislyuk:
Any feedback?

kyleknap:
Hi @kislyuk sorry about the long wait for a reply. We needed a long discussion on how we would like to proceed about this.

First of all, this is something that we definitely want because it would improve the experience in coding with botocore/boto3. Catching a general ClientError and introspecting the error code is cumbersome. So thank you for starting the conversation. Here is some feedback that we have on this:

* We think that the errors should be namespaced to services because services often have overlapping exceptions which could cause confusion if you accidentally caught an exception from a different service that you did not intend for.
* It would be great if exceptions could be scoped to sessions. Right now client classes across sessions are different, that is ``session1.client('s3'), session2.client('s3'``) produce different classes. Putting classes in a module level (global) errorfactory means there's no granularity across sessions which is inconsistent from the rest of botocore.
* How would you propose handling API versions? Across the various API versions, the types exceptions available may differ.
* It would be nice if we could warn when you are importing an exception that does not exist in the model and possibly be able to enumerate/document them. This would really help with discoverability as right now there are two ways (neither of them ideal) you can determine the valid exceptions: look at the model or try calling the operation and running into exception yourself)

Another idea that we have had offline discussions about is expose exceptions as such:
```py
try:
    ec2.describe_instances(...)
except ec2.meta.exceptions.InvalidInstanceId as e:
    pass
```
Dangling it off the ``meta`` property is nice for a few reasons:
* It is scoped to the client/session
* It handles API versions
* It is namespaced based on service name
* It would not be difficult to use the service model to enumerate exceptions.

The main downside is that you would have to reference to the client/resource to be able to catch the exception. Let us know what you think about that idea?

I realize that this is a lot of feedback, but I think having a good discussion about this is important as getting the functionality working in a way that covers all of the edge cases and has an easy to use interface is very important for this feature.

kislyuk:
Thanks @kyleknap, this all sounds reasonable and makes sense. I like the concept of accessing exceptions through `meta`. How would you handle enumerating known exceptions though? I haven't seen them in service models, are they available somewhere?

kyleknap:
Cool. So exceptions are marked in two places in the ``service-2.json`` models:

* Under the ``"errors"`` key for the operation definition. Here is an example for [ACM](https://github.com/boto/botocore/blob/develop/botocore/data/acm/2015-12-08/service-2.json#L21). I could see this being useful for documentation purposes of what operations can throw what exceptions.
* Looking at the shapes, if a shape has the key ``"exception: true"`` than that is a possible exception that can be thrown. Here is an example in [ACM](https://github.com/boto/botocore/blob/develop/botocore/data/acm/2015-12-08/service-2.json#L427). I could see this useful in enumeration and warning if you are importing an exception that does not exist.

Another important thing to note is that some services do not define exceptions, like EC2, so you would not be able to use the model for some of these services to enumerate/document exceptions. But in general, most services will have exceptions defined, especially the newer services.

Let me know if that helps.

kislyuk:
OK I see, so the reason I thought there were no exceptions in service models is that EC2 doesn't have them.

I suggest in cases where the service doesn't define exceptions, we use the behavior that I proposed (to auto-vivify the exception by name blindly).

I'll update the PR later today or tomorrow, and let you know if I get stuck.

kyleknap:
Sounds good. Yeah that would be the best that we could do if exceptions are not defined.

kislyuk:
@kyleknap, here is the initial iteration of the design you discussed. I wired up the models, and introspection now works (`client.meta.exceptions.<TAB>` produces expected results). Please take a look and let me know your feedback / things to finish up.

I would like to restrict the scope of the initial merge to produce subclasses for only those errors in the model, and push subclassing in the blind in a separate PR.

I ran into an annoying inconsistency - IAM and presumably some other services name their errors with the "Exception" suffix, e.g. NoSuchEntityException, but send `NoSuchEntity` on the wire instead. I've put in a secondary lookup hack to deal with this behavior, let me know if you can think of a better way.

kyleknap:
Awesome. I will take a look at it. On a side note, it looks like Python 2.6 tests are failing so it would be great if we can get those fixed.

I like the idea of doing these in increments. It makes it a lot easier to review and increment. I wonder if it makes sense to set up a feature branch up for this?

The only reason I ask is that being able to blindly get an exception class is for backwards compatibility reasons. While we try to avoid it and does not happen often, sometimes a service model gets updated that removes an exception shape. So being able to blindly get these exceptions is important for the use case of: a user writing code that tried catching a particular exception, a model is updated that removes the exception, the user upgrades, and then starts hitting an attribute error.

The Ruby SDK has very similar behavior: https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/lib/aws-sdk-core/errors.rb#L119. Also as a side note, I am not sure how knowledgeable of ruby you are but it may also be worth looking at there logic to see if there is any missing edge cases.

Let me know what you think?

kislyuk:
Hi @kyleknap, I went ahead and added blind error extraction and a logger warning when that happens. Please have a look.

I'm not entirely sure why the 2.6 tests failed; `dir(instance)` on 2.6 doesn't seem to call `__dir__()`, though the docs say it should. I disabled that check on 2.6.

I took a look at the Ruby code; one nice thing they do is take out a mutex before creating the class, so a race among threads doesn't result in multiple copies of the class and `isinstance` failures.

The name mangling that the Ruby library does actually doesn't seem to make sense, so I would suggest going ahead with what we got here.

kislyuk:
(Oh and I added a threading.Lock around class creation in the factory, inspired by the Ruby code.)

kislyuk:
@kyleknap any feedback?

kyleknap:
@kislyuk I'm looking at it again now. I should have feedback by the end of the day.

botocore/client.py
```diff
@@ -491,7 +492,16 @@ def _make_api_call(self, operation_name, api_params):
...
         if http.status_code >= 300:
-            raise ClientError(parsed_response, operation_name)
+            try:
+                error_code = parsed_response["Error"]["Code"]
+                if "." in error_code:
+                    error_code = error_code.replace(".", "")
+                if not error_code[0].isalpha():
+                    error_code = "ClientError" + error_code
+                error_class = getattr(self.meta.exceptions, error_code)
+            except Exception:
```
kyleknap: I am wondering if we should have this general try catch around the logic as it may cover up edge cases that we missed. I think I would prefer to remove the try catch or if not that at least log what exception triggered the except clause.

botocore/errorfactory.py
```diff
@@ -0,0 +1,38 @@
...
+    def _error_shapes(self):
+        shapes = {}
+        for op_name in self._service_model.operation_names:
+            op_model = self._service_model.operation_model(op_name)
+            for shape in op_model.error_shapes:
+                shapes[shape.name] = shape
+        return shapes
+
+    def __getattr__(self, attr):
+        if attr.startswith("_"):
```
kyleknap: What was the reasoning for adding this validation? I do realize that underscores are typically reserved for private api, but just curious.

botocore/errorfactory.py
```diff
@@ -0,0 +1,38 @@
...
+import threading
+
+from botocore.utils import CachedProperty
+from botocore.exceptions import ClientError
+
+logger = logging.getLogger(__name__)
+
+
+class ServiceErrorFactory:
+    ClientError = ClientError
```
kyleknap: Is this class variable used anywhere?

tests/unit/test_errorfactory.py
```diff
@@ -0,0 +1,30 @@
...
+        'shapes': {
+            'NoSuchResourceException': {
+                'type': 'structure',
+                'members': {}
+            }
+        }
+    }
+    factory = ServiceErrorFactory(ServiceModel(model))
+    members = dir(factory)
+    if sys.version_info >= (2, 7):
```
kyleknap: This seems a little fishy. Ideally the tests should be able to handle any python version we support, and if there is no way to support the test then we skip it. I saw your note of ``dir()`` not working on Python2.6. What did ``dir()`` return instead? It would be great if we can figure out why there are issues with it.

tests/unit/test_client.py
```diff
@@ -509,18 +521,27 @@ def test_client_signs_call(self, signer_mock):
...
-        }
-        self.endpoint.make_request.return_value = (
-            mock.Mock(status_code=400), error_response)
+        error_responses = [
+            {'Error': {'Code': 'code', 'Message': 'error occurred'}},
+            {'Error': {'Code': '404', 'Message': 'Not Found'}},
+            {'Error': {'Code': 'InvalidAddress.NotFound', 'Message': ''}},
+            {'Error': {'Code': 'TestOperationException', 'Message': ''}},
+            {'Error': {'Message': 'error occurred'}},
+            {'Error': {}}
```
kyleknap: How do cases when the ``Code`` is an empty string get handled? It may be good to have a test specifying that behavior.

tests/unit/test_client.py
```diff
@@ -509,18 +521,27 @@ def test_client_signs_call(self, signer_mock):
             'test_operation', request)
 
     def test_client_makes_call_with_error(self):
-        error_response = {
-            'Error': {'Code': 'code', 'Message': 'error occurred'}
-        }
-        self.endpoint.make_request.return_value = (
-            mock.Mock(status_code=400), error_response)
+        error_responses = [
```
kislyuk: Done

tests/unit/test_client.py
kislyuk: Added

tests/unit/test_errorfactory.py
kislyuk: I don't have a Python 2.6 runtime handy and the tests take a while to run so I'm at a loss. Here is the error from Travis:
```
FAIL: tests.unit.test_errorfactory.test_errorfactory
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/boto/botocore/tests/unit/test_errorfactory.py", line 24, in test_errorfactory
    assert_equals(dir(factory), ['NoSuchResourceException'])
AssertionError: ['__dir__', '__doc__', '__getattr__', '__init__', '__module__', '_error_shapes', '_service_model'] != ['NoSuchResourceException']

```

kislyuk: Turns out I do have Python 2.6 (os x comes with it) and here's what I found: Python 2.6 uses an earlier version of the attribute enumeration protocol that second-guesses `__dir__` using (since deprecated) `__members__` and `__methods__` attributes as described in https://www.python.org/dev/peps/pep-0252/. Also, Python 2.6 wants attribute names to be bytestrings and discards them if they are unicode strings (as they are when loaded from the model). Fixed both issues (although IMO Python 2.6 should not be accommodated).

kislyuk:
@kyleknap?

tests/unit/test_errorfactory.py
kyleknap: Awesome thanks for the deep dive on that! Yeah I am not a big fan of accommodating python 2.6 as well, but it is a backwards compatibility thing; since botocore started off officially supporting python2.6, we need to continue supporting it until we make a stance to stop supporting 2.6, which in itself will take a decent amount of time.

botocore/errorfactory.py
kyleknap: Makes sense

botocore/errorfactory.py
kyleknap: Cool

botocore/client.py
```diff
@@ -491,7 +492,16 @@ def _make_api_call(self, operation_name, api_params):
         )
 
         if http.status_code >= 300:
-            raise ClientError(parsed_response, operation_name)
+            if parsed_response.get("Error", {}).get("Code"):
+                error_code = parsed_response["Error"]["Code"]
+                if "." in error_code:
+                    error_code = error_code.replace(".", "")
+                if not error_code[0].isalpha():
+                    error_class = ClientError
```
kyleknap: I still think that a ``404`` exception will be created with this logic. You may want to hoist it up to ``parsed_response...`` if statement.

tests/unit/test_client.py
```diff
@@ -1323,6 +1321,50 @@ def return_mock_tuple(**kwargs):
...
+            service_client.test_operation(Foo='one', Bar='two')
+
+    def test_client_makes_call_with_error(self):
+        error_data = {'Error': {'Code': 'code', 'Message': 'error occurred'}}
+        self.assert_error_response(error_data, 'ClientError')
+        self.assert_error_response(error_data, 'code')
+
+    def test_error_with_no_wire_code(self):
+        error_data = {'Error': {'Code': '404', 'Message': 'Not Found'}}
+        self.assert_error_response(error_data, 'ClientError')
```
kyleknap: So for these tests, they should always be ``ClientError`` as any generated exception subclasses from ``ClientError``. I think it would be great to make assertions that differentiates what the exact class are as well such as asserting the name of the class of the exception that is thrown. For example, when I print out the name of the class of this exception, I get ``404`` when it should be client exception.

So I could see always making the assertion the exception thrown is of ``ClientError`` and then making an additional assertion that name of the class is what we except whether it is a specific one or it did not resolve to anything, making it a general ``ClientError``.

kislyuk: Yes, that was a mistake in the test assertion. Fixed.

botocore/client.py
kislyuk: Yes, that was a logic error, thanks for the catch. I think I got this fixed and the test assertion works as intended now, please take a look.

kyleknap:
Great! Looks good. I did some manual testing myself and it worked out pretty well. Here is what I did for reference for whoever else looks at this:
```py
import boto3
from botocore.exceptions import ClientError


def catch_specific_exception():
    dynamodb = boto3.client('dynamodb')

    try:
        dynamodb.describe_table(TableName='some-table-that-does-not-exist')
    except dynamodb.meta.exceptions.ResourceNotFoundException as e:
        print('Caught exception of type %s' % e.__class__.__name__)


def catch_general_exception():
    dynamodb = boto3.client('dynamodb')
    try:
        dynamodb.describe_table(TableName='some-table-that-does-not-exist')
    except dynamodb.meta.exceptions.SomeUnModeledExceptionThatIMadeUp as e:
        print('Caught exception of type %s' % e.__class__.__name__)
    except dynamodb.meta.exceptions.ClientError:
        print('Caught as a general client error')


def catch_general_exception_from_botocore_exceptions():
    dynamodb = boto3.client('dynamodb')
    try:
        dynamodb.describe_table(TableName='some-table-that-does-not-exist')
    except ClientError:
        print('Caught as a general client error')


def catch_unmodeled_exception():
    boto3.set_stream_logger('')
    ec2 = boto3.client('ec2')
    try:
        ec2.describe_instances(InstanceIds=['i-1323434'])
    except ec2.meta.exceptions.InvalidInstanceIDNotFound as e:
        print('Caught exception of type %s' % e.__class__.__name__)


def main():
    catch_specific_exception()
    catch_general_exception()
    catch_general_exception_from_botocore_exceptions()
    catch_unmodeled_exception()
```
catch_specific_exception() output:
```
Caught exception of type ResourceNotFoundException
```
catch_general_exception() output:
```
Caught as a general client error
```
catch_general_exception_from_botocore_exceptions()
```
Caught as a general client error
```
catch_unmodeled_exception()
```
... [redacted] ...
2016-09-28 15:46:52,326 botocore.errorfactory [WARNING] Error "InvalidInstanceIDNotFound" was not found in the service model
Caught exception of type InvalidInstanceIDNotFound
```
Also confirmed it works with tab completion which is awesome:
```py
In [1]: import boto3

In [2]: dynamodb = boto3.client('dynamodb')

In [3]: dynamodb.meta.exceptions.
dynamodb.meta.exceptions.ConditionalCheckFailedException
dynamodb.meta.exceptions.InternalServerError
dynamodb.meta.exceptions.ItemCollectionSizeLimitExceededException
dynamodb.meta.exceptions.LimitExceededException
dynamodb.meta.exceptions.ProvisionedThroughputExceededException
dynamodb.meta.exceptions.ResourceInUseException
dynamodb.meta.exceptions.ResourceNotFoundException

```
Overall, looks good to me. :ship: @jamesls @JordonPhillips should be good to take a look at.

kislyuk:
ping @jamesls

jamesls:
Sorry for the delay, I'm looking at this now.  I'll try to have some comments tomorrow.

botocore/client.py
```diff
@@ -666,6 +677,7 @@ def __init__(self, events, client_config, endpoint_url, service_model,
         self._endpoint_url = endpoint_url
         self._service_model = service_model
         self._method_to_api_mapping = method_to_api_mapping
+        self._error_factory = ServiceErrorFactory(self._service_model)
```
kislyuk: Due to the amount of indirection between session and client, I'm not clear on how to scope an exception factory to a session. Can you show me?

botocore/errorfactory.py
```diff
@@ -0,0 +1,42 @@
...
+            op_model = self._service_model.operation_model(op_name)
+            for shape in op_model.error_shapes:
+                shapes[shape.name] = shape
+        return shapes
+
+    def __getattr__(self, attr):
+        if attr.startswith("_"):
+            raise AttributeError(attr)
+        if attr not in self._error_shapes:
+            msg = 'Error "{0}" was not found in the service model'
```
kislyuk: I think because EC2 and S3 don't model their errors, inability to reference unmodeled errors reduces the utility of this change. I think the path forward should include either getting those services to model their errors, or providing at least an optional way to key their exceptions by name.

botocore/errorfactory.py
```diff
@@ -0,0 +1,42 @@
...
+            logger.warning(msg.format(attr))
+        with self._lock:
```
kislyuk: The hypothetical concern here is a race condition where two threads both say `x = exceptions.X` and both factories decide to create a class for it. Now we have two different identities for the exception class X - one canonical, and one orphaned. The thread whose reference got overwritten will fail to catch the resulting exception with `try .. except X`, because it is no longer an instance of X.

botocore/client.py
```diff
@@ -686,3 +698,7 @@ def config(self):
     @property
     def method_to_api_mapping(self):
         return self._method_to_api_mapping
+
+    @property
+    def exceptions(self):
```
kislyuk: No objections from me - fixed.

botocore/errorfactory.py
```diff
@@ -0,0 +1,42 @@
+import logging
+import threading
+
+from botocore.utils import CachedProperty
+from botocore.exceptions import ClientError
+
+logger = logging.getLogger(__name__)
+
+
+class ServiceErrorFactory:
```
kislyuk: Fixed, thank you. Now I understand why `dir` was behaving weirdly in 2.6. Removed the `__members__` property now that the 2.6 behavior is consistent.

botocore/client.py
```diff
@@ -491,7 +492,17 @@ def _make_api_call(self, operation_name, api_params):
...
         if http.status_code >= 300:
-            raise ClientError(parsed_response, operation_name)
+            if parsed_response.get("Error", {}).get("Code"):
+                error_code = parsed_response["Error"]["Code"]
+                if "." in error_code:
+                    error_code = error_code.replace(".", "")
+                if error_code.isdigit():
+                    error_class = ClientError
+                else:
+                    error_class = getattr(self.meta.exceptions, error_code)
```
kislyuk: @jamesls, I've added the reverse map and error retrieval by code where the model contains that information. Please take a look.

kislyuk:
ping @jamesls

kislyuk:
@jamesls, @kyleknap, is there anything to do to push this PR forward?

kislyuk:
ping @jamesls

kislyuk:
It would be great if there was more responsiveness on this.

HaraldNordgren:
This has been in the queue since August. If the issue is only about unmodelled exceptions than maybe we can push this along and handle that later. This feature will make boto3 so much more usable.

kyleknap:
Sorry for the delay here. I am looking at this now. Let's pick this up and finally get this merged!

@kislyuk As a side note, do you mind if either me or @jamesls push to this branch at any point? I do not know when or if we would do it, but it could help speed up the process.

kislyuk:
@kyleknap, I would love it if you pushed what you think is necessary to get this merged. I think I've put in everything I could per the review above, I'd be happy if you did the rest or continued the review. I've added you as collaborator on https://github.com/kislyuk/botocore if you want to push directly there.

kislyuk:
@kyleknap, at this point I think it would be better if you drive this to completion. I've done major refactoring to this a couple of times already and don't want to get too deeply into another refactor without knowing exactly what you want. Please go for it.

HaraldNordgren:
I feel like at this point we either merge this or let it sit in the queue forever.

Code is never *perfect*. If you didn't write it then might not be exactly the way you would have wanted it kyleknap. As far as I can tell this works, and most importantly: It's available now.

This feature is needed. Parsing ClientError objects is not a nice practice.

kyleknap:
@kislyuk Sounds good. I am working on refactoring your branch so you should be seeing updates soon.

kyleknap:
@kislyuk So I have a branch: https://github.com/kyleknap/botocore/tree/errors-v2 that does the refactoring that I talked about. It is ready for review. However, I rebased off of the develop branch and squashed your commits so I cannot push straight to your branch unless I force push. I was not sure if you cared if I force push so I just opened up another pull request instead. If you want to reuse this current pull request, I can close mine: https://github.com/boto/botocore/pull/1113 and force push to your branch. Let me know.

kislyuk:
@kyleknap, I'm fine either way - I have the branch backed up so please force push if you see fit. It depends on whether you want to keep the long discussion history in this PR or start a fresh thread. Might be better to start a fresh thread to unclutter it.

kyleknap:
@kislyuk Sounds good. Let's use the new thread then to start off clean. The two threads are linked so we can always go back to this one for reference.

Also feel to review the new pull request. It would be great to get any feedback you may have since you did most of the work implementation-wise to make sure I did not miss anything in refactoring it.

Thanks again!

PR #1013, opened 2016-08-18
Allow underscore in endpoint urls

MichaelAquilina:
I dont see any reason why these should be excluded.

A perfectly reasonably hostname to have in your `/etc/hosts` file for example is `dev-dynamodb_1`. Currently this fails with an invalid endpoint url error.

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -604,7 +604,7 @@ def is_valid_endpoint_url(endpoint_url):
     if hostname[-1] == ".":
         hostname = hostname[:-1]
     allowed = re.compile(
-        "^((?!-)[A-Z\d-]{1,63}(?<!-)\.)*((?!-)[A-Z\d-]{1,63}(?<!-))$",
+        "^((?!-)[A-Z\d-_]{1,63}(?<!-)\.)*((?!-)[A-Z\d-_]{1,63}(?<!-))$",
         re.IGNORECASE)
     return allowed.match(hostname)
 
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -541,6 +541,8 @@ class TestIsValidEndpointURL(unittest.TestCase):
     def test_hostname_no_dots(self):
         self.assertTrue(is_valid_endpoint_url('https://foo/'))
 
+    def test_hostname_can_have_underscores(self):
+        self.assertTrue(is_valid_endpoint_url('https://dev-dynamodb_1/'))
 
 class TestFixS3Host(unittest.TestCase):
     def test_fix_s3_host_initial(self):


jamesls:
This has been discussed previously here: https://github.com/boto/botocore/pull/960, and the reason for not allowing this is to adhere to the [RFC](https://tools.ietf.org/html/rfc952)

> The Internet standards (Requests for Comments) for protocols mandate that component hostname labels may contain only the ASCII letters 'a' through 'z' (in a case-insensitive manner), the digits '0' through '9', and the hyphen ('-'). The original specification of hostnames in RFC 952, mandated that labels could not start with a digit or with a hyphen, and must not end with a hyphen. However, a subsequent specification (RFC 1123) permitted hostname labels to start with digits. No other symbols, punctuation characters, or white space are permitted.

MichaelAquilina:
out of curiosity, why force endpoint_urls to adhere to something? Why not let it fail at a lower level? You are essentially restricting the usage possible endpoint names.

jamesls:
For quite a while we didn't have this check, and we kept get getting bug reports about cryptic errors because of incorrectly configured endpoint urls.

Based on the the previous discussion on #960, I think there's two possible options to address your issue:

* Add a config option to disable the endpoint validation.  You'd have to explicitly opt out.
* Consider updating our logic for the pre-existing `parameter_validation` config option to now include not checking endpoint-url.  Technically a breaking change, but a case can be made that thi s is actually a form of param validation.

MichaelAquilina:
It's kind of odd that botocore is handling url validation in my opinion. Feels something that should be handled at lower level. If you feel strongly about keeping it though, a parameter to opt out would definitely be something a lot of people will find useful. Docker container names generated by docker-compose often contain underscores for example.

jamesls:
I think we should stay consistent with our previous discussion and resolution and keep things as is.  That said, I think we would accept a PR that added a client config option to disable this behavior for people who want to turn off this behavior.

Based on the above, I'm going to close out this PR.

inieves:
I realize I am late to this discussion.

Thanks @jamesls for explaining the rationale behind the decision, citing: "bug reports about cryptic errors because of incorrectly configured endpoint urls".  Your decision makes more sense to me now.

Those cryptic errors result from cryptic programs which result from cryptic source code, and possibly a cryptic programming language.  Cryptic errors are all over the place, especially in systems programming.  Especially in Linux and C, and its entire ecosystem.

I don't believe that your decision impacts the situation of cryptic messages, because there are still many of them out there (having nothing to do with botocore), and also because botocore itself outputs such cryptic messages.  I hope that you are seeing less bug reports about cryptic errors, but what about the bug reports you don't see because people don't even write them?  You may get a false sense of the severity of bugs if you are depending only on bug reports.

I strongly suggest that the error message be clearer here, so that regardless of what you believe about RFCs, the developer knows what the problem is and how to address it.  The developer probably cares alot more about that, than RFCs anyways.

mrbrown-09:
I just got bit by this underscore issue using docker generated service names.  One could say "stop docker from generating them this way", but if it resolves, let it resolve and use it.  Let the DNS determine if it's a bad hostname or not.  Why should the program care if the hostname container unprintable characters if it can be resolved?

PR #1015, opened 2016-08-21
Customization of waiter configuration.

### LINKED ISSUE #981: Allow easy customization of waiter delay/iteration count
I'm not sure about others, but sometimes the default waiter delay/count is insufficient for me. For example, the recently added CloudFormation waiters are great (I like that the state modeling and polling machinery is done for me) but the default delay is 30 seconds between polls, which significantly slows things down for my fairly small stacks. In other cases, I've had waiters give up on a particularly slow operation because the max poll count was too low, even though I knew ahead of time it would be slow and would just have increased the max poll time if I could've.

1. Is it even possible to adjust those parameters today?
2. If it is possible, can it be made easier and part of the official (documented) interface for waiters?

coady:
Fixes #981.


JordonPhillips:
This is causing tests to fail on all python versions we support, you will need to fix those before we can look at this.

coady:
Fixed doc test.

botocore/client.py
```diff
@@ -651,36 +651,41 @@ def can_paginate(self, operation_name):
...
+    def _waiter_config(self):
+        try:
+            return self._loader.load_service_model(
+                self._service_model.service_name,
+                'waiters-2',
+                self._service_model.api_version)
+        except DataNotFoundError:
+            return {}
+
+    def get_waiter(self, waiter_name, **config):
```
JordonPhillips: I would prefer to be explicit here rather than accepting any keyword arguments. Specifically we should only allow setting `delay` and `max_attempts` for now.

tests/unit/test_client.py
```diff
@@ -737,6 +737,10 @@ def test_service_has_waiter_configs(self):
         self.assertEqual(sorted(service_client.waiter_names),
                          sorted(['waiter_1', 'waiter_2']))
         self.assertTrue(hasattr(service_client.get_waiter('waiter_1'), 'wait'))
+        self.assertEqual(service_client.get_waiter('waiter_1', delay=10)
```
JordonPhillips: These should be in their own tests

jamesls:
Just wanted to follow up here.  We recently merged a PR that added a public API for customizing waiter configs: https://github.com/boto/botocore/pull/1267

Does this meet your use case?

PR #1018, opened 2016-08-22
Port paginator checks from translate.py to functional test

jamesls:
The motivation for this PR is two fold:

* When reviewing #1003, there was an error in the paging config that did not result in a test failure.  In order to help review these pagination pull requests quicker, we should have automated tests in place to validate as much as we can.
* We already had this code in `translate.py`.  This code was used back in the day to do translation and validation of the various JSON model files.  This means that the bulk of the work was done, it just wasn't being called from anywhere.

So I've added two commits:

* Convert pagination validation from translate.py to functional tests
* Update translate.py to remove dead code that hasn't been used in years.

The only thing we still use in translate.py is the code for building retry configs.  I double checked that nothing in botocore, aws-cli, or boto3 is using the translate module.  There's a very small possibility that people could have been using this code, but given this code requires versions of models that haven't been used for years, I think it's reasonable to remove this code.

cc @kyleknap @JordonPhillips


tests/functional/test_paginator_config.py
```diff
@@ -0,0 +1,155 @@
...
+    # union of 'result_key', 'output_token', 'more_results', and
+    # 'non_aggregate_keys'.
+    # There's enough state drift (especially with non_aggregate_keys) that
+    # this is no longer a realistic thing to check.  Someone would have to
+    # backport the missing keys to all the paginators.
+    output_shape = service_model.operation_model(operation_name).output_shape
+    output_members = output_shape.members
+    for key_name, output_key in _get_all_page_output_keys(page_config):
+        if _looks_like_jmespath(output_key):
+            _validate_jmespath_compiles(output_key)
```
kyleknap: One idea that may be interesting to pursue but may be a bit of work to do is actually try to validate the jmespath expression against the model. I could imagine using the [ArgumentGenerator](https://github.com/boto/botocore/blob/develop/botocore/utils.py#L502) and trying to apply the jmespath expression to that to see if you can get a match.

This would be super helpful if the PR actually has a jmespath expression because sure even though it is valid, I still will need to look into the model to see that the expression match up to the shape.

What's your thought on this? I do not think we need need it now but may be helpful in the future if we see a lot of paginators needing nested jmespath expressions.

kyleknap:
Looks good. I am fine with removing it from ``translate.py`` as it was not used and doubt others were using it. I only had a comment about a potential future check, but otherwise :ship:

tests/functional/test_paginator_config.py
jamesls: Yeah I had a similar idea as well, I'd need to play around with it and see how reliable it is.

I think this is something we should consider adding in the future.

kyleknap: Yeah we would not be able to do it for some of the more complicated ones like s3 but I feel that is a minority.

PR #1019, opened 2016-08-22
Add waiter NetworkAclExists to waiters.

ParimiDev:
Simple enhancement to add waiter for NetworkACL creation.

### DIFF
--- a/botocore/data/ec2/2016-04-01/waiters-2.json
+++ b/botocore/data/ec2/2016-04-01/waiters-2.json
@@ -348,6 +348,24 @@
         }
       ]
     },
+    "NetworkAclExists": {
+      "operation": "DescribeNetworkAcls",
+      "delay": 20,
+      "maxAttempts": 10,
+      "acceptors": [
+        {
+          "matcher": "path",
+          "expected": true,
+          "argument": "length(NetworkAcls[]) > `0`",
+          "state": "success"
+        },
+        {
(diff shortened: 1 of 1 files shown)


jamesls:
:shipit:  Looks good to me, and is consistent with how other [similar ec2 waiters](https://github.com/boto/botocore/blob/develop/botocore/data/ec2/2016-04-01/waiters-2.json#L297-L314) are written.   Also confirmed this is the correct error code when a network acl does not exist.

kyleknap:
@jamesls is the waiter you linked not defined properly? It uses ``pathAll`` instead of ``path`` for a JMESPath expression that should just evaluate to ``true`` or ``false``. I only noticed it because it is slightly different from this PR (which I think is the correct way to do define it)

However, this change looks good to me. :ship:

jamesls:
Oh man, that's a good catch.  I think it's probably worth investing in better automated tests for these types of waiter errors.  That idea of doing an autogen on the output shape and running the JMESPath expression could catch this issue right?

I might put some time into improving our automated checks.

In the meantime, I'll go ahead and merge.

@ParimiDev thanks for the pull request.

kyleknap:
Yeah I like that idea. It will be a lot easier to pull in waiter changes as we really would only have to worry about whether the matcher is appropriate based on the behavior of the service instead of looking at whether the definition is valid (which is not that easy as you need an understanding of JMESPath and reference to the shape).

PR #1024, opened 2016-08-31
Add retry information to ResponseMetadata

jamesls:
This pulls in https://github.com/boto/botocore/pull/965 with the following changes:

* Replace `Retries` list with the number of retry attempts.
* Update `ClientError` messages to include retry information when max attempts are reached.

More context [here](https://github.com/boto/botocore/pull/965#issuecomment-243618656).

A few examples:

Client error messages:

```
An error occurred (NotFound) when calling the DescribeInstances operation (max_retries=4 reached): The instance ID 'i-1234' does not exist
```

ResponseMetadata on error:

```
 'ResponseMetadata': {'HTTPHeaders': {'connection': 'close',
                                      'date': 'Wed, 31 Aug 2016 17:41:51 GMT',
                                      'server': 'AmazonEC2',
                                      'transfer-encoding': 'chunked'},
                      'HTTPStatusCode': 400,
                      'MaxAttemptsReached': True,
                      'RequestId': 'abcd',
                      'RetryAttempts': 4}}
```

ResponseMetadata on success:

```
 'ResponseMetadata': {'HTTPHeaders': {'content-type': 'text/xml;charset=UTF-8',
                                      'date': 'Wed, 31 Aug 2016 17:42:19 GMT',
                                      'server': 'AmazonEC2',
                                      'transfer-encoding': 'chunked',
                                      'vary': 'Accept-Encoding'},
                      'HTTPStatusCode': 200,
                      'RequestId': '25587b14-963b-4357-bc13-56c0acd52d0e',
                      'RetryAttempts': 0}}
```

cc @kyleknap @JordonPhillips

### DIFF
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -342,16 +342,29 @@ class UnsupportedSignatureVersionError(BotoCoreError):
 class ClientError(Exception):
     MSG_TEMPLATE = (
         'An error occurred ({error_code}) when calling the {operation_name} '
-        'operation: {error_message}')
+        'operation{retry_info}: {error_message}')
 
     def __init__(self, error_response, operation_name):
+        retry_info = self._get_retry_info(error_response)
         msg = self.MSG_TEMPLATE.format(
             error_code=error_response['Error'].get('Code', 'Unknown'),
             error_message=error_response['Error'].get('Message', 'Unknown'),
-            operation_name=operation_name)
+            operation_name=operation_name,
+            retry_info=retry_info,
+        )
         super(ClientError, self).__init__(msg)
         self.response = error_response
         self.operation_name = operation_name
+
+    def _get_retry_info(self, response):
+        retry_info = ''
+        if 'ResponseMetadata' in response:
+            metadata = response['ResponseMetadata']
+            if metadata.get('MaxAttemptsReached', False):
+                if 'RetryAttempts' in metadata:
+                    retry_info = (' (max_retries=%s reached)' %
+                                  metadata['RetryAttempts'])
+        return retry_info
 
 
 class UnsupportedTLSVersionWarning(Warning):
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -362,7 +362,7 @@ class ClientError(Exception):
             metadata = response['ResponseMetadata']
             if metadata.get('MaxAttemptsReached', False):
                 if 'RetryAttempts' in metadata:
-                    retry_info = (' (max_retries=%s reached)' %
+                    retry_info = (' (reached max retries: %s)' %
                                   metadata['RetryAttempts'])
         return retry_info
 
--- a/tests/unit/test_endpoint.py
+++ b/tests/unit/test_endpoint.py
@@ -140,27 +140,6 @@ class TestRetryInterface(TestEndpointBase):
         super(TestRetryInterface, self).setUp()
         self.retried_on_exception = None
 
-    def max_attempts_retry_handler(self, attempts, **kwargs):
-        # Simulate a max requests of 3.
-        self.total_calls += 1
-        if attempts == 3:
-            return None
-        else:
-            # Returning anything non-None will trigger a retry,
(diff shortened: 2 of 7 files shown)


botocore/exceptions.py
```diff
@@ -342,17 +342,30 @@ class UnsupportedSignatureVersionError(BotoCoreError):
 class ClientError(Exception):
     MSG_TEMPLATE = (
         'An error occurred ({error_code}) when calling the {operation_name} '
-        'operation: {error_message}')
+        'operation{retry_info}: {error_message}')
```
kyleknap: Sort of unfortunate that ``retry_info`` has to remember to put spaces, but I do not think there is any way around it if we want to put it right between operation and the colon.

botocore/exceptions.py
```diff
@@ -342,17 +342,30 @@ class UnsupportedSignatureVersionError(BotoCoreError):
...
         self.response = error_response
         self.operation_name = operation_name
 
+    def _get_retry_info(self, response):
+        retry_info = ''
+        if 'ResponseMetadata' in response:
+            metadata = response['ResponseMetadata']
+            if metadata.get('MaxAttemptsReached', False):
+                if 'RetryAttempts' in metadata:
+                    retry_info = (' (max_retries=%s reached)' %
```
kyleknap: This may be a little picky but from the context of the CLI ``max_retries`` seems pretty specific as if it is a parameter they can specify but it is not (it is not even something in botocore you can specify). I wonder if it makes sense to make this more like: ``' (reached max retry attempts: %s)'``?

tests/unit/test_endpoint.py
```diff
@@ -140,27 +140,6 @@ def setUp(self):
         super(TestRetryInterface, self).setUp()
         self.retried_on_exception = None
 
-    def max_attempts_retry_handler(self, attempts, **kwargs):
```
kyleknap: Why did these test get removed?

kyleknap:
Looks good. Just had small questions and comments.

botocore/exceptions.py
jamesls: Yeah.  The first version I had rendered the msg, but before calling `super()` would mess with the msg to inject retry info.  This approach seemed less worse that the first approach.

Also open to other ways to improve this.

botocore/exceptions.py
jamesls: Yeah I played around with a few error message formats initially.  The main thing I was noticing is that specifically for the CLI, if you inject the retry info in the middle, and it's fairly long as you've suggested, the actual service error message is all the way on the right hand side and/or wraps to the next line.  Is their something more succinct we can say instead that disambiguates this from a CLI param?

tests/unit/test_endpoint.py
jamesls: Sorry, meant to add a note about that.  This was part of the original PR in #965.  It doesn't look like the tests use these methods anymore but they were never removed.

I can also add them back and send a general "test_endpoint.py" cleanup PR if it helps.

botocore/exceptions.py
kyleknap: Makes sense. I am fine with how it is now. I cannot think of any other ways.

botocore/exceptions.py
kyleknap: Makes sense. How about ``(reached max retries: %s)``?  That is one additional character. I think the thing that might throw users off is the fact that there is an underscore in ``max_retries``. Underscores I would typically associate with a specific parameter.

tests/unit/test_endpoint.py
kyleknap: No that is fine. We can keep them here. Figured to ask because it was not clear.

botocore/exceptions.py
jamesls: Ok I'll update.

jamesls:
@kyleknap updated.

kyleknap:
Thanks! :ship: assuming the tests pass.

PR #1025, opened 2016-09-01
Stubber: Allow add_client_error to use expected_params

Jc2k:
I'd like to be able to use `expected_params` with `add_client_error` - it looks like this should be enough?

### DIFF
--- a/botocore/stub.py
+++ b/botocore/stub.py
@@ -235,7 +235,7 @@ class Stubber(object):
 
     def add_client_error(self, method, service_error_code='',
                          service_message='', http_status_code=400,
-                         service_error_meta=None):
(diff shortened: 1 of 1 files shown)


kyleknap:
If you can add a test for this, that would be great. I would be in favor of pulling this in because I think it would be useful, but we typically need tests for the new functionality to be able to pull it in.

jamesls:
I'll go ahead and take this and add a test for it.  I think this will be really useful.  Assigning to myself.

Jc2k:
Thanks @jamesls! I've been too snowed to get to it 😭

jamesls:
Merged via #1031.

PR #1026, opened 2016-09-08
Add max_pool_connections to client config

### LINKED ISSUE #766: Support for customize the max connections and max pools

jamesls:
This config value allows you to control the size of the
connection pools used in urllib3.  We were previously using the
hardcoded value of 10 (from requests).

I ended up not touching the `pool_connections` based on the
discussion from #766.

There ends up being quite a noticeable difference whenever
`num_threads > max_pool_connection`, so I think this will
help people running requests with a large number of threads.

As an aside, the implementation for this was much more complicated
than necessary.  I'm thinking on sending a few follow up PRs that
start to clean up some of this technical debt.

Closes #773
Closes #766

cc @kyleknap @JordonPhillips

### DIFF
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -17,6 +17,7 @@ import logging
 import time
 import threading
 
+from botocore.vendored.requests.adapters import HTTPAdapter
 from botocore.vendored.requests.sessions import Session
 from botocore.vendored.requests.utils import get_environ_proxies
 from botocore.vendored.requests.exceptions import ConnectionError
@@ -35,6 +36,7 @@ from botocore import parsers
 
 logger = logging.getLogger(__name__)
 DEFAULT_TIMEOUT = 60
+MAX_POOL_CONNECTIONS = 10
 filter_ssl_warnings()
 
 try:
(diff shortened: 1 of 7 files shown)


botocore/endpoint.py
```diff
@@ -75,7 +77,26 @@ def convert_to_response_dict(http_response, operation_model):
 
 
 class PreserveAuthSession(Session):
+    """Internal session class used to workaround requests behavior.
+
+    This class is intended to be used only by the Endpoint class.
```
kyleknap: I think the class should be renamed. It is no longer just for preventing the rebuilding of auth. It also is used to customize the adapter and the connection pool. Maybe call it ``BotocoreRequestsSession``?

kyleknap:
Looks good. Straightforward change. Just had a question about the existing name.

botocore/endpoint.py
jamesls: I can rename it.  Maybe something like `BotocoreHTTPSession` to disambiguate it from `botocore.session.Session`?

kyleknap: Yeah that sounds good.

jamesls:
@kyleknap renamed.

kyleknap:
Thanks. Looks good. :ship:

alanhamlett:
Close the loop... could you link to docs on this and how to use it from boto3?

PR #1027, opened 2016-09-09
Added partion to client meta object

Posnet:
I found that I needed access to the partition being used for a particular client, so that I can write tools that work in both normal AWS regions as well as GovCloud and China.
I view this as a feature similar to the region_name property of the client meta object.

I have added the partition as a property on the client meta object.
I have added unit tests for new meta property.

The one part I am not sure about to what to do if no partition is set for an endpoint_config.
For the moment I have set it to be None, which seems like a reasonable fallback, though I am happy for advice on a better default value. Possibly just setting 'aws' as the fallback partition.


jamesls:
Can you give me a little background info on how you'd use the partition information for a given client?  I'd just like to understand the use case a little better.

Posnet:
This is specifically for a stage in our tooling where we enhance cloudtrails information with data we manually pull from the API.

At the moment we have a hard coded mapping from regions to partition so that our scraping tool (similar to netflix edda) is able to populate the cloudtrails record with more data before forwarding it to our log aggregation platform.

This works fine for us, but it seemed like a cleaner solution to have it available from the client object. Our current mapping code actually uses the botocore data files to lookup the region names and work out the partition, very similar to my submitted pull request.

jamesls:
Seems reasonable to me.  Our `ClientMeta` has `region_name` and `endpoint_url`.  Exposing `partition` seems fine.  We'll need to move the tests over to `tests/functional`.  I think that's why codecov is complaining.  We don't run integration tests during our travis builds.

@kyleknap @JordonPhillips any concerns?

kyleknap:
@jamesls I am fine with this and yes let's get these moved over from integration tests to unit or functional.

Posnet:
I have moved the tests to the unit tests

jamesls:
Looks good to me.  I think I'd still like to also see a functional test that uses a real client via `session.create_client(...)` to ensure everything's hooked up properly.

Posnet:
I have added functional tests as requested.

Posnet:
Any update on whether this can be merged?

Posnet:
Is there anything else I need to do for this pull request?

jamesls:
Could you remove the `tests/coverage.xml` file?  Looks like it was added by accident.

Otherwise, the test updates look good to me, let's just get one more reviewer to look at it. 

cc @kyleknap @JordonPhillips

Posnet:
I have now removed tests/coverage.xml, sorry about that.

jamesls:
Merged, thanks for the pull request.

PR #1028, opened 2016-09-09
Update waiters-2.json

przemyslawj:
Adds a waiter for EMR cluster termination. The waiter waits for up to 300 minutes.

### DIFF
--- a/botocore/data/emr/2009-03-31/waiters-2.json
+++ b/botocore/data/emr/2009-03-31/waiters-2.json
@@ -41,7 +41,7 @@
     "ClusterTerminated": {
       "delay": 30,
       "operation": "DescribeCluster",
-      "maxAttempts": 600,
+      "maxAttempts": 60,
       "acceptors": [
         {
           "state": "success",
--- a/botocore/data/emr/2009-03-31/waiters-2.json
+++ b/botocore/data/emr/2009-03-31/waiters-2.json
@@ -37,6 +37,25 @@
           "expected": "TERMINATED_WITH_ERRORS"
         }
       ]
+    },
+    "ClusterTerminated": {
+      "delay": 30,
+      "operation": "DescribeCluster",
+      "maxAttempts": 600,
+      "acceptors": [
+        {
+          "state": "success",
+          "matcher": "path",
+          "argument": "Cluster.Status.State",
+          "expected": "TERMINATED"
+        },
+        {
+          "state": "failure",
+          "matcher": "path",
+          "argument": "Cluster.Status.State",
+          "expected": "TERMINATED_WITH_ERRORS"
+        }
+      ]
     }
   }
 }


botocore/data/emr/2009-03-31/waiters-2.json
```diff
@@ -37,6 +37,25 @@
           "expected": "TERMINATED_WITH_ERRORS"
         }
       ]
+    },
+    "ClusterTerminated": {
+      "delay": 30,
+      "operation": "DescribeCluster",
+      "maxAttempts": 600,
```
jamesls: Did you mean 60 here instead of 600?  Otherwise, this will block for up to 30 * 600 = 18000 seconds (5 hours).

JordonPhillips: Yeah, that seems like a lot.

przemyslawj:
I agree it's long imeout, but I didn't want to comstrain the usage of this waiter. It's an EMR after all, so clients could be executing some pretty long running tasks. For my particular use case, one hour is enough, but it's difficult to come up with something which will work in general, as I don't think it's possible to customize this?

jamesls:
We do plan on allowing you to customize waiter values, but until that time, it seems like a reasonable default choice is to be consistent with what `ClusterRunning` uses which is `30/60`.

How are you using this waiter such that is requires an hour of wait time?  Are you calling the waiter before calling terminate clusters?  It would help for us to understand your use case better.

przemyslawj:
My code starts an EMR cluster with several steps to run. The cluster is configured to terminate on job completion or error. Immediately after the call to start the cluster it calls _ClusterTerminated_ waiter. When waiter reports success EMR results which were uploaded to S3 are processed.

jamesls:
Ah, I see.  I think that's a good case for being able to customize the waiter values because those values are highly dependent on the specific job you're running.  I think for the default case we should probably go with a more conservative value.  There's an existing pull request for configuring waiter values that I'll take a look at.

jamesls:
:shipit:  I also verified it passed the latest waiter functional tests as well.

botocore/data/emr/2009-03-31/waiters-2.json
```diff
@@ -37,6 +37,25 @@
...
+      "maxAttempts": 60,
+      "acceptors": [
+        {
+          "state": "success",
+          "matcher": "path",
+          "argument": "Cluster.Status.State",
+          "expected": "TERMINATED"
+        },
+        {
+          "state": "failure",
```
kyleknap: @jamesls I am not sure what you are thoughts about this but it seemed interesting so figured to point this out. So I agree it is a good thing to raise any detected errors to the user, but the purpose of this waiter is to wait until terminated. So should we be raising an error if we reached the ``TERMINATED_WITH_ERRORS`` as the cluster did terminate? Honestly, I can see users wanting both behaviors with one side not caring that there were errors and another wanting to know if there were errors when terminating. I think it is fine with this approach, but I can see another waiter added that differentiates between the two.

jamesls: > So should we be raising an error if we reached the TERMINATED_WITH_ERRORS as the cluster did terminate? 

Yes I think we should be (i.e I agree with how it's currently written).

The most common use case here is to call `terminate` and then wait until the cluster is shutdown cleanly.  I think the fact that the service differentiates between `TERMINATED` and `TERMINATED_WITH_ERRORS` suggests we should not be lumping them together in our waiter.

I can't think of a reasonable use case where someone would want to terminate the cluster and wait for it to be terminated where they *wouldn't* want to differentiate the two states.  If they just wanted to terminate a cluster and didn't care about success/failure, I would expect them to not wait on it at all.

Additionally, a user can already get this behavior if they *really* wanted this:

```
try:
   wait()...
except WaiterError:
  pass
```

kyleknap: Yeah that make sense with me.

PR #1032, opened 2016-09-15
Improve functional test for waiter config validation

jamesls:
This improves the waiter config validation to check a few additional cases, primarily related to JMESPath including:

* Verify the JMESPath expression matches something in the output
* Verify the JMESPath expression returns list types for `pathAll` and `pathAny`
* Verify the operation referenced actually produces output.

In order to do this, I made one change to the `ArgumentGenerator` utility class.  In order to rely on the "truth" value of the jmespath search result, I didn't want empty strings being generated.  To account for this, I added the ability for the placeholder strings to optionally use the member names for the values.

As part of this change, I had to update one of the waiter configs as it was now failing the tests:

```
AssertionError: Attempted to use 'pathAll' matcher in waiter 'KeyPairExists' with \
 non list result in JMESPath expression: length(KeyPairs[].KeyName) > `0`
```

I also moved around when we create a waiter so we now get better error messages from pre-existing validation.  Here's what some of the error messages now look like:

```
AssertionError: JMESPath expression did not match anything for \
  waiter 'ConversionTaskCancelled': ConversionTasks[].WillNeverMatch

AssertionError: Could not create waiter 'bundle_task_complete': \
  Invalid jmespath expression: Incomplete expression: "BadJMESPath("

AssertionError: Could not create waiter 'nat_gateway_available': \
  Error processing waiter config: Unknown acceptor: pathUnknown

AssertionError: Waiter 'UserExists' has JMESPath expression with \
  no output shape: OperationModel(name=DeleteUser)
```

cc @kyleknap @JordonPhillips

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -503,12 +503,13 @@ class ArgumentGenerator(object):
     """Generate sample input based on a shape model.
 
     This class contains a ``generate_skeleton`` method that will take
-    an input shape (created from ``botocore.model``) and generate
-    a sample dictionary corresponding to the input shape.
+    an input/output shape (created from ``botocore.model``) and generate
+    a sample dictionary corresponding to the input/output shape.
 
-    The specific values used are place holder values. For strings an
-    empty string is used, for numbers 0 or 0.0 is used.  The intended
-    usage of this class is to generate the *shape* of the input structure.
+    The specific values used are place holder values. For strings either an
+    empty string or the member name can be used, for numbers 0 or 0.0 is used.
+    The intended usage of this class is to generate the *shape* of the input
+    structure.
 
     This can be useful for operations that have complex input shapes.
     This allows a user to just fill in the necessary data instead of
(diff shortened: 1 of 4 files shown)


botocore/utils.py
```diff
@@ -524,8 +525,9 @@ class ArgumentGenerator(object):
         print("Sample input for dynamodb.CreateTable: %s" % sample_input)
 
     """
-    def __init__(self):
-        pass
+    def __init__(self, use_member_names=False):
+        self._use_member_names = use_member_names
```
kyleknap: Is there a particular reason why you made this ``use_member_names`` other than to get a truthy value? If so I wonder if it makes sense to expose a mapping of shape to default value or ``use_truthy_defaults`` optional parameter that defaults to ``True``?

botocore/utils.py
```diff
@@ -524,8 +525,9 @@ class ArgumentGenerator(object):
...
+        self._max_stack_depth = 1
```
kyleknap: Is ``self._max_stack_depth`` used anywhere?

botocore/utils.py
jamesls: It's also useful if you're generating a sample output response and want to test out JMESPath queries.  Using the member name gives you enough context about what you're matching.  Similar to what I use for my re:Invent talk here: https://youtu.be/TnfqJYPjD9I?t=971

botocore/utils.py
jamesls: No I'll remove.  Initially I was going to use it to allow you to specify how many stack levels to recurse down for recursive shapes, but it ended up not being needed so I removed it.

botocore/utils.py
kyleknap: Alright that makes sense

PR #1034, opened 2016-09-19
Add `readlines()` generator to response body streams

davidthornton:
This interface allows file streams to be converted to streams delimited by specified delimiters, instead of the deafult fixed width byte streams returned by `read()`.

See http://stackoverflow.com/a/16890018/119765 for the motivation for this pull request and use-case justification.

### DIFF
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -80,6 +80,21 @@ class StreamingBody(object):
             self._verify_content_length()
         return chunk
 
+    def readlines(self, delimiter='\n'):
+        """Generator that streams lines. The default delimiter
+        is the newline character.
+
+        If the delimiter argument is specified, it will
+        override the default.
+        """
+        line_buffer = ''
+        for byte in self.read():
+            byte = line_buffer + str(byte)
+            lines = byte.split(delimiter)
+            line_buffer = lines.pop()
+            for line in lines:
+                yield line
+
     def _verify_content_length(self):
         # See: https://github.com/kennethreitz/requests/issues/1855
         # Basically, our http library doesn't do this for us, so we have
--- a/tests/unit/test_response.py
+++ b/tests/unit/test_response.py
@@ -46,6 +46,13 @@ class TestStreamWrapper(unittest.TestCase):
         stream = response.StreamingBody(body, content_length=10)
         self.assertEqual(stream.read(), b'1234567890')
 
+    def test_streaming_wrapper_generates_line_streams(self):
+        body = six.BytesIO(b'1234567890\n1234567890')
+        stream = response.StreamingBody(body, content_length=21)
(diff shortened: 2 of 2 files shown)


botocore/response.py
```diff
@@ -80,6 +80,21 @@ def read(self, amt=None):
...
+    def readlines(self, delimiter='\n'):
+        """Generator that streams lines. The default delimiter
+        is the newline character.
+
+        If the delimiter argument is specified, it will
+        override the default.
+        """
+        line_buffer = ''
+        for byte in self.read():
+            byte = line_buffer + str(byte)
```
jamesls: This is creating a new copy of a str each time and creates O(n^2) behavior).  What about reading contents in a fixed chunk size (say 8/16k) and then splitting the lines for each buffer?

jamesls:
I think adding a `readlines()` method is a good addition.  However, I'm concerned about the runtime characteristics of this algorithm.  I think one of the benefits of a `readlines()` over a `.read().splitlines()` is that `readlines()` yields lines without having to read the entire contents into memory.  I would prefer we read a fixed set of data and yield lines within each buffer.

davidthornton:
I'm glad you agree with the idea of this pull request 😀 In respect of the time complexity of this algorithm, I was promised (see the stack overflow link) that this would not be O(n^2) in reality because the generator evaluates the loops lazily.

However, I do not know enough about the Python runtime (together with the context of the `read()` call) to fully confirm this.

neozenith:
@jamesls and @davidthornton, I have opened https://github.com/boto/botocore/pull/1055 to address these issues mentioned in this PR.

tests/unit/test_response.py
```diff
@@ -46,6 +46,13 @@ def test_streaming_wrapper_validates_content_length(self):
         stream = response.StreamingBody(body, content_length=10)
         self.assertEqual(stream.read(), b'1234567890')
 
+    def test_streaming_wrapper_generates_line_streams(self):
+        body = six.BytesIO(b'1234567890\n1234567890')
+        stream = response.StreamingBody(body, content_length=21)
+        lines = stream.readlines()
```
duboviy: - Please do not use the methods `stream.readline()` and `stream.readlines()` for iterate over the stream/file:
```python
# Bad
while True:
    line = your_stream.readline()
    ...

for line in your_stream.readlines():
    ...


# Better
for line in your_stream:
    ...
```

nateprewitt:
This feature ended up being added as part of https://github.com/boto/botocore/commit/a624815eabac50442ed7404f3c4f2664cd0aa784. Now that StreamingBody inherits from IOBase it should have the standard operations most Python users expect around readable/file-like objects.

PR #1044, opened 2016-10-03
Fix exceptions picklability

amw:
Currently exceptions created by botocore fail when sent through `pickle.loads(pickle.dumps(err))`. `pickle` is a module commonly used for serialization of python objects sent between runs or processes. One notable usage example is in `multiprocessing` module which in turn is used in Django's parallel test suite.

Any user who currently performs boto3 operations inside Django's test runner risks that his test suite blows up without running appropriate teardown operations. I've filed a separate fix for Django in django/django#7325.

These commits fix picklability of botocore exception classes. The culprit seemed to be calling parent `Exception.__init__` for defining exception's string representation instead of implementing our own more specialized `__str__` method.

One unrelated thing that surprised me while going through the code is that `ClientError` does not inherit `BotoCoreError`. I was 100% expecting to be able to catch all AWS-related errors with a single error class, but it seems that I would have to write something like this instead:

```
try:
    s3_object.load()
except (BotoCoreError, ClientError) as e:
    ...
```

Is the reason for this to shift blame between AWS/Client? If so I think there would be a room here for another error class that would be parent for both of these.

### DIFF
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -24,9 +24,11 @@ class BotoCoreError(Exception):
     fmt = 'An unspecified error occurred'
 
     def __init__(self, **kwargs):
-        msg = self.fmt.format(**kwargs)
-        Exception.__init__(self, msg)
+        self.message = self.fmt.format(**kwargs)
         self.kwargs = kwargs
+
+    def __str__(self):
+        return self.message
 
 
 class DataNotFoundError(BotoCoreError):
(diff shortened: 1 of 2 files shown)


amw:
Two issues with my PR:

1. Turns out my fix does not work with Python 2.x. I'm not sure how to work with Python2's pickling.

2. I've accidentally omitted `ConnectionClosedError` which has yet another inheritance tree (from "requests" package) and also fails pickling.

JordonPhillips:
Thanks for the pr! Unfortunately we will not be able to accept this until it works on all of our supported python versions.

amw:
I've started to think that aiming for picklability of all errors is pointless especially when you don't catch and translate errors from other libraries that you depend on. Libraries like Django test suite that use `multiprocessing` should just handle errors more gracefully. I'm closing this PR.

PR #1048, opened 2016-10-11
Allow both dualstack and accelerate

JordonPhillips:
cc @kyleknap @jamesls


JordonPhillips:
Capturing some offline discussion:

It would be great if we could make this more generic such that adding new url-based features is painless. This is vastly more complicated than it should be because we actually modify the provided endpoint url and force path style addressing if it looks like an accelerate url. However, there is little reason for us modify the url and we should be able to force path style separately. The next step forward should be to not modify a given endpoint url.

JordonPhillips:
Alright it should be good to look at again. There was some major refactoring on how we set the addressing style that went into this.

botocore/client.py
```diff
@@ -114,10 +114,73 @@ def _register_retries(self, service_model):
...
+
+    def _is_s3_accelerate(self, endpoint_url, s3_config):
+        if s3_config is not None and s3_config.get('use_accelerate_endpoint'):
+            return True
+
+        if endpoint_url is None:
+            return False
+
+        parts = urlsplit(endpoint_url).netloc.split('.')
+        return any(part == 's3-accelerate' for part in parts)
```
jamesls: This will trigger false positives for bucket names that contain "s3-accelerate" (and when an endpoint_url is specified by the user).

botocore/client.py
```diff
@@ -114,10 +114,73 @@ def _register_retries(self, service_model):
...
+        # or not the bucket looks dns compatible.
+        if style == 'virtual':
+            return switch_to_virtual_host_style
+
+        # If path style is configured, no additional steps are needed. If
+        # endpoint_url was specified, don't default to virtual. We could
+        # potentially default provided endpoint urls to virtual hosted
+        # style, but for now it is avoided.
+        if style == 'path' or endpoint_url is not None:
+            return None
```
kyleknap: Question so for the behavior before, if an ``endpoint_url`` was provided did we keep ``fix_s3_host`` handler?

botocore/client.py
JordonPhillips: It won't trigger false positives for containing bucket names since it's an equality check and not an in check. Supporting a specified endpoint url is deliberate because that's what we do now. I think we could do away with blowing out user provided endpoint urls though since we force addressing style separately and that was the real concern anyway.

kyleknap: Sort of agree with @jamesls on this. To be fair there will only ever be one bucket named ``s3-accelerate``, but I think we can make this a bit more robust especially if endpoint_url was using ``s3-accelerate``.

How about adding these checks:
    * Make sure the last two components are ``amazonaws.com``
    * Checking if ``s3-accelerate`` comes directly before ``amazonaws.com`` or two components before ``amazonaws.com``
   * If it comes two components before ``amazonaws.com`` that specific component must be part of a whitelist consisting of just ``dualstack`` for now.

I think if you go with this approach you could also reuse this in the ``s3_accelerate`` handler when deciding what components to use or remove as I was going to have a similar comment there.

JordonPhillips:
@jamesls Would you be supportive of moving them to `ClientArgsCreator`? That's where I initially had them.

botocore/client.py
JordonPhillips: Ah, it could have issues if it's a dns-incompatible bucket name

botocore/client.py
JordonPhillips: No, we removed it [here](https://github.com/boto/botocore/pull/1048/files#diff-e311a009f6abb5ae5c6a861cb37edf53L180). This just doesn't add it in the first place.

kyleknap: Cool

tests/functional/test_s3.py
```diff
@@ -497,6 +513,35 @@ def test_correct_url_used_for_s3():
...
+        region='us-east-1', bucket='bucket', key='key',
+        # Even whitelisted parts cannot be duplicated.
+        customer_provided_endpoint=(
+            'https://s3-accelerate.dualstack.dualstack.amazonaws.com'),
+        expected_url=(
+            'https://s3-accelerate.dualstack.dualstack'
+            '.amazonaws.com/bucket/key'))
+    yield t.case(
+        region='us-east-1', bucket='bucket', key='key',
+        # More than two extra parts is not allowed.
```
kyleknap: If this for more than two extra parts maybe not use ``dualstack`` as you already have a test for double ``dualstack``.

botocore/client.py
```diff
@@ -175,14 +180,38 @@ def _get_s3_addressing_handler(self, endpoint_url, s3_config,
...
             return False
 
-        parts = urlsplit(endpoint_url).netloc.split('.')
-        return any(part == 's3-accelerate' for part in parts)
+        # Accelerate is only valid for Amazon endpoints.
+        netloc = urlsplit(endpoint_url).netloc
+        if not netloc.endswith('amazonaws.com'):
+            return False
+
+        # The first part of the url should always be s3-accelerate
```
kyleknap: I do not think this needs to change, but I had to double check closely that there would be no indexing errors. I think you could simplify with a length check base approach:

```python
parts = netloc.split('.')

if len(parts) not in [3, 4]:
    return False

if parts[0] != 's3-accelerate':
    return False

if len(parts) == 4:
    return parts[1] in ['dualstack']

return True
```

Feel free to use it, but I think what you are doing has no missing edge cases so it should be fine.

tests/functional/test_s3.py
JordonPhillips: The idea is to show that it's being rejected because the number rather than the part not being whitelisted.

botocore/client.py
JordonPhillips: That wouldn't weed out duplicates, and it's not very clear why you're testing for 3 and 4. Actually, is there any reason to constrain based on length? So long as we make sure that each part is in the whitelist and there are no duplicates we'll catch everything anyway and that will remove the need to update this if the url feature matrix gets bigger.

kyleknap: Oh the full code sample would be:
```python
if not netloc.endswith('amazonaws.com'):
    return False

parts = netloc.split('.')

if len(parts) not in [3, 4]:
    return False

if parts[0] != 's3-accelerate':
    return False

if len(parts) == 4:
    return parts[1] in ['dualstack']

return True
```
The reason we would check for length 3 and 4 was the case of ``s3-accelerate.amazonaws.com`` and ``s3-accelerate.{whitelisted-componet}.amazonaws.com``. This approach just makes the logic a bit simpler, but I am fine with what you have right now.

botocore/utils.py
```diff
@@ -774,10 +775,12 @@ def switch_host_s3_accelerate(request, operation_name, **kwargs):
     # before it gets changed to virtual. So we are not concerned with ensuring
     # that the bucket name is translated to the virtual style here and we
     # can hard code the Accelerate endpoint.
-    whitelist = ['dualstack', 'amazonaws', 'com']
     parts = urlsplit(request.url).netloc.split('.')
-    parts = [p for p in parts if p in whitelist]
-    endpoint = 'https://s3-accelerate.' + '.'.join(parts)
+    parts = [p for p in parts if p in S3_ACCELERATE_WHITELIST]
+    endpoint = 'https://s3-accelerate.'
+    if len(parts) > 0:
```
kyleknap: And duplicates are not accounted for because of the initial check for duplication when the handler is registered.

JordonPhillips: Yes. I could make it a set, but I don't know that we really need to do validation at this point.

PR #1050, opened 2016-10-17
Stubber: add requests history

adamchainz:
This allows tests to run code and make simpler assertions on the params that were used rather than having to pass everything to `expected_params`.

### DIFF
--- a/botocore/stub.py
+++ b/botocore/stub.py
@@ -164,6 +164,7 @@ class Stubber(object):
         :param client: The client to add your stubs to.
         """
         self.client = client
+        self._requests = []
         self._event_id = 'boto_stubber'
         self._expected_params_event_id = 'boto_stubber_expected_params'
         self._queue = deque()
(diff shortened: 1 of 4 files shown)


adamchainz:
This is incomplete, but I've opened the PR to discuss the idea. My initial use case was when testing SES calls, I had several tests that really just need to check the different values for HTML body, but because it's a deeply nested parameter it required specifying many other parameters. It would have been easier for me to check it after the stubber with a simple assert as I've semi-documented here.

botocore/stub.py
```diff
@@ -362,3 +388,13 @@ def _validate_response(self, operation_name, service_response):
...
                     "Service response should only contain ResponseMetadata."))
+
+
+class StubRequest(object):
+    """
+    A tracked request that was made to a stubbed client. Has two attributes,
+    ``operation_name`` and ``params``.
+    """
+    def __init__(self, operation_name, params):
+        self.operation_name = operation_name
```
JordonPhillips: These shouldn't be settable after creation, I would rather have them as read-only properties. I would also like for the service name to be added in so that if we create a stubber that works on the session we won't have breakages when adding service name.

adamchainz: There's no such thing as a read-only attribute in python, just magic to mask it. I could do `_operation_name` proxied by an `@property` function that returns `self._operation_name` but then one can still modify `_operation_name`...

JordonPhillips: Maybe I should say "read-only". People can of course still access those attributes, but it's generally understood that they're internal and should not be used. You could make it more forceful with `__operation_name`, but there's no need to go so far.

adamchainz: done

JordonPhillips:
The only thing left (for me) is to add `service_name` to `StubRequest`.

adamchainz:
What should that contain? It's not clear to me

JordonPhillips:
@adamchainz the service name of the client, e.g. `s3`. You can source that from `client.meta.service_name`.

adamchainz:
Tadaaa

JordonPhillips:
I was holding off on this because of codecov, but I just looked closer and it's clearly wrong. Curious. I'm gonna restart it to see if it clears up.

adamchainz:
Curious

adamchainz:
> Can we add a test with multiple stubbed calls?

Good point, added!

> I don't see a whole lot of value of making these things `_internal` and then proxying to them via `@property`

Neither do I but @JordonPhillips made me do it

PR #1052, opened 2016-10-17
Add generated files to git ignore

adamchainz:
These files are generated by running `make html` for the docs or `tox` for tests.


JordonPhillips:
Thanks for the pr!

PR #1053, opened 2016-10-17
Add argument passing to tests

adamchainz:
This allows e.g. `tox -e py27 -- functional/test_stub.py` to just run that one test file, making the red-green-refactor cycle faster.

### DIFF
--- a/scripts/ci/run-tests
+++ b/scripts/ci/run-tests
@@ -4,17 +4,34 @@
 # binary package not from the CWD.
 
 import os
+import sys
 from subprocess import check_call
 
 _dname = os.path.dirname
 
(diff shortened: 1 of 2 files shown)


jamesls:
I might be missing something here, but during development, is there a reason you wouldn't just run `nosetests` directly?  i.e `nosetests tests/functional/test_stub.py`.  That's usually what I do, so I'm just curious.

adamchainz:
When I come to a new project I don't know how to set it up, but if I see `tox` I know it will do it for me, so I prefer to keep it as then TOOWTDI. Then I don't need to make a virtualenv myself, or maintain it, plus I can run tests on multiple envs every time.

PR #1054, opened 2016-10-17
Stubber: made ANY usable for nested parameters

adamchainz:
Use the same compare-all class that `unittest.mock.ANY` uses to enable deep comparisons to 'just work'.

### DIFF
--- a/.changes/next-release/feature-Stub-87309.json
+++ b/.changes/next-release/feature-Stub-87309.json
@@ -1,0 +1,5 @@
+{
+  "category": "Stub",
+  "type": "feature",
+  "description": "Made ANY usable for nested parameters"
+}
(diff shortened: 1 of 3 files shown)


adamchainz:
This enables requests with nested parameters that have variable data to be tested easily with the stubber, and also is less surprising since `ANY` reminds users of `unittest.mock.ANY`, anyway.

Could also be used to solve the problem I had in #1050, but it would still require specifying `ANY` for each parameter/nested parameter that needs ignoring, which can be a lot for some service methods.

adamchainz:
ok sure

adamchainz:
Added

JordonPhillips:
Awesome, thanks again!

PR #1055, opened 2016-10-18
Add readline() method to S3 streaming

neozenith:
Add the functionality to read a delimited stream of data from S3.

Stackoverflow answer justifying use case:
http://stackoverflow.com/a/16890018/622276

Related PR from David Thornton that got the ball rolling:
https://github.com/boto/botocore/pull/1034

Addressed in this PR as well:
- Buffering a max of 16kb before performing split and yield operations rather than checking on every byte appended to the buffer from the stream
- Prior solution did not yield final result if it wasn't terminated by delimiter.

### DIFF
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -90,7 +90,7 @@ class StreamingBody(object):
         line_buffer = ''
         BUFFER_SIZE = 2**14 # 16k
         for byte in self.read():
-            line_buffer = line_buffer + byte
+            line_buffer = line_buffer + str(byte)
             # Where possible accumulate BUFFER_SIZE bytes before performing
             # Split and yield
             if len(line_buffer) >= BUFFER_SIZE:
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -80,6 +80,33 @@ class StreamingBody(object):
             self._verify_content_length()
         return chunk
 
+    def readlines(self, delimiter='\n'):
+        """Generator that streams lines. The default delimiter
+        is the newline character.
+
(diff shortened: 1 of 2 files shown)


neozenith:
Ok going to have to circle back and address how that failing test works. Generating a large repeating string is different between python 2 and 3 it seems.

neozenith:
Actually it is struggling with difference in strings and byte arrays between Python 2 and 3.

botocore/response.py
```diff
@@ -80,6 +80,33 @@ def read(self, amt=None):
...
 
+    def readlines(self, delimiter='\n'):
+        """Generator that streams lines. The default delimiter
+        is the newline character.
+
+        If the delimiter argument is specified, it will
+        override the default.
+        """
+        line_buffer = ''
+        BUFFER_SIZE = 2**14 # 16k
```
allan-simon: can we have this as an argument, or as a constant at the module level, so it can overrided if needed ?

botocore/response.py
```diff
@@ -80,6 +80,33 @@ def read(self, amt=None):
...
+        for byte in self.read():
```
allan-simon: isn't read without parameter going to read all the data at once ?

neozenith:
After fighting with the differences between python 2 and 3 it is clear that getting it to interpret the byte stream and then correctly delimit for whatever token (which could be a variable amount of bytes in utf-8) is more than I have time for at the moment so I am going to close this Pull Request but thanks for the review points.

Issue #1059, opened 2016-10-22
Blank params in base URL not preserved in SigV4 Presigned URL 


eukaryote:
Generating a sigv4 presigned URL when the base URL contains a query parameter without a value results in a presigned URL that doesn't contain the parameter at all, because parse_qs strips blank parameters by default. 

For example, 'https://s3.amazonaws.com/mybucket/mykey?uploads' is a base URL for creating a multipart upload URL, and the resulting presigned URL that is generated by the following code no longer has an 'uploads' param when SigV4 is used:

```
s3.generate_presigned_url(
    'create_multipart_upload',
    Params={
        'Bucket': 'mybucket',
        'Key': 'mykey',
    }
)
```

JordonPhillips:
Lets' go ahead and move all discussion to the PR

PR #1060, opened 2016-10-22
sigv4 presigned urls should preserve blank params

eukaryote:
A unit test and fix for #1059, which is a sigv4 bug related to presigned urls where the base url contains a query parameter without a value, such as '/bucket/key?uploads' for an s3 create_multipart_upload url.

The first commit contains a unit test that fails, and the second commit updates SigV4QueryAuth._modify_request_before_signing to use the keep_blank_values param of parse_qs in order to preserve parameters with no value.

### DIFF
--- a/.changes/next-release/bugfix-s3-27146.json
+++ b/.changes/next-release/bugfix-s3-27146.json
@@ -1,0 +1,5 @@
+{
+  "category": "s3", 
+  "type": "bugfix", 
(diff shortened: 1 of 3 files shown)


JordonPhillips:
Nice catch! Thanks for the PR

eukaryote:
Added changelog entry. Thanks for reviewing so quickly.

eukaryote:
Is there anything I can do to help get this into a release?

JordonPhillips:
Sorry, this slipped through the cracks.

PR #1062, opened 2016-10-25
Cloudformation documentation update

### LINKED ISSUE #1058: get_template returns string or dict?
The documentation at http://boto3.readthedocs.io/en/latest/reference/services/cloudformation.html#CloudFormation.Client.get_template says this about the response syntax of `get_template`:
```
TemplateBody (string) --
Structure containing the template body. (For more information, go to Template Anatomy in the AWS CloudFormation User Guide.)

AWS CloudFormation returns the same template that was used when the stack was created.
```
From the `(string)` part, I was expecting to get a single string representation of the template, but when I try to use it, I get a `dict` object:
```
>>> import boto3
>>> region='region'
>>> stack_name='stack-name'
>>> session = boto3.session.Session(region_name=region)
>>> client = session.client('cloudformation')
>>> template = client.get_template(StackName = stack_name)
>>> type(template['TemplateBody'])
<type 'dict'>
```
Am I missing something or is the documentation due to be updated?

stealthycoin:
Updated cloud formation documentation, changes get_template return type from string to dict.

Fixes #1058 

cc @kyleknap @jamesls @JordonPhillips

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -586,14 +586,13 @@ def document_cloudformation_get_template_return_type(section, event_name, **kwar
         template_body_section = section.get_section('TemplateBody')
         type_section = template_body_section.get_section('param-type')
         type_section.clear_text()
-        type_section.write(':type TemplateBody: dict')
+        type_section.write('(*dict*) --')
     elif 'response-example' in event_name:
         parent = section.get_section('structure-value')
         param_line = parent.get_section('TemplateBody')
         value_portion = param_line.get_section('member-value')
         value_portion.clear_text()
         value_portion.write('{}')
-        
 
 def switch_host_machinelearning(request, **kwargs):
     switch_host_with_param(request, 'PredictEndpoint')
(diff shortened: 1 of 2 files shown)


botocore/handlers.py
```diff
@@ -581,6 +581,27 @@ def document_glacier_tree_hash_checksum():
...
+        template_body_section = section.get_section('TemplateBody')
+        type_section = template_body_section.get_section('param-type')
+        type_section.clear_text()
+        type_section.write(':type TemplateBody: dict')
+    elif 'response-example' in event_name:
+        parent = section.get_section('structure-value')
+        param_line = parent.get_section('TemplateBody')
+        value_portion = param_line.get_section('member-value')
+        value_portion.clear_text()
+        value_portion.write('{"AWSTemplateFormatVersion" : "version date", '
```
JordonPhillips: Lets leave this as an empty dict. Since this is modeled as a string the structure could get updated without us knowing. There's a link in the description to the actual structure so it should be okay.

stealthycoin:
Old documentation:
before

New documentation:
after

PR #1064, opened 2016-10-28
Always populate ResponseMetadata

JordonPhillips:
This makes sure that, so long as the response was able to be parsed
into a dict, the ResponseMetadata key is always provided. This is
important because we use that for more than just what the service
directly returns.


JordonPhillips:
This is an edge case, but some services are reported to occasionally omit the data we would need to have been putting the response in previously. Since some customers are beginning to rely on additional data we add to that metadata, we should still be putting it in.

PR #1068, opened 2016-11-03
Emit service-data-loaded event on client creation

JordonPhillips:
Loading service data via the Session currently emits an event, but
the same event isn't emitted when loading that same data during
client creation. This PR updates that event so that it can be
emitted during client creation. This will allow us to perform any
transformations to the service models upon loading.

This also has the side-benefit of reducing code duplication, as
that event is used to register retry handlers for the CLI, but since
it wasn't being fired for botocore clients we had to add them in
manually.


botocore/handlers.py
```diff
@@ -210,36 +210,38 @@ def _needs_s3_sse_customization(params, sse_member_prefix):
...
+        return
+
     retry_config = translate.build_retry_config(
         endpoint_prefix, original_config['retry'],
         original_config.get('definitions', {}))
     return retry_config
 
 
-def _register_for_operations(config, session, service_name):
+def _register_for_operations(config, event_emitter, service_name):
```
jamesls: Interesting.  Per operation retries don't appear to be explicitly dealt with in client configs.  Are you sure this is exactly the same functional behavior as before?

JordonPhillips:
Closing this out as it isn't needed anymore

PR #1071, opened 2016-11-08
Fix retrys for API gateway so they will work reasonably by default

michael-dev2rights:
This change makes API gateway work properly for me.  Without it I get timeouts.  Please see 

https://github.com/boto/boto3/issues/876

### DIFF
--- a/botocore/data/_retry.json
+++ b/botocore/data/_retry.json
@@ -89,7 +89,7 @@
         "max_attempts": 10,
         "delay": {
           "type": "exponential",
-          "base": 5,
+          "base": 0.25,
           "growth_factor": 2
         }
       }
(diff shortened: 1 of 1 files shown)


botocore/data/_retry.json
```diff
@@ -84,6 +84,16 @@
           "too_many_requests": {"$ref": "too_many_requests"}
       }
     },
+    "apigateway": {
+      "__default__": {
```
JordonPhillips: This changes the strategy for every apigateway operation, not just the operations specifically impacted.

jamesls:
I think this may need to be something people will opt in to.  This has come up a few times, but at a rate of 2 calls per minute, this gets very close just looking like botocore has hung (or at least exhibits the same type of behavior).

I would propose we get the configurable retries support added to botocore so that people can tune these numbers as they need.

PR #1075, opened 2016-11-09
Enforce GMT == UTC

### LINKED ISSUE #1070: Timezone parsing error in develop branch
I'm running on Ubuntu 16.04 LTS.  I have python I have pulled and checked out develop branch at commit

e079c0db2e2fc28684c4f6a87298d0cf592e6e29

I installed the dependencies in requirements.txt (see attached file)

[pip-list.txt](https://github.com/boto/botocore/files/578672/pip-list.txt)

When I run 

`nosetests tests/unit/`

Then the output ends 

```
FAIL: test_parse_rfc822 (tests.unit.test_utils.TestParseTimestamps)
  File "/home/mikedd/dev/botocore/tests/unit/test_utils.py", line 249, in test_parse_rfc822
    datetime.datetime(2002, 10, 2, 13, 0, tzinfo=tzutc()))
AssertionError: datetime.datetime(2002, 10, 2, 13, 0, tzinfo=tzlocal()) != datetime.datetime(2002, 10, 2, 13, 0, tzinfo=tzutc())

Ran 1615 tests in 21.648s

FAILED (failures=1)

```

actual output log attached.  

[nose-unit-run-botocore-develop-e079c0db-log.txt](https://github.com/boto/botocore/files/578670/nose-unit-run-botocore-develop-e079c0db-log.txt)

JordonPhillips:
In the UK the time switches from GMT to BST and back as part of
their daylight savings time. time.tzname will therefore report
both time zones. dateutil sees that the time zone is a local time
zone and so parses it as local time, but it ends up being BST
instead of GMT. To remedy this issue we can provide a time zone
context which will enforce GMT == UTC.

Fixes #1070

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -331,7 +331,10 @@ def parse_timestamp(value):
         except (TypeError, ValueError):
             pass
     try:
-        return dateutil.parser.parse(value)
+        # In certain cases, a timestamp marked with GMT can be parsed into a
+        # different time zone, so here we provide a context which will
+        # enforce that GMT == UTC.
+        return dateutil.parser.parse(value, tzinfos={'GMT': tzutc()})
     except (TypeError, ValueError) as e:
         raise ValueError('Invalid timestamp "%s": %s' % (value, e))
 
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -248,6 +248,18 @@ class TestParseTimestamps(unittest.TestCase):
             parse_timestamp('Wed, 02 Oct 2002 13:00:00 GMT'),
             datetime.datetime(2002, 10, 2, 13, 0, tzinfo=tzutc()))
 
+    def test_parse_gmt_in_uk_time(self):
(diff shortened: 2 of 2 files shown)


michael-dev2rights:
Confirmed that this branch fixes #1070 and now all tests pass locally for me.

jamesls:
We're sure that even within regions that are in BST, GMT is still always treated as UTC correct?  

If so then :shipit:

JordonPhillips:
@jamesls According to [RFC822](https://tools.ietf.org/html/rfc822#section-5):

> Time zone may be indicated in several ways. "UT" is Universal Time (formerly called "Greenwich Mean Time"); "GMT" is permitted as a reference to Universal Time.

This seems very clear: GMT == Universal Time. The RFC defines several other valid timezones, but BST is not one of them.

As for ISO8601, I do not have access to the full spec (it's behind a paywall). However everything I've been able to find has indicated that you must specify timezones as an offset to utc.

PR #1077, opened 2016-11-11
Support loading json extra files

JordonPhillips:
These files will define extra data to be added into the service model
upon loading. These will be used to make additive changes to the
model which reflect sdk-specific information that is not sent over
the wire.

The deep merge is, of course, quite expensive. In a worst case scenario
(a very large service model and an equally sized extras file) client creation
takes ~4x as long on python 2 and ~2x as long on python 3. However, these extra
files will never be more than adding a few parameters. The impact given a
small extra file is negligible.

### DIFF
--- a/botocore/loaders.py
+++ b/botocore/loaders.py
@@ -16,6 +16,7 @@ This module provides the classes that are used to load models used
 by botocore.  This can include:
 
     * Service models (e.g. the model for EC2, S3, DynamoDB, etc.)
+    * Service model extras which customize the service models
     * Other models associated with a service (pagination, waiters)
     * Non service-specific config (Endpoint data, retry config)
 
@@ -24,6 +25,7 @@ Loading a module is broken down into several steps:
     * Determining the path to load
     * Search the data_path for files to load
     * The mechanics of loading the file
+    * Searching for extras and applying them to the loaded file
 
 The last item is used so that other faster loading mechanism
 besides the default JSON loader can be used.
@@ -75,6 +77,7 @@ this structure for service models::
       |       |-- paginators-1.json
       |       |-- service-2.json
       |       |-- waiters-2.json
+      |       |-- service-2.sdk-extras.json
 
 
 That is:
@@ -92,6 +95,11 @@ the ``version`` key within the model, this version is also part of the filename
 so that code does not need to load the JSON model in order to determine which
 version to use.
 
+The ``sdk-extras`` and similar files represent extra data that needs to be
+applied to the model after it is loaded. Data in these files might represent
+information that doesn't quite fit in the original models, but is still needed
+for the sdk. For instance, additional operation parameters might be added here
+which don't represent the actual service api.
 """
 import os
 import logging
@@ -100,6 +108,7 @@ from botocore import BOTOCORE_ROOT
 from botocore.compat import json
 from botocore.compat import OrderedDict
 from botocore.exceptions import DataNotFoundError, UnknownServiceError
+from botocore.utils import deep_merge
 
 
 logger = logging.getLogger(__name__)
@@ -205,9 +214,11 @@ class Loader(object):
     # For convenience we automatically add ~/.aws/models to the data path.
     CUSTOMER_DATA_PATH = os.path.join(os.path.expanduser('~'),
                                       '.aws', 'models')
+    BUILTIN_EXTRAS_TYPES = ['sdk']
 
     def __init__(self, extra_search_paths=None, file_loader=None,
-                 cache=None, include_default_search_paths=True):
+                 cache=None, include_default_search_paths=True,
+                 include_default_extras=True):
         self._cache = {}
         if file_loader is None:
             file_loader = self.FILE_LOADER_CLASS()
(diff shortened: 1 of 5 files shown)


botocore/loaders.py
```diff
@@ -205,9 +207,11 @@ class Loader(object):
     # For convenience we automatically add ~/.aws/models to the data path.
     CUSTOMER_DATA_PATH = os.path.join(os.path.expanduser('~'),
                                       '.aws', 'models')
+    BUILTIN_TOOL_EXTRAS = ['sdk', 'botocore']
```
jamesls: I agree on removing the `tool` part.  It's really up to the higher level layers to decide the semantics of the file.  From the loaders perspective they're just files that get deep merged on top of other files.  I'd prefer something like extras or extras_types as well.  

I'd also prefer to just have a single built in extra type (either botocore or sdk) for now unless we really need to have two.

botocore/loaders.py
```diff
@@ -205,9 +207,11 @@ class Loader(object):
...
 
     def __init__(self, extra_search_paths=None, file_loader=None,
-                 cache=None, include_default_search_paths=True):
+                 cache=None, include_default_search_paths=True,
+                 additional_tool_extras=None):
```
jamesls: I don't think we need `additional_tool_extras` for now.  I'd prefer to just have the equivalent of `include_default_search_paths` which can toggle on/off the behavior for the .extra.json files.

botocore/loaders.py
```diff
@@ -319,7 +331,8 @@ def list_api_versions(self, service_name, type_name):
         return sorted(known_api_versions)
 
     @instance_cache
-    def load_service_model(self, service_name, type_name, api_version=None):
+    def load_service_model(self, service_name, type_name, api_version=None,
+                           load_extras=True):
```
jamesls: I don't think we need this `load_extras` argument added here.  This is something we can't remove in the future so unless we actually need this now, I'd prefer to just have a toggle at the `__init__` that allow you to turn this behavior on/off.

botocore/loaders.py
```diff
@@ -355,7 +372,32 @@ def load_service_model(self, service_name, type_name, api_version=None):
...
+            return model
+
+        # Load in all the extras
+        for tool_name in self._tool_extras:
+            self._load_extras(
+                model, service_name, type_name, api_version, tool_name)
+
+        return model
+
+    def _load_extras(self, service_model, service_name, type_name,
```
jamesls: This doesn't seem like the right verb.  In the loaders, "load" is actually finding something on disk and deserializing it.  Here we'll really merging/updating an existing model.  I'd prefer to have that be named distinctly from "load".  Something like `self._merge_extras`.

botocore/loaders.py
```diff
@@ -355,7 +372,32 @@ def load_service_model(self, service_name, type_name, api_version=None):
...
+        full_path = os.path.join(service_name, api_version, extras_type_name)
+
+        try:
+            extra_data = self.load_data(full_path)
+        except DataNotFoundError:
+            return
+
+        merge_data = extra_data.get('merge')
+        if merge_data is not None:
+            deep_merge(service_model, merge_data)
```
jamesls: I don't think you're testing the case where the `merge` key does not exist.

FWIW, I would prefer that the "processing" logic is delegated to another class.  Even in the simple case of a single merge key, it's somewhat difficult to just isolate the "merge" logic, and I can only see this logic growing in the future (e.g `add`, `remove`, etc).

For now though we should at least get a test added for `merge` not existing (or make it required)

tests/unit/test_utils.py
```diff
@@ -1016,6 +1017,54 @@ def test_uses_dualstack(self):
             'https://s3-accelerate.dualstack.amazonaws.com/foo/key.txt')
 
 
+class TestDeepMerge(unittest.TestCase):
+    def test_simple_merge(self):
```
jamesls: We should add tests for other types, particularly for lists so we can capture the existing behavior.

JordonPhillips:
Summary of changes:

* Made the tests more comprehensive
* tool_extras -> extras_types
* Only builtin extras type is `sdk`
* Boolean flag to disable extras loading moved to loader init
* Processing extras delegated to another class

botocore/loaders.py
```diff
@@ -397,3 +439,23 @@ def _potential_locations(self, name=None, must_exist=False,
...
+
+class ExtrasProcessor(object):
+    """Processes data from extras files into service models."""
+    def process(self, service_data, extras_data):
+        """Processes data from a list of loaded extras files into a model
+
+        :type service_data: dict
+        :param service_data: The service model to load all the extras into.
+
+        :type extras_data: iterable of dict
```
kyleknap: Based on the name of ``extras_data``, I would have expected this to be a dict because ``service_data`` is a dict. I wonder if it makes sense to rename this to something else? Something that makes it seem like a list or maybe making it ``*extra_data`` in the signature?

kyleknap: I like it. Let's do ``extra_models`` and ``original_model``.

PR #1081, opened 2016-11-14
Idempotent param

stealthycoin:
cc @kyleknap @JordonPhillips @jamesls

### DIFF
--- a/botocore/docs/method.py
+++ b/botocore/docs/method.py
@@ -180,8 +180,8 @@ def document_model_driven_method(section, method_name, operation_model,
     context = {
         'special_shape_types': {
             'streaming_input_shape': operation_model.get_streaming_input(),
-            'streaming_output_shape': operation_model.get_streaming_output()
-        }
+            'streaming_output_shape': operation_model.get_streaming_output(),
+        },
     }
 
     if operation_model.input_shape:
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -22,6 +22,7 @@ import xml.etree.cElementTree
 import copy
 import re
 import warnings
+import uuid
 
 from botocore.compat import unquote, json, six, unquote_str, \
     ensure_bytes, get_md5, MD5_AVAILABLE
@@ -107,6 +108,14 @@ def decode_console_output(parsed, **kwargs):
             parsed['Output'] = value
         except (ValueError, TypeError, AttributeError):
             logger.debug('Error decoding base64', exc_info=True)
+
+
+def generate_idempotent_uuid(params, model, **kwargs):
+    for name in model.idempotent_members:
+        if name not in params:
+            params[name] = str(uuid.uuid4())
+            logger.debug("injecting idempotency token (%s) into param '%s'." %
+                         (params[name], name))
 
 
 def decode_quoted_jsondoc(value):
@@ -594,6 +603,7 @@ def document_cloudformation_get_template_return_type(section, event_name, **kwar
         value_portion.clear_text()
         value_portion.write('{}')
 
+
 def switch_host_machinelearning(request, **kwargs):
     switch_host_with_param(request, 'PredictEndpoint')
 
@@ -764,6 +774,8 @@ BUILTIN_HANDLERS = [
     ('after-call.cloudformation.GetTemplate', json_decode_template_body),
     ('after-call.s3.GetBucketLocation', parse_get_bucket_location),
 
+    ('before-parameter-build', generate_idempotent_uuid),
+
     ('before-parameter-build.s3', validate_bucket_name),
 
     ('before-parameter-build.s3.ListObjects',
@@ -864,7 +876,7 @@ BUILTIN_HANDLERS = [
     # Cloudformation documentation customizations
     ('docs.*.cloudformation.GetTemplate.complete-section',
      document_cloudformation_get_template_return_type),
-    
+
     # UserData base64 encoding documentation customizations
     ('docs.*.ec2.RunInstances.complete-section',
      document_base64_encoding('UserData')),
--- a/botocore/model.py
+++ b/botocore/model.py
@@ -49,7 +49,8 @@ class Shape(object):
     SERIALIZED_ATTRS = ['locationName', 'queryName', 'flattened', 'location',
                         'payload', 'streaming', 'timestampFormat',
                         'xmlNamespace', 'resultWrapper', 'xmlAttribute']
-    METADATA_ATTRS = ['required', 'min', 'max', 'sensitive', 'enum']
+    METADATA_ATTRS = ['required', 'min', 'max', 'sensitive', 'enum',
+                      'idempotencyToken']
     MAP_TYPE = OrderedDict
 
     def __init__(self, shape_name, shape_model, shape_resolver=None):
@@ -128,6 +129,7 @@ class Shape(object):
             * enum
             * sensitive
             * required
+            * idempotencyToken
 
         :rtype: dict
         :return: Metadata about the shape.
@@ -406,6 +408,16 @@ class OperationModel(object):
             return None
         return self._service_model.resolve_shape_ref(
             self._operation_model['output'])
+
+    @CachedProperty
+    def idempotent_members(self):
+        input_shape = self.input_shape
+        if not input_shape:
+            return []
+
+        return [name for (name, shape) in input_shape.members.items()
+                if 'idempotencyToken' in shape.metadata and
+                shape.metadata['idempotencyToken']]
 
     @CachedProperty
     def has_streaming_input(self):
--- a/tests/functional/test_ec2.py
+++ b/tests/functional/test_ec2.py
@@ -1,0 +1,70 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from tests import unittest
+from tests.functional.docs import BaseDocsFunctionalTest
+from botocore.stub import Stubber, StubAssertionError, ANY
+import botocore.session
+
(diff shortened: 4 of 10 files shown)


tests/functional/test_ec2.py
```diff
@@ -0,0 +1,66 @@
...
+        self.stubber.add_response(
+            function_name, service_response, expected_params_notoken)
+        self.stubber.add_response(
+            function_name, service_response, expected_params_token)
+
+        self.stubber.activate()
+
+        # Check the ClientToken parameter exsits
+        # after injection and before the call
+        self.client.meta.events.register_first(
```
kyleknap: This is sort of confusing on what is happening here. For this functional test, here is what I recommend:


1) Split these tests out to two test methods: one where the user provides the parameter and one where you inject it for the user. In general if you are making two somewhat unrelated assertions, it is a sign you should break them up into separate tests because if the first assertion fails you will not be able to see if the second assertion fails as well.
2) For the case where you are injecting the idempotency token for the user, I would recommend instead of registering that lambda event, use ``botocore.stub.ANY``
3) For stubbers, feel free to use the context manager to activate and deactivate the stubber.

botocore/handlers.py
```diff
@@ -109,6 +110,24 @@ def decode_console_output(parsed, **kwargs):
             logger.debug('Error decoding base64', exc_info=True)
 
 
+def generate_idempotent_uuid(params, model, **kwargs):
```
kyleknap: It would be great to have a unit test here. Mainly to test that this function properly injects a uuid. You would not have to guess the exact value of the uuid but make sure it looks like an uuid.

botocore/handlers.py
```diff
@@ -764,6 +784,8 @@ def _replace_content(self, section):
     ('after-call.cloudformation.GetTemplate', json_decode_template_body),
     ('after-call.s3.GetBucketLocation', parse_get_bucket_location),
 
+    ('before-parameter-build.*.*', generate_idempotent_uuid),
```
stealthycoin: Hmm without the *.* it didn't get run as far as I could tell.

tests/functional/test_ec2.py
stealthycoin: 1) Sounds good
2) I tried that first and it threw fit because I think the token gets inserted after the stubber looks at it? I'm not sure I can try that way again and fiddle with it more. I just figured the stubber didn't support what I wanted to do so I went around it.
3) Will do.

tests/functional/test_s3.py
```diff
@@ -11,15 +11,11 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 from tests import unittest, mock, BaseSessionTest, create_session
-import os
-import nose
```
stealthycoin: Yes, its in the commit message.

botocore/handlers.py
```diff
@@ -109,6 +110,24 @@ def decode_console_output(parsed, **kwargs):
             logger.debug('Error decoding base64', exc_info=True)
 
 
+def generate_idempotent_uuid(params, model, **kwargs):
+    input_shape = model.input_shape
+    if not input_shape or not hasattr(input_shape, 'members'):
+        return
+    # Might be a mock model from a test that doesn't care about this
+    if not hasattr(input_shape.members.items(), '__iter__'):
```
JordonPhillips: Why would all tests need that? This should only be running in functional / integration tests aside from the unit tests directly related to it.

tests/functional/test_ec2.py
JordonPhillips: Personally I prefer the context manager. It was added later on, which is why it isn't used in some tests.

botocore/model.py
```diff
@@ -408,6 +408,26 @@ def output_shape(self):
             self._operation_model['output'])
 
     @CachedProperty
+    def has_idempotent_members(self):
+        input_shape = self.input_shape
```
JordonPhillips: It would probably be better to just use `get_idempotent_members` here.

tests/unit/test_idempotency.py
```diff
@@ -0,0 +1,32 @@
...
+from botocore.handlers import generate_idempotent_uuid
+
+
+class TestIdempotencyInjection(unittest.TestCase):
+    def setUp(self):
+        self.uuid_pattern = re.compile(
+            '^[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}$',
+            re.I)
+
+    def test_injection(self):
```
JordonPhillips: You should also add a test for when the param is already provided.

JordonPhillips:
I think all that's really missing now is documentation for idempotent parameters.

tests/functional/test_ec2.py
```diff
@@ -0,0 +1,77 @@
...
+            'PurchaseRequests': [
+                {'PurchaseToken': 'foo',
+                 'InstanceCount': 123}],
+            'ClientToken': ANY
+        }
+        self.stubber.add_response(
+            self.function_name, self.service_response, expected_params)
+
+        with self.stubber:
+            try:
```
kyleknap: I do not think you need this try, except, finally for a couple reasons:

1) The ``self.params_seen`` will always get reset to an empty list because it is defined in ``setUp`` and ``setUp`` gets called for each test.
2) There does not appear to be too much benefit for catching the StubAssertionError and then re-raising it. It seems unneeded.

Let me know what you think.

tests/functional/test_ec2.py
```diff
@@ -0,0 +1,77 @@
...
+        self.client = self.session.create_client(
+            'ec2', self.region)
+        self.stubber = Stubber(self.client)
+        self.service_response = {}
+        self.params_seen = []
+
+        # Record all the parameters that get seen
+        self.client.meta.events.register_first(
+            'before-call.*.*',
+            lambda model, params, **kwargs: self.params_seen.extend(
```
kyleknap: I typically try to avoid lambda's as they are harder to debug when things start going wrong. I usually define a function on the TestCase and register the method to the event.

tests/functional/test_ec2.py
```diff
@@ -75,3 +76,10 @@ def test_insert_idempotency_token(self):
                 self.fail(str(e))
             finally:
                 self.params_seen = []
+
+
+class TestEC2Docs(BaseDocsFunctionalTest):
```
JordonPhillips: This should be moved to `tests/functional/docs/test_ec2.py`

botocore/docs/method.py
```diff
@@ -207,6 +208,19 @@ def document_model_driven_method(section, method_name, operation_model,
                 request_params_section, operation_model.input_shape,
                 include=include_input, exclude=exclude_input)
 
+
+    # Find idempotent params and append documentation
+    if 'request-params' in section.available_sections:
+        param_section = section.get_section('request-params')
+        # if operation_model.name == 'PurchaseScheduledInstances':
+        #     pass #import pdb; pdb.set_trace();
```
JordonPhillips: These comments should be removed

stealthycoin:
screen shot 2016-11-29 at 2 45 09 pm

tests/functional/test_ec2.py
JordonPhillips: I think this is the last thing that needs to be addressed

botocore/docs/method.py
```diff
@@ -207,6 +208,17 @@ def document_model_driven_method(section, method_name, operation_model,
                 request_params_section, operation_model.input_shape,
                 include=include_input, exclude=exclude_input)
 
+
+    # Find idempotent params and append documentation
+    if 'request-params' in section.available_sections:
```
kyleknap: Make sure you run flake8. These lines look way too long.

botocore/docs/method.py
```diff
@@ -207,6 +208,17 @@ def document_model_driven_method(section, method_name, operation_model,
...
+        param_section = section.get_section('request-params')
+        for param_name in param_section.available_sections:
+            if param_name in context.get('special_shape_types', {}).get('idempotent_members', []):
+                docs = param_section.get_section(param_name).get_section('param-documentation')
+                docs.writeln('This field is autopopulated if not provided.')
```
kyleknap: So based on the example you had. It looks like a block quote, which does not look good. Could remove ``writeln`` and use ``write`` instead to append on the value?

botocore/docs/method.py
```diff
@@ -207,6 +208,17 @@ def document_model_driven_method(section, method_name, operation_model,
                 request_params_section, operation_model.input_shape,
                 include=include_input, exclude=exclude_input)
 
+
+    # Find idempotent params and append documentation
+    if 'request-params' in section.available_sections:
+        param_section = section.get_section('request-params')
+        for param_name in param_section.available_sections:
```
kyleknap: I would prefer we move this in ``params.py`` and ``shape.py``.  It seems weird to be adding parameter specific logic not in one of these files, especially with the use of the context (it seems unneeded right now). You can probably follow the same logic as streaming input shape.

tests/functional/test_ec2.py
kyleknap: Any update on this?

stealthycoin: Whoops I missed this one...

botocore/docs/method.py
stealthycoin: Yea I had automatic flake8 on all the time and then I turned that off so I forgot to run it manually.

stealthycoin:
I changed the wording to match elsewhere in the docs. 
screen shot 2016-11-30 at 2 37 47 pm

Moved everything to `shape.py` and `params.py` I found that the metadata dict is populated with the idempotency property, so it was easier than I thought it would be. Still probably needs to be made more general.

botocore/docs/shape.py
```diff
@@ -83,6 +83,17 @@ def traverse_and_document_shape(self, section, shape, history,
                     section=section)
             history.pop()
 
+    def _append_special_param_docstring(self, shape, section):
+        special_docs = {
+            'idempotencyToken':
+            'Please note this parameter is automatically populated '
```
JordonPhillips: This formatting is confusing, and I'm a bit surprised it works. Initially I thought this was a list.

botocore/docs/shape.py
```diff
@@ -83,6 +83,17 @@ def traverse_and_document_shape(self, section, shape, history,
...
+            'if it is not provided. Including this parameter is not required\n'
```
JordonPhillips: Trailing period is missing

stealthycoin:
screen shot 2016-11-30 at 4 13 58 pm

PR #1086, opened 2016-11-16
Add Route53 action: CreateVPCAssociationAuthorization

herbygillot:
To permit the association between a Route53 hosted zone and a VPC from another account.

I am aware that there is a chance that this PR may be rejected due to the contribution guidelines.  That said, I'm hoping that this API call makes it in. :) Thank you.


JordonPhillips:
As you expected, we cannot accept changes to the service models. Unfortunately I cannot comment on when these changes might be released.

herbygillot:
Well, worth a shot. Thanks anyway.

agassner:
Looking forward to be able to make calls to CreateVPCAssociationAuthorization.
I'm keen to help if possible.

JordonPhillips:
@agassner This operation has since been added: http://boto3.readthedocs.io/en/latest/reference/services/route53.html#Route53.Client.create_vpc_association_authorization

agassner:
Thanks @JordonPhillips I got that. Basically the new version of https://github.com/boto/botocore/ has it.

PR #1088, opened 2016-11-18
Add support disabling subsection parsing

kyleknap:
This useful for retrieving the literal value of an option instead of possibly treating it as its own subsection.


kyleknap:
I updated the documentation so hopefully that reads better?

JordonPhillips:
Much better!

PR #1092, opened 2016-11-23
Add support for timestamp in ArgumentGenerator

kyleknap:
Uses epoch time as a ``datetime`` for its default value. Right now it defaults to ``None`` which is in appropriate representation of the ``timestamp`` type in our models.

This is needed for this PR: https://github.com/aws/aws-cli/pull/2293

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -563,6 +563,8 @@ class ArgumentGenerator(object):
                 return 0.0
             elif shape.type_name == 'boolean':
                 return True
+            elif shape.type_name == 'timestamp':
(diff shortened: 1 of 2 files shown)


kyleknap:
So I cannot actually merge this pull request until this one: https://github.com/aws/aws-cli/pull/2293 is ready. Datetimes cannot be handled when dumping to json so it would break ``--generate-cli-skeleton`` if I merge this without merging the referenced pull request. The referenced pull request fixes that issue for ``--generate-cli-skeleton``.

kyleknap:
Merging this as I am getting close to getting the other PR merged.

PR #1093, opened 2016-11-29
fix attribute error in sigv2 auth

vharitonsky:
head_object method will fail with sigv2:

  File "/usr/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "/work/pixel/lib/s3.py", line 60, in head_key
    await client.head_object(Bucket=bucket, Key=key)
  File "/usr/local/lib/python3.5/dist-packages/aiobotocore/client.py", line 79, in _make_api_call
  File "/usr/local/lib/python3.5/dist-packages/aiobotocore/endpoint.py", line 180, in _send_request
    request = self.create_request(request_dict, operation_model)
  File "/usr/local/lib/python3.5/dist-packages/botocore/endpoint.py", line 150, in create_request
    operation_name=operation_model.name)
  File "/usr/local/lib/python3.5/dist-packages/botocore/hooks.py", line 227, in emit
  File "/usr/local/lib/python3.5/dist-packages/botocore/hooks.py", line 210, in _emit
    response = handler(**kwargs)
  File "/usr/local/lib/python3.5/dist-packages/botocore/signers.py", line 90, in handler
    return self.sign(operation_name, request)
  File "/usr/local/lib/python3.5/dist-packages/botocore/signers.py", line 147, in sign
    auth.add_auth(request)
  File "/usr/local/lib/python3.5/dist-packages/botocore/auth.py", line 109, in add_auth
    params = request.param

The reason is request has params attribute, not param

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -107,7 +107,7 @@ class SigV2Auth(BaseSigner):
             params = request.data
         else:
             # GET
-            params = request.param
+            params = request.params
         params['AWSAccessKeyId'] = self.credentials.access_key
         params['SignatureVersion'] = '2'
         params['SignatureMethod'] = 'HmacSHA256'


jamesls:
Thanks for the pull request.  Not opposed to to this change (though we'll need to get a test added first), but I'm curious how you're able to get S3 configured with v2.

The only signature versions that are going to work with S3 are `HmacV1Auth (type='s3')` and `S3SigV4Auth (type='s3v4')`.

Even with this fix pulled in, using v2 against S3 doesn't work for me (which is the expected behavior).

vharitonsky:
That explains why our local Riak CS returns 500s! Signature version is passed as string and I could not find proper documentation for what are variations of those strings, so I just passed s3v2 and spotted this error. After monkey-patching the error, I discovered that I could not get a response with this setting. So I relied on defaults and called it a day. Thanks:)

PR #1094, opened 2016-11-29
Use member names in lists for ArgumentGenerator

kyleknap:
Use the member name for strings in lists. Catches and edge case pointed out by @jamesls in this PR: https://github.com/aws/aws-cli/pull/2293


kyleknap:
Alright tests now pass. Should be good to look at.

PR #1095, opened 2016-12-02
Rename opsworkscm to opsworks-cm

JordonPhillips:


botocore/loaders.py
```diff
@@ -215,10 +215,12 @@ class Loader(object):
     CUSTOMER_DATA_PATH = os.path.join(os.path.expanduser('~'),
                                       '.aws', 'models')
     BUILTIN_EXTRAS_TYPES = ['sdk']
+    BUILTIN_SERVICE_RENAMES = {'opsworkscm': 'opsworks-cm'}
```
stealthycoin: It's not really a rename its more of an additional name.

PR #1100, opened 2016-12-06
add health service pagination

### LINKED ISSUE #1099: missing paginators on health
all the health service apis support pagination minus describe event aggregates per 
http://boto3.readthedocs.io/en/latest/reference/services/health.html

but the paginators.json file is currently missing from the service data directory.

kapilt:
Closes #1099


JordonPhillips:
Thanks for the PR! Looks like you had at least one mis-model though, so I've sent up a separate pr with fixes.

PR #1101, opened 2016-12-06
First draft of exec-env appending to user-agent

stealthycoin:
Attach the environment variable AWS_EXECUTION_ENV to the end of the user agent.

cc @kyleknap @JordonPhillips @jamesls

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -16,7 +16,6 @@ This module (and all function/classes within this module) should be
 considered internal, and *not* a public API.
 
 """
-import os
 import copy
 import logging
 
@@ -108,8 +107,6 @@ class ClientArgsCreator(object):
                 user_agent = client_config.user_agent
             if client_config.user_agent_extra is not None:
                 user_agent += ' %s' % client_config.user_agent_extra
-        if os.environ.get('AWS_EXECUTION_ENV') is not None:
-            user_agent += ' exec-env/%s' % os.environ.get('AWS_EXECUTION_ENV')
 
         # Create a new client config to be passed to the client based
         # on the final values. We do not want the user to be able
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -16,6 +16,7 @@ This module (and all function/classes within this module) should be
 considered internal, and *not* a public API.
 
 """
+import os
 import copy
 import logging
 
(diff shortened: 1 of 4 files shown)


botocore/args.py
```diff
@@ -107,6 +108,8 @@ def compute_client_args(self, service_model, client_config,
                 user_agent = client_config.user_agent
             if client_config.user_agent_extra is not None:
                 user_agent += ' %s' % client_config.user_agent_extra
+        if os.environ.get('AWS_EXECUTION_ENV') is not None:
+            user_agent += ' exec-env/%s' % os.environ.get('AWS_EXECUTION_ENV')
```
stealthycoin: Should this still be appended if a custom client_config is provided?

tests/unit/test_args.py
```diff
@@ -125,3 +126,31 @@ def test_max_pool_from_client_config_forwarded_to_endpoint_creator(self):
...
+
+    def test_execution_env_append(self):
+        with mock.patch('os.environ.get') as m:
+            m.return_value = 'FooEnv'
+            result = self.args_create.compute_client_args(mock.MagicMock(),
+                                                          None,
+                                                          mock.MagicMock(),
+                                                          'us-west-2',
+                                                          '', False, {})
+            self.assertNotEqual(re.match('.+ exec-env/FooEnv$',
```
JordonPhillips: I would prefer not to use regex. You can use `.endswith` to achieve what you want.

tests/unit/test_args.py
```diff
@@ -14,6 +14,7 @@
 import botocore.config
 from tests import unittest
 import mock
+import re
```
stealthycoin: I took this out... in case there are more changes I didn't push it yet.

botocore/session.py
```diff
@@ -472,6 +472,9 @@ def user_agent(self):
                                           platform.release())
         if self.user_agent_extra:
             base += ' %s' % self.user_agent_extra
+        if os.environ.get('AWS_EXECUTION_ENV') is not None:
```
kyleknap: I think we would want to move this before the user_agent_extra.

botocore/session.py
```diff
@@ -472,6 +472,9 @@ def user_agent(self):
...
+            base += ' exec-env/%s' % os.environ.get('AWS_EXECUTION_ENV')
```
kyleknap: Also it would be good to add this to the doc string about the ``user_agent`` as all of the other components are documented..

PR #1103, opened 2016-12-08
Add compat function to ensure a type is text like

stealthycoin:
Not entirely sure this is the right approach in both versions.


stealthycoin:
Moving to cli.

PR #1109, opened 2016-12-15
Implemented NoSuchBucketError

HaraldNordgren:
This will be very useful when doing error handling in boto3. Instead of catching the generic ClientError and trying to parse the error code, I can now catch NoSuchBucketError and be sure that I caught the specific error.

I designed this by starting from what the test_s3_error_response() is testing.

And the how ClientError forms its msg compared to what I receive from boto3: `botocore.exceptions.ClientError: An error occurred (NoSuchBucket) when calling the ListObjects operation: The specified bucket does not exist`

### DIFF
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -19,6 +19,7 @@ from botocore.awsrequest import prepare_request_dict
 from botocore.docs.docstring import ClientMethodDocstring
 from botocore.docs.docstring import PaginatorDocstring
 from botocore.exceptions import ClientError, DataNotFoundError
+from botocore.exceptions import NoSuchBucketError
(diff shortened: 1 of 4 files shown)


HaraldNordgren:
Related to [boto3 #167](https://github.com/boto/boto3/issues/167)

jamesls:
Thanks for the pull request.

There's another PR (https://github.com/boto/botocore/issues/1005) that implements this more generally for all error types.  I agree this is something we want to improve, but I think we should try to solve this in a general manner instead of just for S3.

I'd propose closing this in favor of #1005.

HaraldNordgren:
Okay, let's close this. I just hope there is some progress on #1005 soon.

PR #1112, opened 2016-12-20
Add cross links to service API

jamesls:
You'll see that part of the diff is changing the target to `html`.  botocore/boto3 never set the target so it was defaulting to man page.  This meant that for certain style calls in bcdoc, the proper refs/links weren't being generated.

In order to use the new `external_link` method, the target needed to be html.

One option was to only turn on target html when generating html docs, but the `man` target doesn't make much sense for boto3/botocore.

image

### DIFF
--- a/botocore/docs/method.py
+++ b/botocore/docs/method.py
@@ -18,6 +18,9 @@ from botocore.docs.example import ResponseExampleDocumenter
 from botocore.docs.example import RequestExampleDocumenter
 
 
+AWS_DOC_BASE = 'http://docs.aws.amazon.com/goto/WebAPI'
+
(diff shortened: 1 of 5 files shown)


botocore/docs/method.py
```diff
@@ -18,6 +18,9 @@
 from botocore.docs.example import RequestExampleDocumenter
 
 
+AWS_DOC_BASE = 'http://docs.aws.amazon.com/goto/WebAPI'
```
kyleknap: Probably should make this https again.

jamesls: Fixed

PR #1113, opened 2016-12-21
Add modeled client exceptions

kyleknap:
Refactor of this PR: https://github.com/boto/botocore/pull/1005 that adds modeled exceptions to clients.

The main difference is that the exception classes are generated upfront as opposed to using ``__getattr__`` magic.

I did some performance testing with this script:
```python
import time
import boto3


iterations = 200
total = 0

total_start = time.time()
for _ in range(iterations):
    start = time.time()
    c = boto3.client('iam')
    end = time.time()
    total += end - start

print 'Average: ', total/iterations
print 'Total: ', time.time() - total_start
```

With the way I have it now, the client creation time is not affected as exception classes are lazily generated on accessing the ``exceptions`` property on the client.

cc @kislyuk @jamesls @JordonPhillips @stealthycoin @dstufft


botocore/errorfactory.py
```diff
@@ -0,0 +1,83 @@
...
+        :type code_to_exception: dict
+        :param code_to_exception: Mapping of error codes (strings) to exception
+            class that should be raised when encountering a particular
+            error code.
+        """
+        self._code_to_exception = code_to_exception
+        if code_to_exception is None:
+            self._code_to_exception = {}
+
+    def from_code(self, error_code):
```
kyleknap: In the previous pull request, this was an internal method. I actually think this is quite useful as it allows users to determine how ``exception.parsed_response['Error']['Code']`` gets mapped to an exception class, which would make it easier to convert legacy style error catching to this of inspecting the dictionary to this format. Would like to get thoughts on this.

jamesls:
Looking at the code now, but quick question.  What's the perf look like when you actually need to generate the exception mapping?  I get that client instantiation is largely unaffected because of the lazy generation, but what about the case where it needs to generate the exceptions?  Just curious.  Something like:

```
# in develop branch
try:
   client.someting()
except ClientError as e:
   pass

vs. 
# in this PR
try:
  client.someting()
except client.exceptions.Foo as e:
  pass

File: botocore/errorfactory.py [hunk 2]
```diff
@@ -0,0 +1,83 @@
...
+                    code = shape.metadata.get("error", {}).get("code")
+                    cls_props[exception_name] = exception_cls
+                    if code:
+                        code_to_exception[code] = exception_cls
+                    else:
+                        # Use the exception name if there is no explicit code
+                        # modeled
+                        code_to_exception[exception_name] = exception_cls
+        cls_name = str(get_service_module_name(service_model) + 'Exceptions')
+        client_exceptions_cls = type(
```
jamesls: What's the rationale for this needing to be typed to the service module.  Are we expecting to type check against `.exception`?  I can't really think of a reason why you'd want to do that.

tests/integration/test_client.py
```diff
@@ -148,15 +148,37 @@ def test_client_raises_exception_invalid_region(self):
...
 
+    def test_client_modeled_exception(self):
+        client = self.session.create_client(
+            'dynamodb', region_name='us-west-2')
+        with self.assertRaises(client.exceptions.ResourceNotFoundException):
+            client.describe_table(TableName="NonexistentTable")
+
+    def test_client_modeleded_exception_with_differing_code(self):
+        client = self.session.create_client('iam', region_name='us-west-2')
+        # The NoSuchEntityException should be rasied on NoSuchEntity error
```
jamesls: typo

botocore/errorfactory.py
```diff
@@ -0,0 +1,83 @@
...
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from botocore.exceptions import ClientError
+from botocore.utils import get_service_module_name
+
+
+class BaseClientExceptions(object):
```
jamesls: What do you think about a better `AttributeError` message that gives you the list of possible exception names?  Probably via `__getattr__`.

botocore/errorfactory.py
```diff
@@ -0,0 +1,83 @@
...
+    ClientError = ClientError
+
+    def __init__(self, code_to_exception=None):
```
kyleknap: It does not. I can make it required. I just made it optional for conveniency if you do not want to provide a mapping, in which case everything maps to ClientError

botocore/errorfactory.py
```diff
@@ -0,0 +1,83 @@
...
+        :returns: The exceptions object of a client that can be used
+            to grab the various different modeled exceptions.
+        """
+        cls_props = {}
+        code_to_exception = {}
+        for op_name in service_model.operation_names:
+            op_model = service_model.operation_model(op_name)
+            for shape in op_model.error_shapes:
+                exception_name = str(shape.name)
+                # An exception can be in multiple operations.
```
kyleknap: I think that should be fine. Went through all of the models to make sure that all exception shapes will have ``exceptions`` marked as ``true``. I just went this route to preserve logic from the previous pull request.

botocore/errorfactory.py
kyleknap: Sure that seems reasonable to me.

botocore/errorfactory.py
kyleknap: I am not entirely sure what you are asking?

I mainly did it to be more consistent with how client's use the full service name in their class name. I mean sure I could just make the class name ``ClientExceptions`` but I am not sure how much of the difference in functionality will be if I prefix it with the full service name as I am creating a new ClientExceptions object for each new service. We aren't looking to share exception types across services right?

The way I have it now (it has not been pushed up yet) is the exceptions factory is hoisted to the session. So we could expose the exceptions object dangled on the client in the future via the session without requiring the client. So it may be a little inconsistent if we are returning a bunch of ``ClientExceptions`` that share the same name but are of different classes if you do an ``is`` check where instead for the client we at least include the full service name in the client's class name.

Other than that not too attached to the name. Let me know what you think

jamesls: > So we could expose the exceptions object dangled on the client in the future via the session without requiring the client. So it may be a little inconsistent if we are returning a bunch of ClientExceptions...

Yeah, that's a pretty reasonable point.  I was commenting that accessing it via the client `ec2.exceptions` implies that the exceptions are ec2 specific so creating a brand new class (vs instantiating an existing class) seemed heavy weight if we weren't going to use that functionality anywhere.  But I agree that if these eventually do make it up to the session, it would be weird to have different different `ClientExceptions` with different attrs.

nueverest:
Here is a link to the docs as it helped me a lot. http://botocore.readthedocs.io/en/latest/client_upgrades.html#error-handling

PR #1116, opened 2016-12-27
Fix failing rds test

jamesls:
Our travis tests are failing due to a recent RDS functional test that was added.  I added two things:

1. Fix the test (tests that need credentials should subclass BaseSessionTest)
2. Add package setup/teardown that more closely resembles the travis environment.  This likely went unnoticed because if you have credentials configured locally the test will pass.  With this change, the original test would have failed on a dev laptop.

### DIFF
--- a/tests/functional/__init__.py
+++ b/tests/functional/__init__.py
@@ -10,3 +10,31 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import os
+import uuid
+
+_ORIGINAL = os.environ.copy()
+# These are environment variables that allow users to control
+# the location of config files used by botocore.
+_CONFIG_FILE_ENV_VARS = [
+    'AWS_CONFIG_FILE',
+    'AWS_SHARED_CREDENTIALS_FILE',
+    'BOTO_CONFIG',
+]
+_CREDENTIAL_ENV_VARS = [
+    'AWS_ACCESS_KEY_ID',
+    'AWS_SECRET_ACCESS_KEY',
+    'AWS_SESSION_TOKEN',
+]
+
+
+def setup_package():
+    random_file = str(uuid.uuid4())
+    for varname in _CONFIG_FILE_ENV_VARS:
+        os.environ[varname] = random_file
+    for credvar in _CREDENTIAL_ENV_VARS:
+        os.environ.pop(credvar, None)
+
+
+def teardown_package():
+    os.environ = _ORIGINAL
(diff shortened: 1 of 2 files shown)


tests/functional/__init__.py
```diff
@@ -10,3 +10,31 @@
...
+]
+_CREDENTIAL_ENV_VARS = [
+    'AWS_ACCESS_KEY_ID',
+    'AWS_SECRET_ACCESS_KEY',
+    'AWS_SESSION_TOKEN',
+]
+
+
+def setup_package():
+    random_file = str(uuid.uuid4())
```
dstufft: Is this file going to be created? If so would it make more sense to use the tempfile module?

dstufft: Also if this file is going to be created on disk ever (on phone so can't verify easily) cleanup should probably ensure to clean it up. Otherwise it's fine.

tests/functional/__init__.py
```diff
@@ -10,3 +10,31 @@
...
+    for varname in _CONFIG_FILE_ENV_VARS:
+        os.environ[varname] = random_file
```
dstufft: Isn't this whole thing going to get ignored because BaseSessionTest monkeypatches os.environ in its setUp method? It also appears to already do something similar.

tests/functional/__init__.py
jamesls: No, the intent is to set it to some path that's not going to exist.  `/does/not/exist` would also be more clear and probably (?) won't exist.  I was using a random file name to ensure it won't exist.

tests/functional/__init__.py
jamesls: > Isn't this whole thing going to get ignored because BaseSessionTest monkeypatches os.environ in its setUp method?

For the specific test I fixed, yes.

The intent of these test fixtures is to address how this can caught in the future.  That is, how can people know they need to use `BaseSessionTest` or face potential travis failures?  If you have credentials configured in your dev setup (very likely) you'll never see this issue until travis runs the tests.  This forces the issue locally by setting all the normal cred env vars to bogus values so you'll see similar behavior to what travis sees.

tests/functional/__init__.py
dstufft: Alright, that sounds reasonable to me.

tests/functional/__init__.py
dstufft: Makes sense, and sounds reasonable to me. I guess my only suggestion is a comment to mention that, so that 6 months from now we don't come back and try to figure out why this code seemingly duplicates ``BaseSessionTest``.

jamesls: Sounds good, I'll add a comment before merging.

PR #1130, opened 2017-01-24
Move non_aggregate_keys to sdk extras

JordonPhillips:

### DIFF
--- a/botocore/data/dynamodb/2012-08-10/paginators-1.json
+++ b/botocore/data/dynamodb/2012-08-10/paginators-1.json
@@ -14,9 +14,6 @@
         "Items",
         "Count",
         "ScannedCount"
-      ],
-      "non_aggregate_keys": [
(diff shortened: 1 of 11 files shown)


JordonPhillips:
This was the only customization I was able to find after programmatically searching all files. Since there were only five instances, I didn't think it would be worth scripting. Since I don't forsee us having to do this again, I don't know that it's worth creating one.

jamesls:
In general, moving these out to a separate file seems fine, but another high level question: How are we validating the correctness of these files?  We have one functional test that's more of a "smoke" test, but I'd like something that is more rigorous in validating the extra.json files.

One of the things I remember from when we had the old .extra.json files was that we'd miss a level of nesting or typo something.  I think if we're going to move these out we should have some sort of validation in place.

jamesls:
After discussion, I think we're going to hold off on this PR for now right?

PR #1132, opened 2017-01-24
Update models

JordonPhillips:
This updates paginators, waiters, and examples for a large number of
services. This has been broken out into three commits to make it
somewhat easier to digest, but there are still quite a lot of changes.

cc @kyleknap @jamesls @dstufft @stealthycoin

PR #1133, opened 2017-01-25
Sync up paginators

kyleknap:
This is a similar to @JordonPhillips PR: https://github.com/boto/botocore/pull/1132 except this is just a subset of it. I am only adding in paginators. In terms of differences, there is only one functional difference which is mine adds a ``limit_key`` in one of the paginators for ``config`` when the referenced PR does not.

cc @jamesls @JordonPhillips @stealthycoin @dstufft


JordonPhillips:
We should probably add a changelog entry

kyleknap:
@JordonPhillips added the changelog entry for the paginators

PR #1134, opened 2017-01-25
Update waiters

JordonPhillips:

### DIFF
--- a/botocore/data/elb/2012-06-01/waiters-2.json
+++ b/botocore/data/elb/2012-06-01/waiters-2.json
@@ -1,0 +1,31 @@
+{
+    "version":2,
+    "waiters":{
+        "AnyInstanceInService":{
+            "acceptors":[
+                {
+                    "argument":"InstanceStates[].State",
+                    "expected":"InService",
+                    "matcher":"pathAny",
+                    "state":"success"
+                }
+            ],
+            "delay":15,
+            "maxAttempts":40,
+            "operation":"DescribeInstanceHealth"
+        },
+        "InstanceInService":{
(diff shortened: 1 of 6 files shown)


botocore/data/elb/2012-06-01/waiters-2.json
```diff
@@ -0,0 +1,49 @@
+{
+    "version":2,
+    "waiters":{
+        "InstanceDeregistered": {
```
kyleknap: I wonder if we should hold off on this one. I am still not sure if this one is valid in terms of when the ``InvalidInstance`` exception is thrown when multiple instances are provided.

jamesls: Yeah, I think we should hold off on this for now until we can look into this waiter more in depth.

botocore/data/s3/2006-03-01/waiters-2.json
```diff
@@ -12,6 +12,16 @@
           "state": "success"
         },
         {
+          "expected": 301,
+          "matcher": "status",
+          "state": "success"
+        },
+        {
+          "expected": 403,
```
jamesls: Yeah, I'm on the fence with this one as well.

Technically it's correct.  An auth error could mean the bucket exists but you can't access it.  However, in practice  I think the most common workflow is:

```
create_bucket()        # 1.
wait_for_it_to_exist() # 2.
operate_on_bucket()    # 3.
```

In which case if something happens between (1) and (2) where you can no longer access the bucket, I think this should be an error rather than a waiter success.

PR #1138, opened 2017-01-26
Fix order of credential providers

sczizzo:
According to the [Credentials Configuration documentation for Boto 3](https://boto3.readthedocs.io/en/latest/guide/configuration.html):

```
The order in which Boto3 searches for credentials is:

  1. Passing credentials as parameters in the boto.client() method
  2. Passing credentials as parameters when creating a Session object
  3. Environment variables
  4. Shared credential file (~/.aws/credentials)
  5. AWS config file (~/.aws/config)
  6. Assume Role provider
  7. Boto2 config file (/etc/boto.cfg and ~/.boto)
  8. Instance metadata service on an Amazon EC2 instance that has an IAM role configured.
```

However, looking at the definition of [`create_credential_resolver()`](https://github.com/boto/botocore/blob/4c6a23cbe39a3c54f96e3574278b3a164f2e4d30/botocore/credentials.py#L44-L82), the `AssumeRoleProvider` (_#_*6* above) is loaded _before_ the `SharedCredentialProvider` (_#_*4*) and `ConfigProvider` (_#_*5*). This PR fixes the ordering of these providers to reflect the documentation.

Supposing you're using temporary credentials through AWS's STS/assume-role facilities, before this change you may be prompted for an MFA code (even when valid credentials exist):
```
$ python
>>> import boto3, botocore
>>> botocore.session.Session().get_credentials()
Enter MFA code:
```

After this change, credentials are loaded automatically from the shared file, as expected:
```
$ python
>>> import boto3, botocore
>>> botocore.session.Session().get_credentials()
<botocore.credentials.Credentials object at 0x105192d90>
>>>
```

Environment information:
```
$ env | grep AWS
AWS_PROFILE=foobar
$ python --version
Python 2.7.13
$ pip --version
pip 9.0.1 from /usr/local/lib/python2.7/site-packages (python 2.7)
$ pip list | grep boto
boto (2.45.0)
boto3 (1.4.4)
botocore (1.5.5)
```

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -82,7 +82,6 @@ def create_credential_resolver(session, cache=None):
     )
     providers = [
         env_provider,
-        assume_role_provider,
         SharedCredentialProvider(
             creds_filename=credential_file,
             profile_name=profile_name
@@ -92,6 +91,7 @@ def create_credential_resolver(session, cache=None):
         # The new config file has precedence over the legacy
         # config file.
         ConfigProvider(config_filename=config_file, profile_name=profile_name),
+        assume_role_provider,
         OriginalEC2Provider(),
         BotoProvider(),
         container_provider,


jamesls:
Hi, thanks for the pull request.  I'm having trouble following your example scenario.  Why would you be prompted for MFA in one case but not the other?  Are you saying you have both assume_role configuration *and* `aws_access_key_id`/`aws_secret_access_key` configured in the same profile?  Just want to make sure I understand the problem.

sczizzo:
So we have two accounts for _prod_ and _non-prod_. We have some tooling that takes care of creating temporary credentials and writing them to the shared credentials file and updating `AWS_PROFILE`.

For non-prod access:

- I have an IAM user (with access keys) in our non-prod account
- This user is required to authenticate with MFA for API access
- We use STS get-session-token to create temporary non-prod credentials
- These temporary non-prod credentials are written to `~/.aws/credentials`
- My `AWS_PROFILE` is set to `non-prod-mfa` (refer to config below)

These temporary non-prod credentials are picked up from the shared credentials file by various tools like the AWS CLI. As long as those credentials have not expired, I'm not prompted for an MFA code (even by botocore). We're good here.

Now, to get access to the prod account:

- My non-prod user is trusted to assume a particular prod role
- This user is required to authenticate with MFA for API, assume-role access
- We use STS assume-role to elevate access, create temporary prod credentials
- These temporary prod credentials are written to `~/.aws/credentials`
- My `AWS_PROFILE` is set to `prod` (refer to config below)

Again these temporary prod credentials in my shared credentials file are picked up by various tools. As long as the credentials haven't expired, I _should not_ be prompted for an MFA code. Indeed, the AWS CLI picks up these credentials, and I'm able to execute API requests without prompting. This is where botocore does not work as expected, as it prompts me to enter an MFA code.

Concretely, after gaining access to the prod, with a fresh set of prod credentials loaded in my credentials file and my `AWS_PROFILE=prod`, running AWS CLI commands does not prompt me for an MFA code, but botocore does:

```
$ aws iam list-users
{
  "Users": [ ... ]
}
$ python
>>> import boto3, botocore
>>> botocore.session.Session().get_credentials()
Enter MFA code:
```

(To be clear, the interpreter is waiting for user input after `Enter MFA code:`)

I expect `get_credentials()` here to simply grab the appropriate credentials for the given profile from the shared credentials, i.e. something like this:

```
$ python
>>> import boto3, botocore
>>> botocore.session.Session().get_credentials()
<botocore.credentials.Credentials object at 0x105192d90>
>>>
```

As far as I can tell, this behavior is a result of the improper ordering of credential providers: The `SharedCredentialProvider` should take precedence over `AssumeRoleProvider`, and this is actually exactly what the [boto3 documentation declares](https://boto3.readthedocs.io/en/latest/guide/configuration.html#configuring-credentials).

For reference, a stripped-down AWS config:

```
$ cat ~/.aws/config
[default]
region = us-east-1
output = json

[profile non-prod]

[profile non-prod-mfa]
source_profile = non-prod

[profile prod]
mfa_serial = arn:aws:iam::1234567890:mfa/my-user
role_arn = arn:aws:iam::0987654321:role/prod-user
source_profile = non-prod
```

And my shared credentials file after creating temporary credentials for
both the non-prod and prod accounts:

```
$ cat ~/.aws/credentials
[non-prod]
aws_access_key_id = shh-secret-foo1
aws_secret_access_key = shh-secret-bar1

[non-prod-mfa]
aws_access_key_id = shh-secret-foo2
aws_secret_access_key = shh-secret-bar2
aws_security_token = shh-secret-baz2
aws_session_token = shh-secret-bat2

[prod]
aws_access_key_id = shh-secret-foo3
aws_secret_access_key = shh-secret-bar3
aws_security_token = shh-secret-baz3
aws_session_token = shh-secret-bat3
```

sczizzo:
@jamesls I see this is still marked as `response-needed`, is there any more context I can provide?

sczizzo:
Rebased off `develop`

sczizzo:
Rebased off `develop`

sczizzo:
Rebased off `develop`

vboginskey:
Hi @jamesls, @stealthycoin. This PR represents a very minor change that simply swaps the order of credential providers to match the documentation. Is there an issue that's preventing this from being merged? And if so, should we be updating the documentation to reflect what the code is actually doing instead?

sczizzo:
I'd love to see this merged, but I'm not sure there's any more information I can provide, and it's just been languishing here for over a year and a half. Closing.

PR #1139, opened 2017-01-26
Sync Examples

JordonPhillips:
This adds basic linting for examples and pulls in several hundred
examples.


tests/functional/docs/test_shared_example_config.py
```diff
@@ -0,0 +1,38 @@
...
+        )
+        examples = example_config.get("examples", {})
+        for operation, operation_examples in examples.items():
+            for example in operation_examples:
+                yield _lint_single_example, operation, example, service_model
+
+
+def _lint_single_example(operation_name, example_config, service_model):
+    # There should be no non-ascii characters
+    json.dumps(example_config, ensure_ascii=False).encode('ascii')
```
jamesls: It would be nice if we have good error message here (for both conditions you checked).

Things like:

1) What service/operation we failed on.
2) The actual character/value that failed  (I think we do this same check for the .rst example files in the CLI)  You may be able to reuse some of that code.

3) The name of the operation that does not exist.

kyleknap:
One thing I am wondering about is if we want to support non-ascii characters in the examples? So that we are a bit more robust in pulling the examples in. I know we are able to handle them in the regular docs that we generate from the service model.

JordonPhillips:
@kyleknap I wasn't aware we supported non-ascii in the service docs. Looking into it.

kyleknap:
It used to be an issue in the old boto3 doc generator. Its been a while, but I believed I fixed that issue by ensuring we are always using utf-8 encoding.

JordonPhillips:
Looks like the issue was an explicit `str` casting, so the fix was easy.

JordonPhillips:
I found another issue that was shadowed by the unicode issue, namely that there is at least one invalid timestamp.

JordonPhillips:
Found even more problems that weren't causing build failures

JordonPhillips:
This will need another look

tests/functional/docs/test_shared_example_config.py
```diff
@@ -0,0 +1,107 @@
...
+
+def assert_valid_values(service_name, operation_model, example_config):
+    try:
+        example_input = example_config.get('input')
+        if example_input is not None:
+            _assert_valid_values(operation_model.input_shape, example_input)
+
+        example_output = example_config.get('output')
+        output_shape = operation_model.output_shape
+        if example_output is not None and output_shape is not None:
```
jamesls: Is there validation elsewhere if there's example output but not output shape?  Same question for input.  If not should that be a failure?

JordonPhillips:
Updated

tests/functional/docs/test_shared_example_config.py
```diff
@@ -0,0 +1,124 @@
...
+        path = []
+    path.append(shape.name)
+
+    if shape.type_name == 'timestamp':
+        _assert_valid_timestamp(example_value, path)
+    elif shape.type_name == 'structure':
+        _assert_valid_structure_values(shape, example_value, path)
+    elif shape.type_name == 'list':
+        _assert_valid_list_values(shape, example_value, path)
+
```
kyleknap: You probably want to handle maps as well because you can recurse through those as well.

tests/functional/docs/test_shared_example_config.py
```diff
@@ -0,0 +1,124 @@
...
+        raise AssertionError(
+            "Invalid value in example for %s from %s with id %s: %s" % (
+                operation_model.name, service_name, example_id, e
+            ))
+
+
+def _assert_valid_values(shape, example_value, path=None):
+    if path is None:
+        path = []
+    path.append(shape.name)
```
kyleknap: Might be a little nit-picky but not sure if we want to be using shape names for the info on the path because it could be hard to determine what member is the issue as members in a structure can have the same shape and the shapes name could literally be something like ``"String"``. I think it would be better if we can incorporate member names into this. Maybe by changing the logic to be:

1) Pass in the output/input shape name as the shape name
2) As you recurse structures, append the member name to the path of each branch that needs to be further recursed.
3) As you recurse lists, you can just use append the shape name of its member as that does not have a member name.

Let me know what you think.

JordonPhillips:
Updated to lint maps and make the path prettier

PR #1144, opened 2017-01-31
Decode file paths in config loader

### LINKED ISSUE #2395: Expose the exception type (and other helpful information) from configparser

I am a user of AWS Organizations, as well as having multiple AWS-using freelance clients, and my own personal AWS account. As a result I have over 160 profiles in my `~/.aws/config` file.

Today, after adding  a couple dozen more profiles for recently-provisioned accounts, I started getting the following error from the AWS CLI.

```plain
Unable to parse config file: /Users/{me}/.aws/config
```

However, there was no additional information to help me debug. Was it an issue with line endings? Spacing around `=` signs? A typo somewhere?

After a few hours of poking around, I tracked it down to this line in `botocore`: https://github.com/boto/botocore/blob/6451ae1fad57f4453af97649e7ed9192b0f623be/botocore/configloader.py#L147-L152

The issue was that I'd inadvertently duplicated a key (similar to #2311).

Looking at [the read() method of configparser](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read), and in-turn [configparser exceptions](https://docs.python.org/3/library/configparser.html#configparser.DuplicateSectionError), I'd like to see a little more information exposed from the underlying `configparser` class when an exception occurs, so that it doesn't take so long to track down the source of the problem.

With AWS continuing to get exponentially more complex over the years, I'm certainly not the only one responsible for managing hundreds of accounts.

Thank you.

JordonPhillips:
This decodes the path in config parser before throwing errors.

Fixes aws/aws-cli#2395

### DIFF
--- a/botocore/configloader.py
+++ b/botocore/configloader.py
@@ -14,8 +14,9 @@
 import os
 import shlex
 import copy
+import sys
 
-from six.moves import configparser
+from botocore.compat import six
 
 import botocore.exceptions
 
@@ -142,12 +143,13 @@ def raw_config_parse(config_filename, parse_subsections=True):
         path = os.path.expandvars(path)
         path = os.path.expanduser(path)
         if not os.path.isfile(path):
-            raise botocore.exceptions.ConfigNotFound(path=path)
-        cp = configparser.RawConfigParser()
+            raise botocore.exceptions.ConfigNotFound(path=_unicode_path(path))
+        cp = six.moves.configparser.RawConfigParser()
         try:
             cp.read(path)
-        except configparser.Error:
-            raise botocore.exceptions.ConfigParseError(path=path)
+        except six.moves.configparser.Error:
+            raise botocore.exceptions.ConfigParseError(
+                path=_unicode_path(path))
         else:
             for section in cp.sections():
                 config[section] = {}
@@ -161,9 +163,15 @@ def raw_config_parse(config_filename, parse_subsections=True):
                             config_value = _parse_nested(config_value)
                         except ValueError:
                             raise botocore.exceptions.ConfigParseError(
-                                path=path)
+                                path=_unicode_path(path))
                     config[section][option] = config_value
     return config
+
+
+def _unicode_path(path):
+    if isinstance(path, six.text_type):
+        return path
+    return path.decode(sys.getfilesystemencoding(), 'replace')
 
 
 def _parse_nested(config_value):
--- a/botocore/configloader.py
+++ b/botocore/configloader.py
@@ -146,7 +146,7 @@ def raw_config_parse(config_filename, parse_subsections=True):
             raise botocore.exceptions.ConfigNotFound(path=_unicode_path(path))
         cp = six.moves.configparser.RawConfigParser()
         try:
-            cp.read(path)
+            cp.read([path])
         except six.moves.configparser.Error:
             raise botocore.exceptions.ConfigParseError(
                 path=_unicode_path(path))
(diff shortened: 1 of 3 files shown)


jamesls:
What's the rationale for not just fixing it at the source, given we know where this error originates from?  Otherwise, we have to be sure there's no breakage here given you're possibly affecting every botocore exception.

galopin:
@JordonPhillips I hope your pull request will be approved… or tweaked. Kudos! :+1:

@jamesls @kyleknap whatever it takes, Boto development team should address this blocking issue as soon as possible. I lost a few hours trying to make AWS Command Line Interface work on my Windows set-up until I realized the problem originated from my (poorly) chosen username and some stupid UTF-8 handling issue within Boto.

jamesls:
@galopin Yes, completely agree we should fix this.  We just need to be careful to not fix this while unintentionally breaking other people as a result of this change.

@JordonPhillips To elaborate on my point more, while file system encoding in general is flaky, I don't think we can always assume it's utf-8.  Additionally, failure to decode to unicode when we can't load a config shouldn't itself be a failure, even using replacement chars seems to be better than failing on a decode in this specific scenario.

So my suggestion:

1. Move the decoding into the configloader module when we raise an exception.
2. Use similar logic to what the S3 code in the CLI uses (`sys.getfilesystemencoding()` I think).
3. Use `'replace'` when decoding so it's not a hard failure (I don't think we want to mask the original error message if we can't decode the file).

JordonPhillips:
@jamesls From the perspective of the config loader, there's nothing wrong with the strings we're using. It finds the right paths and loads the files just fine when they're available. The bug occurs in the exception itself. While we can fix the specific issue by making a change to the config loader, we can't guarantee that it will be fixed elsewhere. I think the point about file system encoding is a good one though. In this specific case, the ambiguity of the file system's encoding probably make it make more sense to fix at the config loader level.

JordonPhillips:
Updated

botocore/configloader.py
```diff
@@ -142,12 +143,13 @@ def raw_config_parse(config_filename, parse_subsections=True):
...
         if not os.path.isfile(path):
-            raise botocore.exceptions.ConfigNotFound(path=path)
-        cp = configparser.RawConfigParser()
+            raise botocore.exceptions.ConfigNotFound(path=_unicode_path(path))
+        cp = six.moves.configparser.RawConfigParser()
         try:
-            cp.read(path)
-        except configparser.Error:
-            raise botocore.exceptions.ConfigParseError(path=path)
+            cp.read([path])
```
kyleknap: Note that the list was added because of this logic:
https://github.com/python/cpython/blob/master/Lib/configparser.py#L690-L692

If you pass bytes to the ``read``, it will not wrap the filename in a list and just iterate through the individual bytes in the bytes object...

jamesls: @JordonPhillips What's the rationale for this?

This is unrelated to your change which should just be decoding to unicode for config errors.

If this is because of some other bug let's address that as a separate PR with a separate set of tests.  I really just want to address the actual customer pain here which is masking errors due to not properly decoding issues.   This change is not directly related to that.

JordonPhillips: This was needed to be able to write tests that pass in bytes. `read` takes in a list of file paths, but if you provide a single string it will wrap that in a list. But python 3 treats strings and bytes differently, and in the code specifically for `read` it doesn't take bytes into account (which is probably a bug). But since a `bytes` object is iterable, it just silently does the wrong thing.

This is needed so that tests can be written that pass in raw bytes and have them function on both python 2 and python 3.

galopin:
@jamesls @kyleknap what's up, doc? Is there any hope to have this conundrum solved anytime soon?

botocore/configloader.py
jamesls: Ok, seems reasonable as long as we're confident this won't break anything.

PR #1145, opened 2017-02-01
Coerce headers stored in dict to be lowercase.

Fedorof:
Make sure that the headers when converted to `dict` are in lower case.

This change is needed to support case-insensitive `CIMultiDict` that is provided with `aiohttp` responses. When `dict()`'ed, the keys of `CIMultiDict` are in Camel-Case.

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -215,7 +215,8 @@ class ResponseParser(object):
         if isinstance(parsed, dict):
             response_metadata = parsed.get('ResponseMetadata', {})
             response_metadata['HTTPStatusCode'] = response['status_code']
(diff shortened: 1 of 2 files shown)


JordonPhillips:
I don't think it would be possible to make this change because it will break anybody currently using those headers.

thehesiod:
this change is actually wrong, headers are already in lower-case per https://github.com/boto/botocore/blob/develop/botocore/vendored/requests/packages/urllib3/_collections.py#L216

thehesiod:
@Fedorof based on my findings in aiobotocore want to close this?

Fedorof:
Yes. Thank you.

PR #1154, opened 2017-02-16
Iam string to dict

stealthycoin:
https://github.com/boto/boto3/issues/978

We decode the string for the AssumeRolePolicyDocument field in the response from get_role in iam.
https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L428-L456

The documentation currently cites its type as a string

cc @dstufft @kyleknap @jamesls @JordonPhillips 

### Changes to example:
screen shot 2017-02-15 at 5 37 06 pm

### Changes to parameter docs:
screen shot 2017-02-15 at 5 29 32 pm

### DIFF
--- a/botocore/docs/shape.py
+++ b/botocore/docs/shape.py
@@ -16,6 +16,21 @@
 # ``traverse_and_document_shape`` method called directly. It should be
 # inherited from a Documenter class with the appropriate methods
 # and attributes.
+
+# Some shapes have their types remapped for before being exposed to the
+# user for convenience. A common example would be a shape that has a string
+# type being remapped to a dictionary because the string is simply a serialized
+# json object. This dictionary overrides a particular shape's documented type.
+# In addition shape names are not unique between services, there are over 2200
+# duplicated shape names between services, so this mapping needs to take into
+# account which service the shape being remapped belongs to.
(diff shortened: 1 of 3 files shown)


JordonPhillips:
It looks like this is a problem with all the returned iam policy docs, so it would be better to apply that to all return structures with `policyDocumentType`.

For reference, [here](https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L428) is where we decode it.

wchresta:
We ran into this documentation issue today. What is needed to get this merged? How can I help?

chrispblink:
I think it's worth adding that this breaks [boto3-stubs[iam]](https://pypi.org/project/mypy-boto3-iam/) as they are automatically generated from boto docs.

tim-finnigan:
After bringing this up for discussion the consensus was that [this comment](https://github.com/boto/botocore/pull/1154#issuecomment-282800323) from earlier still applies:

>It looks like this is a problem with all the returned iam policy docs ([another example](http://boto3.readthedocs.io/en/latest/reference/services/iam.html#IAM.Client.get_user_policy)), so it would be better to apply that to all return structures with `policyDocumentType`.

Issue #1148, opened 2017-02-06
role_arn credentials are not cached


eric-tucker:
Unlike the CLI, credentials are not cached on disk under `~/.aws/cli/cache` and as such I have to type in my MFA every time I run a python script using boto.  This is even more annoying with codes only usable once because I now have to wait 30 seconds between script executions.

Please have the credentials cached on disk like the awscli does.

stealthycoin:
Marking this as a feature request.

gene1wood:
@jamesls , I see back in 2014 [you added](https://github.com/aws/aws-cli/blame/02ead4f0f2f5be1e91903aaf97a9669db228bc49/awscli/customizations/assumerole.py#L43-L88) the persistently cached credentials to awscli and not botocore intentionally ( https://github.com/aws/aws-cli/commit/22932e53c7085bd8e53b22904b326c426e2e60fc ).

Have you had any further thought on adding this to botocore? I currently use a patched version of botocore and boto3 to introduce your persistent cached credentials from awscli into botocore so that awscli and boto3 share the cached credentials store. If you'd like I can make a PR that introduces this functionality into botocore.

eric-tucker:
I cannot understand why we wouldn't want cached MFA credentials when using boto3 (or any of the SDKs for that matter).  MFA = human interaction = I want SOME level of caching.

gene1wood:
I created a PR to fix this bug ( #1157 ) a couple weeks back. I've been using it locally since then and it works great. Any guidance on what would be needed to get this merged? Any additions or changes?

gene1wood:
[Here is the final disposition](https://github.com/boto/botocore/pull/1157#issuecomment-387580482) of PR #1157 and how it relates to #1338 including how to use the botocore cache.

PR #1156, opened 2017-02-23
Adding persistent caching of temporary credentials

gene1wood:
Adding persistent caching of temporary credentials from aws/aws-cli@22932e53c7085bd8e53b22904b326c426e2e60fc

This PR addresses #1148 by porting the code present in awscli into botocore to make it available to boto3. Having the code in both awscli and botocore until it's removed from awscli doesn't appear to have any negative impact.

One thing I'd like guidance on is:

* The `RefreshableCredentials` class stores the `_expiry_time` natively in `datetime` format.
* The `AssumeRoleProvider` class expects the possibility that the `response['Credentials']['Expiration']` value which may come from the cache may need to be [parsed from a string back into a `datetime`](https://github.com/boto/botocore/blob/8ec77d0b1683cd01e1b62506c4c1ea53ccda6daf/botocore/credentials.py#L904-L905)
* By my searches it appears to me that `awscli` doesn't have any special logic (not present in botocore) to ensure that the expiration `datetime` is serialized to a string before it's passed to `JSONFileCache`
* When I use the existing [`assumerole` customization](https://github.com/aws/aws-cli/blob/02ead4f0f2f5be1e91903aaf97a9669db228bc49/awscli/customizations/assumerole.py) in `awscli` the expiration `datetime` is correctly serialized to a string when stored in the `JSONFileCache`
* When I use this new functionality that I've moved into botocore in this PR *without* and explicit serialization call, the expiration `datetime` is passed to `JSONFileCache` un-serialized (which `JSONFileCache` rightly rejects). As a result I've [added a serialization step](https://github.com/boto/botocore/compare/develop...gene1wood:add-persistent-credential-cache?expand=1#diff-dcd72ae45553af71f1ceafb2d4ee5dcfR903) and I don't understand why this step isn't also required for awscli to work.

@jamesls Can you provide any insight on this question about serialization? If there's any way I can remove that serialization call and have the expiration `datetime` serialized through whatever method awscli accomplishes it I'd prefer it.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -17,6 +17,7 @@ import logging
 import os
 import getpass
 import threading
+import json
 from collections import namedtuple
 
 from dateutil.parser import parse
@@ -61,7 +62,7 @@ def create_credential_resolver(session):
         AssumeRoleProvider(
             load_config=lambda: session.full_config,
             client_creator=session.create_client,
-            cache={},
+            cache=JSONFileCache(),
             profile_name=profile_name,
         ),
         SharedCredentialProvider(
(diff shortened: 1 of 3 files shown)


gene1wood:
Actually, my use of `_serialize_if_needed` isn't correct as I incorrectly used it on the entire cached object instead of the specific value in the object that needs to be serialized.

I'd originally implemented this by adding `_serialize_if_needed` as the [`default` argument to the `json.dumps`](https://docs.python.org/2/library/json.html#basic-usage) which worked but seemed to be in opposition to the intent of the `JSONFileCache` based on the [`test_only_accepts_json_serializable_data` test](https://github.com/aws/aws-cli/blob/develop/tests/unit/customizations/test_assumerole.py#L84-L87).

I'll close this PR out as it doesn't work and see if I can figure out the right way to do the serialization. Ideally it would work exactly the same way as the serialization in awscli works but I haven't been able to figure that out yet. And ideally it would both serialize correctly to the file *and* allow the existing `test_only_accepts_json_serializable_data` test to continue to pass.

PR #1157, opened 2017-02-24
Adding persistent caching of temporary credentials with serialization

gene1wood:
Adding persistent caching of temporary credentials from aws/aws-cli@22932e53c7085bd8e53b22904b326c426e2e60fc with serialization

This PR addresses #1148 by porting the code present in awscli into botocore to make it available to boto3. Having the code in both awscli and botocore until it's removed from awscli doesn't appear to have any negative impact.

One thing I'd like guidance on is:

* The `RefreshableCredentials` class stores the `_expiry_time` natively in `datetime` format.
* The `AssumeRoleProvider` class expects the possibility that the `response['Credentials']['Expiration']` value, which may come from the cache, may need to be [parsed from a string back into a `datetime`](https://github.com/boto/botocore/blob/8ec77d0b1683cd01e1b62506c4c1ea53ccda6daf/botocore/credentials.py#L904-L905)
* In my searches of the `awscli` code it appears to me that `awscli` doesn't have any special logic (logic not present in botocore) to ensure that the expiration `datetime` is serialized to a string before it's passed to `JSONFileCache`
* When I use the existing [`assumerole` customization](https://github.com/aws/aws-cli/blob/02ead4f0f2f5be1e91903aaf97a9669db228bc49/awscli/customizations/assumerole.py) in `awscli` the expiration `datetime` is correctly serialized to a string when stored in the `JSONFileCache`
* When I use this new functionality that I've moved into botocore in this PR *without* an explicit serialization call the expiration `datetime` is passed to `JSONFileCache` un-serialized (which `JSONFileCache` rightly rejects). As a result I've [set `json.dumps` to serialize the `datetime`](https://github.com/boto/botocore/compare/develop...gene1wood:persistent-credential-cache-with-serialization?expand=1#diff-dcd72ae45553af71f1ceafb2d4ee5dcfR185) but I don't understand why this step isn't also required for awscli to work.

I've implemented the serialization by adding `_serialize_if_needed` as the [`default` argument to the `json.dumps`](https://docs.python.org/2/library/json.html#basic-usage) which works but seems to be in opposition to the intent of the `JSONFileCache` based on the [`test_only_accepts_json_serializable_data` test](https://github.com/aws/aws-cli/blob/develop/tests/unit/customizations/test_assumerole.py#L84-L87).

@jamesls Can you provide any insight on this question about serialization? If there's any way I can remove that `default` argument to `json.dumps` which calls `_serialize_if_needed` and have the expiration `datetime` serialized through whatever method awscli accomplishes it I'd prefer it.

Note : I originally submitted PR #1156 with serialization outside of `JSONFileCache` but I implemented it in a way that didn't work so I closed out that PR.

To try this PR out you can install it by running

    pip install git+https://github.com/gene1wood/botocore.git@persistent-credential-cache-with-serialization


andrewkrug:
This works great!  Please merge.

eric-tucker:
Please merge this and upload to pip!

botocore/credentials.py
```diff
@@ -152,6 +153,51 @@ def _refresher():
...
+            file_content = json.dumps(value, default=_serialize_if_needed)
+        except (TypeError, ValueError):
+            raise ValueError("Value cannot be cached, must be "
+                             "JSON serializable: %s" % value)
+        if not os.path.isdir(self._working_dir):
+            os.makedirs(self._working_dir)
+        with os.fdopen(os.open(full_key,
+                               os.O_WRONLY | os.O_CREAT, 0o600), 'w') as f:
+            f.truncate()
+            f.write(file_content)
```
dstufft: Hmm, should this be written to a temporary location and then renamed into place so that it is written in one atomic action?

dstufft:
One implementation comment, but otherwise this looks ~ok to me. I'd like another reviewer though!

botocore/credentials.py
gene1wood: @jamesls Can you speak to @dstufft 's question about this section of your commit aws/aws-cli@22d4cea5ab7e2e8d82cbbda4e10f4f271dfeb013

gene1wood: @dstufft I brought this code block over directly from @jamesls 's code in `aws-cli`. I would imagine if we wanted to implement your suggestion (which I think is good) we should do so in aws-cli and boto. If we defer that change it could potentially be made only in botocore once aws-cli no longer duplicates this cache persistence logic. @dstufft Do you think we should change this here in this PR now or wait on fixing it in both `botocore` and `aws-cli`?

eric-tucker:
Can we please oh pretty please merge this?

gene1wood:
@dstufft 

Any thoughts on [my comment above](https://github.com/boto/botocore/pull/1157#discussion_r108290138)?

@dstufft or @stealthycoin 

Anything you can think of that I can do to enable/encourage @jamesls @JordonPhillips or @kyleknap to review this PR?

eric-tucker:
I know everyone is probably very busy, but please can we prioritize merging this?  We've stopped writing Python scripts for AWS CLI calls because of how annoying having to put MFA in every single time has become.

eric-tucker:
Guys this is getting ridiculous.  This is an insanely needed feature and there's been no activity on it.  Why can't this be merged?

botocore/credentials.py
gene1wood: @dstufft Could you let me know if you'd like me to implement your suggestion in light of my note in the [comment above ](https://github.com/boto/botocore/pull/1157#discussion_r108290138). Really, any thoughts you have would be helpful here.

jamesls:
I don't mind adding the json file cache class here in botocore, but I'm hesitant to change the defaults to start writing out to `~/.aws/cli/cache` in botocore, which isn't actually tied to the CLI.  In botocore, the default implementations are typically in memory representations.  I'd prefer to make this an opt-in configurable thing (likely via client config objs), in addition to allowing a user to specify the cache directory so it can be shared with the CLI.

I think that should be a relatively small change.  Thoughts?

eric-tucker:
If the issue is putting it in botocore then why not put it into boto3?

eric-tucker:
I've resorted to applying a patch based on this pull request to my local botocore package folder.  It's been working without issue for weeks.  I just don't see why anyone would not want this behaviour.  Am I missing something?  Windows support issues?

dstufft:
@eric-tucker I think the comment by @jamesls stands even if we moved to it boto3 instead of botocore, since I don't think that boto3 has any default implementations that aren't in memory either. Would having a default in memory implementation with the way to configure it work for you? Does it *need* to be on by default?

eric-tucker:
@dstufft I think there should be some synergy between boto3 and the CLI since they're both powered by the same engine (sort of?)

Considering that all SDKs (I think) are cable of using ~/.aws contents I see no reason that by default the SDKs shouldn't also utilize the exact same mechanisms for caching MFA credentials as the CLI.  Consistency is key.

I believe that it's because this wasn't all done in a consistent manner across SDKs from day one that we're in the situation we're in now.  Consider that the SDKs are only catching up recently to supporting the same techniques that the CLI does, such as assumed role profiles.

I could be way off here, of course.

nskitch:
I'm just here as a cheerleader. Almost a month since last comment.  I'd love to see this functionality restored. Even if it means getting it working again, and assigning future work to follow the best practice suggestions.
typing an mfa code for each iteration of my python/boto script testing is slowing me down.   appreciate everybody's work!

rvandegrift:
@gene1wood are you still interested in this PR?  If not, I'm happy to pick it up and implement the changes that @jamesls suggested.

gene1wood:
74a061e87141c7f4615068461cad4b881ddff769 implements [the suggestion above](https://github.com/boto/botocore/pull/1157#issuecomment-304376764)

This change necessarily requires a PR against boto3 to update the [configuration documentation](https://github.com/boto/boto3/blob/develop/docs/source/guide/configuration.rst).

This uses two configuration settings, one to set the type of cache to be used, and one to set the directory to store the cache in the filesystem if a file based cache is used. This seeks to follow the intent of the `cache` argument to `credentials.AssumeRoleProvider` by leaving it open to additional cache systems being added down the road.

The default behavior is to function as botocore does currently by using an in memory cache.

The default file cache directory is a new directory, `~/.aws/cache`. This is different from the cache directory of awscli (`~/.aws/cli/cache`) but the user can set it to be the same so that botocore and awscli share cache files.

Feel free to suggest different configuration setting names or environment variable names, suggest words other than `memory` and `file` to select the cache system, suggest a different default directory or anything else.

I did not add a unit test for the session `conversion_func` values for `credential_cache` and `credential_cache_directory` which use `string.lower` and `os.path.exanduser` respectively because it felt like overkill but I'm happy to if it' needed.

gene1wood:
Well, that didn't work, lemme fix it.

et304383:
I see no reason to not use the same cache directory by default. If I'm switching between cli and boto I would prefer not to have to enter MFA twice. 

I'm in favour of defaults that support making user's lives easier.

gene1wood:
@et304383 I think @jamesls 's point was that the *default* behavior should be distinct from awscli and as a consequence shouldn't use the same directory but that the user should be able to set the directory to be the same.

@jamesls I'd like some clarification on your thought on this.

gene1wood:
@rvandegrift Thank you for the offer, I just hadn't gotten a moment to commit an update with jamesls's suggestions. I've done so now (but my unit tests are failing). If you have any interest in taking a look at 74a061e87141c7f4615068461cad4b881ddff769 and the unit tests, I could use the help.

rvandegrift:
@gene1wood Thanks for the update!  I see two issues:
- according to your test, `JSONFileCache` should store its working dir with user path expanded
- the test class you're adding to isn't actually using the session config the tests setup

With these fixed, the tests are all passing locally.  For some reason github won't let me open a PR with your branch as the base.  I've pushed these changes here: https://github.com/rvandegrift/botocore/tree/persistent-credential-cache-with-serialization

gene1wood:
@rvandegrift 

Thanks so much for digging into it!

> For some reason github won't let me open a PR with your branch as the base

Not sure, I'll just incorporate your changes

> the test class you're adding to isn't actually using the session config the tests setup

Aha! Bingo. Good catch.

> according to your test, JSONFileCache should store its working dir with user path expanded

Indeed, the expansion occurs in the [`conversion_func` of the `SESSION_VARIABLES`](https://github.com/boto/botocore/commit/74a061e87141c7f4615068461cad4b881ddff769#diff-29bf4883de496fd71c0deda976cc480fR110) and it looks like the problem is that since `fake_get_config_variable` gets it's values from `session_instance_vars` the `conversion_func` logic in `session.py` doesn't get run. As it looks like the generalized `conversion_func` behavior is [already tested](https://github.com/boto/botocore/blob/81f8c4cd042187fcd5e1d2a1f5cad2bbbbefa1d3/tests/unit/test_session.py#L148-L159) I'll just add the path expansion to the test.

gene1wood:
Aha and @andrewkrug pointed out that `string.lower` isn't present in Python 3. I'll correct this (sorry for all this noise, hopefully if this gets accepted it can be squash merged)

gene1wood:
[reposting comment from earlier now that the noise regarding unit tests has resolved into passing tests]

74a061e87141c7f4615068461cad4b881ddff769 93514847059154b3c5009bb36d93a2d57f2dc760 and 5eb91547e9c3ba3369ca9541f55a8adc93aaab82 implement [the suggestion above](https://github.com/boto/botocore/pull/1157#issuecomment-304376764)

This change necessarily requires a PR against boto3 to update the [configuration documentation](https://github.com/boto/boto3/blob/develop/docs/source/guide/configuration.rst).

This uses two configuration settings, one to set the type of cache to be used, and one to set the directory to store the cache in the filesystem if a file based cache is used. This seeks to follow the intent of the `cache` argument to `credentials.AssumeRoleProvider` by leaving it open to additional cache systems being added down the road.

The default behavior is to function as botocore does currently by using an in memory cache.

The default file cache directory is a new directory, `~/.aws/cache`. This is different from the cache directory of awscli (`~/.aws/cli/cache`) but the user can set it to be the same so that botocore and awscli share cache files. @et304383 [advocates](https://github.com/boto/botocore/pull/1157#issuecomment-317260983) for setting this default to `~/.aws/cli/cache`. @jamesls I'd like to get your thoughts on what this value should be.

Feel free to suggest different configuration setting names or environment variable names, suggest words other than `memory` and `file` to select the cache system, suggest a different default directory or anything else.

I did not add a unit test for the session `conversion_func` values for `credential_cache` and `credential_cache_directory` which use `lower` and `os.path.exanduser` respectively because it felt like overkill but I'm happy to if it' needed.

As a few of these commits involve bugfixing new unit tests, if this PR is accepted you may want to squash it.

Looking for thoughts from @jamesls and others.

et304383:
I can get why you might want to default caching to memory as you have to think about boto3 being used in web applications.  I get that.

But if I choose file cache it's almost certain that I'm running this interactively and thus caching should default to the same place as the CLI.  I just don't envision a scenario where a user wouldn't want that as the default behaviour.

Please, help me understand a scenario where I would not want to use the same cached credentials if flipping back and forth between the CLI and boto3 using the same profile.

rvandegrift:
@et304383 IMO you're correct about user expectations, but I think @jamesls is concerned about correctness.  Here's the kind of thing I'd be concerned about.

If botocore uses the same cache as awscli, the cache is now shared global state - that brings the risk of concurrent update bugs.  Keeping botocore's cache somewhere else is simpler than working out a robust, cross-platform locking scheme.  And doesn't awscli + patched botocore trigger this race condition?  awscli uses botocore to get temp creds, and both will race to update the cache.  (AFAICT nothing ensures that the cache update is finished being written when the result is returned to the caller.)

So the new location is safer.  If this is accepted, I imagine the awscli cache should be removed.

gene1wood:
> awscli uses botocore to get temp creds, and both will race to update the cache

I don't think this is the case in that awscli doesn't write to the cache itself, it merely [tells boto to do so](https://github.com/aws/aws-cli/blob/345df8555b8e5a40bbf2418484458ad9fe8ff77e/awscli/customizations/assumerole.py#L39-L40) so this change in `botocore` would just cause botocore to use the cache (in the same way asserting `credential_cache = True` in the config would do). So no race condition but it would make sense after this is merged to remove the then-redundant code in the awscli `assumerole.py` customization

> If this is accepted, I imagine the awscli cache should be removed.

Yes, all that `awscli` would need to do after this is to remove the [assume role customization](https://github.com/aws/aws-cli/blob/develop/awscli/customizations/assumerole.py) and add an awscli default cache location of `~/.aws/cli/cache`

gene1wood:
@jamesls Does this update satisfy [your concerns](https://github.com/boto/botocore/pull/1157#issuecomment-304376764)

rvandegrift:
> >awscli uses botocore to get temp creds, and both will race to update the cache
> 
> I don't think this is the case in that awscli doesn't write to the cache itself, it merely tells [boto to do so](https://github.com/aws/aws-cli/blob/345df8555b8e5a40bbf2418484458ad9fe8ff77e/awscli/customizations/assumerole.py#L39-L40)

Is that the right link?  The code there isn't asking boto to cache, it has a local copy of `JSONFileCache` that directly modifies the json files in `~/.aws/cli/cache`.  Am I missing something?

gene1wood:
I could be mistaken but I think that the existing awscli code 
* first [creates a `CredentialProvider`](https://github.com/aws/aws-cli/blob/345df8555b8e5a40bbf2418484458ad9fe8ff77e/awscli/customizations/assumerole.py#L39), which, with this PR merged, will either by default use an in memory cache or if set in the config use [the `JSONFileCache` in botocore](https://github.com/boto/botocore/pull/1157/files#diff-dcd72ae45553af71f1ceafb2d4ee5dcfR165)
* then [replaces the cache in the CredentialProvider]((https://github.com/aws/aws-cli/blob/345df8555b8e5a40bbf2418484458ad9fe8ff77e/awscli/customizations/assumerole.py#L39)) with the [`JSONFileCache` in awscli](https://github.com/aws/aws-cli/blob/345df8555b8e5a40bbf2418484458ad9fe8ff77e/awscli/customizations/assumerole.py#L43-L88)

In between those two steps I don't see any race condition. On line 39 the CredentialProvider would set `self.cache` to the botocore `JSONFileCache`, then on the subsequent line it would change `self.cache` to use the awscli `JSONFileCache`

@rvandegrift to help me understand, can you walk through a scenario where the race condition you're thinking of would be triggered?

rvandegrift:
Ah, now I get it - I missed that the cache provider is replaced.  Makes perfect sense now, sorry for being dense!

et304383:
OK so if the concerns around race conditions on file cache are satisfied, can we please merge this PR with the option to enable file cache in a **persistent** manner?  I don't want to have to set an additional option in code to enable this.  It should be a user environment setting.

gene1wood:
@et304383 This PR allows a user to enable file cache in a persistent manner by setting the [`credential_cache` and `credential_cache_directory` config settings](https://github.com/boto/botocore/pull/1157#issuecomment-317563584). No in-code settings are needed.

@jamesls Does this update satisfy [your concerns](https://github.com/boto/botocore/pull/1157#issuecomment-304376764)

gene1wood:
@jamesls Does this update satisfy [your concerns](https://github.com/boto/botocore/pull/1157#issuecomment-304376764)

gwkunze:
While we were eagerly anticipating the merging of this PR, the workaround we now use in our code is to use the caching mechanism of the awscli itself. This does require making the awscli a dependency of your app but at least it alleviates the problem of constantly having to wait for (and enter) new MFA codes.

```
        session = boto3.session.Session()
        # UGLY Hack to use credentials caching from awscli in this application
        # Awaiting https://github.com/boto/botocore/pull/1157 to be merged
        try:
            from awscli.customizations.assumerole import inject_assume_role_provider_cache
            from awscli.customizations.scalarparse import add_scalar_parsers
            inject_assume_role_provider_cache(session._session)
            add_scalar_parsers(session._session)
        except:
            pass
```

gene1wood:
@jamesls Does this update satisfy [your concerns](https://github.com/boto/botocore/pull/1157#issuecomment-304376764)

et304383:
Why on earth can't we merge this?  :(

joguSD:
The conflicts seem pretty minimal, mostly just imports and a declaration moved.

gene1wood:
Oh, didn't even notice that master had moved causing conflicts. Lemme take care of that.

JordonPhillips:
Thanks for opening the PR and driving it as far as you have! I've talked with the rest of the team about this, and here's the interface we'd like to go for:

```python
from botocore.credentials import create_credential_resolver
from botocore.credentials import JSONFileCache


sess = Session()
cache = JSONFileCache()
resolver = create_credential_resolver(session, cache)
sess.register_component('credential_resolver', resolver)
```

There are several reasons for this. This allows a customer to specify whatever cache implementation they want instead of specifically tying them to our built-in cache implementations. We didn't simply add a `credential_cache` session component because that would only ever be used once. The way the session works, the credential chain is generated exactly once. So if you set `credential_cache` after the chain had already been generated, your new cache would never be used. We can't simply re-generate the chain because customers can provide their own, so we would effectively be overwriting their custom chain with our default chain without warning.

With this approach customers can use whatever cache they want and the process for setting it will result in the behavior they would expect out of the `Session` whenever they decide to set it. It still leaves us the ability to add config file settings later on.

I'll be taking this on unless you really want to do it yourself. Let me know what you think.

et304383:
@JordonPhillips are you suggesting we'd have to enable proper caching of MFA on a script by script basis?

I would highly suggest providing a flexible way for users to set this on a system level.  MFA is for interactive execution of scripts.  Please tell me a scenario where I, as a user, would want to cache MFA for one script and not for another.

gene1wood:
@JordonPhillips Thanks!

et304383:
OK how do we restore the functionality of this PR?  Having to type MFA in every time I run a python script is driving me nuts.

gene1wood:
Here's a summary of where the session caching functionality in `botocore` is at

* The session caching in this PR was merged into `botocore` via #1338 and made available in [`botocore` 1.8.14](https://github.com/boto/botocore/blob/develop/CHANGELOG.rst#1814).
* So as to not interfere with `awscli`, the default on disk location of the cache was set to [`~/.aws/boto/cache`](https://github.com/boto/botocore/pull/1338/files#diff-dcd72ae45553af71f1ceafb2d4ee5dcfR192). This means that, by default, cached credentials are not shared between `awscli` and `botocore`.
* The ability to enable session caching in botocore via a [config setting](https://github.com/boto/botocore/pull/1157/commits/74a061e87141c7f4615068461cad4b881ddff769) was not included in #1338 and was not merged into `botocore`. I'm looking for more guidance on when or how we'd be able to get this configuration setting functionality in #1461.

In order to use this new session caching functionality, you'll need to both
* enable it in your code instead of your user's config
* set the `working_dir` where the on disk cache files are stored to be the same as where `awscli` stores it's cache. This is optional. With it `awscli` and `botocore` share a single set of cached sessions reducing the frequency with which you have to type your MFA code.

Here's a [code snippet](https://github.com/boto/botocore/pull/1338#issuecomment-368472031) from @mixja showing how to use the new functionality

```
from botocore import credentials
import botocore.session
import boto3
import os

# Change the cache path from the default of ~/.aws/boto/cache to the one used by awscli
working_dir = os.path.join(os.path.expanduser('~'),'.aws/cli/cache')

# Construct botocore session with cache
provider = session.get_component('credential_provider').get_provider('assume-role')
provider.cache = credentials.JSONFileCache(working_dir)

# Create boto3 client from session
client = boto3.Session(botocore_session=session).client('s3')

print(client.list_buckets())
```

PR #1159, opened 2017-02-27
Revert "Decode file paths in config loader"

JordonPhillips:
Reverts boto/botocore#1144

The tests were failing on our Windows test boxes. Reverting until we can get it fixed.

### DIFF
--- a/botocore/configloader.py
+++ b/botocore/configloader.py
@@ -14,9 +14,8 @@
 import os
 import shlex
 import copy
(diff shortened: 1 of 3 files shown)


galopin:
@JordonPhillips bummer! This Unicode decoding issue seems trickier to solve than anticipated. Hope you will eventually find a fix that will work alike on Linux, *BSD and Windows boxes…

PR #1160, opened 2017-03-08
Generalize container credential provider

jamesls:
You can now provide a full url in addition to a relative url to retrieve credentials.
I've added hostname validation to only allow the ECS IP and localhost.
Anything else results in an exception being raised.

cc @kyleknap @JordonPhillips @stealthycoin @dstufft 
cc @jasdel


kyleknap:
You may also want to include a changelog entry before you merge. Make it an ``enhancement``?

PR #1161, opened 2017-03-09
Added paginators for the Organizations service

Posnet:
I added the pagination definitions for the organizations service.


dstufft:
Thanks for this PR! This will need extra review before we can marge it, so I'm going to go ahead and tag it for review.

PR #1164, opened 2017-03-14
Fix for date parsing in AssumeRoleProvider _is_expired check

richarpw:


dstufft:
Thanks! I *think* that this is pretty safe to add, even if there are cases when it is infact a string it should still parse it. However it needs additional tests to ensure that everything is covered.

dstufft:
Nevermind on the tests, the issue was the test file got marked ``+x`` which caused nose to skip it.

PR #1170, opened 2017-03-23
Decode config file paths

JordonPhillips:
Fixes aws/aws-cli#2395

This was originally committed in #1144 but the tests failed on Windows due to some file system issues. This was resolved by creating the file on the fly and cleaning it up after.

I've run the tests (including integration) on several python versions on OSX, Linux, and Windows.

### DIFF
--- a/botocore/configloader.py
+++ b/botocore/configloader.py
@@ -14,8 +14,9 @@
 import os
 import shlex
 import copy
+import sys
 
-from six.moves import configparser
+from botocore.compat import six
 
 import botocore.exceptions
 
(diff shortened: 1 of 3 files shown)


botocore/configloader.py
```diff
@@ -142,12 +143,13 @@ def raw_config_parse(config_filename, parse_subsections=True):
         path = os.path.expandvars(path)
         path = os.path.expanduser(path)
         if not os.path.isfile(path):
-            raise botocore.exceptions.ConfigNotFound(path=path)
-        cp = configparser.RawConfigParser()
+            raise botocore.exceptions.ConfigNotFound(path=_unicode_path(path))
+        cp = six.moves.configparser.RawConfigParser()
```
dstufft: I *think* this is going to fail depending on what the current state of ``sys.modules`` is since it's relying on something having munged ``sys.modules["six"]`` to add the moves attribute. You likely want to do ``from botocore.compat.six import moves as six_moves`` and then ``six_moves.configparser.RawConfigParser()``.

JordonPhillips: Not sure I understand. Moves is added [here](https://github.com/boto/botocore/blob/develop/botocore/vendored/six.py#L300), what would be overriding that?

JordonPhillips: Not sure I understand. Moves is added [here](https://github.com/boto/botocore/blob/develop/botocore/vendored/six.py#L300), what would be overriding that?

dstufft: Ah, I missed that. I was just looking at the metapath importer from the bottom.

JordonPhillips:
Ran through the tests again, merging.

galopin:
@JordonPhillips congrats!

galopin:
@jamesls is there an ETA on when this bug fix is going to be released? 🚀

kyleknap:
@galopin It should have got released in yesterday's release: https://github.com/boto/botocore/blob/develop/CHANGELOG.rst#1530. Upgrading to the latest version of the CLI/botocore should fix this for you.

galopin:
@kyleknap ah, got it. Last time I checked it was still version 1.5.29… Cannot believe `awscli` is finally working on my _ageing_ system after all those years 😄

PR #1171, opened 2017-03-23
Explicitly reference methods in client docs

JordonPhillips:
Fixes boto/boto3#991

Our generated docs were using loose references to client methods,
which was causing some issues with generated links referencing
methods in other clients. This makes those references more explicit.


JordonPhillips:
Whoops, that was an incorrect reference. Fixed it.

PR #1172, opened 2017-03-31
Fix mturk method name

JordonPhillips:
Fix naming on `list_hits_for_qualification_type` for mturk. The old method name will still be supported.

This will need to be merged along with aws/aws-cli#2522


botocore/handlers.py
```diff
@@ -781,11 +781,25 @@ def _replace_content(self, section):
         section.write(updated_content)
 
 
+class CommandAlias(object):
+    def __init__(self, actual_name):
+        """ Aliases a non-extant method to an existing method.
```
stealthycoin: existent

JordonPhillips: http://lmgtfy.com/?q=define%3A+extant

botocore/handlers.py
```diff
@@ -781,11 +781,25 @@ def _replace_content(self, section):
         section.write(updated_content)
 
 
+class CommandAlias(object):
```
kyleknap: Sort of a strange class name because of the name ``Command`` since there is no command abstraction. Maybe call this ``ClientMethodAlias``?

botocore/handlers.py
```diff
@@ -781,11 +781,25 @@ def _replace_content(self, section):
...
+    def __init__(self, actual_name):
+        """ Aliases a non-extant method to an existing method.
+
+        :param actual_name: The name of the method that actually exists on
+            the client.
+        """
+        self._actual = actual_name
+
+    def alias_command(self, client, **kwargs):
```
kyleknap: Or we can just make this a ``__call__``? Using the command notation is a little confusing for me.

tests/unit/test_client.py
```diff
@@ -1368,6 +1368,23 @@ def return_mock_tuple(**kwargs):
...
         self.assertFalse(self.endpoint.make_request.called)
 
+    def test_getattr_emits_event(self):
+        emitter = mock.Mock()
+        emitter.emit_until_response.return_value = (None, None)
+
+        creator = self.create_client_creator(event_emitter=emitter)
+        service_client = creator.create_client('myservice', 'us-west-2')
+
+        emitter.emit_until_response.assert_not_called()
```
jamesls: I don't follow why we're asserting the thing isn't called before we actually try to access an unknown attribute?

tests/unit/test_client.py
```diff
@@ -1368,6 +1368,23 @@ def return_mock_tuple(**kwargs):
...
+
+        with self.assertRaises(AttributeError):
+            service_client.attribute_that_does_not_exist
+
+        emitter.emit_until_response.assert_called_once_with(
+            'getattr.myservice.attribute_that_does_not_exist',
+            client=service_client
+        )
+
```
jamesls: We should have a unit test here that asserts that an event handler that returns a response will be used instead of raising an attribute error (i.e a positive test case).

tests/unit/test_client.py
JordonPhillips: I should probably add a comment to that. It's to assert that we're not emitting the event at client creation.

JordonPhillips: I should probably add a comment to that. It's to assert that we're not emitting the event at client creation.

JordonPhillips:
Good to look at again

PR #1174, opened 2017-04-10
Fixed percent_encode when handing it bytes.

morganwahl:
`percent_encode` fails when handed a string that's already been encoded into bytes. Have it just URL-quote that instead.

This fixes an error I'm running into when using S3 via the "django storages" package. I have a few keys with non-ASCII characters. Django storages handles this by encoding them as UTF-8 and handing the bytes off to boto. On Python 2 this fails with a UnicodeDecodeError, and on Python 3 I don't think it will raise an exception, but will return a url-encoded repr of the string passed in (which is probably not what you would want).


morganwahl:
Oops, wrong branch. Will rebase.

PR #1176, opened 2017-04-11
utils.percent_encode fix (with test)

morganwahl:
`percent_encode` fails when handed a string that's already been encoded into bytes. Have it just URL-quote that instead.

This fixes an error I'm running into when using S3 via the "django storages" package. I have a few keys with non-ASCII characters. Django storages handles this by encoding them as UTF-8 and handing the bytes off to boto. On Python 2 this fails with a UnicodeDecodeError, and on Python 3 I don't think it will raise an exception, but will return a url-encoded repr of the string passed in (which is probably not what you would want).

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -318,10 +318,18 @@ def percent_encode(input_str, safe=SAFE_CHARS):
     producing a percent encoded string, this function deals only with
     taking a string (not a dict/sequence) and percent encoding it.
 
+    If given the binary type, will simply URL encode it. If given the
+    text type, will produce the binary type by UTF-8 encoding the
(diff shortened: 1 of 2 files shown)


botocore/utils.py
```diff
@@ -19,7 +19,7 @@
 import functools
 import weakref
 
-from six import string_types, text_type
+from six import string_types, text_type, binary_type
```
JordonPhillips: While you're at it, could you change this to import from `botocore.compat.six`?

jamesls: Given my changes from https://github.com/boto/botocore/pull/1180, I'll just go ahead and modify the import before merging.

PR #1182, opened 2017-04-20
Fix bug with sigv4 signer with consecutive spaces

jamesls:
See: http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html

Added integration test that fails without this change.
I verified that this affected glacier vault descriptions as well as s3
object metadata.

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -239,7 +239,7 @@ class SigV4Auth(BaseSigner):
         #
         # The Trimall function removes excess white space before and after
         # values, and converts sequential spaces to a single space.
-        return ' '.join(value.strip().split())
+        return ' '.join(value.split())
 
     def signed_headers(self, headers_to_sign):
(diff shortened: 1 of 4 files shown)


botocore/auth.py
```diff
@@ -228,11 +228,19 @@ def canonical_headers(self, headers_to_sign):
...
             headers.append('%s:%s' % (key, value))
         return '\n'.join(headers)
 
+    def _header_value(self, value):
+        # From the sigv4 docs:
+        # Lowercase(HeaderName) + ':' + Trimall(HeaderValue)
+        #
+        # The Trimall function removes excess white space before and after
+        # values, and converts sequential spaces to a single space.
+        return ' '.join(value.strip().split())
```
dstufft: Small nit, but the ``.strip()`` is extraneous here, the ``.split()`` will give the correct answer on its own.

jamesls: Should be fixed now.

PR #1185, opened 2017-04-26
Fix RDS Test for 2.6

JordonPhillips:
This injects a scheme into some compared urls as urlsplit will fail
to properly split if the scheme is absent in some versions.


JordonPhillips:
Just to make sure I also ran the tests on Windows and they pass there as well

PR #1186, opened 2017-04-27
Support paginating with tokens that contain bytes

JordonPhillips:
Some services (such as dynamodb) may use bytes in their pagination tokens, so this adds fallback support for that.

The solution in this PR is to attempt to encode as before, but if it fails to traverse the token dict encoding any bytes found and noting the path to those values.

Another option considered was using pickle, but the concern there was that we couldn't easily limit it to json+bytes and therefore we'd be exposing ourselves to code execution issues.

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -27,6 +27,152 @@ from botocore.utils import set_value_from_jmespath, merge_dicts
 log = logging.getLogger(__name__)
 
 
+class TokenEncoder(object):
+    """Encodes dictionaries into opaque strings.
+
+    This for the most part json dumps + base64 encoding, but also supports
+    having bytes in the dictionary in addition to the types that json can
+    handle by default.
+
+    This is intended for use in encoding pagination tokens, which in some
+    cases can be complex structures and / or contain bytes.
+    """
+
+    def encode(self, token):
+        """Encodes a dictionary to an opaque string.
+
+        :type token: dict
+        :param token: A dictionary containing pagination information,
+            particularly the service pagination token(s) but also other boto
+            metadata.
+
+        :rtype: str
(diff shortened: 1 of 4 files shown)


botocore/paginate.py
```diff
@@ -27,6 +27,149 @@
...
+        :returns: An opaque string
+        """
+        try:
+            # Try just using json dumps first to avoid having to traverse
+            # and encode the dict. In 99.9999% of cases this will work.
+            json_string = json.dumps(token)
+        except (TypeError, UnicodeDecodeError):
+            # If normal dumping failed, go through and base64 encode all bytes.
+            encoded_token, encoded_keys = self._encode(token, [])
+            encoded_token['boto_encoded_keys'] = encoded_keys
```
kyleknap: And the boto_encoded_keys is safe to do because no service will have boto_encoded_keys as a parameter? Might be good to add a comment about that.

botocore/paginate.py
```diff
@@ -27,6 +27,149 @@
...
+
+    def _path_get(self, data, path):
+        """Return the nested data at the given path.
+
+        For instance:
+            data = {'foo': ['bar', 'baz']}
+            path = ['foo', 0]
+            ==> 'bar'
+        """
+        # jmespath isn't used here because it would be difficult to actually
```
kyleknap: Yeah... it would be great if reverse jmespath was a thing. This is good for now.

JordonPhillips:
So it can have multiple bytes values, the paths just get appended to the list. I'll add in tests for that case.

botocore/paginate.py
JordonPhillips: Yeah, it's the same justification we use for the truncation key.

PR #1187, opened 2017-05-01
Support env var credential expiry

JordonPhillips:
This adds support for environment credential expiration via the `AWS_CREDENTIAL_EXPIRATION` environment variable.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -508,6 +508,7 @@ class EnvProvider(CredentialProvider):
     # The token can come from either of these env var.
     # AWS_SESSION_TOKEN is what other AWS SDKs have standardized on.
     TOKENS = ['AWS_SECURITY_TOKEN', 'AWS_SESSION_TOKEN']
+    EXPIRY_TIME = 'AWS_CREDENTIAL_EXPIRATION'
 
     def __init__(self, environ=None, mapping=None):
         """
@@ -533,6 +534,7 @@ class EnvProvider(CredentialProvider):
             var_mapping['access_key'] = self.ACCESS_KEY
             var_mapping['secret_key'] = self.SECRET_KEY
             var_mapping['token'] = self.TOKENS
+            var_mapping['expiry_time'] = self.EXPIRY_TIME
         else:
             var_mapping['access_key'] = mapping.get(
                 'access_key', self.ACCESS_KEY)
@@ -542,6 +544,8 @@ class EnvProvider(CredentialProvider):
                 'token', self.TOKENS)
             if not isinstance(var_mapping['token'], list):
                 var_mapping['token'] = [var_mapping['token']]
+            var_mapping['expiry_time'] = mapping.get(
+                'expiry_time', self.EXPIRY_TIME)
         return var_mapping
 
     def load(self):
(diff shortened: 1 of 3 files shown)


JordonPhillips:
RIP codecov

tests/unit/test_credentials.py
```diff
@@ -211,6 +211,29 @@ def test_can_override_partial_env_var_mapping(self):
         self.assertEqual(creds.secret_key, 'bar')
         self.assertEqual(creds.token, 'baz')
 
+    def test_can_override_expiry_env_var_mapping(self):
```
kyleknap: One more test that might be good to add is one that checks that it does not refresh the credentials when it is not going to expire any time soon.

botocore/credentials.py
```diff
@@ -550,19 +554,51 @@ def load(self):
...
-            access_key, secret_key = self._extract_creds_from_mapping(
-                self.environ, self._mapping['access_key'],
-                self._mapping['secret_key'])
-            token = self._get_session_token()
-            return Credentials(access_key, secret_key, token,
-                               method=self.METHOD)
+            fetcher = self._create_credentials_fetcher()
+            credentials = fetcher()
+            expiry_time = credentials['expiry_time']
+            if expiry_time is not None:
```
kyleknap: So for the very first time you load credentials and you do not have an expiry time set, it will not ever try to refresh if it you add the expiry time variable later on? I'm just curious about the behavior. Have not really thought out what should be desired.

JordonPhillips: Correct. I'm not sure how else we would do it without hacks

JordonPhillips:
Updated

botocore/credentials.py
kyleknap: Yeah I cannot think of other ways to do this. I guess another question I have is it possible to transition from refershable environment credentials to non-refreshable? So say I had an expiry time set and then unset the expiry time variable before the next refresh what would happen?

JordonPhillips: It would fail as it expects it to be there upon refresh

kyleknap: Ok cool. That at least ensures the behavior is consistent for both directions (i.e. you can't do it).

botocore/credentials.py
```diff
@@ -550,19 +554,51 @@ def load(self):
...
+                expiry_time = parse(expiry_time)
+            return RefreshableCredentials(
```
jamesls: I don't think we should rely on no expiry time meaning non-refreshable credentials.  I would prefer we're explicit.  If we get expiry time use RefreshableCredentials, otherwise use Credentials.

I'm guessing that code was put in place to protect against malformed responses and not necessarily as part of its contract.  It's always confusing to have RefreshableCredentials not actually refresh if you don't provide expiry.

botocore/credentials.py
jamesls: I think it would good to have tests for these cases (fixed->refreshable, refreshable->fixed are not allowed).  Unlike other refreshable providers were it was unlikely to have missing keys across subsequent responses, this is the first refreshable provider that seems more prone to refresh errors.

botocore/credentials.py
JordonPhillips: 👍 updated

koreno:
How should this feature be used? how would boto refresh the credentials once it finds they've expired?

PR #1190, opened 2017-05-03
Update s3.Bucket CreationDate docstring

kmosher:
Adding this as followup to boto/boto3#1064.

Existing docs are a little inconsistent in their usage of `null` vs `None`, but `null` seems slightly more in keeping with existing style?

(Ref boto/boto3#1063)


JordonPhillips:
Unfortunately we don't accept changes to our service models as they are generated upstream. This would need to be implemented in boto3 as a doc event handler that appends that information.

kmosher:
Oh, I hadn't seen those. If you don't mind, could you link an example where one is used in boto3?

PR #1191, opened 2017-05-04
Make ClientError a subclass of BotoCoreError and allow pickling of exception classes

### LINKED ISSUE #834: botocore exceptions are not picklable
they need to take a "msg" argument, and all other arguments must be optional: 

http://stackoverflow.com/questions/16244923/how-to-make-a-custom-exception-class-with-multiple-init-args-pickleable.  I've attempted fixing this in PR: https://github.com/boto/botocore/pull/835

simple testcase:

``` python
try:
    raise botocore.exceptions.NoCredentialsError
except Exception as e:
    b = pickle.dumps(e)
    e = pickle.loads(b)
```

### LINKED ISSUE #899: Exception classes should all extend a common botocore exception
Right now, if I want to handle all classes of botocore exceptions, I can't easily do so. For example, `ClientError` extends `Exception` directly. If they all extended from `BotoCoreError` or something, then I could easily handle any botocore exception.

jezdez:
Fix #834, fix #899.

### DIFF
--- a/botocore/errorfactory.py
+++ b/botocore/errorfactory.py
@@ -10,12 +10,15 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
-from botocore.exceptions import ClientError
+import sys
+import types
+
+from botocore import exceptions
 from botocore.utils import get_service_module_name
 
 
 class BaseClientExceptions(object):
-    ClientError = ClientError
+    ClientError = exceptions.ClientError
 
     def __init__(self, code_to_exception):
         """Base class for exceptions object on a client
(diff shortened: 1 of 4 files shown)


jezdez:
FTR, being able to pickle exceptions is usful for example when using boto3/botocore with Celery tasks whose exception handling can use pickle to store the exception in case one is raised when a task is executed.

jezdez:
@kyleknap Any time to review this?

jezdez:
Or @dstufft (hello!) for that matter?

JordonPhillips:
I don't think we can pull this in because it would be a breaking change for anybody catching `BotocoreError`.

jezdez:
@JordonPhillips I'm not sure I follow, `BotocoreError` continues to exist. Would you mind elaborating the case for which this would be a breaking change?

JordonPhillips:
Anywhere a user is catching `BotocoreError` they would suddenly be catching a whole new class of errors. Anywhere a user is only catching a `ClientError` there would be no issue. Anywhere a user is catching both and performing different actions there may be a breaking change depending on the ordering of the catches. Something like the following would break:

```python
try:
    client.service_call()
except BotocoreError:
    # panic
except ClientError:
    # show error to user
```

wimglenn:
@jezdez   He means you try for a way to make `ClientError` picklable, without making it a subclass of `BotocoreError`.

wimglenn:
IMO it's a mistake that `ClientError` and `BotocoreError` don't have some common, picklable ancestor.  We should have something like this instead

```
Exception
└── BotocoreErrorForRealsThisTime
    ├── BotocoreError
    └── ClientError
```

dstufft:
I like the idea of making a common ancestor for ``BotocoreError`` and ``ClientError``. Perhaps something like ``BaseBotocoreError`` or something.

peterbe:
My problem is that `EndpointConnectionError` exceptions can't be pickled. 
Is that an entirely separate problem? ...considering that `EndpointConnectionError` inherits from `BotoCoreError` and `ClientError` inherits from `Exception`.

peterbe:
For anybody else stumbling on this and can't wait, I just did the following:

```python
class OwnClientError(ClientError):  # XXX Replace "Own" with "Picklable" ?
    """Because the botocore.exceptions.EndpointConnectionError can't be
    pickled, if this exception happens during task work, celery
    won't be able to pickle it. So we write our own.

    See https://github.com/boto/botocore/pull/1191
    """

    def __reduce__(self):
        return (
            self.__class__,
            (self.response, self.operation_name),
            {},
        )

def reraise_clienterrors(f):
    """Decorator whose whole job is to re-raise any ClientError
    exceptions raised to be OwnClientError because those
    exceptions are "better". In other words, if, instead an
    OwnClientError exception is raised by the task
    celery can then pickle the error. And if it can pickle the error
    it can apply its 'autoretry_for' magic.
    """

    @wraps(f)
    def wrapper(*args, **kwargs):
        try:
            return f(*args, **kwargs)
        except ClientError as exception:
            raise OwnClientError(exception.response, exception.operation_name)
    return wrapper

@shared_task(autoretry_for=(
    ConnectionError,
    OwnClientError,
))
@reraise_clienterrors
def upload_inbox_upload(upload_id):
    # ...the celery task...
```

Seems to work for now.

thehesiod:
btw, can you make sure the callstack gets pickled as well?  This is an issue with the version I created where the callstack doesn't get printed with `logger.exception` from the parent process.

tedivm:
Is this ever going to get resolved? It makes working with boto3 and celery rather frustrating.

ryansb:
Perhaps a way to allow those chains to remain intact would be to add a new class that's mixed in to BotoCoreError, ClientError, etc, such as `AnyBotoCoreError` that would leave existing `except ClientError / except BotoCoreError` chains intact while letting new code have a catchall.

lightblu:
+1 for the picklability: The same issue is present when using Python3's [concurrent.futures.ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor). Letting boto3's exceptions raise in futures executed by a ProcessPoolExecutor leads to a broken process pool, because those exceptions cannot be pickled and sent back to the submitting process.

jezdez:
Closing since #1596 at least fixed the pickling issue.

PR #1192, opened 2017-05-10
Added exponential backoff functionality for elastic mapreduce clusters

aiell0:
I was experiencing throttling exceptions using the boto3 EMR API. I noticed no configuration in `_retry.json` for EMR, so I added it.

Issue: https://github.com/boto/botocore/issues/1193

### DIFF
--- a/botocore/data/_retry.json
+++ b/botocore/data/_retry.json
@@ -125,6 +125,26 @@
         }
       }
     },
+    "elasticmapreduce": {
+      "__default__": {
+        "max_attempts": 10,
+        "delay": {
+          "type": "exponential",
+          "base": 0.8,
+          "growth_factor": 2
+        },
+        "policies": {
+          "request_limit_exceeded": {
+            "applies_when": {
+              "response": {
+                "service_error_code": "ThrottlingException",
+                "http_status_code": 400
+              }
+            }
+          }
+        }
+      }
+    },
     "cloudsearch": {
       "__default__": {
         "policies": {


JordonPhillips:
This is already covered by the default configuration.

JordonPhillips:
This is already covered by the default configuration [here](https://github.com/boto/botocore/blob/develop/botocore/data/_retry.json#L82)

aiell0:
Jordon,

While this may be true, I find the default configuration for that to be too liberal for the EMR API. I still get throttled if I'm querying for many EMR clusters, and put that configuration in there to allow for more of a backoff. From talking to AWS, the API limits for EMR are:

API 	                       Bucket Size 	         Refill Rate (per second)
=== 	                =========== 	========================
DescribeCluster 	      10    	                       1.0
AddTags                        5                              .5

And since DescribeCluster must be called for each EMR, those rates are hit pretty quickly.

aiell0:
Bumping for an update. Will this be merged?

dstufft:
Looking at this, it appears that the primary difference between the default and the overridden retry is a max attempts of 10 (instead of 5) and a hardcoded base of 0.8 (instead of random). That would give us a retry time of... like 6 minutes if we start getting close to 10? That feels way too aggressive for a backoff to me.

At a minimum I think that this would need some mechanism to prevent us from having multiple minutes between retries. Thinking out loud, I'm also wondering if maybe a rate limit exception should have a higher number of retry counts by default then something like a socket error. The rationale being that we know sooner or later the rate limit exception should resolve itself, but a socket error might never do so.

I don't feel comfortable merging this as is, and I'd want at least @JordonPhillips or one of the other teams to weigh in on the above thoughts before telling you to go make changes.

dstufft:
From https://github.com/boto/botocore/issues/1199:

> For EMR, calling `DescribeCluster` for each EMR in an account gets throttled after 4 max retries. I created a PR for a fix, but it has not yet been approved.

-- -

> From test results, I call DescribeClusters on a total of 27 clusters, and I am getting rate limited on those with the default configuration. This is **after** I filtered out terminated or terminating clusters, so this is just a list of active and running clusters. This is not acceptable.

aiell0:
Bumping this, as I am sure this will come up again in the future. If anyone else wishes to chime in, I can go ahead and make the change.

Ips1986:
Added exponential backoff functionality for elastic mapreduce clusters #1192

I don't see exponential back off implemented for EMR. I was getting throttling exception and failed in 25 seconds. Why this pull request is not merged?

aiell0:
@dstufft This PR is really old. Can we either merge or delete? I'm attempting to clean up my profile a bit.

aiell0:
@dstufft Bump once more.

PR #1195, opened 2017-05-15
Add line iterator to StreamingBody

sujaymansingh:
This way, we can readlines from a streaming body (without having to load
all bytes into memory!)

This will allow us to use StreamingBody in places where a Python
file-like object is expected.
(E.g. csv.reader!)

### DIFF
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -80,6 +80,53 @@ class StreamingBody(object):
             self._verify_content_length()
         return chunk
 
+    def __iter__(self):
+        """Return an iterator to yield lines from the raw stream.
+        """
+        # Reading 1024 bytes at a time seems a sane choice [citation needed].
+        # If the user needs to specify something different they can always call
+        # streaming_body.iter_lines(<other-chunk-size>)
+        default_chunk_size = 1024
+        return self.iter_lines(default_chunk_size)
+
+    def iter_lines(self, chunk_size):
+        """Return an iterator to yield lines from the raw stream.
+
+        This is achieved by reading chunk of bytes (of size chunk_size) at a
+        time from the raw stream, and then yielding lines from there.
+        """
+        pending = None
+        for chunk in self._iter_chunks(chunk_size):
+            if pending is not None:
+                chunk = pending + chunk
+
+            lines = chunk.splitlines()
+
+            chunk_ends_with_newline = chunk[-1] == b"\n"
+
+            if lines and lines[-1] and chunk and not chunk_ends_with_newline:
+                # We might be in the 'middle' of a line. Hence we keep the
+                # last line as pending.
+                pending = lines.pop()
+            else:
+                pending = None
+
+            for line in lines:
+                yield line
+
+        if pending is not None:
+            yield pending
+
+    def _iter_chunks(self, chunk_size):
+        """Return an iterator to yield chunks of chunk_size bytes from the raw
+        stream.
+        """
+        while True:
+            current_chunk = self.read(chunk_size)
+            if current_chunk == b"":
+                raise StopIteration()
+            yield current_chunk
+
     def _verify_content_length(self):
         # See: https://github.com/kennethreitz/requests/issues/1855
         # Basically, our http library doesn't do this for us, so we have
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -122,7 +122,7 @@ class StreamingBody(object):
         while True:
             current_chunk = self.read(chunk_size)
             if current_chunk == b"":
-                raise StopIteration()
+                break
             yield current_chunk
 
     def _verify_content_length(self):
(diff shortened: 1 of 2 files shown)


botocore/response.py
```diff
@@ -80,6 +80,51 @@ def read(self, amt=None):
...
+    def __iter__(self):
+        """Return an iterator to yield lines from the raw stream.
+        """
+        # Reading 1024 bytes at a time seems a sane choice [citation needed].
+        # If the user needs to specify something different they can always call
+        # streaming_body.iter_lines(<other-chunk-size>)
+        default_chunk_size = 1024
+        return self.iter_lines(default_chunk_size)
+
+    def iter_lines(self, chunk_size):
```
dstufft: I think I would prefer to have this as a private method, normal file objects do not have an ``iter_lines`` method.

dstufft:
Overall this looks good to me, just one small comment.

botocore/response.py
sujaymansingh: Thanks @dstufft. I can change it to `_iter_lines`. That seem ok to you?

dstufft: Yea that seems fine to me.

sujaymansingh: @dstufft, I've made the change in @6866ab7
Hope it helps!

mbeacom:
Is this mergeable with the requested changes in? I think it'd be a great addition until the `IOBase`/`RawIOBase` possibilities are explored and tested further.

mcrowson:
Looks like the comment was addressed? There is a group of us really excited for this PR 💯

mrasband:
@dstufft Would it be best to use a `bytearray` here instead of creating a bunch of new `bytes` objects? For larger files (as common in s3) it seems like this could cause a bit of resource thrashing.

sujaymansingh:
@mrasband @dstufft Is the `bytearray` performance the only concern?

mrasband:
@sujaymansingh I don't have any authority to approve/merge - I stumbled on the issue as I wrote my own workaround a while ago and wanted to contribute it back (you already had). Eagerly awaiting this myself.

ranman:
This would be super cool to have for all streaming response types

sburns:
What will it take to get this merged?

W-Ely:
Do it!

ranman:
I would still love to see this!

iolairus:
Looking forward to this one, +1

seddy:
Any chance on getting this merged? It would be really useful for us!

smferguson:
+1. it'd be great to have this in!

billcrook:
+1

sujaymansingh:
@dstufft @mbeacom What should we do about this? At this point it's been over a year since this PR was raised. Presumably there have been billions of changes upstream.

I could rebase or branch off again, but it's only worth doing that if
- there is still a need/desire for this functionality from people
- it's something that is likely to be merged (it'd be terrible to have to wait another year :))

Would you say it's worth doing so? If so, I can create a new branch and submit.
Otherwise, it's probably worth abandoning this PR.

sburns:
> there is still a need/desire for this functionality from people

I hope the continued +1 comments are evidence of this.

sujaymansingh:
Good point @sburns !

jamesls:
Hi everyone sorry for the delay on this.  I'm taking a look at this now.  I'll assign this to myself and make sure this gets merged.

botocore/response.py
```diff
@@ -80,6 +80,51 @@ def read(self, amt=None):
...
+    def __iter__(self):
+        """Return an iterator to yield lines from the raw stream.
+        """
+        # Reading 1024 bytes at a time seems a sane choice [citation needed].
+        # If the user needs to specify something different they can always call
+        # streaming_body._iter_lines(<other-chunk-size>)
+        default_chunk_size = 1024
+        return self._iter_lines(default_chunk_size)
+
+    def _iter_lines(self, chunk_size):
```
jamesls: What are people's thoughts on exposing this method directly as `iter_lines()` instead of having the default behavior of `__iter__` be line based?

This would be similar to what requests exposes via it's [iter_lines](https://github.com/boto/botocore/blob/develop/botocore/vendored/requests/models.py#L705) method.  Looking at what they provide for `__iter__`, it appears to be chunks of 128 bytes.  That's closer to how I'd expect a streaming response object to behave, though I'd bump up the chunks to maybe 1k or 4k.

jamesls:
Given the age of the PR and that this was against your `develop` branch, I was hesitant to push to your fork so I pulled in your PR and incorporated some of the feedback I had. PR here: https://github.com/boto/botocore/pull/1491

botocore/response.py
sburns: +1 to the default iterator being larger chunks (4k but that's just me) and but also being able to get lines explicitly via `iter_lines()`.

sujaymansingh:
That's a great idea @jamesls! Thanks 👍

jamesls:
Merged via https://github.com/boto/botocore/pull/1491

PR #1198, opened 2017-05-16
Minor typo: ConditionExpresssion -> ConditionExpression

andrew-sledge:
it bugged me :)


dstufft:
Hi!

Thanks for the PR, I need to double check whether or not we can accept this or not, it might get overwritten. Let me figure that out before I merge this in. Thanks!

JordonPhillips:
As @dstufft hinted at, we don't accept updates to these models. I'll make sure it gets updated in the right place though. Thanks for pointing it out!

PR #1201, opened 2017-05-17
greatly speedup datetime parsing

thehesiod:
I run 60 queries in parallel per process and noticed that datetime parsing was ~25% of CPU usage.  This changes it such that it's not even noticeable anymore


dstufft:
Hi!

Speeding up botocore is always welcome, however it looks like you're missing coverage and all the tests are failing on py3. We would need to have passing tests and coverage before we're able to merge this.

thehesiod:
ok finally all tests passing, coverage passing, not sure why the above coverage wasn't replaced

PR #1202, opened 2017-05-18
Move all uses of six to the vendored version

jdufresne:
As botocore does not include six as a dependency, it should never
reference the globally installed six. Always use the vendored version
instead.

Helps ensure there is not an ImportError at runtime.


dstufft:
Thanks a lot! This looks good to me.

PR #1208, opened 2017-05-26
Use SigV4 by default for S3

JordonPhillips:
This re-introduces #958 . It has been rebased to work with the current state of develop.

We have since resolved the issues in handling unicode in the headers, integration tests have passed on python 2 and 3.


JordonPhillips:
coverage fixed

botocore/utils.py
```diff
@@ -1085,3 +1085,54 @@ def _get_response(self, full_url, headers, timeout):
 
     def full_url(self, relative_uri):
         return 'http://%s%s' % (self.IP_ADDRESS, relative_uri)
+
+
+def get_configured_signature_version(service_name, client_config,
```
kyleknap: What was the reason for having this pulled out into ``utils`` as public functions as opposed to keeping a ``client`` module level internal function?

tests/unit/test_args.py
```diff
@@ -130,3 +130,48 @@ def test_max_pool_from_client_config_forwarded_to_endpoint_creator(self):
...
+                'metadata': {}
+            }
+        ]
+        client_args = self.args_create.get_client_args(
+            service_model, 'us-west-2', True, 'http://other.com/', True, None,
+            {}, config, bridge)
+        self.assertEqual(
+            client_args['client_config'].region_name, 'us-west-2')
+
+    def test_region_does_not_resolve_if_not_s3_and_endpoint_url_provided(self):
```
kyleknap: I am having trouble following along the purpose of this test. Could you explain what this test is trying to do?

botocore/utils.py
```diff
@@ -1085,3 +1085,54 @@ def _get_response(self, full_url, headers, timeout):
...
+            if version:
+                logger.debug(
+                    "Switching signature version for service %s "
+                    "to version %s based on config file override.",
+                    service_name, version)
+                return version
+    return None
+
+
+def default_s3_presign_to_sigv2(signature_version, signing_name,
```
kyleknap: I think this would be good to have as an internal method of the ``ClientCreator`` class as it does not appear to be used by anything outside of that class.

botocore/utils.py
JordonPhillips: I think at one point it was being used in `args` too

botocore/utils.py
JordonPhillips: It seems a little weird to do things differently here than we do for other handlers

tests/unit/test_args.py
JordonPhillips: It tests that no region is inferred when you provide an endpoint url for non-s3 services.

botocore/utils.py
kyleknap: Would it be different? We currently do it for s3 addressing style: https://github.com/boto/botocore/blob/develop/botocore/client.py#L139-L142. For me, I suggested this because:

1) It is weird to have a public function in `utils` that is only used in one location. Typically if it is in `utils`, it should be used by more than one module.

2) By making it internal and not in the `utils` module, it gives us more flexibility to change the interface. I feel that public functions in `utils` is a gray area to break the public interface.

botocore/utils.py
kyleknap: Would it be possible to add it as a client method? I would prefer we do not add functionality to utils if it is not needed to be there and be public.

botocore/utils.py
JordonPhillips: That linked example isn't the same thing. It calls a function that picks a handler (or not) out of utils. Those handlers are only used in the one location as well. So it would be different. It also makes it harder to test since now we will only be able to have functional tests.

.changes/next-release/feature-s3-78494.json
```diff
@@ -0,0 +1,5 @@
+{
+  "description": "All S3 requests will now use SigV4 by default so that we can take advantage of stronger security algorithms and improved access key isolation.",
+  "type": "feature",
```
JordonPhillips: The question: do we want to minor version bump for this? There precedence for that sort of thing (changing to use s3transfer for instance).

botocore/auth.py
```diff
@@ -230,7 +231,7 @@ def canonical_headers(self, headers_to_sign):
         for key in sorted_header_names:
             value = ','.join(self._header_value(v) for v in
                              sorted(headers_to_sign.get_all(key)))
-            headers.append('%s:%s' % (key, value))
+            headers.append('%s:%s' % (key, ensure_unicode(value)))
```
kyleknap: Sorry I missed this one. Can you explain why this is needed?

JordonPhillips: Because we're tossing unicode into the headers. If we don't make sure it's encoded then we'll get ascii failures down the line.

JordonPhillips: Because we've got a set of header keys that are already unicode and a set of header values that are likely normal strings. When you the percent encoding it wants to decode the string using ascii, which fails. Try running this (which is run by our integration test) on python 2.6:

`'%s:%s' % (u'content-disposition', 'attachment; filename=5小時接力起跑.jpg;')`

Why don't we run into this issue now you might ask. Because sigv2 doesn't sign content disposition. It only signs `content-md5`, `content-type`, `date`, and any header that starts with `x-amz`.

kyleknap: Ok cool. Thanks for the explanation.

botocore/client.py
```diff
@@ -130,6 +134,9 @@ def _register_s3_events(self, client, endpoint_bridge, endpoint_url):
             client.meta.events.register_first(
                 'request-created.s3', switch_host_s3_accelerate)
 
+        self._set_s3_presign_signature_verion(
```
matthiask: `s/verion/version/`

JordonPhillips:
Ran tests across multiple python versions for botocore and the CLI with botocore on linux, windows, and macos. Merging.

JordonPhillips:
All tests have been run across multiple python versions on macos, windows, and linux.

PR #1214, opened 2017-06-03
Bump six

jtdoepke:
This PR bumps the vendored six version to 1.10.0. Pretty much a duplicate of #904, but for different reasons.

I ran into https://github.com/benjaminp/six/issues/98 while using botocore in some multithreaded code. That was fixed in six==1.9.0. Bumped to 1.10.0 because why not? Includes a regression test.


rmacduff:
I just ran into this threading issue with version 1.8.0 of six that ships with botocore. Indeed, bumping it up to 1.10 resolved my issue.

PR #1217, opened 2017-06-11
Fix SigV2Auth for GET requests

ltm:
The add_auth() method references request.param rather than
request.params which causes it to thow an AttributeError when invoked on
a GET request.

This fix is required to access the Alexa Top Sites API using botocore.

```python
from botocore.auth import SigV2Auth
from botocore.awsrequest import AWSRequest
from botocore.endpoint import BotocoreHTTPSession
from botocore.session import get_session

request = AWSRequest()
request.method = 'GET'
request.url = 'https://ats.amazonaws.com'
request.params['Action'] = 'TopSites'
request.params['ResponseGroup'] = 'Country'

credentials = get_session().get_credentials()
auth = SigV2Auth(credentials)
auth.add_auth(request)

session = BotocoreHTTPSession()
session.send(request.prepare())
```

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -109,7 +109,7 @@ class SigV2Auth(BaseSigner):
             params = request.data
         else:
             # GET
-            params = request.param
+            params = request.params
         params['AWSAccessKeyId'] = self.credentials.access_key
         params['SignatureVersion'] = '2'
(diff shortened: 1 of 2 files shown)


dstufft:
Thanks!

PR #1220, opened 2017-06-16
Use assertEqual instead of deprecated assertEquals

jdufresne:
Fixes warnings when running tests of the form:

```
DeprecationWarning: Please use assertEqual instead.
```


dstufft:
Thanks!

PR #1221, opened 2017-06-16
Fix all invalid escape sequences discovered running tests

jdufresne:
Fixes warning of the form:

```
DeprecationWarning: invalid escape sequence \...
```

See:

https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior


dstufft:
Looks good, thanks!

PR #1224, opened 2017-06-17
Enable pip cache in Travis CI builds

jdufresne:
See:

https://docs.travis-ci.com/user/caching/#pip-cache

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: python
 python:
   - "2.6"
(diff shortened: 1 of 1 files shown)


jamesls:
Seems relatively harmless but I'm curious, was this causing a specific issue for you?  Does this speed up builds on travis?  From looking at the last couple builds there doesn't seem to be much of a difference so I'm just curious as to the motivation for this.

jdufresne:
> was this causing a specific issue for you?

Not a specific issue, no. I try to always enable pip cache for projects I contribute to as it helps speed up test initialization without any side effects.

> From looking at the last couple builds there doesn't seem to be much of a difference so I'm just curious as to the motivation for this.

The cache hasn't been enabled yet. The cache is only saved once merged into the base branch. If you look at my Travis runs, you'll see pip is still downloading dependencies. When the cache is only enabled in the PR, it isn't saved for the next use.

The test suite is quite large, so its runtime may overshadow the initialization time.

My motivation is:

* Reduce runtime of tests
* Reduce stress on PyPI by reusing cache

JordonPhillips:
Seems reasonable. Even if it doesn't reduce time by much, every little bit helps.

PR #1228, opened 2017-06-22
Allow proxies to be set through the Config object.

timuralp:
Allows setting the proxy servers to be used through the Config object
when creating the client.

Refs: boto/boto3#338


dstufft:
Looks good, thanks!

omkarkulkarni:
Can we please get to see a sample code on how it works with a sample proxy. 
(I am trying for sqs)

PR #1230, opened 2017-06-23
Make copying UNSIGNED return the same instance

JordonPhillips:
botocore.UNSIGNED is meant as a sentinal value like None, so copying
it should always return the same instance. This implements that
behavior.

Fixes boto/boto3#1144

### DIFF
--- a/botocore/__init__.py
+++ b/botocore/__init__.py
@@ -62,7 +62,7 @@ BOTOCORE_ROOT = os.path.dirname(os.path.abspath(__file__))
 
 
 # Used to specify anonymous (unsigned) request signature
-class _UNSIGNED(object):
+class UNSIGNED(object):
     def __copy__(self):
         return self
 
@@ -70,7 +70,7 @@ class _UNSIGNED(object):
         return self
 
 
-UNSIGNED = _UNSIGNED()
+UNSIGNED = UNSIGNED()
 
 
 def xform_name(name, sep='_', _xform_cache=_xform_cache,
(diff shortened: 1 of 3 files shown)


botocore/__init__.py
```diff
@@ -60,8 +60,17 @@ def emit(self, record):
 
 BOTOCORE_ROOT = os.path.dirname(os.path.abspath(__file__))
 
+
 # Used to specify anonymous (unsigned) request signature
-UNSIGNED = object()
+class _UNSIGNED(object):
```
dstufft: More of a nit than anything, but would it be reasonable to do:

```python
class UNSIGNED(object):
    ...

UNSIGNED = UNSIGNED()
```

That would remove the ``_UNSIGNED`` class from the namespace as well.

JordonPhillips: Seems a bit odd, but then again there's no reason we need to keep the class around.

botocore/__init__.py
```diff
@@ -60,8 +60,17 @@ def emit(self, record):
...
 BOTOCORE_ROOT = os.path.dirname(os.path.abspath(__file__))
 
+
 # Used to specify anonymous (unsigned) request signature
-UNSIGNED = object()
+class UNSIGNED(object):
+    def __copy__(self):
+        return self
+
+    def __deepcopy__(self, memodict={}):
```
kyleknap: Not sure if we should be instantiating a dictionary in the method signature. Is this a special case for `__deepcopy__()`?

JordonPhillips: Yeah that's just the signature for deepcopy

PR #1231, opened 2017-06-24
Release HTTP connection back to the pool after making an API request

jdufresne:
Fixes warnings in application code:

```
  ResourceWarning: unclosed <ssl.SSLSocket fd=X, family=AddressFamily.AF_INET, type=2049, proto=6, laddr=('...', ...), raddr=('52.95.145.10', 443)>
```

The warning normally only occurs for 4XX responses.

Fixes boto/boto3#454

### DIFF
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -12,6 +12,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 
+import contextlib
 import os
 import logging
 import time
@@ -224,14 +225,15 @@ class Endpoint(object):
             logger.debug("Exception received when sending HTTP request.",
                          exc_info=True)
             return (None, e)
-        # This returns the http_response and the parsed_data.
-        response_dict = convert_to_response_dict(http_response,
-                                                 operation_model)
-        parser = self._response_parser_factory.create_parser(
-            operation_model.metadata['protocol'])
-        parsed_response = parser.parse(
-            response_dict, operation_model.output_shape)
-        return (http_response, parsed_response), None
+        with contextlib.closing(http_response):
+            # This returns the http_response and the parsed_data.
+            response_dict = convert_to_response_dict(http_response,
+                                                     operation_model)
+            parser = self._response_parser_factory.create_parser(
+                operation_model.metadata['protocol'])
+            parsed_response = parser.parse(
+                response_dict, operation_model.output_shape)
+            return (http_response, parsed_response), None
 
     def _looks_like_dns_error(self, e):
         return 'gaierror' in str(e) and e.request is not None


botocore/endpoint.py
```diff
@@ -231,6 +231,7 @@ def _get_response(self, request, operation_model, attempts):
             operation_model.metadata['protocol'])
         parsed_response = parser.parse(
             response_dict, operation_model.output_shape)
+        http_response.close()
```
dstufft: I think this would be better written as:

```python
with http_response:
    response_dict = ...
    parser = ...
    parsed_response = ...
    return (http_response, parsed_response), None
```

It has the same effect, but it ensures it gets closed even if there is a exception here.

jamesls: Is this safe?  I don't think this accounts for streaming responses does it?

dstufft:
Talking offline, this is probably going to require a lot more work to handle streaming responses. We pass the http_response as part of the object we return in some cases so we can't close it until *after* the user has consumed the response.

jdufresne:
I've modified the code to use a context manager.

What do you think about this alternative implementation which moves the closing to `BaseClient._make_api_call()` so as to delay calling close until after the `'after-call.{endpoint_prefix}.{operation_name}'` event is emitted. 


```patch
diff --git a/botocore/client.py b/botocore/client.py
index e7a81c5..5ec6974 100644
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -10,6 +10,7 @@
+import contextlib
 import logging
 import functools
 
@@ -585,20 +586,21 @@ class BaseClient(object):
             http, parsed_response = self._endpoint.make_request(
 
-        self.meta.events.emit(
-            'after-call.{endpoint_prefix}.{operation_name}'.format(
-                endpoint_prefix=self._service_model.endpoint_prefix,
-                operation_name=operation_name),
-            http_response=http, parsed=parsed_response,
-            model=operation_model, context=request_context
-        )
+        with contextlib.closing(http):
+            self.meta.events.emit(
+                'after-call.{endpoint_prefix}.{operation_name}'.format(
+                    endpoint_prefix=self._service_model.endpoint_prefix,
+                    operation_name=operation_name),
+                http_response=http, parsed=parsed_response,
+                model=operation_model, context=request_context
+            )
 
-        if http.status_code >= 300:
-            error_code = parsed_response.get("Error", {}).get("Code")
-            error_class = self.exceptions.from_code(error_code)
-            raise error_class(parsed_response, operation_name)
-        else:
-            return parsed_response
+            if http.status_code >= 300:
+                error_code = parsed_response.get("Error", {}).get("Code")
+                error_class = self.exceptions.from_code(error_code)
+                raise error_class(parsed_response, operation_name)
+            else:
+                return parsed_response
 
     def _convert_to_request_dict(self, api_params, operation_model,
                                  context=None):
```

jdufresne:
I'm motivated to fix this runtime warning as it causes a log of noise in my logs. If the original approach isn't workable, I'm very willing to investigate other solution or incorporate necessary feedback.

kdaily:
The solution presented here as a fix for boto/boto3#454 was discussed recently as to why we would not be able to accept this change:

https://github.com/boto/boto3/issues/454#issuecomment-1047277122

Closing this out, please check on boto/boto3#454 for updates on proposed solutions. Thanks!

Issue #1205, opened 2017-05-22
Configurable exponential timeout


Spektor0307:
I am trying to use boto for a various set of operations for which I would like to have the same logic in terms of retry attempts. After looking through the documentations, with the exception of being able to set a maximum number of linear retries for downloading data, it does not seem that this logic is exposed at all to the user. Is your documentation out of date, or is there a reason why this is not exposed?

dstufft:
There are a number of places where botocore internally has retry capabilities for but you're correct that we don't expose it to end users. I'm going to go ahead and flag this as a feature request. Thanks!

eric-weaver:
This is also functionality I am currently looking for. Specifically being able to lower the retry count on failed s3 upload requests. The default of 5 retries in `_retry.json` takes ~9s to throw a ClientError which is a fairly long time to wait.

jantman:
#1260 implemented configurable max_attempts for retries. As far as I can tell, overriding the ``growth_factor`` on the [delay_config](https://github.com/boto/botocore/blob/ed311bd4c9b4b630912ca148e541b7ad2dcbe448/botocore/retryhandler.py#L89) should be a very similar change.

taraslayshchuk:
@jantman https://github.com/boto/botocore/pull/891#issuecomment-321910539

jantman:
:( ok

declanwilson-lw:
Any further thoughts to allowing a base_delay to be configurable, as part of the config users can make on the client, subsequently passed on through to the 
I believe IAM endpoints quota replenish operates on discreet 1 second windows.  Having retries within the current 1 second boundary only adds to the problem, for the iam client boto3.

PR #1236, opened 2017-07-07
Allow max_attempts to be set through the Config object.

eric-weaver:
This change allows for configuration of the [max_attempts](https://github.com/boto/botocore/blob/develop/botocore/data/_retry.json#L77) variable through the config object. Currently the default limit of 5 retries before throwing a ClientError takes too long under certain circumstances and it would be helpful to be able to lower this. This seems to be mentioned as a `feature-request` in this botocore issue https://github.com/boto/botocore/issues/1205 and this boto3 issue https://github.com/boto/boto3/issues/1104

If I am missing anything please let me know as I am not the most familiar with the internal workings of this library.

If this functionality gets merged in I would be happy to extend the functionality over to boto3 if needed.


kyleknap:
@eric-weaver Thank you for the contribution. We recently merged a similar PR: https://github.com/boto/botocore/pull/1260 that allows you to set maximum retry attempts. Functionally, it is the same of what your PR does, but it just exposes the parameter a little differently. For that reason, I am going to close out this PR, but thanks again for sending the PR!

eric-weaver:
No problem! Glad to see the functionality making it into botocore!

PR #1244, opened 2017-07-13
set charset parameter to utf-8 for simpledb queries

memory:
Addresses https://github.com/boto/boto3/issues/354

This is my first time mucking about in models.py so feedback is definitely appreciated.

### DIFF
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -138,6 +138,11 @@ class Endpoint(object):
     def make_request(self, operation_model, request_dict):
         logger.debug("Making request for %s (verify_ssl=%s) with params: %s",
                      operation_model, self.verify, request_dict)
+        if self._endpoint_prefix == 'sdb':
+            logger.debug("Forcing charset to UTF-8 for SimpleDB request")
+            # https://github.com/boto/boto3/issues/354
+            request_dict['headers']['Content-Type'] = (
+                'application/x-www-form-urlencoded; charset=utf-8')
         return self._send_request(request_dict, operation_model)
 
     def create_request(self, params, operation_model=None):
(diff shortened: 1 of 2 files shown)


memory:
Note that the failure in `tests.functional.test_endpoints.test_service_name_matches_endpoint_prefix` appears to also be present in develop/HEAD and is not AFAICT related to this PR.

memory:
Ugh, okay.  Any suggestions on where might be a reasonable place to shim this in?

memory:
okay, doing this the sleazy way: special-casing SDB inside endpoint.py.

memory:
(I'm on vacation this week; please don't close this just yet -- will deal with the coverage issues when I get back.)

memory:
@JordonPhillips I believe this is ready for review now.

memory:
Poke? :)

jamesls:
Assigning to @stealthycoin.  This might be an issue with our query serializers.

memory:
PR https://github.com/boto/botocore/pull/1285 addresses this in a more systematic fashion; closing this as there is not ever going to be a reason to merge it.  Thanks @stealthycoin!

woodrow:
@memory can you help me understand how to reproduce this error? I am wondering if #1285 is not the right fix (I don't know that the header needs to be signed) but I want to do more research before proposing a fix.

memory:
@woodrow see the commentary at boto/boto3#354 -- updating a simpledb item attribute with any xml-valid utf-8 character should produce the behavior with a pre-#1285 boto3, e.g.:

```
sdb.put_attributes(DomainName = 'foo',
                   ItemName = 'foo',
                   Attributes = [{'Name': 'foo', 'Value': 'bär', 'Replace': True}])
```

PR #1245, opened 2017-07-14
reading 0 bytes from StreamingBody does not return IncompleteReadError

cclaassen3:
I kept running into unexplained IncompleteReadErrors and realized the source came from reading 0 bytes from a StreamingBody object. I cross referenced with the Python documentation [here](https://docs.python.org/2/library/stdtypes.html#file.read) and [here](https://docs.python.org/2/tutorial/inputoutput.html#methods-of-file-objects), and Python does not throw an error for this, so I don't think botocore should either (especially since it supports negative num bytes as an argument as well, so zero is the only exception).

I just added a quick AND statement to the code to encompass this use case & wrote a test case to assert the error is no longer thrown.

### DIFF
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -73,7 +73,7 @@ class StreamingBody(object):
         """
         chunk = self._raw_stream.read(amt)
         self._amount_read += len(chunk)
-        if not chunk or amt is None:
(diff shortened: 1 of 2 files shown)


joguSD:
Closed via #1312

Issue #1063, opened 2016-10-28
Route53 param validation for PageSize


manuel-sousa:
This might be related to #243 

From cli:
aws route53  list-resource-record-sets --page-size 15 --hosted-zone-id 123

Parameter validation failed:
Invalid type for parameter MaxItems, value: 1, type: <class 'int'>, valid types: <class 'str'>

Tested this in ubuntu 16.10, before and after updateing boto,  and latest amazon ami.

Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/awscli/clidriver.py", line 186, in main
    return command_table[parsed_args.command](remaining, parsed_args)
  File "/usr/lib/python2.7/dist-packages/awscli/clidriver.py", line 379, in **call**
    return command_table[parsed_args.operation](remaining, parsed_globals)
  File "/usr/lib/python2.7/dist-packages/awscli/clidriver.py", line 549, in **call**
    call_parameters, parsed_globals)
  File "/usr/lib/python2.7/dist-packages/awscli/clidriver.py", line 674, in invoke
    self._display_response(operation_name, response, parsed_globals)
  File "/usr/lib/python2.7/dist-packages/awscli/clidriver.py", line 683, in _display_response
    formatter(command_name, response)
  File "/usr/lib/python2.7/dist-packages/awscli/formatter.py", line 68, in __call__
    response_data = response.build_full_result()
  File "/usr/lib/python2.7/dist-packages/botocore/paginate.py", line 290, in build_full_result
    for response in self:
  File "/usr/lib/python2.7/dist-packages/botocore/paginate.py", line 102, in **iter**
    response = self._make_request(current_kwargs)
  File "/usr/lib/python2.7/dist-packages/botocore/paginate.py", line 174, in _make_request
    return self._method(**current_kwargs)
  File "/usr/lib/python2.7/dist-packages/botocore/client.py", line 159, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/lib/python2.7/dist-packages/botocore/client.py", line 470, in _make_api_call
    api_params, operation_model, context=request_context)
  File "/usr/lib/python2.7/dist-packages/botocore/client.py", line 523, in _convert_to_request_dict
    api_params, operation_model)
  File "/usr/lib/python2.7/dist-packages/botocore/validate.py", line 270, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())
ParamValidationError: Parameter validation failed:
[...]

PR #1247, opened 2017-07-20
Fix Route 53 pagination when string MaxItems is expected

joguSD:
Route 53 has defined their `MaxItems` pagination field as a string which leads to issues when passing the `page-size` flag to the cli as it will convert the value to an integer. This PR adds a hook to automatically convert the `MaxItems` field back into a string in the case that the operation expects it to be a string.

Fixes: https://github.com/boto/botocore/issues/1063

@JordonPhillips @stealthycoin @kyleknap @jamesls @dstufft

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -545,7 +545,8 @@ class PageIterator(object):
 class Paginator(object):
     PAGE_ITERATOR_CLS = PageIterator
 
-    def __init__(self, method, pagination_config):
+    def __init__(self, method, pagination_config, model):
+        self._model = model
         self._method = method
         self._pagination_cfg = pagination_config
         self._output_token = self._get_output_tokens(self._pagination_cfg)
@@ -623,11 +624,17 @@ class Paginator(object):
             max_items = int(max_items)
         page_size = pagination_config.get('PageSize', None)
         if page_size is not None:
-            if self._pagination_cfg.get('limit_key', None) is None:
+            if self._limit_key is None:
                 raise PaginationError(
                     message="PageSize parameter is not supported for the "
                             "pagination interface for this operation.")
-            page_size = int(page_size)
+            input_members = self._model.input_shape.members
+            limit_key_shape = input_members.get(self._limit_key)
+            if limit_key_shape.type_name == 'string':
+                if not isinstance(page_size, six.string_types):
+                    page_size = str(page_size)
+            else:
+                page_size = int(page_size)
         return {
             'MaxItems': max_items,
             'StartingToken': pagination_config.get('StartingToken', None),
--- a/tests/functional/test_route53.py
+++ b/tests/functional/test_route53.py
@@ -1,0 +1,51 @@
+# Copyright 2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from tests import unittest
+
+import botocore.session
+from botocore.stub import Stubber
+
+class TestRoute53Pagination(unittest.TestCase):
+    def setUp(self):
+        self.session = botocore.session.get_session()
+        self.client = self.session.create_client('route53', 'us-west-2')
+        self.stubber = Stubber(self.client)
+        # response has required fields
+        self.response = {
+            'HostedZones': [],
+            'Marker': '',
+            'IsTruncated': True,
+            'MaxItems': '1'
+        }
+        self.operation_name = 'list_hosted_zones'
+
+    def test_paginate_with_max_items_int(self):
+        # Route53 has a string type for MaxItems.  We need to ensure that this
+        # still works with integers as the cli auto converts the page size
+        # argument to an integer.
+        self.stubber.add_response(self.operation_name, self.response)
+        paginator = self.client.get_paginator('list_hosted_zones')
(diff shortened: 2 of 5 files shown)


botocore/handlers.py
```diff
@@ -734,6 +735,17 @@ def convert_body_to_file_like_object(params, **kwargs):
...
 
 
+def fix_route53_max_items(params, model, **kwargs):
+    # Converts the MaxItems param from an integer to a string.
+    # Route53 has defined their MaxItems field for pagination as a string.
+    # This causes issues when the cli automatically converts the page size
+    # to an integer as it will fail validation.
+    if 'MaxItems' in params:
+        if (isinstance(params['MaxItems'], six.integer_types) and
+            isinstance(model.input_shape.members['MaxItems'], StringShape)):
```
JordonPhillips: This should be indented over to not share the same level as the next logical line. If that makes things too long you can also flatten all of it out with guard clauses.

joguSD: I actually researched this yesterday and found PEP8 has no stance. I'll update the indentation to match what exists already.

tests/functional/test_route53.py
```diff
@@ -0,0 +1,38 @@
...
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from tests import unittest
+
+import botocore.session
+from botocore.stub import Stubber, ANY
```
JordonPhillips: I don't believe you're using `ANY`

kyleknap:
So one concern that I have with this approach is that while it fixes the the issue with Route53, it may still be an issue for other services that have a string type for their page size. An audit would need to be done to see if there other services do this.

However looking into the botocore code, this is the [line](https://github.com/boto/botocore/blob/develop/botocore/paginate.py#L630) that is causing the issue as we blindly convert the page size into an integer.

Unfortunately, I am not sure if the solution is as easy as removing that line because while we document that page size as an integer, a user could provide a string and it will still work. Thus, removing that line would potentially cause issues.

I think in order to put in a more general solution, we will need to:

1) Remove the line that always converts the page size configuration value to an integer.
2) Include the operation model in the paginator so the types are converted appropriately if on the provided pagination configuration values.

I think I would prefer this approach as it will ensure that any existing or future services do not hit this issue. The only downside with this approach is that the paginators will now have reference to the operation model, but that should not be too big of an issue as the model is already loaded prior to the `get_paginator()` call.

Let me know what you think.

joguSD:
I totally agree that this implementation is not very future proof. I had thought about pushing this fix down to pagination code as you mention but was unsure if the target type would be available at that point. Adding a reference to the shape in the paginators is definitely viable, but would require us to default to int in the case that it's not there, or requiring it and thus having to retroactively update all paginator definitions (and possibly breaking any custom paginators customers may have). I'll look into what's possible.

JordonPhillips:
@joguSD I think he means adding the shapes to the actual `Paginator` python object, not as a new key in the paginator spec.

joguSD:
There actually already is a unit test for this exact situation. It can be seen [here](https://github.com/boto/botocore/blob/develop/tests/unit/test_paginate.py#L722). However, a unit test is insufficient to catch this as it's not an issue with the pagination or validation code independently but both of them combined, hence the functional test.

botocore/paginate.py
```diff
@@ -623,11 +625,17 @@ def _extract_paging_params(self, kwargs):
...
         if page_size is not None:
-            if self._pagination_cfg.get('limit_key', None) is None:
+            if self._limit_key is None:
                 raise PaginationError(
                     message="PageSize parameter is not supported for the "
                             "pagination interface for this operation.")
-            page_size = int(page_size)
+            input_members = self._model.input_shape.members
+            limit_key_shape = input_members.get(self._limit_key)
+            if isinstance(limit_key_shape, StringShape):
```
joguSD: Hmm. I'm not sure how I feel about this. Either way this code will have a dependency on `StringShape`. The way you propose will remove the import but the dependency on `StringShape`'s `type_name` being `'string'` is still there. Can you elaborate on why you want to avoid the import?

kyleknap:
@joguSD I don't think that test applies to what you are trying to fix. The issue you are trying to fix is when `PageSize` is provided in the pagination config as a string as it is modeled as a string in service model. As to a counter argument to just relying on a functional test, I see at least a couple of benefits out of adding the unit tests:

1) If tests start fail, it is easier to determine that it was a change in the pagination code that is causing the issue if there is a unit test for this because a unit test will only start failing if there is an issue with the pagination code. With just a functional test, it is entirely possible that the service definition changed (i.e. changing the type of Route53's `MaxItems` to an integer) and that would cause failures as well, making it a bit more difficult to figure out why it failed

2) If the SDK made a major version bump and for whatever reason the service model definition changed in the major version bump (i.e. making the `MaxItems` an integer), it is entirely possible that the test becomes invalid, we remove it, but then no longer have coverage for this edge case.

In the end, it never really hurts to have the extra bit of testing especially since it should not be that much more effort to add.

botocore/paginate.py
kyleknap: I prefer `type_name` over `StringShape` for the following reasons:

1) It is more consistent throughout the botocore codebase. Just grepping through the botocore codebase `type_name` is used considerably more than `StringShape`. When grepping for `StringShape`, I see it only being used a couple of times in the `model` module

2) It is entirely possible for us to change the class name of `StringShape`, but it is way less possible for the `type_name` to ever change. This value is directly pulled from the JSON model and if the literal value `"string"` was to ever change it would break not just our SDK, but probably all of the other SDKs.

3) It seems weird that we would import `StringShape` just for an `isinstance` check when we can just use `type_name`. Relying on an import for instance checks seems to defeat the purpose of having all of the shape classes from the `model` module subclassing from the `Shape` class and having a `type_name` property.

joguSD: Good points, I'll change this over.

PR #1253, opened 2017-07-28
Fix escaped characters in shared example doc generation

joguSD:
Some shared examples have escape characters in them that get interpolated and rendered into the generated documentation which causes issues when we hand it off to sphinx.

Before:
screen shot 2017-07-28 at 2 03 44 pm

After:
screen shot 2017-07-28 at 2 10 04 pm

### DIFF
--- a/botocore/docs/utils.py
+++ b/botocore/docs/utils.py
@@ -10,6 +10,7 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import re
 from collections import namedtuple
 
 
@@ -177,3 +178,20 @@ class AppendParamDocumentation(object):
             description_section = section.get_section(
                 'param-documentation')
             description_section.writeln(self._doc_string)
+
+
+_CONTROLS = {
+    '\n': '\\n',
+    '\r': '\\r',
+    '\t': '\\t',
+    '\b': '\\b',
+    '\f': '\\f',
+}
+# Combines all CONTROLS keys into a big or regular expression
+_ESCAPE_CONTROLS_RE = re.compile('|'.join(map(re.escape, _CONTROLS)))
+# Based on the match get the appropriate replacement from CONTROLS
+_CONTROLS_MATCH_HANDLER = lambda match: _CONTROLS[match.group(0)]
+
+
+def escape_controls(value):
+    return _ESCAPE_CONTROLS_RE.sub(_CONTROLS_MATCH_HANDLER, value)
--- a/botocore/docs/sharedexample.py
+++ b/botocore/docs/sharedexample.py
@@ -13,6 +13,7 @@
 import re
 import numbers
 from botocore.utils import parse_timestamp
+from botocore.docs.utils import escape_controls
 from botocore.compat import six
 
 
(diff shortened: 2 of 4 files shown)


botocore/utils.py
```diff
@@ -48,6 +48,20 @@
 RETRYABLE_HTTP_ERRORS = (requests.Timeout, requests.ConnectionError)
 S3_ACCELERATE_WHITELIST = ['dualstack']
 
+CONTROLS = {
```
kyleknap: I think I would prefer this lives in botocore/docs/utils.py because it is really only used in the documentation and the `botocore/docs/utils.py` is more internal than `botocore/utils.py` module

botocore/utils.py
```diff
@@ -48,6 +48,20 @@
...
+    '\n': '\\n',
+    '\r': '\\r',
+    '\t': '\\t',
+}
+# Combines all CONTROLS keys into a big or regular expression
+ESCAPE_CONTROLS_RE = re.compile('|'.join(map(re.escape, CONTROLS)))
```
kyleknap: Also it would be good to make these constants being with an underscore and only make `escape_controls` public. This just further restricts the interface we are exposing.

botocore/utils.py
```diff
@@ -48,6 +48,20 @@
...
+# Based on the match get the appropriate replacement from CONTROLS
+CONTROLS_MATCH_HANDLER = lambda match: CONTROLS[match.group(0)]
```
kyleknap: Just as a heads up, we usually try to avoid using `lambdas` and defer to defining a function. While for this case it is not bad, the main reason is that following tracebacks in lambdas can be hard.

botocore/utils.py
joguSD: I was actually unsure of where to put this as it's definitely not needed (for now) outside of doc generation. Thanks.

botocore/utils.py
joguSD: Yeah that's how I was planning on doing it initially but ended up following the regex example above it. I'll go ahead and mark it as private when I move it to `docs/utils.py`

botocore/utils.py
joguSD: Got it.

botocore/utils.py
```diff
@@ -48,6 +48,20 @@
 RETRYABLE_HTTP_ERRORS = (requests.Timeout, requests.ConnectionError)
 S3_ACCELERATE_WHITELIST = ['dualstack']
 
+CONTROLS = {
+    '\n': '\\n',
+    '\r': '\\r',
+    '\t': '\\t',
```
JordonPhillips: Are these the only control characters that will get borked by sphinx? What if the string happens to contain `\a`? Or something that looks like bytes like `\xe2`?

botocore/docs/utils.py
```diff
@@ -177,3 +178,18 @@ def append_documentation(self, event_name, section, **kwargs):
             description_section = section.get_section(
                 'param-documentation')
             description_section.writeln(self._doc_string)
+
+
+_CONTROLS = {
+    '\n': '\\n',
```
JordonPhillips: Are these the only characters that will cause sphinx to do weird things? What about other control characters like `\a`? What about things that look like bytes such as `\xe2`?

joguSD: I *think* it's just characters that mess with whitespace. I'll try the examples you gave.

joguSD: According to json.org the following are control characters:
```
\b
\f
\n
\r
\t
```

I'll add all of these just in case, but as far as I can tell the three there currently are the only ones that will break sphinx rendering.

Issue #482, opened 2015-03-05
More retrying for CreateNetworkAclEntry?


Jc2k:
It's quite easy to hit RequestLimitExceeded for CreateNetworkAclEntry operations. Can we increase the number of retries please?

kyleknap:
We can look into it. Is it possible if you could give us an example (a set of steps to follow would be great) of how to hit this exception? We do exponential backoff on retries, but we could also look into increasing the sleep time between retries.

Jc2k:
I create 5 or 6 subnets each with a set of network act's. I configure the acl's to all traffic between the subnets. Obviously its stateless not stateful so each rule turns into an inbound and outbound rule in both groups. For the load balancer network acl i limit it to IP's retrieved from the json file that Amazon provides with all the CloudFront IP's in it. This adds up to quite a lot of calls to CreateNetworkAclEntry and it fails 1 in 2 times.

rayluo:
Documenting a recent conversation here. We decide that rather than increase the retry limits globally, a more flexible way is to expose an api for caller to customize the retry limit.

2rs2ts:
Is there a way to customize the retry limit yet?

PR #1260, opened 2017-08-10
Add ability to set max retry attempts

kyleknap:
This adds the ability for users to set the max retry attempts for any client call through the `botocore.config.Config` object. So for example:

```python

s = Session()
# Will retry any method call at most 10 times
c = s.create_client('s3', config=Config(retries={'max_attempts': 10}))
```
It addresses the following issues:

Fixes https://github.com/boto/botocore/issues/482
Fixes https://github.com/boto/botocore/issues/882
Fixes https://github.com/boto/boto3/issues/1104

There are a couple of related PR's for this as well:
https://github.com/boto/botocore/pull/1236
https://github.com/boto/botocore/pull/891

I think I would prefer to go with the approach proposed in this PR because it exposes a little less of the implementation details of how retries work in botocore (so we can swap the strategy out potentially in the future if needed) and allows for retry configuration to be expanded by making a general `retries` parameter in the `Config` object.


botocore/client.py
```diff
@@ -70,6 +70,7 @@ def create_client(self, service_name, region_name, is_secure=True,
             service_model, region_name, is_secure, endpoint_url,
             verify, credentials, scoped_config, client_config, endpoint_bridge)
         service_client = cls(**client_args)
+        self._register_retries(service_client)
```
kyleknap: Might be worth double checking that moving when the retries gets registered is fine. Now, it is now getting registered to the client's event emitter after the client is created instead of on any service model load and to the client creator's event emitter. I made the change because it is technically more correct (in terms of what event emitter the handler should be registered to), the code flow was better (we are using the computed client config instead of what is provided directly to the `create_client()` call), and it should not affect any of the other registered handlers.

kyleknap:
Hmm looks like there is a bad patch somewhere as an unrelated test is failing. Going to have to figure that out as I use a context manager for the patches I use.

kyleknap:
Alright tests now pass. Also, I found a bug in our retry config logic. So I fixed that while I was looking into the failures as they were related. Should be good to look at.

botocore/config.py
```diff
@@ -157,6 +173,12 @@ def _validate_s3_configuration(self, s3):
                 raise InvalidS3AddressingStyleError(
                     s3_addressing_style=addressing_style)
 
+    def _validate_retry_configuration(self, retries):
```
kyleknap: Sounds good. I can add that.

kyleknap:
I checked for that handler. I did not see it being used when I sent it up the PR. I will look to see if I can remove it. It should be legacy code as that bit of code has roots to when botocore was based on services and operations and not just clients.

botocore/exceptions.py
```diff
@@ -400,13 +400,21 @@ class InvalidS3AddressingStyleError(BotoCoreError):
...
-class InvalidRetryConfiguration(BotoCoreError):
+class InvalidRetryConfigurationError(BotoCoreError):
     """Error when invalid retry configuration is specified"""
     fmt = (
         'Cannot provide retry configuration for "{retry_config_option}". '
         'Valid retry configuration options are: \'max_attempts\''
     )
 
+
+class InvalidMaxRetryAttemptsError(BotoCoreError):
```
JordonPhillips: This should probably subclass from `InvalidRetryConfigurationError`

tests/functional/test_retry.py
```diff
@@ -0,0 +1,97 @@
...
+            'codecommit', self.region, config=Config(
+                retries={'max_attempts': 1}))
+
+        # Make another client that has no custom retry configured.
+        client = self.session.create_client('codecommit', self.region)
+        # It should use the default max retries, which should be four retries
+        # for this service.
+        self.assert_will_retry_n_times(client.list_repositories, 4)
+
+    def test_service_specific_defaults_do_not_clobber(self):
```
jamesls: Can you explain what this is actually testing?  It looks like it's just verifying we're able to load service specific config.  I don't really get the clobbering part.

botocore/translate.py
```diff
@@ -11,22 +11,46 @@
...
+def _merge_client_retry_config(retry_config, client_retry_config):
+    max_retry_attempts_override = client_retry_config.get('max_attempts')
+    if max_retry_attempts_override is not None:
+        # In the retry config, the max_attempts refers to the maximum number
+        # of requests in general will be made. However, for the client's
+        # retry config it refers to how many retry attempts will be made at
+        # most. So to translate this number from the client config, one is
+        # added to convert it to the maximum number request that will be made
+        # by including the initial request.
+        retry_config['__default__'][
```
jamesls: We don't support this now, but if we ever support `max_attempts` at the operation level, we'll need to revisit how we do this.  Might be worth adding a note about that.

botocore/translate.py
```diff
@@ -11,22 +11,46 @@
...
     # We want to merge the global defaults with the service specific
     # defaults, with the service specific defaults taking precedence.
     # So we use the global defaults as the base.
-    final_retry_config = {'__default__': retry_model.get('__default__', {})}
+    #
+    # A deepcopy is done on the retry defaults because it ensures the
+    # retry model has no chance of getting mutated when the service specific
+    # configuration or client retry config is merged in.
+    final_retry_config = {
+        '__default__': copy.deepcopy(retry_model.get('__default__', {}))
```
kyleknap: Yeah there was a bug. The only time it would be executed though is if you created a dynamodb client and then created other types of clients after as the other clients would inherit the 10 max_attempts of dynamodb. I added the `test_service_specific_defaults_do_not_clobber` to make sure that does not happen again.

botocore/translate.py
kyleknap: I can do that.

tests/functional/test_retry.py
kyleknap: This is to test the case:
```python
c = session.create_client('dynamodb')
c2 = session.create_client('ec2')
```
The second client for ec2 would actually have the max retries of dynamodb, even though it is configured for 5 max_attempts.

botocore/exceptions.py
kyleknap: Sure that's fine with me.

tests/functional/test_retry.py
kyleknap: I also see what I can do to make the name clearer.

jamesls: yeah or just a comment or something to explain it.

JadyNews:
There is an issue with sts
__init__() got an unexpected keyword argument 'aws_session_token'

`
import boto3
import botocore
from boto3.session import Session

s = Session()
# Will retry any method call at most 10 times
sts_client = s.create_client('sts', config=Config(retries={'max_attempts': 10}))

def assumerole(arn):

    try:
        assumedRoleObject = sts_client.assume_role(
            RoleArn=arn,
            RoleSessionName=socket.gethostname()
        )

        credentials = assumedRoleObject['Credentials']
        session = Session (
            aws_access_key_id = credentials['AccessKeyId'],
            aws_secret_access_key = credentials['SecretAccessKey'],
            aws_session_token = credentials['SessionToken'],
        )
        return session`

Python/2.7.10 
boto3==1.5.25
botocore==1.8.39

PR #1262, opened 2017-08-15
Add keyring support to config credential provider

mwgamble:
This adds support for the keyring module to the config credential provider. This allows users to store just the access key in a config file while keeping the secret key in a more secure storage facility. The excellent keyring module handles all of the cross-platform code required to make this work.

There have been numerous discussions about this in the past, including some currently open feature requests. Support for it was in boto2, but never made it across to boto3/botocore. Despite a lot of work being put in for #72, that PR was never merged due to the belief that it would make more sense for the
keyring to be set up as an external credential provider plugin.

I don't believe this is the optimal approach. The point of having keyring support at all is to automate the process of retrieving the credentials while still keeping the secret key in a secure storage facility. This process still requires some configuration - namely, the profile name and the access key. This
configuration still has to go somewhere, and the AWS config file is the best choice as far as I can tell. It follows that integrating keyring support into the config credential provider makes the most sense.

The other reason for having this go into the core config credential provider rather than act as a third party plugin is down to the fact that a lot of the time users are not interacting with botocore directly. Rather, they are using the AWS CLI tool, or through ansible's various modules, or some other
integration on top of botocore.

The issue of integrating keyring support so it can be used automatically by AWS CLI was raised again [last year](https://github.com/boto/boto3/issues/629#issuecomment-221467941). Unfortunately, it appears no answer was provided by any of the boto maintainers. I definitely agree with @handlerbot that having this integrated into one of the projects as core functionality would make everyones' lives easier.

Please note that at this stage I have only added support for the AWS config file, due to the documentation stating that the boto config file is mainly for backwards compatibility with boto2. However I would be happy to update the PR and extend to the boto config file too if that is desired.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -713,7 +713,8 @@ class ConfigProvider(CredentialProvider):
     def load(self):
         """
         If there is are credentials in the configuration associated with
-        the session, use those.
+        the session, use those. Can optionally load the secret key from
+        the user's keyring, if it is configured to do so.
         """
         try:
             full_config = self._config_parser(self._config_filename)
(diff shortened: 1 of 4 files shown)


mwgamble:
I've updated this PR with some additional error handling, and some unit tests. I'd really appreciate any feedback on the implementation.

mwgamble:
Is there any chance of an update on this?

jamesongithub:
looks like coverage @mwgamble

lloydhazlett:
@stealthycoin Wondering what the "large" label applied to this PR indicates, is it possible to have some feedback on anything else that could be done to increase the chances of this being looked at in the near(er) future?

joguSD:
I'm curious on your thoughts about [sourcing credentials from an external process](http://docs.aws.amazon.com/cli/latest/topic/config-vars.html#sourcing-credentials-from-external-processes) addressing some of your concerns about getting keyring support in the CLI as well.

WhyNotHugo:
Code here looks perfect. Any reason why this has had no review in all these years?
I hate the fact that AWS credentials must be stored in plain text and unprotected.

tim-finnigan:
Hi all, thanks for your patience. I brought this PR up for discussion with the team and was advised that if this feature is still needed, it should be tracked as an issue in our cross-SDK repository (https://github.com/aws/aws-sdk). Any proposed changes that involve credentials need to be considered at a cross-SDK level in order to maintain consistency.

PR #1264, opened 2017-08-16
Add functional test to valid result_key types

joguSD:
See https://github.com/boto/botocore/pull/1263 for context.

### DIFF
--- a/tests/functional/test_paginator_config.py
+++ b/tests/functional/test_paginator_config.py
@@ -16,12 +16,15 @@ import jmespath
 from jmespath.exceptions import JMESPathError
 
 import botocore.session
+from botocore.compat import six
+from botocore.utils import ArgumentGenerator
 
 
 KNOWN_PAGE_KEYS = set(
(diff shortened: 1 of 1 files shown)


jamesls:
So what's the plan with this?  Are we going to fix the paginators (is that even possible?) or are we going to whitelist these 3 failing testes?

joguSD:
@jamesls Strictly speaking the current paginators are broken and are a regression over not having them. Since we make no attempt to aggregate the results it's the same as getting the first page with no next token. 

Here are some options:
* Remove the broken paginators (backwards incompatible)
* Support aggregate maps in pagination which is hard as different services expect different aggregation behavior. It should be clear that support for these existing paginators is an exception
* White list these paginators in the test (not ideal, the paginators are still broken) 

I think the ideal behavior is: provide the correct aggregation for paginators we already 'support' and whitelist them in the tests. I think map based pagination is strange enough that we still want to be failing tests when they are introduced in the future.

PR #1267, opened 2017-08-17
Expose waiter configuration

joguSD:
Adding so we can properly address https://github.com/aws/aws-cli/pull/2761

This will allow `WaiterConfig` to be passed to `waiter.wait()` calls allowing customers to adjust waiter's `maxAttempts` and `delay` settings. This is based on how `PaginationConfig` works. The current keys are formatted to match the waiter models.

```
waiter.wait(WaiterConfig={
    'delay': 5,
    'maxAttempts': 5,
})
```

### DIFF
--- a/botocore/docs/waiter.py
+++ b/botocore/docs/waiter.py
@@ -11,6 +11,8 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 from botocore import xform_name
+from botocore.compat import OrderedDict
+from botocore.docs.utils import DocumentedShape
 from botocore.utils import get_service_module_name
 from botocore.docs.method import document_model_driven_method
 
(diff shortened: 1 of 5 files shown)


botocore/waiter.py
```diff
@@ -287,9 +287,11 @@ def __init__(self, name, config, operation_method):
     def wait(self, **kwargs):
         acceptors = list(self.config.acceptors)
         current_state = 'waiting'
-        sleep_amount = self.config.delay
+        # pop the invocation specific config
+        config = kwargs.pop('WaiterConfig', {})
+        sleep_amount = config.get('delay', self.config.delay)
+        max_attempts = config.get('maxAttempts', self.config.max_attempts)
```
JordonPhillips: s/maxAttempts/max_attempts/

joguSD: I did this to be consistent with the `PaginateConfig` formatting. The keys are capitalized words there. The waiter config uses mixed case keys. No matter what the key is called it will break consistency in some way.

kyleknap: Little picky, but the casing should be pascal casing where the first letter should be capitalized. That is how pagination does it.

kyleknap:
Approach looks fine. But I agree with Jordon. We need docs for this.

botocore/waiter.py
joguSD: I'm okay with `MaxAttempts` and `Delay`.

JordonPhillips:
It's also worth considering exposing this in the CLI

joguSD:
Updated to pascal casing and added the docs in a similar manner to pagination.
screen shot 2017-08-23 at 10 06 37 am

botocore/docs/waiter.py
```diff
@@ -78,6 +80,27 @@ def document_wait_method(section, waiter_name, event_emitter,
     :param include_signature: Whether or not to include the signature.
         It is useful for generating docstrings.
     """
+    waiter_config_members = OrderedDict()
+
+    waiter_config_members['Delay'] = DocumentedShape(
+        name='Delay', type_name='integer',
+        documentation=(
+            '<p>The amount of time in seconds to wait between attempts.</p>'))
```
kyleknap: I know this is in the top level description, but it might be good to say what the defaults are for `Delay` and `MaxAttempts` for these operations as well.

kyleknap:
Also make sure you add a changelog entity for this.

joguSD:
Added defaults inline.
screen shot 2017-08-25 at 1 57 04 pm

PR #1269, opened 2017-08-22
Make links anonymous

### LINKED ISSUE #1268: API Gateway Documentation Duplicate AWS CLI targets
This is coming from aws/aws-cli#2786. Looking over the botocore service files for API gateway, there are multiple links with the "AWS CLI" target including the `get-method` method. Needs to be something more specific to avoid sphinx target errors on the `aws-cli` side. Since I'm anticipating that these service files are auto generated I'm holding off on any sort of pull request.

JordonPhillips:
Sphinx will throw warnings if you have two uri's with the same name
but different targets because it helpfully assumes your intent was
to make an embedded uri global. This add an additional underscore to
the end of those links to make Sphinx treat them as 'anonymous'.

The links do still work properly without this change, but the CLI
will spit out warnings without it, which is not a great customer
experience.

Fixes #1268
Fixes aws/aws-cli#2786

### DIFF
--- a/botocore/docs/bcdoc/style.py
+++ b/botocore/docs/bcdoc/style.py
@@ -227,11 +227,11 @@ class ReSTStyle(BaseStyle):
                 # case then there was no link text.  We should just
                 # use an inline link.  The syntax of this is
                 # `<http://url>`_
-                self.doc.push_write('`<%s>`_' % self.a_href)
+                self.doc.push_write('`<%s>`__' % self.a_href)
             else:
                 self.doc.push_write(self.a_href)
                 self.doc.hrefs[self.a_href] = self.a_href
-                self.doc.write('`_')
+                self.doc.write('`__')
             self.a_href = None
         self.doc.write(' ')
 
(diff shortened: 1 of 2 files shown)


botocore/docs/bcdoc/style.py
```diff
@@ -220,7 +220,7 @@ def end_a(self):
                 if ':' in last_write:
                     last_write = last_write.replace(':', r'\:')
                 self.doc.push_write(last_write)
-                self.doc.push_write(' <%s>`_' % self.a_href)
+                self.doc.push_write(' <%s>`__' % self.a_href)
```
kyleknap: There is a couple more places in the method where only a single `_` is used. Do those need to be updated as well?

JordonPhillips: They can, but I'm not sure it's strictly needed. I may have found another issue. It seems rst may not be able to do hyperlinked literals like [`this`](https://example.com/).

PR #1270, opened 2017-08-24
Allow Zero for passwordUsePreventionType minimum value

Ry4an:
Zero is a valid value for PasswordReusePrevention. It disables reuse checking and is the default.  From line 6011 in the same file:
> Specifies the number of previous passwords that IAM users are prevented from reusing. The default value of 0 means IAM users are not prevented from reusing previous passwords. Default value: 0.


joguSD:
We don't accept pull requests that modify these files as we get them from upstream. I think they intended for omitting the field to mean 0 and if you include a value it should be greater than 1. Either way I'll forward your feedback to the service team.

PR #1284, opened 2017-09-28
Change result_key for kinesis to StreamDescription

tapiralec:
I'm not sure if this change is preferred behavior, but it should fix this issue: https://github.com/aws/aws-cli/issues/2001

When result_key is StreamDescription.Shards, you can't query for the ARN with "--output text" without "--no-paginate" but you can with "--output json", so I figured this was a paginating issue.


joguSD:
Unfortunately, I don't think this will actually work. Setting the `result_key` to `StreamDescription` might let you apply the query you want this will actually break the merging of the pages (the last page will win I believe) as the behavior for merging two dictionaries is currently undefined. The `result_key` must point to a list.

PR #1287, opened 2017-09-30
Add retry for Glacier RequestTimeoutException

LiptonB:
According to http://docs.aws.amazon.com/amazonglacier/latest/dev/api-error-responses.html, Glacier replies with a RequestTimeoutException when the upload takes too long. Since this could be due to network conditions, it makes sense to retry in this case.

### DIFF
--- a/botocore/data/_retry.json
+++ b/botocore/data/_retry.json
@@ -198,6 +198,20 @@
         }
       }
     },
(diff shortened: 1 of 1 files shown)


stealthycoin:
Thanks for the submission!

PR #1288, opened 2017-10-06
Add Lifecyle Waiters for Autoscaling

homelinen:
Add Waiters for `enter_standby` and `exit_standby` results.

### DIFF
--- a/botocore/data/autoscaling/2011-01-01/waiters-2.json
+++ b/botocore/data/autoscaling/2011-01-01/waiters-2.json
@@ -1,0 +1,31 @@
+{
+  "version": 2,
+  "waiters": {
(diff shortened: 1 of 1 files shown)


botocore/data/autoscaling/2011-01-01/waiters-2.json
```diff
@@ -0,0 +1,31 @@
+{
+  "version": 2,
+  "waiters": {
+    "LifecycleStateInService": {
+      "operation": "DescribeAutoScalingInstances",
+      "maxAttempts": 40,
+      "delay": 15,
```
homelinen: This seemed to be the default in EC2, can change, I guess Autoscaling rate limits might be lower?

PR #1293, opened 2017-10-13
Log a warning when calling a deprecated operation

joguSD:
Example log:
```
2017-10-13 16:54:49,583 botocore.client [DEBUG] Warning: s3.GetBucketLifecycle() is deprecated
```

### DIFF
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -565,6 +565,10 @@ class BaseClient(object):
 
     def _make_api_call(self, operation_name, api_params):
         operation_model = self._service_model.operation_model(operation_name)
+        service_name = self._service_model.service_name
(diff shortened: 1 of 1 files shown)


botocore/client.py
```diff
@@ -565,6 +565,10 @@ def _service_model(self):
 
     def _make_api_call(self, operation_name, api_params):
         operation_model = self._service_model.operation_model(operation_name)
+        service_name = self._service_model.service_name
+        if operation_model.deprecated:
+            logger.debug('Warning: %s.%s() is deprecated!',
```
jamesls: The exclamation point isn't needed, we typically don't add those to our logs.

PR #1296, opened 2017-10-16
Sanitize sigv4 canonical host header

### LINKED ISSUE #2883: Error when installing the last version of botocore 1.29.87 with poetry. Hashes not found
### Describe the bug

When installing the latest available version of botocore(1.29.87) with poetry, it throws the following error:
image

**_Note: The version 1.29.87 is installed correctly._**

### Expected Behavior

The package should install successfully.

### Current Behavior

image

### Reproduction Steps

Try to install or update botocore with poetry with the last version of this package(1.29.87)
```sh
poetry add botocore=1.29.87
```

### Possible Solution

_No response_

Poetry version: 1.4.0

### SDK version used

1.29.87

MacOs Ventura

JordonPhillips:
This strips out the port on the host header if using the default port
80. It also strips out the http username and password. That was being
included because urlsplit returns it as part of the netloc.

Fixes aws/aws-cli#2883

### DIFF
--- a/tests/unit/auth/test_signers.py
+++ b/tests/unit/auth/test_signers.py
@@ -559,6 +559,24 @@ class TestSigV4(unittest.TestCase):
         headers = auth.canonical_headers(original)
         self.assertEqual(headers, 'foo:leading and trailing')
 
+    def test_strips_default_port(self):
+        request = AWSRequest()
+        request.url = 'https://s3.us-west-2.amazonaws.com:80/'
+        request.method = 'GET'
+        auth = self.create_signer('s3', 'us-west-2')
+        actual = auth.headers_to_sign(request)['host']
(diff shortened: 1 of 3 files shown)


tests/unit/auth/test_signers.py
```diff
@@ -559,6 +559,24 @@ def test_trims_leading_trailing_spaces(self):
         headers = auth.canonical_headers(original)
         self.assertEqual(headers, 'foo:leading and trailing')
 
+    def test_strips_default_port(self):
```
stealthycoin: What about both?

joguSD:
Do we need to be worried about any other ports (HTTPS, for example)?

JordonPhillips:
@joguSD updated, and I also manually tested to make sure that it works on live

Issue #1248, opened 2017-07-20
A urllib3 connection is not put back in the pool after HEAD


timuralp:
Issuing a HEAD request results in the connection not being put back into the urllib3 pool. ~~This is tied to this issue in urllib3: https://github.com/shazow/urllib3/issues/673. The problem is resolved in urllib3 >= 1.11. However, botocore vendors requests 2.7.0, which vendors urllib3 1.10.4. Updating the vendored requests library to >= 2.8.0 would resolve this issue.~~

Relevant issue in urllib3: https://github.com/shazow/urllib3/issues/1234

timuralp:
Here's an example to reproduce the issue:
```
import boto3

s = boto3.session.Session(aws_access_key_id=<AWS_ACCESS_KEY_ID>,
                          aws_secret_access_key=<AWS_SECRET_ACCESS_KEY>)
client = s.client('s3')
bucket = <Bucket>
try:
    client.put_object(Bucket=bucket, Key='foo')
    urllib3_pool = client._endpoint.http_session.adapters['https://'].poolmanager.pools._container.values()[0].pool
    print 'Size after PUT:', urllib3_pool.qsize()
    client.head_object(Bucket=bucket, Key='foo')
    print 'Size after HEAD 200:', urllib3_pool.qsize()
    client.head_object(Bucket=bucket, Key='bar')
except:
    pass
print 'Size after HEAD 404:', urllib3_pool.qsize()
```

I can reproduce it after HEAD returns 404, but not after 200.

timuralp:
Spent more time debugging this. It's not actually fixed in urllib3 and requests as I original thought. I will open an issue with urllib3 and then with requests about this.

timuralp:
Filed bugs in urllib3 (https://github.com/shazow/urllib3/issues/1234) and requests (https://github.com/requests/requests/issues/4199), along with a PR to fix this in urllib3 (https://github.com/shazow/urllib3/pull/1235).

There is a packaging issue for botocore. New versions of requests do not vendor urllib3 anymore. Does botocore want to continue vendoring urllib3 (previously indirectly through requests)? That would require adding a hack to register it so that requests can find it. Or if boto does not vendor urllib3, the dependency can be set in requirements.

joguSD:
Upgrading `requests`/`urllib3` may fix: https://github.com/boto/boto3/issues/1053

joguSD:
[...]

PR #1300, opened 2017-10-19
Fix urllib3 connection leak after HEAD request

genben:
Fix issue #1248


timuralp:
For what it's worth, my employer had to apply the same patch to make progress. It's disappointing that the boto team does not consider this bug a priority.

Andreas5739738:
Ping to botocore maintainers - any reason why we wouldn't want to merge this?

joguSD:
We have un-vendored requests/urllib3 and are now using urllib3 directly as of `botocore` v1.11.0.

PR #1305, opened 2017-10-27
Switch from absolute imports to relative imports

levi-rs:
This PR replaces the existing absolute imports of botocore modules with relative imports. This was done to
* Use a consistent approach to importing local modules
* Increase portability and vendoring of botocore code

By switching to a relative import scheme, botocore is considerably more portable, particularly when vendoring botocore code within an application.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -19,10 +19,10 @@ considered internal, and *not* a public API.
 import copy
 import logging
 
-import botocore.serialize
-from botocore.signers import RequestSigner
-from botocore.config import Config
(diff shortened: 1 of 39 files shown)


stealthycoin:
This is a very invasive change and in general absolute imports are preferred. Why do you need to vendor botocore? It is designed to be used with a package manager and I do not think we will be making large invasive changes for a use case we do not support.

levi-rs:
Fair enough ¯\_(ツ)_/¯

PR #1306, opened 2017-11-03
Allow recent Sphinx versions

andrey-utkin:
Bump upper boundary of allowed version of Shpinx to 1.7.

requirements-docs.txt (which hasn't been updated since 2015) declares
that Sphinx of version older than 1.3 should be used. However, Sphinx
1.6.5 builds docs successfully.

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -1,2 +1,2 @@
-Sphinx>=1.1.3,<1.3
+Sphinx>=1.1.3,<1.7
 guzzle_sphinx_theme>=0.7.10,<0.8


stealthycoin:
What is the use case for this? I would be hesitant to upgrade a dependency without a compelling reason to do so. I don't see any reason in the changelog between 1.3 and 1.7 that would compel me to upgrade.

andrey-utkin:
@stealthycoin you don't have to upgrade if you don't want to.
The point is to declare that it is supposed to work with recent Sphinx versions as well as with e.g. 1.3. For example, previously botocore Gentoo Linux package would require you to install stale 1.2.x version of Sphinx if you want to have botocore docs built and installed. If a person has recent version of Sphinx installed for whatever reason, this requires the user to **downgrade** Sphinx on their system, which is pretty unpleasant and is totally not necessary from technical perspective, yet this is what upstream metadata dictates.

In Gentoo Linux botocore (and boto3) packages, I have already unbound required Sphinx version from what upstream requirements.txt dictates, but I hope you get the point why such an update in requiremnents.txt would be nice to have.

tim-finnigan:
Thanks for creating this PR. However it is now outdated due to recent changes to the documentation. Here are the current requirements: https://github.com/boto/botocore/blob/develop/requirements-docs.txt. Please create a new GitHub issue for any issues related to this going forward.

Issue #1309, opened 2017-11-08
Error when trying to read 0 byte from StreamingBody


youhyunkim:
Referring to the read method of `StreamingBody`:
https://github.com/boto/botocore/blob/c632931a6cc5eab4113c976d430bcb9c059f829f/botocore/response.py#L69-L81

If anyone asks for 0 bytes from a StreamingBody, the conditional on line 76 will pass because chunk is empty (since 0 bytes were asked for) and amount was set to 0 (not None). This leads to the content length verification, which will fail because you've read 0 bytes so far out of the entire content.

Might be an odd use case, but I feel like is a valid use case.
In fact, I ran into this issue when trying to use the `ijson` package [link](https://pypi.python.org/pypi/ijson).
That library uses `.read(0)` in order to figure out what type of encoding the stream reader should use. Whether that's the best way to do it or not, I'm not entirely sure. But I feel like `.read(0)` should still be supported.

If you guys agree that it should be supported, maybe considering a condition like this:
```
if (not chunk and amt > 0) or amt is None:
```

joguSD:
Thanks for the PR @youhyunkim. I've cleaned up the PR to pass all unit tests and added a test to validate this change in #1312.

PR #1310, opened 2017-11-08
support for 0 byte read for StreamingBody

youhyunkim:
Added support for 0 byte read for StreamingBody (issue [#1309](https://github.com/boto/botocore/issues/1309))


joguSD:
I've fixed this PR to pass the unit tests and added an additional test to verify this change.

Closing in favor of https://github.com/boto/botocore/pull/1312

PR #811, opened 2016-02-26
create a reusable AssumeRoleProvider

jstewmon:
This PR is still a WIP. I am interested in getting feedback from the botocore maintainers on the direction, and am willing to invest more time in iterating on the solution.

The existing AssumeRoleProvider was a direct port of the implementation from the awscli, and was not created with the intent to be reused in cases where another application might need to use an assumed role. Thus far, implementers have created their own versions of AssumeRoleProvider, injecting it into a Session object, which can then be used in an application.

This PR defines a new BaseAssumeRoleProvider (a better name probably exists), which can be used by the awscli (or other applications that want to reuse its config files) and by applications that use botocore directly. Also, a convenience method is added to Session to easily allow the creation of a new session, which may only use the credentials provided by an assumed role provider. (Fixes #761)

The config loading strategy I used allows recursive role assumption from the awscli, which can be convenient for users who have access to one role through another role. For example:

```
$ cat ~/.aws/credentials
[default]
aws_access_key_id = foo
aws_secret_access_key = bar

[role1]
role_arn = role1arn
source_profile = default

[role2]
role_arn = role2arn
source_profile = role1

$ aws --profile role2 ...
```

PR #1313, opened 2017-11-11
Support additional sources for assume role

JordonPhillips:
This adds support for credential sources aside from profiles with
hard-coded credentials. Now customers will be able to point to
another assume role profile with `source_profile`. Additionally,
they will be able to specify a new `credential_source` parameter
to load source credentials from other credential providers. This
currently has three valid values:

* `Environment` - This loads source credentials from environment
variables.
* `Ec2InstanceMetadata` - This loads source credentials from EC2's
instance metadata service.
* `EcsContainer` - This loads source credentials from ECS's
container credentials.

Resolves aws/aws-cli#2938
Resolves aws/aws-cli#2590
Resolves aws/aws-cli#1390
Closes #946 
Closes #851 

Addresses some, but not all, of #811


botocore/credentials.py
```diff
@@ -814,6 +991,10 @@ def __init__(self, load_config, client_creator, cache, profile_name,
         :param prompter: A callable that returns input provided
             by the user (i.e raw_input, getpass.getpass, etc.).
 
+        :type credential_sourcer: CredentialSource
```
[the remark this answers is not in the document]
JordonPhillips: Updated

JordonPhillips:
Cool. I'm gonna leave this up for a couple of days to give people a chance to try it out / leave feedback.

botocore/credentials.py
```diff
@@ -436,11 +444,169 @@ def get_frozen_credentials(self):
...
+        self._cache_key = self._create_cache_key()
+
+    def _create_cache_key(self):
+        """Create a predictable cache key for the current configuration.
+
+        The cache key is intended to be compatible with file names.
+        """
+        # On windows, ':' is not allowed in file names, so we'll
+        # replace them with '_' instead.
+        cache_key = self._role_arn.replace(':', '_').replace(os.path.sep, '_')
```
jamesls: Interesting.  The cache key will no longer have the profile name as part of its cache key which means you will share cached credentials across different profiles if they happen to assume the same role.  e.g

```
[profile one]
role_arn=MYROLE
[profile two]
role_arn=MYROLE
```

This isn't a safe assumption because different profiles could have different arguments but still assume the same role (external_id, mfa_serial, etc).

botocore/credentials.py
```diff
@@ -436,11 +444,169 @@ def get_frozen_credentials(self):
...
+    def fetch_credentials(self):
+        return self._get_cached_credentials()
+
+    def _get_cached_credentials(self):
+        """Get up-to-date credentials.
+
+        This will check the cache for up-to-date credentials, calling assume
+        role if none are available.
+        """
+        response = self._load_from_cache()
```
jamesls: The original version had debug logs about whether or not we load from cache that don't appear to be ported over with this change (line 856 in the original file).

It would be helpful to log when we load from the cache and when we load from cache but they're expired.

botocore/credentials.py
```diff
@@ -835,161 +1016,295 @@ def __init__(self, load_config, client_creator, cache, profile_name,
...
+
+    def _validate_credential_source(self, parent_profile, credential_source):
+        if self._credential_sourcer is None:
+            raise InvalidConfigError(error_msg=(
+                'The credential_source "%s" is specified in profile "%s", '
+                'but no source provider was configured.' % (
+                    credential_source, parent_profile)
+            ))
+        if not self._credential_sourcer.is_supported(credential_source):
+            raise InvalidConfigError(error_msg=(
```
jamesls: It would be helpful if we can also surface the list of canonical names with the `_credential_sourcer` knows what they are.  I ran into this my mistakenly using `EC2InstanceMetadata` instead of `Ec2InstanceMetadata`

botocore/credentials.py
```diff
@@ -835,161 +1016,295 @@ def __init__(self, load_config, client_creator, cache, profile_name,
...
-            return None
-
-    def _is_expired(self, credentials):
-        end_time = _parse_if_needed(credentials['Credentials']['Expiration'])
-        now = datetime.datetime.now(tzlocal())
-        seconds = total_seconds(end_time - now)
-        return seconds < self.EXPIRY_WINDOW_SECONDS
+    def _load_creds_via_assume_role(self, profile_name):
+        role_config = self._get_role_config(profile_name)
+        source_credentials = self._resolve_source_credentials(
```
jamesls: This resolution is now happening before the cache checks.  This results in an unnecessary requirement that the entire chain of assumed profiles is not expired.

However, these intermediate results aren't actually used if a subsequent role *is* cached.  For example, consider the role chain `A->B->C`.  Let's say C  and B are expired in the cache, but A's entry in `~/.aws/cli/cache` is up to date.  This means the sequence of calls you'll make is:

* `AssumeRole(C)`
* `AssumeRole(B)`
* Used cache credentials from A

It seems like the ideal behavior would be if we're trying to assume the role in profile `A`, as long as it's cache entry is up to date we should just use it.

You can test this behavior by just deleting the cache file for `B` and `C` in `~/.aws/cli/cache`

botocore/credentials.py
JordonPhillips: That's a good point. How about doing a hash of all of `self._assume_role_kwargs`? It would also serve to make the generation much simpler.

botocore/credentials.py
JordonPhillips: 👍

botocore/credentials.py
JordonPhillips: Looks like this crept in when moving from assume role credentials to the fetcher since I didn't have a test for it. The problem is that `RefreshableCredentials` needs to have initial credentials.

botocore/credentials.py
```diff
@@ -1064,19 +1076,18 @@ def _load_creds_via_assume_role(self, profile_name):
...
-        expiration = _parse_if_needed(base_credentials['expiry_time'])
+        # The initial credentials are empty and the expiration time is set
+        # to now so that we can delay the call to assume role until it is
+        # strictly needed.
         return RefreshableCredentials(
-            access_key=base_credentials['access_key'],
-            secret_key=base_credentials['secret_key'],
-            token=base_credentials['token'],
-            expiry_time=expiration,
+            access_key='',
```
kyleknap: Taked to you offline about this as well. But I would prefer we have a explicit value as the access and secret key (i.e. `NOT_SET`, `NEEDS_REFRESH`) to indicate on the first load of these we need to do the refresh logic.

botocore/credentials.py
```diff
@@ -477,6 +454,30 @@ def get_frozen_credentials(self):
...
+    def __init__(self, refresh_using, method, time_fetcher=_local_now):
+        self._refresh_using = refresh_using
+        self._access_key = None
+        self._secret_key = None
+        self._token = None
+        self._expiry_time = None
+        self._time_fetcher = time_fetcher
+        self._refresh_lock = threading.Lock()
+        self.method = method
+        self._frozen_credentials = None
```
kyleknap: Just thinking out loud here. I understand why you didn't just use super here because you do not want `_normalize()` to be called. I'm a little concerned that these internal variables may grow out of sync with the `RefreshableCredentials`, but given that we hardly touch this code and we now have a lot of test, I'm not too concerned about it.

surkls:
Hi guys, I'm using this new feature so far with great success, thanks!  This made it so much easier than doing a cli assume-role call.  I do have a question though, I'm trying to add the assumed role as a principle in a bucket policy.  Using this new feature, how do I know what the role-session-name is?

[https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html](url)

sarathkumar21:
so now i can try without using access and secret key

PR #1314, opened 2017-11-14
Finer grained stub error handling

claudenm:
This PR introduces an `UnStubbedResponseError` class which is raised when an instance of the Stubber class's  stub queue is exhausted.  This came about as a result of writing tests, and receiving an unhelpful message informing me that there was an unexpected API call.  

I found this unhelpful because it masked the fact that the `stubber` instance had no additional requests it was expecting.  This change gives a more specific error message as to why the request failed.

### DIFF
--- a/botocore/stub.py
+++ b/botocore/stub.py
@@ -16,7 +16,7 @@ from pprint import pformat
 
 from botocore.validate import validate_parameters
 from botocore.exceptions import ParamValidationError, \
-    StubResponseError, StubAssertionError
+    StubResponseError, StubAssertionError, UnStubbedResponseError
 from botocore.vendored.requests.models import Response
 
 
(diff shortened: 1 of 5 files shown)


botocore/stub.py
```diff
@@ -317,10 +317,10 @@ def assert_no_pending_responses(self):
 
     def _assert_expected_call_order(self, model, params):
         if not self._queue:
-            raise StubResponseError(
+            raise UnStubbedResponseError(
                 operation_name=model.name,
-                reason=('Unexpected API Call: called with parameters:\n%s' %
-                        pformat(params)))
+                reason=('Unexpected API Call: A call was made but no additional calls expected. \
+                Either the API Call was not stubbed or it was called multiple times. '))
```
JordonPhillips: I would prefer to omit escaping the newline and instead use a multiline string:

```python
reason = (
    'foo'
    'bar'
)
```

claudenm:
Thanks @JordonPhillips.  Addressed your comments in 9983470 and afc9ee7.

JordonPhillips:
Thanks again @claudenm !

PR #1316, opened 2017-11-21
Add support for process-based credential providers

JordonPhillips:
This allows users to configure an external process to provide
credentials using the `credential_process` config variable.

Note that this is built on top of the assume role provider since
there is rebasing necessary.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -17,6 +17,8 @@ import logging
 import os
 import getpass
 import threading
+import subprocess
+import shlex
 from collections import namedtuple
 from copy import deepcopy
 from hashlib import sha256
@@ -82,6 +84,8 @@ def create_credential_resolver(session):
             creds_filename=credential_file,
             profile_name=profile_name
         ),
+        ProcessProvider(profile_name=profile_name,
+                        load_config=lambda: session.full_config),
         # The new config file has precedence over the legacy
         # config file.
         ConfigProvider(config_filename=config_file, profile_name=profile_name),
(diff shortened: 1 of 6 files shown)


botocore/credentials.py
```diff
@@ -474,8 +705,79 @@ def _extract_creds_from_mapping(self, mapping, *key_names):
         return found
 
 
+class ProcessProvider(CredentialProvider):
```
jamesls: We should probably work to get a canonical name added for this provider (though I don't think it should block the merge).

botocore/credentials.py
```diff
@@ -82,6 +84,8 @@ def create_credential_resolver(session):
             creds_filename=credential_file,
             profile_name=profile_name
         ),
+        ProcessProvider(profile_name=profile_name,
```
stell-aelsabbahy: Can the `ProcessProvider` take precedence over the `SharedCredentialProvider` or would this cause issues?

I can submit a PR with the change if there's no obvious reasons not to do it. It'll allow me to re-use a saml provider CLI tool more efficiently.

JordonPhillips: Existing providers always take precedence over new providers (except for EC2 metadata, which is always last). This is to make sure that people's config files continue to work the same as they had before if they don't change them.

Regardless, all you will need to do to make sure you're using this provider is to make a new profile where it's the only one set and use that profile explicitly.

stell-aelsabbahy: Guess where it feels a bit inconsistent to me is `source_profile`/`role_arn` win over hardcoded credentials:
```
[profile]
source_profile = rd
role_arn       = arn:blah:blah
# These will be ignored and it will use assumeRole provider
aws_access_key_id = ...
aws_secret_access_key = ...
aws_session_token = ...
```

But when using the `credential_process` it's the other way around, the hardcoded credentials win over the config option.

JordonPhillips: The ability to have hard-coded credentials in a profile along with static credentials was never intended behavior. We won't break that (because people rely on it now), but we also don't want to continue doing that for other providers.

PR #1318, opened 2017-11-22
Always strip / from cache file names

JordonPhillips:
Previously we had just been stripping colon and path sep, but forward
slash will also break on Windows.

Fixes aws/aws-cli#2978

cc @kyleknap


JordonPhillips:
codecov is complaining due to the windows-only code that isn't run on travis

PR #1322, opened 2017-11-27
Reduce cache key size

JordonPhillips:
This reduces the cache key size by switching the hashing algorithm
to sha1 and removing the arn.

Closes aws/aws-cli#2983


JordonPhillips:
codecov is just complaining because the number of lines went down

JordonPhillips:
It would be nice to have the arn in the filename, but it's not worth exacerbating the issue over. The arn can be something like 2000 characters long, which is far beyond what many filesystems can handle. Plus, the arn is inside the file anyway so you can still look at the contents to see what role it's for.

PR #1328, opened 2017-11-29
Do not generate 100-continue expectation with no body.

harshavardhana:
HTTP RFC explicitly states that 100-continue should
not be set when there is no message body.

https://tools.ietf.org/html/rfc7231#section-5.1.1

> A client MUST NOT generate a 100-continue expectation in
> a request that does not include a message body.

This is a mandatory requirement, this PR Fixes the current
implementation behavior by not sending `100-continue`
when content-length is '0' which means there is no message
body

Fixes boto/boto3#1341

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -115,6 +115,12 @@ class AWSConnection(object):
                 break
 
     def _send_request(self, method, url, body, headers, *args, **kwargs):
+        if headers.get('Content-Length') == '0':
+            # From RFC: https://tools.ietf.org/html/rfc7231#section-5.1.1
+            # Requirement for clients:
+            # - A client MUST NOT generate a 100-continue expectation
+            #   in a request that does not include a message body.
+            headers.pop('Expect', None)
         self._response_received = False
         if headers.get('Expect', b'') == b'100-continue':
             self._expect_header_set = True
@@ -230,7 +236,7 @@ class AWSConnection(object):
 
     def send(self, str):
         if self._response_received:
-            logger.debug("send() called, but reseponse already received. "
+            logger.debug("send() called, but response already received. "
                          "Not sending data.")
             return
         return super(AWSConnection, self).send(str)
--- a/tests/unit/test_awsrequest.py
+++ b/tests/unit/test_awsrequest.py
@@ -436,6 +436,14 @@ class TestAWSHTTPConnection(unittest.TestCase):
         response = conn.getresponse()
         self.assertEqual(response.status, 200)
 
+    def test_no_expect_header_set_no_body(self):
+        s = FakeSocket(b'HTTP/1.1 200 OK\r\n')
+        conn = AWSHTTPConnection('s3.amazonaws.com', 443)
+        conn.sock = s
+        conn.request('PUT', '/bucket/foo', b'')
+        response = conn.getresponse()
+        self.assertEqual(response.status, 200)
+
     def test_tunnel_readline_none_bugfix(self):
         # Tests whether ``_tunnel`` function is able to work around the
         # py26 bug of avoiding infinite while loop if nothing is returned.


joguSD:
Can we got some tests to verify this behavior?
@jamesls any thoughts on this?

harshavardhana:
> Can we got some tests to verify this behavior?
> @jamesls any thoughts on this?

Sure can, but can you point me to correct location @joguSD to add this test? functional, unit?

joguSD:
I think a unit test in `tests/unit/test_awsrequest.py` showing that in the case of an empty body we don't send the 100-continue expect is a good start.

harshavardhana:
> I think a unit test in tests/unit/test_awsrequest.py showing that in the case of an empty body we don't send the 100-continue expect is a good start.

Thanks let me do that.

harshavardhana:
Added a test @joguSD PTAL

JordonPhillips:
Looks like your test is failing on python 2.6

harshavardhana:
> Looks like your test is failing on python 2.6

Sure let me fix @JordonPhillips

harshavardhana:
Build issues are fixed @joguSD @JordonPhillips PTAL

harshavardhana:
> Build issues are fixed @joguSD @JordonPhillips PTAL

Can @joguSD @JordonPhillips PTAL ?

bancek:
Any progress on this?

harshavardhana:
> Any progress on this?

Maybe not :1st_place_medal:

Hernrup:
This will probably be solved by #1466 if they ever get around to merging it as it is fixed natively in a newer version of requests.

joguSD:
@harshavardhana I don't think this is the right fix. This still sends the `Expects: 100-continue` header, it just modifies the client behavior to not expect a response to the header. Your test also passes with or without this change. Ideally, this header wouldn't be added at all in the case of an empty body but this can be a little tricky to determine as the header is only added to requests when we receive a file-like object to upload to s3. The point at which we inject the expects header there may or may not already be a content-length present on the request. 

@Hernrup As far as I can tell `requests` and `urllib3` still don't support the expects 100-continue behavior so it's unlikely that #1466 would fix this.

harshavardhana:
The point is that to not set expect continue when content-length is not present to be RFC compliant here. Otherwise some other part of code will hang expecting a 100-continue from server. This is the issue this PR fixes. 

I don't follow why it is not a right fix.

joguSD:
@harshavardhana This will still send the `Expect: 100-continue` header in the request when there's an empty body. This is what the RFC is saying must not be done.

harshavardhana:
Yes this does on purpose because the code which sets this value, I had a hard time refactoring. The only simpler solution I found was don't wait for expect on the socket when there is no content length.

harshavardhana:
> @harshavardhana This will still send the Expect: 100-continue header in the request when there's an empty body. This is what the RFC is saying must not be done.

@joguSD I made a slight change, can you check now?

harshavardhana:
> @joguSD I made a slight change, can you check now?

@joguSD can you take a look ?

harshavardhana:
> @joguSD can you take a look ?

Can this PR be looked at?

joguSD:
I think this would fail the case where the content length is 0 but the expect header is not set.

harshavardhana:
> I think this would fail the case where the content length is 0 but the expect header is not set.

What case that would be? I tested this with AWS S3 and it works perfectly fine.

joguSD:
@harshavardhana 
If content length is set to 0 and there is no expect header the `pop` will fail.
S3 is the *only* service that the expect header is set for.
It's possible that a request for another service ends up having a content length of 0 and no expect header.

harshavardhana:
> @harshavardhana
> If content length is set to 0 and there is no expect header the `pop` will fail.
> S3 is the _only_ service that the expect header is set for.
> It's possible that a request for another service ends up having a content length of 0 and no expect header.

@joguSD so server implementations should honor this anyways, are you saying that services don't honor this? 

`Expect` header is not a requirement for any API operation whatsoever even with `Expect` headers I see that all services run fine. The problem is the HTTP client in botocore is violating the RFC by unnecessarily doing things which are not expected for an HTTP client to perform. 

```
https://tools.ietf.org/html/rfc7231#section-5.1.1

A client MUST NOT generate a 100-continue expectation in
a request that does not include a message body.
```

So I don't see why this would fail, can you provide me concrete examples on where this would fail? @joguSD

tyrdavis:
Unsure if this is the right place to make this comment so apologies if not.

My app was running up against this problem, so I monkey patched this change set in to test if it resolves the issue. I saw no behavior change, however. In order to get the expected result, I needed to change the test from `is '0'` to `== '0'`, and also needed to mitigate the issue @joguSD mentioned by allowing the `pop` to pass silently when the header isn't present: `pop('Expect', None)`

harshavardhana:
@joguSD addressed your concerns PTAL - thanks @tyrdavis incorporated your suggestions as well.

harshavardhana:
> @joguSD addressed your concerns PTAL - thanks @tyrdavis incorporated your suggestions as well.

ping @joguSD can you PTAL ?

harshavardhana:
Ping @joguSD can you PTAL ?

harshavardhana:
> Ping @joguSD can you PTAL ?

ping @joguSD can you PTAL?

harshavardhana:
ping @joguSD can you PTAL

harshavardhana:
ping @joguSD can you PTAL

daviddavis:
This fixed a MissingHeaderBodySeparatorDefect warning I get when I run my app against minio using boto3. Any chance we can get this merged?

PR #1334, opened 2017-12-06
Add support for service name aliases and alias sagemaker runtime

joguSD:
Must be merged in conjunction with: https://github.com/aws/aws-cli/pull/3019

This renames `runtime.sagemaker` to `sagemaker-runtime` and adds the ability to alias service names for backwards compatibility reasons.


tests/functional/test_service_alias.py
```diff
@@ -0,0 +1,37 @@
...
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import botocore.session
+
+# service_name => alias
+ALIAS_CASES = {
+    'sagemaker-runtime': 'runtime.sagemaker'
```
jamesls: What about making `_SERVICE_NAME_ALIASES` public and dynamically creating the reverse map here?

Otherwise we'll have to remember to update this test if we add a new alias.

tests/unit/test_client.py
```diff
@@ -119,6 +119,14 @@ def setUp(self):
     def tearDown(self):
         self.endpoint_creator_patch.stop()
 
+    def create_mock_emitter(self, responses=None):
+        if responses is None:
+            responses = []
+
+        emitter = mock.Mock()
```
jamesls: Can you add a `spec=` here?

joguSD: Yep.

tests/functional/test_service_alias.py
joguSD: My initial thought was that having at least a small change to the tests would be good, but I can also see the merit in getting the tests automatically too. And rather than reversion the map I think I'll just make that one public and import it. While it should never happen a service name could have multiple aliases so `alias => service_name` makes more sense.

tests/unit/test_client.py
joguSD: Looks like this causes issues on Python 2, gonna leave the mock without the spec as that's how it was before.

joguSD: Did some more digging on why this was happening. We apply a `copy.copy` on the event_emitter [here](https://github.com/joguSD/botocore/blob/3774b8d4488fa5567cc589cb33df6dd9ad122866/botocore/args.py#L62) which seems to cause issues on Python 2. What appears to happen is the copy becomes an actual instance of the specified class that never had its constructor called (and thus can be missing attributes).

PR #1336, opened 2017-12-07
Remove content-type from presigned query requests

JordonPhillips:
This deletes the content-type header from query requests which are
being presigned. Specifically it removes the auto-set value
"application/x-www-form-urlencoded; charset=utf-8" because that
value is not valid for presigned urls.

### DIFF
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -575,7 +575,8 @@ def generate_presigned_url(self, ClientMethod, Params=None, ExpiresIn=3600,
 
     # Prepare the request dict by including the client's endpoint url.
     prepare_request_dict(
-        request_dict, endpoint_url=self.meta.endpoint_url)
+        request_dict, endpoint_url=self.meta.endpoint_url
+    )
 
     # Generate the presigned url.
     return request_signer.generate_presigned_url(
--- a/tests/functional/test_sts.py
+++ b/tests/functional/test_sts.py
@@ -1,0 +1,45 @@
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from datetime import datetime
+
+import mock
+
+from tests import BaseSessionTest
+from tests import assert_url_equal
+from botocore.stub import Stubber
+
+
+class TestSTSPresignedUrl(BaseSessionTest):
+    def setUp(self):
+        super(TestSTSPresignedUrl, self).setUp()
+        self.client = self.session.create_client('sts', 'us-west-2')
+        # Makes sure that no requests will go through
+        self.stubber = Stubber(self.client)
+        self.stubber.activate()
+
+    def test_presigned_url_contains_no_content_type(self):
+        timestamp = datetime(2017, 3, 22, 0, 0)
+        with mock.patch('botocore.auth.datetime') as _datetime:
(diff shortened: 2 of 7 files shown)


JordonPhillips:
rebased

botocore/signers.py
```diff
@@ -575,7 +575,9 @@ def generate_presigned_url(self, ClientMethod, Params=None, ExpiresIn=3600,
 
     # Prepare the request dict by including the client's endpoint url.
     prepare_request_dict(
-        request_dict, endpoint_url=self.meta.endpoint_url)
+        request_dict, endpoint_url=self.meta.endpoint_url,
+        context={'protocol': self.meta.service_model.protocol}
```
kyleknap: Yeah it is unfortunate that this appears to be really the only option we have given the signers do not receive any models in order to do the signing logic. One thing I do wonder is if we should put the `service_model` instead into the context? I only suggest this because the context could get a little bloated if we ever decide to add another handler that needs something other than the protocol from the model

joguSD:
Looks good to me. Though it does still bother me a little that we're making a customization to remove a header that we manually add to the request. Have we thought about how viable the opposite approach is? (Adding the header in a customization instead of unconditionally).

JordonPhillips:
@joguSD we could, though it would be a bit weird since we would be moving something we always need for normal requests out of the serializer in favor of the special case of pre-signing.

JordonPhillips:
@jamesls The signer was rejecting the request because the header wasn't included in the request. When you actually do send the header, you get a different error:

```python
import botocore

sess = botocore.session.get_session()
sns = sess.create_client('sts')
url = sns.generate_presigned_url('get_caller_identity', {})
response = requests.post(
    url, headers={
        'content-type': 'application/x-www-form-urlencoded; charset=utf-8'
    }
)
print(response.text)
```

```xml
<ErrorResponse xmlns="https://sts.amazonaws.com/doc/2011-06-15/">
  <Error>
    <Type>Sender</Type>
    <Code>IncompleteSignature</Code>
    <Message>When Content-Type:application/x-www-form-urlencoded, URL cannot include query-string parameters (after '?'): '/?Action=GetCallerIdentity&amp;Version=2011-06-15&amp;X-Amz-Algorithm=AWS4-HMAC-SHA256&amp;X-Amz-Credential=redacted%2F20171212%2Fus-east-1%2Fsts%2Faws4_request&amp;X-Amz-Date=20171212T220209Z&amp;X-Amz-Expires=3600&amp;X-Amz-SignedHeaders=content-type%3Bhost&amp;X-Amz-Signature=signature'</Message>
  </Error>
  <RequestId>1a16614a-df88-11e7-af03-a144f0113aaf</RequestId>
</ErrorResponse>
```

While that error does come from the remote signer, it is fundamentally a serialization error.

JordonPhillips:
One alternative is to move it up into `generate_presigned_url`. That would eliminate the need for a handler as well as the need for adding to the context.

kyleknap:
The above idea is growing on me. I think this whole remove content type header should be the concern of the client layer and not necessarily the general signer. Not sure what everyone else thinks?

tests/functional/test_sts.py
```diff
@@ -0,0 +1,44 @@
...
+        with mock.patch('botocore.auth.datetime') as _datetime:
+            _datetime.datetime.utcnow.return_value = timestamp
+            url = self.client.generate_presigned_url('get_caller_identity', {})
+
+        expected_url = (
+            'https://sts.amazonaws.com/?Action=GetCallerIdentity&'
+            'Version=2011-06-15&X-Amz-Algorithm=AWS4-HMAC-SHA256&'
+            'X-Amz-Credential=access_key%2F20170322%2Fus-east-1%2Fsts%2F'
+            'aws4_request&X-Amz-Date=20170322T000000Z&X-Amz-Expires=3600&'
+            'X-Amz-SignedHeaders=host&X-Amz-Signature=767845d2ee858069a598d5f'
```
jamesls: How are we verifying there's no content type?  Is it because we'd get a different `X-Amz-Signature`?  It's probably worth clarifying.

JordonPhillips: Yes. We would also see it in `X-Amz-SignedHeaders`

JordonPhillips:
test added

PR #1337, opened 2017-12-12
Add iterator support to streaming body

mikepk:
This will allow the common pattern of iterating over lines
like in file like objects. 

For example:
```
for line in streaming_body:
    # do something with line
```

Implemented using the underlying raw stream
readline. This can clean up code looking to read newline delimited
data while keeping the streaming behavior (rather than reading bytes
and implementing newline detection).

This change also implements the read length verification from the read()
method as well to mirror the behavior.

### DIFF
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -15,6 +15,7 @@
 import sys
 import xml.etree.cElementTree
 import logging
+from botocore.compat import six
 
 from botocore import ScalarTypes
 from botocore.hooks import first_non_none_response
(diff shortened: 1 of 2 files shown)


mikepk:
This seems like a pretty small / easy change and adds a lot of convenience for reading line delimited records (a pretty common case). Would love feedback on this PR.

mikepk:
The total lack of interest and engagement has made me lose interest in contributing to this project.

PR #1338, opened 2017-12-13
Move JSONFileCache to botocore

JordonPhillips:
This moves the file cache to botocore and allows users to set their
own cache via `create_credential_resolver`.

Future work: allow setting this via the config file and/or some other
easy method of configuration.

Closes #1157

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -48,7 +48,7 @@ ReadOnlyCredentials = namedtuple('ReadOnlyCredentials',
                                  ['access_key', 'secret_key', 'token'])
 
 
-def create_credential_resolver(session):
+def create_credential_resolver(session, cache=None):
     """Create a default credential resolver.
 
     This creates a pre-configured credential resolver
@@ -61,6 +61,8 @@ def create_credential_resolver(session):
     config_file = session.get_config_variable('config_file')
     metadata_timeout = session.get_config_variable('metadata_service_timeout')
     num_attempts = session.get_config_variable('metadata_service_num_attempts')
+    if cache is None:
+        cache = {}
 
     env_provider = EnvProvider()
     container_provider = ContainerProvider()
@@ -72,7 +74,7 @@ def create_credential_resolver(session):
     assume_role_provider = AssumeRoleProvider(
         load_config=lambda: session.full_config,
         client_creator=session.create_client,
-        cache=JSONFileCache(),
+        cache=cache,
         profile_name=profile_name,
         credential_sourcer=CanonicalNameCredentialSourcer([
             env_provider, container_provider, instance_metadata_provider
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -17,6 +17,7 @@ import logging
 import os
 import getpass
 import threading
+import json
 import subprocess
 from collections import namedtuple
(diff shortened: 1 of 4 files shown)


botocore/credentials.py
```diff
@@ -178,6 +181,51 @@ def __call__(self):
...
 
+class JSONFileCache(object):
+    """JSON file cache.
+    This provides a dict like interface that stores JSON serializable
+    objects.
+    The objects are serialized to JSON and stored in a file.  These
+    values can be retrieved at a later time.
+    """
+
+    CACHE_DIR = os.path.expanduser(os.path.join('~', '.aws', 'cli', 'cache'))
```
kyleknap: I think we should consider using a different directory than what the CLI uses. I have not thought through whether there will be edge cases to this, but it may be surprising if you are running a script to learn that your temp credentials were being cached under a `cli` directory. I wonder if it makes sense to call it `boto` or `python` instead?

JordonPhillips: sounds reasonable

JordonPhillips: done

JordonPhillips:
Will do. Also gonna put up a PR for the cli to switch to using this.

et304383:
Can we get a quick explanation of how to use this?  Is it possible to cache the MFA globally yet?

mixja:
@et304383 - here is an example of leveraging the AWS CLI cache.  Note you could create your own custom cache location if you wanted.

```
from botocore import credentials
import botocore.session
import boto3
import os

# By default the cache path is ~/.aws/boto/cache
cli_cache = os.path.join(os.path.expanduser('~'),'.aws/cli/cache')

# Construct botocore session with cache
session.get_component('credential_provider').get_provider('assume-role').cache = credentials.JSONFileCache(cli_cache)

# Create boto3 client from session
client = boto3.Session(botocore_session=session).client('ecs')
```

bgabrhelik:
If you need to specify a profile you have to specify it in botocore.Session.  It's too late to specify it in boto3.Session as AssumeRoleProvider instantiates with wrong 'default' profile.  
So the code above will be
```
from botocore import credentials
import botocore.session
import boto3
import os

# By default the cache path is ~/.aws/boto/cache
cli_cache = os.path.join(os.path.expanduser('~'),'.aws/cli/cache')

# Construct botocore session with cache
session = botocore.session.Session(profile='prod')
session.get_component('credential_provider').get_provider('assume-role').cache = credentials.JSONFileCache(cli_cache)

# Create boto3 client from session
client = boto3.Session(botocore_session=session, profile_name = 'prod').client('ecs')
```

mixja:
@bgabrhelik - it's generally better practice to rely on the external environment to configure your credentials, rather than specify an AWS profile name in code.  This approach has benefit of allowing the code to work in EC2/ECS environments where you may be using EC2 instance profile or ECS task role.

My code snippet works in a local desktop environment assuming you have set profile name as environment variable - e.g. `export AWS_PROFILE=prod`

mikehas:
Does anyone know if there's a feature request to implement this as a configuration option, like @JordonPhillips mentioned in initial description?

> Future work: allow setting this via the config file and/or some other
> easy method of configuration.

PR #1339, opened 2017-12-14
tests: pass all env vars to cmd-runner

andrey-utkin:
cmd-runner was started with no environment variables inherited.
This breaks tests when run with custom PYTHONPATH, which is useful for
testing botocore while not being installed in standard locations.

One case when this is important is performing tests before installing
the package in Gentoo Linux.

Link: https://bugs.gentoo.org/640726


andrey-utkin:
> Because it has been longer than two years since the last update on this,

This is the case because the repo maints never came in to handle this.
That's OK actually.

> and in the absence of more information

There was no request for more information.

> please feel free to reopen it

Github doesn't let me do that.

Issue #978, opened 2016-07-11
s3 config with use_accelerate_endpoint and generate_presigned_post does not work


jnoring:
This code:

```
    s3 = boto3.client('s3', config=Config(s3={'use_accelerate_endpoint': True}))
    s3.generate_presigned_post(
        Bucket = 'some bucket,
        Key = 'some key'
    )
```

...will not result in an accelerated URL being generated for a presigned post operation.  The bug is in this code, in `client.py`:

```
        # Enable accelerate if the configuration is set to to true or the
        # endpoint being used matches one of the Accelerate endpoints.
        if s3_accelerate or S3_ACCELERATE_ENDPOINT in self._endpoint.host:
            # Amazon S3 accelerate is being used then always use the virtual
            # style of addressing because it is required.
            self._force_virtual_style_s3_addressing()
            # Also make sure that the hostname gets switched to
            # s3-accelerate.amazonaws.com
            self.meta.events.register_first(
                'request-created.s3', switch_host_s3_accelerate)
```

...`request-created.s3` is never called for a presigned post because no request to s3 is made, and thus `switch_host_s3_accelerate` is not called.

kyleknap:
I can confirm. That looks to be the issue as well.

kevinbui001:
Has this been fixed?

johnd08:
Issue is still there in version 1.4.4 - are there any plans to fix this?

joshma:
It's been 16 months, has anyone found a fix or workaround? Looks like `url.replace(S3_HOST, 's3-accelerate')` works, but that's pretty hacky.

PR #1340, opened 2017-12-18
Hook more events into presign functionality

kyleknap:
Specifically ensure the following:

* The emission of `provide-client-params` event to hook in injection of parameters
* The emission of `before-parameter-build` event event to hook in internal customizations related to parameters such as SSE-C MD5 injection.
* Registering s3 accelerate handler to `before-sign` so s3 accelerate presigned url's can be created. It is important to note that I decided not to move the `request-created` event to the presigner because the signers are already hooked up to that event so re-emitting that event will cause the signers to get triggered again. Furthermore, putting it at `before-sign` is consistent with the handlers we use to change addressing style.

Fixes https://github.com/boto/botocore/issues/978

### DIFF
--- a/tests/functional/test_sts.py
+++ b/tests/functional/test_sts.py
@@ -23,6 +23,9 @@ class TestSTSPresignedUrl(BaseSessionTest):
     def setUp(self):
         super(TestSTSPresignedUrl, self).setUp()
         self.client = self.session.create_client('sts', 'us-west-2')
+        # Makes sure that no requests will go through
+        self.stubber = Stubber(self.client)
+        self.stubber.activate()
 
     def test_presigned_url_contains_no_content_type(self):
         timestamp = datetime(2017, 3, 22, 0, 0)
--- a/tests/functional/test_sts.py
+++ b/tests/functional/test_sts.py
@@ -23,9 +23,6 @@ class TestSTSPresignedUrl(BaseSessionTest):
     def setUp(self):
         super(TestSTSPresignedUrl, self).setUp()
         self.client = self.session.create_client('sts', 'us-west-2')
-        # Makes sure that no requests will go through
-        self.stubber = Stubber(self.client)
-        self.stubber.activate()
 
     def test_presigned_url_contains_no_content_type(self):
         timestamp = datetime(2017, 3, 22, 0, 0)
--- a/tests/unit/test_signers.py
+++ b/tests/unit/test_signers.py
@@ -815,6 +815,22 @@ class TestGenerateUrl(unittest.TestCase):
             ]
         )
 
+    def test_generate_presign_url_emits_is_presign_in_context(self):
+        emitter = mock.Mock(HierarchicalEmitter)
+        emitter.emit.return_value = []
+        self.client.meta.events = emitter
+        self.client.generate_presigned_url(
+            'get_object', Params={'Bucket': self.bucket, 'Key': self.key})
+        kwargs_emitted = [
+            emit_call[1] for emit_call in emitter.emit.call_args_list
+        ]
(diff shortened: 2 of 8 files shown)


tests/functional/test_sts.py
```diff
@@ -23,9 +23,6 @@ class TestSTSPresignedUrl(BaseSessionTest):
     def setUp(self):
         super(TestSTSPresignedUrl, self).setUp()
         self.client = self.session.create_client('sts', 'us-west-2')
-        # Makes sure that no requests will go through
```
kyleknap: This was removed because the stubber hooks into the parameter building related event for validating API/parameter calls, but since we did not previously add any expected API calls in test, the stubber will now complain about an unexpected API call.

It seems strange that a stubber would be used for presigned URL's in the first place, but I could see it potentially affect downstream tests though if someone used a stubber to stub out an API call and used the same stubbed client to generate a presigned url in their tests. What people's thoughts are on this?

tests/unit/test_signers.py
```diff
@@ -727,7 +728,7 @@ def test_generate_presigned_url(self):
             'query_string': {},
             'url_path': u'/mybucket/mykey',
             'method': u'GET',
-            'context': {}}
+            'context': mock.ANY}
```
kyleknap: This was switched to `mock.ANY` because we are now emitting the client parameter related events that inject values into the context. Thus using the context's exact value for these tests will be a maintenance burden if anymore customizations are added that inject into the context.

tests/functional/test_sts.py
joguSD: Is there another even the stubber could rely on?

joguSD:
We should also add functional tests to verify that we correctly generate presigned URLs for the two cases this fixes.

tests/functional/test_sts.py
jamesls: Yeah I'd prefer if we could preserve this behavior somehow.  The fact that we're changing a functional test suggests it's possible others will be impacted, especially given the stubber is a public API.

What are the other options here?

tests/unit/test_signers.py
jamesls: I don't think review comments on your own PR are the best place for these type of contextual comments.  Either a) A comment in the code or b) in the comment message.

tests/functional/test_sts.py
kyleknap: @joguSD don't think there is another event the stubber could rely on. This PR is having the presigners emit`provide-client-params` and `before-parameter-build` which are the events that would contain the unserialized parameters. Any of the other events down the stack use the serialized version of the parameters. So if we were to hook into another event, we would need to create a new one which we would not want to do.

@jamesls One idea initially that I have is to inject a value into the `context` object a flag like `is_presign_request` in the presigner logic and then in the Stubber class we can check for this flag and pass over asserting the request if the presign flag is present in the context. What's your thoughts on this? Not sure if there really any better options given the stubber currently does not have a way to disambiguate between a presign request and a normal client request.

kyleknap:
@jamesls @joguSD I updated the PR to add back the behavior to ignore presign requests for stubbing. Let me know what you think about that approach.

joguSD:
I think this looks good now.

tests/functional/test_sts.py
jamesls: Sorry didn't see this until now.  I'm not a huge fan on this because it unnecessarily couples the stubber and signers.  Ideally the stubber shouldn't care about (or know) whether a request is presigned or not, it would just have better events it could hook into.

However, I'm not sure it's worth the trouble at this point to introduce new events so I think this is fine for now.

PR #1345, opened 2017-12-20
Add paginators for several services

JordonPhillips:
This adds paginators for a number of services where it was possible
to conclusively determine what the result key was. This only covers
cases where the result key was the only list in the operation
response and there were no structures or maps in the response.

### DIFF
--- a/botocore/data/logs/2014-03-28/paginators-1.json
+++ b/botocore/data/logs/2014-03-28/paginators-1.json
@@ -38,12 +38,6 @@
         "events",
         "searchedLogStreams"
       ]
-    },
-    "GetLogEvents": {
-      "result_key": "events",
-      "output_token": "nextForwardToken",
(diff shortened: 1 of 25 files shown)


botocore/data/logs/2014-03-28/paginators-1.json
```diff
@@ -38,6 +38,12 @@
         "events",
         "searchedLogStreams"
       ]
+    },
+    "GetLogEvents": {
```
jamesls: You'll need to remove this, see: https://github.com/aws/aws-cli/blob/develop/tests/functional/logs/test_get_log_events.py

Might be a good idea to run through the CLI/boto3 tests to make sure there aren't other cases like this.

JordonPhillips:
I ran all the functional / unit tests for all three packages, and that logs paginator was the only one that triggered an error.

PR #1347, opened 2017-12-21
Improve error message when source profile is misconfigured

joguSD:
We attempt to blindly treat a profile as an assume role if the profile doesn't have static credentials. This adds extra validation to ensure that a source profile has either static credentials or a role.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1317,39 +1317,40 @@ class AssumeRoleProvider(CredentialProvider):
             self._has_assume_role_config_vars(profile),
         ])
 
-    def _validate_source_profile(self, parent_profile, source_profile):
+    def _validate_source_profile(self, parent_profile_name,
+                                 source_profile_name):
         profiles = self._loaded_config.get('profiles', {})
-        if source_profile not in profiles:
+        if source_profile_name not in profiles:
             raise InvalidConfigError(
                 error_msg=(
                     'The source_profile "%s" referenced in '
                     'the profile "%s" does not exist.' % (
-                        source_profile, parent_profile)
+                        source_profile_name, parent_profile_name)
                 )
             )
 
-        profile = profiles[source_profile]
+        source_profile = profiles[source_profile_name]
 
         # Ensure the profile has valid credential type
-        if not self._source_profile_has_credentials(profile):
+        if not self._source_profile_has_credentials(source_profile):
             raise InvalidConfigError(
                 error_msg=(
                     'The source_profile "%s" must specify either static '
                     'credentials or an assume role configuration' % (
-                        source_profile)
+                        source_profile_name)
                 )
             )
 
         # Make sure we aren't going into an infinite loop. If we haven't
         # visited the profile yet, we're good.
-        if source_profile not in self._visited_profiles:
+        if source_profile_name not in self._visited_profiles:
             return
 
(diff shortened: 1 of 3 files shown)


botocore/credentials.py
```diff
@@ -1207,12 +1207,14 @@ def __init__(self, load_config, client_creator, cache, profile_name,
...
+        profile = profiles.get(self._profile_name, {})
+        if self._has_assume_role_config_vars(profile):
             return self._load_creds_via_assume_role(self._profile_name)
 
-    def _has_assume_role_config_vars(self):
-        profiles = self._loaded_config.get('profiles', {})
-        return self.ROLE_CONFIG_VAR in profiles.get(self._profile_name, {})
+    def _has_assume_role_config_vars(self, profile):
+        assume_role_keys = [self.ROLE_CONFIG_VAR]
+        return any(key in profile for key in assume_role_keys)
```
JordonPhillips: What is the point of this?

botocore/credentials.py
```diff
@@ -1321,6 +1323,18 @@ def _validate_source_profile(self, parent_profile, source_profile):
                 )
             )
 
+        profile = profiles[source_profile]
+
+        #ensure the profile has valid credential type
+        if not self._has_static_credentials(profile) and \
+           not self._has_assume_role_config_vars(profile):
```
JordonPhillips: It would be nice to get these into a `_has_credentials` or similar function so that when we add support for other profile types we won't have to make this if statement worse.

botocore/credentials.py
```diff
@@ -1321,6 +1323,18 @@ def _validate_source_profile(self, parent_profile, source_profile):
...
+            raise InvalidConfigError(
+                error_msg=(
+                    'The source_profile "%s" must specify either static '
+                    'credentials or a role arn to assume.' % (source_profile)
```
JordonPhillips: I would say "assume role configuration" rather than just mentioning the role arn.

botocore/credentials.py
joguSD: Just to check if the profile has a `role_arn`. I based it on the [_has_static_credentials](https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L1348-L1350) method. Originally I had this as another method but merged it with `_has_assume_role_config_vars` which is plural so I left it open for extension by leaving it as a list of keys).

botocore/credentials.py
joguSD: I'll break this out.

botocore/credentials.py
joguSD: Sounds good.

botocore/credentials.py
```diff
@@ -1321,6 +1328,18 @@ def _validate_source_profile(self, parent_profile, source_profile):
                 )
             )
 
+        profile = profiles[source_profile]
```
kyleknap: I was originally a little confused because `source_profile`/`parent_profile` refer to the name of the profile while `profile` refers to the configuration and I made the assumption they would all be the same type. I wonder if it makes sense to just not create the `profile` variable and use `profiles[source_profile]` directly or call the variable something like `source_profile_config` or rename the other two parameters to `source_profile_name` and `parent_profile_name`?

joguSD: Yeah, I'll see what I can do about this.

Issue #1348, opened 2017-12-23
credential_process should connect subprocess stderr to caller stderr


bwalding:
When the botocore is configured to use a credential_process to acquire AWS credentials, the sub-process stderr is never sent to the user.

(In our case, the `stderr` is prompting for the MFA details to be entered)

`stdin` is mapped correctly - and if I know that I'm waiting on MFA I can enter it and the process completes successfully.

What I'd hope for - `stderr` is passed from the credential_process to the stderr on the `aws` CLI process.

JordonPhillips:
Off the top of my head I can't think of anything that this behavior change would break. I was initially unsure about using stderr as a prompting source, but when stdout is being used entirely then there's not much other choice. The library we use in our sample implementation uses getpass, which has [stderr as a fallback](https://github.com/python/cpython/blob/master/Lib/getpass.py#L65).

Incidentally, getpass is a possible workaround since it tries to write/read from the tty directly first. Most cases that should work, though there is the potential downside of a non-echoing input.

pcolmer:
Part of the drawback to using getpass to handle this requirement is where the credential process needs to output something more complicated than a single prompt.

For example, in the code I've submitted to enhance awsprocesscreds to support MFA under Okta, the code lists out all of the MFA options available to the user. That *could* be built up as a single string prompt but it isn't nice for the user to then parse it visually.

pcolmer:
There is also the scenario where the credential process needs to provide the user with information about the progress of the authentication process. For example, if the user chooses a push notification operation, my code currently emits some text to let the user know that the code is waiting for the result of the push notification. Granted, I could remove that but I think it is more user-friendly if the user is kept informed about what is happening ...

giles-jason:
This thread has been stale for a while but it would be nice to know if this is gaining any traction. We have federated logins for AWS and a company policy that requires MFA at login. This means that we need more interaction with the user than just prompting for a password. Not being able to use `stdout` for anything but the JSON payload, and not being able to use `stderr` at all limits what you can do with this integration.

edulop91:
FWIW AWS libraries in other languages already do this:
https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/processcreds/provider.go#L347

ryan-gerstenkorn-sp:
Wanted to link this PR here as well as add that this would be useful for use with aws-vault.

https://github.com/boto/botocore/pull/1835

ryan-gerstenkorn-sp:
[...]

PR #1349, opened 2017-12-23
#1348 - map subprocess stderr to caller stderr

bwalding:


jamesls:
Can you share the motivation for this PR?  The stderr is included in the exception that's raised, I'm a little hesitant to change this without a good reason.

lorengordon:
Linking the issue for the requested info on motivation... https://github.com/boto/botocore/issues/1348

lorengordon:
@jamesls @JordonPhillips Another use case where stderr getting swallowed makes it much harder to process MFA prompts (particularly when called via `credential_process`), https://github.com/segmentio/aws-okta/issues/143. Anything we can do to unblock this pr?

twooster:
Plus one to this. I just wrote a similar patchset for this before checking the PRs, after spending several hours writing something similar to segment's `aws-okta` command mentioned above. I would love to see this in the aws cli to simplify MFA workflows.

My patchset checks if `stdin/stderr` are ttys, and only then pass them through. Would this be an acceptable solution to the maintainers?

edulop91:
FWIW AWS libraries in other languages already do this:
https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/processcreds/provider.go#L347

moritzheiber:
Can we please merge this PR? This is literally breaking pretty much all of the tooling that needs to request more details before providing credentials. If there's anything that still needs to be done I'm happy to volunteer.

bwalding:
Closing because this problem doesn't impact me any more (and having it in my dashboard was becoming irksome) - if someone else wants to clone it - please do.

PR #1352, opened 2018-01-05
Enforce ordering on CloudFormation get_template TemplateBody

stealthycoin:
The TemplateBody is loaded from a string into a JSON dictionary which
did not preserve order. Instead use an OrderedDict to prserve the
template ordering.

cc @jamesls @kyleknap @dstufft @JordonPhillips @joguSD

### DIFF
--- a/.changes/next-release/enhancement-cloudformationgettemplatetemplatebodyordering-86054.json
+++ b/.changes/next-release/enhancement-cloudformationgettemplatetemplatebodyordering-86054.json
@@ -1,0 +1,5 @@
+{
+  "category": "cloudformation get_template template body ordering",
+  "type": "enhancement",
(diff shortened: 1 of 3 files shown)


.changes/next-release/enhancement-cloudformationgettemplatetemplatebodyordering-86054.json
```diff
@@ -0,0 +1,5 @@
+{
+  "category": "cloudformation get_template template body ordering", 
+  "type": "enhancement", 
+  "description": "fixes boto/boto3`#1378 <https://github.com/boto/botocore/issues/1378>`__"
```
kyleknap: I think the link is messed up here. The url is pointing to `botocore` (which does not exist for that issue number) but really needs to point to `boto3`.

PR #1353, opened 2018-01-06
unquote string value in parse_timestamp

rsalmaso:
Hi, I have a django project which uses django-storages with boto3 backend to handle images on s3, and sometime botocore raise an exception because the 'expires' header is quoted.

```python
{
  'body': b'',
  'headers': {'content-length': '1874986', 'expires': 'Thu,%2031%20Dec%202099%2020:00:00%20GMT'},
  'status_code': 200,
}
```

For this case I try to be sure that the value is validated before trying to convert to a datetime object.

(
As a micro optimization could be done only in case of error, but I don't think it is really worth it

```python
    try:
        # enforce that GMT == UTC.
        return dateutil.parser.parse(unquote(value), tzinfos={'GMT': tzutc()})
        raise ValueError('Invalid timestamp "%s": %s' % (value, e))
```

)

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -27,7 +27,7 @@ import botocore
 from botocore.exceptions import InvalidExpressionError, ConfigNotFound
 from botocore.exceptions import InvalidDNSNameError, ClientError
 from botocore.exceptions import MetadataRetrievalError
-from botocore.compat import json, quote, zip_longest, urlsplit, urlunsplit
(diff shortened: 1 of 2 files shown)


rsalmaso:
Uhm, I tracked a bit and found that S3 is right: it's metadata that is quoted. Probably it was boto or the django-storage boto backend that for some reason sometime quote this value and sometime not, and it was hiding this problem.
Feel free to close this PR if there is no general advantage other than me, I'll keep a patched version in this case.

PR #1358, opened 2018-01-09
Add waiters for athena

cfchou:
I added two waiters for athena(https://github.com/boto/boto3/issues/1212).

For single query, `query_succeeded` only comes back without error iff query succeeded. Otherwise, `WaitError` is raised and one may call `get_query_execution` to examine the details.

For multiple queries, `queries_finished` is used. When it comes back, it means none of the queries is in `RUNNING` state. One should call `batch_get_query_execution` to examine the details.

### DIFF
--- a/botocore/data/athena/2017-05-18/waiters-2.json
+++ b/botocore/data/athena/2017-05-18/waiters-2.json
@@ -1,0 +1,43 @@
+{
+  "version": 2,
+  "waiters": {
+    "QuerySucceeded": {
+      "delay": 15,
+      "operation": "GetQueryExecution",
+      "maxAttempts": 40,
+      "acceptors": [
+        {
+          "expected": "SUCCEEDED",
+          "matcher": "path",
+          "state": "success",
+          "argument": "QueryExecution.Status.State"
+        },
+        {
+          "expected": "FAILED",
+          "matcher": "path",
+          "state": "failure",
+          "argument": "QueryExecution.Status.State"
+        },
+        {
+          "expected": "CANCELED",
+          "matcher": "path",
+          "state": "failure",
+          "argument": "QueryExecution.Status.State"
+        }
+      ]
+    },
+    "QueriesFinished": {
+      "delay": 15,
+      "operation": "BatchGetQueryExecution",
+      "maxAttempts": 40,
+      "acceptors": [
+        {
+          "expected": 0,
+          "matcher": "path",
+          "state": "success",
+          "argument": "length(QueryExecutions[?Status.State == `RUNNING`])"
+        }
+      ]
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


jamesjrg:
I assume delay=15 means a delay of 15 seconds, that seems quite a long time for cases where queries return relatively quickly?

cfchou:
@jamesjrg you are right. While the default values work for to my cases, they may not be practical for many. I'd love to change them if that makes more sense.

Fortunately, the default values can be overwritten if needed:
```
waiter.wait(WaiterConfig={
            'Delay':5,
 })
```

jamesjrg:
I'm just some random internet passer-by, but 5 seconds sounds like a reasonable default to me

xysr89:
+1 on 5 secs. in our use case, 1 sec seems too much API calls and 15 secs are just too long especially you have a chain of lambda functions that execute different athena queries

jamesjrg:
I've actually now gone for 2 seconds, though depending on the use case I don't know if that would work for others. A few extra HTTP calls don't really cost very much and I don't like waiting for results!

davoscollective:
This is awesome, looks like code coverage reduced, maybe there are example tests for other waiters that could provide an example of how to test it. That's probably the main reason why its not been merged.

pswenson:
is this feature abandoned?

neilmcguigan:
import boto3
from botocore.waiter import WaiterModel
from botocore.waiter import create_waiter_with_client

my_waiter_model = WaiterModel({
    "version":2,
    "waiters":{
        "Executed": {
            "operation":"get_query_execution",
            "delay": 2,
            "maxAttempts": 5,
            "acceptors": [
                {
                    "matcher": "path",
                    "argument": "QueryExecution.Status.State",
                    "expected": "SUCCEEDED",
                    "state":"success"
                },
                {
                    "matcher": "path",
                    "argument": "QueryExecution.Status.State",
                    "expected": "RUNNING",
                    "state": "retry"
                },
                {
                    "matcher": "path",
                    "argument": "QueryExecution.Status.State",
                    "expected": "FAILED",
                    "state":"failure"
                },
                {
                    "matcher": "path",
                    "argument": "QueryExecution.Status.State",
                    "expected": "QUEUED",
                    "state":"retry"
                },
                {
                    "matcher": "path",
                    "argument": "QueryExecutionId.Status.State",
                    "expected": "CANCELLED",
                    "state":"success"
                }
            ]
        }
    }
})

athena = boto3.client("athena")

waiter = create_waiter_with_client("Executed", my_waiter_model, athena)

waiter.wait(QueryExecutionId = "whatever")
`

stobrien89:
Hi all,

Thanks again for the request. We are now submitting requests for paginators and waiters upstream so that all AWS SDKs benefit from a consistent implementation. I will pass on the request for these waiters to the service team. We will also update the guidance in the contributing guide to better reflect this policy. I've added an issue to track this here:

aws/aws-sdk#80

Thanks again for bringing this to our attention!

PR #1369, opened 2018-01-20
Add role_arn keyword argument to Session()

akdor1154:
Currently Session() can be instantiated with a manual `profile` argument, which forces the use of a profile in your local config, and can be used to get boto to assume a role for you.
This PR takes this one step further and adds a `role_arn` argument which can be used the same way, but does not require config files to exist on the machine.
If `role_arn` is provided, a session will be attempted to be created by assuming `role_arn` from the otherwise default credentials (env vars, config, instance role, etc).

This was done by modifying the AssumeRoleProvider.

Questions:
 - is this an acceptable feature to add? It would eliminate the need to call sts.assume_role and manually instantiate a session in any usage of botocore that I've ever come across.
 - Is this an OK way to go about it? AssumeRoleProvider isn't named like it should be restricted to only assuming roles from profiles, but this does add a few different paths through it. It also has had to be extended a little as it was written with the assumption that it would never have to guess an appropriate credential source.

if the above two questions are fine, then see my github comments in the code itself for more specific questions.

Still to do:
 - [ ] integration tests
 - [ ] functional tests
 - [ ] more unit tests to get coverage up (or to ensure quality code, or something I guess :) )
 - [ ] docs
 - [ ] allow for role_session_name as well

I just want to wait on some indication that this is acceptable before I undertake these.

As is, botocore with this pr can be used by:
```python
   # default session
    botocore.session.Session()

    # assume a role with the credentials inferred as with the default session:
    botocore.session.Session(role_arn='arn:aws:iam::00000000:role/test-role'),

    # assume a role on top of a manually selected profile (I expect this is a corner case)
    botocore.session.Session(profile='some-profile', role_arn='arn:aws:iam::00000000:role/test-role')
```

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -61,6 +61,7 @@ def create_credential_resolver(session, cache=None):
     config_file = session.get_config_variable('config_file')
     metadata_timeout = session.get_config_variable('metadata_service_timeout')
     num_attempts = session.get_config_variable('metadata_service_num_attempts')
+    role_arn = session.get_config_variable('role_arn')
     if cache is None:
         cache = {}
 
@@ -71,18 +72,8 @@ def create_credential_resolver(session, cache=None):
             timeout=metadata_timeout,
             num_attempts=num_attempts)
     )
-    assume_role_provider = AssumeRoleProvider(
-        load_config=lambda: session.full_config,
-        client_creator=session.create_client,
-        cache=cache,
-        profile_name=profile_name,
-        credential_sourcer=CanonicalNameCredentialSourcer([
-            env_provider, container_provider, instance_metadata_provider
-        ])
-    )
+
     providers = [
-        env_provider,
-        assume_role_provider,
         SharedCredentialProvider(
             creds_filename=credential_file,
             profile_name=profile_name
@@ -98,9 +89,22 @@ def create_credential_resolver(session, cache=None):
         instance_metadata_provider
     ]
 
+    assume_role_provider = AssumeRoleProvider(
+        load_config=lambda: session.full_config,
+        client_creator=session.create_client,
+        cache=cache,
+        profile_name=profile_name,
+        role_arn=role_arn,
+        credential_sourcer=CanonicalNameCredentialSourcer([env_provider]+providers)
+    )
+
+    providers.insert(0, assume_role_provider)
+
     explicit_profile = session.get_config_variable('profile',
                                                    methods=('instance',))
-    if explicit_profile is not None:
+    if explicit_profile is None:
+        providers.insert(0, env_provider)
+    else:
         # An explicitly provided profile will negate an EnvProvider.
         # We will defer to providers that understand the "profile"
         # concept to retrieve credentials.
@@ -116,7 +120,6 @@ def create_credential_resolver(session, cache=None):
         # This means that the only way a "profile" would win is if the
         # EnvProvider does not return credentials, which is what we want
         # in this scenario.
-        providers.remove(env_provider)
         logger.debug('Skipping environment variable credential check'
                      ' because profile name was explicitly set.')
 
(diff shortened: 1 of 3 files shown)


botocore/credentials.py
```diff
@@ -1207,19 +1217,44 @@ def __init__(self, load_config, client_creator, cache, profile_name,
...
+        role_config = self._get_role_config_from_role_arn(role_arn)
+        # TODO: is this OK?
+        if self._profile_name is not None and self._profile_name != 'default':
+            self._visited_profiles.append(self._profile_name)
+            source_credentials = self._resolve_credentials_from_profile(self._profile_name)
+        else:
+            source_credentials = self._credential_sourcer.get_first_successful_credentials()
+
+        if source_credentials is None:
+            raise PartialCredentialsError(
```
akdor1154: What Error should be raised here? Or should I just return None?

botocore/credentials.py
```diff
@@ -1295,30 +1367,48 @@ def _get_role_config(self, profile_name):
...
-                'but no source provider was configured.' % (
-                    credential_source, parent_profile)
-            ))
-        if not self._credential_sourcer.is_supported(credential_source):
-            raise InvalidConfigError(error_msg=(
-                'The credential source "%s" referenced in profile "%s" is not '
-                'valid.' % (credential_source, parent_profile)
-            ))
+            if parent_profile is not None:
+                error_msg = (
```
akdor1154: This is a lot of code and branches just to get nice error messages, is there any other way you might have me write this?

botocore/credentials.py
```diff
@@ -1295,30 +1367,48 @@ def _get_role_config(self, profile_name):
...
+                    'but no source provider was configured.' % (
+                        credential_source, parent_profile)
+                )
+            else:
+                error_msg = (
+                    'The credential_source "%s" was going to be used, but no '
+                    'source provider was configured.' % (credential_source)
+                )
+            raise InvalidConfigError(error_msg=error_msg)
+        if parent_profile is None and credential_source == 'auto':
```
akdor1154: see HACK above.

botocore/credentials.py
```diff
@@ -1253,8 +1288,35 @@ def _load_creds_via_assume_role(self, profile_name):
...
         )
+    
+    def _get_role_config_from_role_arn(self, role_arn, profile_name=None):
+        """Retrieves and validates the role configuration from a specified role_arn."""
+        if profile_name is not None:
+            source_profile = profile_name
+            credential_source = None
+        else:
+            source_profile = None
+            credential_source = 'auto' # TODO
```
akdor1154: HACK:
currently AssumeRoleProvider is written with the assumption that the credential_source is always set, and that it should fail if it is incorrect. This PR requires that we do some guesswork, though. By doing this, I can use the same _validate(role_config) path for both the _from_role_arn and _from_profile() paths, which is really nice. Can you think of a way to modify AssumeRoleProvider in a nice way such that we don't go duplicating all the validation logic? Thoughts?

botocore/credentials.py
```diff
@@ -1398,6 +1488,20 @@ def _resolve_credentials_from_source(self, credential_source,
 class CanonicalNameCredentialSourcer(object):
     def __init__(self, providers):
         self._providers = providers
+    
+    def get_first_successful_credentials(self):
```
akdor1154: Is this OK to add here?

botocore/credentials.py
```diff
@@ -98,10 +89,24 @@ def create_credential_resolver(session, cache=None):
         instance_metadata_provider
     ]
 
+    assume_role_provider = AssumeRoleProvider(
+        load_config=lambda: session.full_config,
+        client_creator=session.create_client,
+        cache=cache,
+        profile_name=profile_name,
+        role_arn=role_arn,
+        credential_sourcer=CanonicalNameCredentialSourcer([env_provider]+providers)
```
akdor1154: This is so the AssumeRole provider will attempt to assume role from whatever the otherwise default credentials are. There is a small side effect of this, though: it allows me to write 
```
[profile some-profile]
credential_source = [something that isn't Environment, Ec2InstanceMetadata, or EcsContainer], e.g. Ec2Config
```
Is this acceptable?

akdor1154:
still happy to work on this if I was to get even a hint of interest from a maintainer.

PR #1371, opened 2018-01-23
Close the connection in StreamingBody.close().

timuralp:
If StreamingBody.close() is called and the connection is closed, boto
should also close the underlying socket if any data is buffered.


timuralp:
Calling `close()` on `HTTPResponse` in recent urllib3 versions closes the underlying connection (if one is present).

PR #1374, opened 2018-01-30
Validate waiters against schema

JordonPhillips:
This validates waiter files against a json schema to ensure that they
are at least structurally valid.

### DIFF
--- a/tests/functional/test_waiter_config.py
+++ b/tests/functional/test_waiter_config.py
@@ -11,16 +11,94 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import jmespath
+from jsonschema import Draft4Validator
 
 import botocore.session
+from botocore.exceptions import UnknownServiceError
 from botocore.utils import ArgumentGenerator
+
+
+WAITER_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "version": {"type": "number"},
+        "waiters": {
+            "type": "object",
+            "additionalProperties": {
+                "type": "object",
+                "properties": {
+                    "type": {
+                        "type": "string",
+                        "enum": ["api"]
+                    },
+                    "operation": {"type": "string"},
+                    "description": {"type": "string"},
+                    "delay": {
+                        "type": "number",
+                        "minimum": 0,
+                    },
+                    "maxAttempts": {
+                        "type": "integer",
+                        "minimum": 1
+                    },
+                    "acceptors": {
+                        "type": "array",
+                        "items": {
+                            "type": "object",
+                            "properties": {
+                                "state": {
+                                    "type": "string",
+                                    "enum": ["success", "retry", "failure"]
+                                },
+                                "matcher": {
+                                    "type": "string",
+                                    "enum": [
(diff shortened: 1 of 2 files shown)


tests/functional/test_waiter_config.py
```diff
@@ -53,6 +131,12 @@ def _lint_single_waiter(client, waiter_name, service_model):
         _validate_acceptor(acceptor, op_model, waiter.name)
 
 
+def _validate_schema(validator, waiter_json):
```
kyleknap: Looks like this function is not used anywhere.

tests/functional/test_waiter_config.py
```diff
@@ -11,16 +11,94 @@
...
+                                    "enum": ["success", "retry", "failure"]
+                                },
+                                "matcher": {
+                                    "type": "string",
+                                    "enum": [
+                                        "path", "pathAll", "pathAny",
+                                        "status", "error"
+                                    ]
+                                },
+                                "argument": {"type": "string"},
```
kyleknap: Not sure if it is possible in JSON schema, but is it possible to mark argument as required if matcher's value is `path`, `pathAll` or `pathAny`? Don't think it is a big deal as we have tests later on that assume argument is existing if any of those three matchers are used.

tests/functional/test_waiter_config.py
```diff
@@ -11,16 +11,94 @@
...
+            # We use the loader directly here because we need the entire
+            # json document, not just the portions exposed (either
+            # internally or externally) by the WaiterModel class.
+            loader = session.get_component('data_loader')
+            waiter_model = loader.load_service_model(
+                service_name, 'waiters-2')
+        except UnknownServiceError:
+            # The service doesn't have waiters
+            continue
+        validator.validate(waiter_model)
```
kyleknap: Forgot a comment... We may want to put this in a yield statement to make sure the entire test does not immediately fail for a single waiter. Also what does the error message look like for something a waiter config that fails this test?

tests/functional/test_waiter_config.py
JordonPhillips: ah yes

JordonPhillips: No, conditional requirements aren't supported.

tests/functional/test_waiter_config.py
JordonPhillips:
```json
{
  "version": 2,
  "waiters": {
    "BucketExists": {
      "delay": 5,
      "operation": "HeadBucket",
      "maxAttempts": 20,
      "acceptors": [
        {
          "expected": 200,
          "matcher": "status",
          "state": "success"
        },
        {
          "expected": 404,
          "matcher": "status",
          "state": "retry"
        }
      ]
    }
  }
}
```

```
  File "/Users/phjordon/.pyenv/versions/3.6.2/envs/bc/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/phjordon/workspace/botocore/tests/functional/test_waiter_config.py", line 137, in _validate_schema
    raise AssertionError('\n'.join(errors))
AssertionError: Additional properties are not allowed ('foo' was unexpected)
```

PR #1377, opened 2018-02-06
Sync waiters with latest versions

jamesls:


dstufft:
Did you mean to set ``botocore/data/dynamodb/2012-08-10/waiters-2.json`` to executable?

jamesls:
Pushed a commit that removed the executable bit from all our json files.

botocore/data/appstream/2016-12-01/waiters-2.json
```diff
@@ -0,0 +1,55 @@
+{
+  "version": 2,
+  "waiters": {
+    "FleetStarted": {
```
kyleknap: Are these correct? Looking at the [docs](https://docs.aws.amazon.com/appstream2/latest/APIReference/API_Fleet.html#AppStream2-Type-Fleet-State). It looks like the states should be `RUNNING` and `STOPPED`. There are also no `PENDING_*` states.

botocore/data/elbv2/2015-12-01/waiters-2.json
```diff
@@ -0,0 +1,100 @@
...
+      "maxAttempts": 40,
+      "acceptors": [
+        {
+          "state": "success",
+          "matcher": "pathAll",
+          "argument": "LoadBalancers[].State.Code",
+          "expected": "active"
+        },
+        {
+          "state": "retry",
```
kyleknap: This state is unneeded, but should not block pulling it in.

botocore/data/elbv2/2015-12-01/waiters-2.json
```diff
@@ -0,0 +1,100 @@
...
+        }
+      ]
+    },
+    "LoadBalancersDeleted": {
+      "delay": 15,
+      "operation": "DescribeLoadBalancers",
+      "maxAttempts": 40,
+      "acceptors": [
+        {
+          "state": "retry",
```
kyleknap: Another unneeded acceptor. Would be nice to fix but should not block pulling this in.

botocore/data/rds/2014-10-31/waiters-2.json
```diff
@@ -98,20 +86,87 @@
         }
       ]
     },
-    "DBSnapshotCompleted": {
```
kyleknap: This looks like a breaking change as we are renaming the waiter from `DBSnapshotCompleted` to `DBSnapshotAvaliable`. However, I am not sure how useful it was given its configuration was this:
```
    "DBSnapshotCompleted": {
      "delay": 15,
      "operation": "DescribeDBSnapshots",
      "maxAttempts": 40,
      "acceptors": [
        {
          "expected": "DBSnapshotNotFound",
          "matcher": "error",
          "state": "success"
        },
        {
          "expected": "available",
          "matcher": "pathAll",
          "state": "success",
          "argument": "DBSnapshots[].Status"
        }
      ]
}
```
Looks like the first acceptor should have been a failure?

botocore/data/appstream/2016-12-01/waiters-2.json
jamesls: Yeah it's correct.

botocore/data/rds/2014-10-31/waiters-2.json
jamesls: Ah, I missed that one.  This is one botocore had that wasn't upstream.  I'm going to just add it back verbatim.

jamesls:
@kyleknap Added back the missing waiter.  Just to clarify the intent of this PR, this is syncing all the waiters we can while avoiding ones that are flat out wrong.  Small cleanups aren't something I'm targeting here.  Those ideally happen upstream.

PR #1389, opened 2018-02-19
Improve reliability of kinesis integ tests

jamesls:
There are two things I fixed:

* Poll up to 1 minute before data becomes available.
* Use unique data for payloads data.

The integ tests were technically wrong because they were checking
the first record's data, but given the test all share the same stream
and shard that data will depend on whatever test ran first.  In only
worked for now because the tests were all using the same first data
payload of "foobar".

### DIFF
--- a/tests/integration/test_kinesis.py
+++ b/tests/integration/test_kinesis.py
@@ -51,7 +51,6 @@ class TestKinesisListStreams(unittest.TestCase):
         self.client.put_record(
             StreamName=self.stream_name, PartitionKey='foo', Data=unique_data)
         # Give it a few seconds for the record to get into the stream.
-        time.sleep(10)
         records = self.wait_for_stream_data()
         self.assert_record_data_contains(records, unique_data.encode('ascii'))
         self.assertTrue(len(records['Records']) > 0)
@@ -67,8 +66,6 @@ class TestKinesisListStreams(unittest.TestCase):
                 'PartitionKey': 'foo'
             }]
         )
-        # Give it a few seconds for the record to get into the stream.
-        time.sleep(10)
         records = self.wait_for_stream_data()
         self.assert_record_data_contains(records, unique_data.encode('ascii'))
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_kinesis.py
```diff
@@ -99,19 +86,30 @@ def test_can_put_records_multiple_blob(self):
...
-
-        records = self.client.get_records(
-            ShardIterator=shard_iterator['ShardIterator'])
-        self.assertTrue(len(records['Records']) == 2)
-        # Verify that both made it through.
+        records = self.wait_for_stream_data()
+        self.assert_record_data_contains(records, b'foobar', b'barfoo')
+
+    def wait_for_stream_data(self, num_attempts=6, poll_time=10):
+        # Poll until we get records returned from get_records().
```
kyleknap: Is there any specific documentation on why we are making the max wait time 1 min? Or is this being chosen because it was working well for that value and waiting a minute is not too bad?

tests/integration/test_kinesis.py
```diff
@@ -46,44 +47,30 @@ def test_list_streams(self):
 
     @attr('slow')
     def test_can_put_stream_blob(self):
+        unique_data = str(uuid4())
         self.client.put_record(
-            StreamName=self.stream_name, PartitionKey='foo', Data='foobar')
+            StreamName=self.stream_name, PartitionKey='foo', Data=unique_data)
         # Give it a few seconds for the record to get into the stream.
         time.sleep(10)
```
kyleknap: Do you think we need to sleep 10 seconds now before going into the wait loop? Or do you think we can just go straight into the waiter logic? Not really a big deal. Just read a little weird when I saw a `sleep()` followed by a `wait_for_stream_data()` as I would have assumed the `sleep()` would not be needed if there was already a wait logic.

tests/integration/test_kinesis.py
jamesls: > Is there any specific documentation on why we are making the max wait time 1 min? Or is this being chosen because it was working well for that value and waiting a minute is not too bad?

No, completely arbitrary value that I picked that was larger than the previous time of 10 seconds.

tests/integration/test_kinesis.py
jamesls: I'll go ahead and remove it.

jamesls:
@kyleknap Ready for another look.  I had to change the order of when we `sleep()` to ensure we still sleep initially for 10 seconds.  The rationale being that we still have to allow for some settling time when we add a record.  The wait() method is just waiting until records exist but that will always be true after the first test runs because they share the same stream/shard.

PR #1392, opened 2018-02-22
Adjust documentation to use floats for timeouts instead of ints

JustinTArthur:
While `int`s work fine here, it's not the native or most precise numeric type available for these. The requests and urllib3 libraries expect `float`.

### DIFF
--- a/botocore/config.py
+++ b/botocore/config.py
@@ -35,12 +35,12 @@ class Config(object):
     :param user_agent_extra: The value to append to the current User-Agent
         header value.
 
-    :type connect_timeout: int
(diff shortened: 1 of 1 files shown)


kyleknap:
Thanks for the PR. I'm fine with including float in the docstring, but I still think we should still document integers as a valid type. Could we instead change the value to be `int/float` to indicate both types are accepted?

JustinTArthur:
@kyleknap sure, will revise.

JustinTArthur:
@kyleknap I've adjusted the text to say _float or int_ and have re-based.

PR #1394, opened 2018-02-22
Allow disabling IMDS via env vars

JordonPhillips:
This allows customers to disable fetching credentials from IMDS by
setting `AWS_EC2_METADATA_DISABLED` to `true`.

### DIFF
--- a/.changes/next-release/enhancement-Credentials-75846.json
+++ b/.changes/next-release/enhancement-Credentials-75846.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Credentials",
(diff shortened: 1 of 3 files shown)


JordonPhillips:
The reason I put it there was to make sure that the environment variable disables all IMDS features. Right now we aren't using that class for anything but credentials, but that might not be the case in the future.

jamesls:
Just to reiterate, I stand by my original feedback.  Curious to see what others think.

cc @kyleknap

jamesls:
Capturing offline discussions, while I'm personally not a fan of this and would prefer this is done at the credential level, given this API is *not* public, it still has the same end result for users.  I think we should take a serious look at changing this if we expose a general IMDS interface.

PR #1402, opened 2018-03-13
Tighten upper bound on python-dateutil version requirement

joguSD:
Version 2.7.0 of `python-dateutil` [dropped support for Python 2.6](https://github.com/dateutil/dateutil/pull/362).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ def find_version(*file_paths):
 
 
 requires = ['jmespath>=0.7.1,<1.0.0',
-            'python-dateutil>=2.1,<3.0.0',
+            'python-dateutil>=2.1,<2.7.0',
             'docutils>=0.10']
 
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 tox>=2.5.0,<3.0.0
-python-dateutil>=2.1,<3.0.0
+python-dateutil>=2.1,<2.7.0
 nose==1.3.0
 mock==1.3.0
 wheel==0.24.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,7 @@ universal = 1
 
 [metadata]
 requires-dist =
-	python-dateutil>=2.1,<3.0.0
+	python-dateutil>=2.1,<2.7.0
 	jmespath>=0.7.1,<1.0.0
 	docutils>=0.10
 	ordereddict==1.1; python_version=="2.6"


setup.py
```diff
@@ -24,7 +24,7 @@ def find_version(*file_paths):
 
 
 requires = ['jmespath>=0.7.1,<1.0.0',
-            'python-dateutil>=2.1,<3.0.0',
+            'python-dateutil>=2.1,<2.7.0',
```
dstufft: Would it make sense to use different version ranges for Python 2.6 and every other version of Python? That'd limit the potential missing out of bugfixes and such only to people using 2.6.

dstufft:
The problem with doing that is we're effectively preventing people from getting updated dateutils then until we drop support for Python 2.6. I'm not 100% sure on what all our uses of dateutil is, but that would also include locking people to an older version of the tzinfo database, which means that they'll get incorrect data when timezones change, like with what Florida is currently attempting to do with it's [Sunshine Protection Act](https://www.flsenate.gov/Session/Bill/2018/1013/BillText/er/PDF). It also locks users out of any bug fixes or performance improvements in dateutil even if they're not using Python 2.6.

jamesls:
That's a fair point, but at this point it fails py2.6 as well as py3.6 for us (the py3.6 we'd still have to figure out if that's on us or dateutil but the tests are failing).  I'd prefer to hold off on taking this update until we can vet it better.

I'd like to get us unblocked for now and come back and figure out a better plan unless there's a better option for today.

dstufft:
Yea, a single pin for today seems perfectly reasonable since 3.6 is failing.

Syeberman:
Hi all!  Our builds are failing because of this change.  

This change could have pinned versions for your CI environment, while **not** impacting downstream users.  I believe this is what [@dstufft](https://github.com/boto/botocore/pull/1402/files#r174229009) mentioned in his review.

setup.py
blueyed: :+1:
https://github.com/boto/botocore/pull/1406

freezegun has this btw: https://github.com/spulec/freezegun/blob/6bb258967769aa1117c00affc3a4bdab5db0ddee/setup.py#L8-L12J

blueyed:
@jamesls 
Tests for py36 do not seem to be failing with https://github.com/boto/botocore/pull/1406 (using `tox -e py36`).
Do you have some reference/build log?

ryanpetrello:
I'll chime in and say it would be nice if this pin was temporary - I'm also working on a downstream project that depends on `boto`, but would also really appreciate some of the bug fixes in 2.7.0 of `python-dateutil`.

blueyed:
@ryanpetrello 
=> #1406.

jhoelzl:
This change is breaking our internal builds because now we can't find a proper version of python_dateutil:

> Could not find a version that matches python_dateutil<2.7.0,==2.7.0,>=2.1,>=2.5.3

pganssle:
Are you sure you need these tight version constraints? `dateutil` version 2.7.0+ specified [`python-requires>=2.7.0`](https://github.com/dateutil/dateutil/blob/2.7.0/setup.py#L44), so `pip` or anything that respects the `Requires-Python` metadata will not download this for Python 2.6.

wimglenn:
This seems backwards - it's inconveniencing the Python 3 world for the benefit of some subset that are have both an outdated version of pip/setuptools and a Python version that is EOL for years now?  Could we just remove the upper bound and let the luddites sort out their own mess?

PR #1405, opened 2018-03-14
Fix race condition bug with deferred credentials

kyleknap:
Related to: https://github.com/aws/aws-cli/issues/3194

The issue occurred for the following scenario:

1) One thread secures the lock and goes through the initial refresh process
2) The thread with the lock sets the `self._access_key` and `self._secret_key`, but not yet `self._frozen_credentials`
3) The other threads call `refresh_needed()` and notice that since `self._access_key` and `self._secret_key` are set and the credential are not expired. As a result, they just return `self._frozen_credentials`.
4) However since `self._frozen_credentials` starts off as None and since `self._frozen_credentials` have not been set by the thread with the lock, the `DeferredRefreshableCredentials` returns None which causes the threads in this to raise a `NoCredentialsError` for those threads.

To fix this, the `refresh_needed()` was updated to use `self._frozen_credentials` to check if credentials have been set.

I added a test that confirmed this was an issue, and I ran it with the fix in place. I ran it in a loop (about 100 iterations) and I would see the test would fail about 10% of the time. With the fix, the test never failed no matter how many times I would try it.

### DIFF
--- a/.changes/next-release/bugfix-Credentials-23327.json
+++ b/.changes/next-release/bugfix-Credentials-23327.json
@@ -1,0 +1,5 @@
+{
+  "category": "Credentials", 
+  "type": "bugfix", 
+  "description": "Fix a race condition related to assuming a role for the first time (`#1405 <https://github.com/boto/botocore/pull/1405>`__)"
+}
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -444,6 +444,8 @@ class RefreshableCredentials(Credentials):
             # set of temporary credentials we have.
             return
         self._set_from_data(metadata)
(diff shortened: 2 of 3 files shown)


joguSD:
If I'm understanding correctly the test added will only fail if we regress ~10% of the time? I'm not sure how I feel about that.

kyleknap:
So I was thinking about that too. I could maybe run this in a for loop? Increasing the thread count does not really help as I tried spinning up hundreds of threads and getting it to fail was still spotty. In general, these stress testing/race condition ones can be difficult to alarm 100% of the time and we catch these because we run the tests very often across different python versions.

PR #1406, opened 2018-03-14
setup.py: do not tighten dateutil for py27 and up

blueyed:
Ref: https://github.com/boto/botocore/pull/1402#discussion_r174229009

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,9 +24,12 @@ def find_version(*file_paths):
 
 
 requires = ['jmespath>=0.7.1,<1.0.0',
-            'python-dateutil>=2.1,<2.7.0',
             'docutils>=0.10']
-
+if sys.version_info <= (2, 7):
+    # python-dateutil 2.7.0 dropped support for py26.
+    requires.append('python-dateutil>=2.1,<2.7.0')
+else:
+    requires.append('python-dateutil>=2.1,<3.0.0')
 
 if sys.version_info[:2] == (2, 6):
     # For python2.6 we have a few other dependencies.
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 tox>=2.5.0,<3.0.0
-python-dateutil>=2.1,<2.7.0
+python-dateutil>=2.1,<3.0.0
 nose==1.3.0
 mock==1.3.0
 wheel==0.24.0
--- a/requirements26.txt
+++ b/requirements26.txt
@@ -1,1 +1,2 @@
 unittest2==0.8.0
+python-dateutil>=2.1,<2.7.0
--- a/scripts/ci/install
+++ b/scripts/ci/install
@@ -14,16 +14,14 @@ def run(command):
     return check_call(command, shell=True)
 
 
+run('pip install -r requirements.txt')
 try:
     # Has the form "major.minor"
     python_version = os.environ['PYTHON_VERSION']
 except KeyError:
     python_version = '.'.join([str(i) for i in sys.version_info[:2]])
-
 if python_version == '2.6':
     run('pip install -r requirements26.txt')
-
-run('pip install -r requirements.txt')
 run('pip install coverage')
 if os.path.isdir('dist') and os.listdir('dist'):
     shutil.rmtree('dist')


joguSD:
Travis will pass for this because the `requirements.txt` wasn't also updated which is what we use when we run travis.

This is something that we're currently looking into doing. I have a branch that will do this (and also updates `setup.cfg` and `requirements.txt` as well). The issue is that we're failing one test on Python 3.6 in travis. I've been testing different permutations to see what exactly is causing the recursion limit to be hit and it seems to be a combination of:
* manipulating stdout (like travis or other CI systems tend to do)
* dateutil logging a warning that gets intercepted by nosetests xunit plugin specifically it's [`Tee` class](https://github.com/nose-devs/nose/blob/7c26ad1e6b7d308cafa328ad34736d34028c122a/nose/plugins/xunit.py#L127-L147)
* Running BOTH the unit and functional tests together (running just the functional tests didn't seem to trigger this failure).

blueyed:
@joguSD 
I've added a fixup to address this (would be nice to have a single point to configure this, i.e. setup.py), and it works locally.

blueyed:
Do you have a link to a failing build?

blueyed:
And to be honest: while I am really on the side of supporting legacy (Python) versions, what is the reason for not dropping py26 here also?

joguSD:
Heres a failing build: https://travis-ci.org/joguSD/botocore/builds/353493633
I've made a few changes as previously the real exception would be swallowed.

In the longer term I think we would like to drop Py26 support but we won't be able to do so for a long while as we still have customers expecting botocore (even more so for the aws-cli). The expectation is that these tools will work on Python 2.6 and without a proper warning and deprecation campaign there's not much we can do at the moment.

blueyed:
I see, thanks for the explanations so far.
btw: can you cancel https://travis-ci.org/boto/botocore/builds/353553219 - it seems like Travis got stuck on the `[ci skip]` (it should not have created a build for it in the first place).

Feel free to close this PR (but hopefully for some strange reason it will work now - it works locally at least, but you say that it is related to Travis only, right?), or include it in your branch.

blueyed:
Regarding the RecursionError in nose: are there plans to use pytest instead of node maybe?

joguSD:
I cancelled that build. Not sure why it got stuck.
We've discussed migrating to pytest and do use it on some of our other projects but there would be a non-trivial amount of effort to migrate our current test suite to be pytest compatible so it hasn't really been a priority.

joguSD:
For what it's worth I haven't been able to reproduce the recursion either, but I have seen it on Travis (and other CI systems). So I'm fairly confident there is a real issue somewhere in our tests or one of our dependencies. Most likely there's a rogue mock/patch on stdout messing everything up.

blueyed:
Well, Travis also does not like this one - will close/re-open.

blueyed:
Maybe https://github.com/boto/botocore/pull/1408 helps in this regard - it is unnecessary to wrap running the tests with shell=True.

joguSD:
I think there's just quite a few branches / PRs queued up on travis that it's taking a while to get to these new ones. Just give it some time and I think they'll run.

blueyed:
It works! :fireworks: 
https://travis-ci.org/boto/botocore/jobs/353572555#L476

fortran-favors-the-old:
+1 to relaxing the `python-dateutil` version restriction for Python 2.7+ (this was breaking our internal builds).  Thanks for the PR @blueyed!

prometheanfire:
yes, please relax this, openstack is not able to update dateutil because of this.

http://logs.openstack.org/06/552806/4/check/requirements-tox-py27-check-uc/4d47d21/job-output.txt.gz#_2018-03-16_15_35_57_233537

`2018-03-16 15:35:57.233537 | ubuntu-xenial | ContextualVersionConflict: (python-dateutil 2.7.0 (/home/zuul/src/git.openstack.org/openstack/requirements/.tox/py27-check-uc/lib/python2.7/site-packages), Requirement.parse('python-dateutil<2.7.0,>=2.1'), set(['botocore']))`

ryanpetrello:
👍, this is also preventing the Ansible AWX project from using the latest `boto` and the latest `python-dateutil`:

https://github.com/ansible/awx/pull/1598

setup.py
```diff
@@ -24,9 +24,12 @@ def find_version(*file_paths):
 
 
 requires = ['jmespath>=0.7.1,<1.0.0',
-            'python-dateutil>=2.1,<2.7.0',
             'docutils>=0.10']
-
+if sys.version_info <= (2, 7):
```
pganssle: I think this switch only works depending on what version of Python you run `setup.py` with, not the version that you're installing this for.

Shouldn't this be something more like:

```python
requires.extend(["python-dateutil>=2.1,<2.7.0;python_version == '2.6'",
                           "python-dateutil>=2.1,<3.0.0;python_version >= '2.7'"])
```

That way it will also work if you build a wheel on Python 3.6 and install it with python 2.6.

blueyed: @pganssle 
Good call.
But `sys.version_info` is also used below, which should be changed then, too.
IIRC this might require a certain version of setuptools etc (I think checking pytest's setup.py might be good).

pganssle: At `dateutil` we ended up doing [this](https://github.com/dateutil/dateutil/blob/master/setup.py#L15) to ensure a minimum requirement, but it might be possible to use `install_requires`.

prometheanfire:
anything left to do here before merge?

pganssle:
@prometheanfire I think the `sys.version_info` stuff needs to be fixed, this will only really work correctly for `sdist` installations.

pganssle:
That said, I'm still dubious as to whether this is necessary *at all*. It seems like a belt-and-suspenders approach. Updating `pip` will solve the problem in the vast majority of cases. Some people on old versions of Artifactory or other mirrors that don't respect `Requires-Python` may also be affected.

prometheanfire:
hmm, ya, just fixing requirements.txt to be the following should work

    python-dateutil>=2.1,<2.7.0;python_version=='2.6'
    python-dateutil>=2.1,<3.0.0;python_version>='2.7'

as an example, see https://github.com/openstack/requirements/blob/master/global-requirements.txt#L75

pganssle:
@prometheanfire What I'm saying that leaving the `requirements.txt` as:

```python
python-dateutil>=2.1,<3.0.0
```
Will *also* work, because `python-dateutil >= 2.7.0` has [this line in its `setup.py`](https://github.com/dateutil/dateutil/blob/master/setup.py#L62), so anyone doing `pip install botocore` on Python 2.6, so long as they are using `pip >= 8.0`, will already get `python-dateutil~=2.6`.

That said, your way is a "belt and suspenders" approach, in that it will also work for people using very old versions of `pip`.

prometheanfire:
True, I'd be happy with either method.  We (gentoo) have pip 7.1.2 available, but pip 9 has been stable for a long time.  The current situation is preventing OpenStack from consuming the latest botocore, so solving it in some way would be nice (and the requisite release).  I can submit another PR to do the requirements update unless someone's going to update this one.

jakul:
What needs to be done to get this merged?

blueyed:
@jakul 
I think this should not get merged as-is, see the previous comment (https://github.com/boto/botocore/pull/1406#issuecomment-377535983).
It's up to maintainers though, so let's just wait for them I guess?!

blueyed:
Closing.
Maintainers should be aware, and otherwise an issue should be created instead probably.
While this PR would work, https://github.com/boto/botocore/pull/1406#issuecomment-377535983 seems to be a better approach.
I could update the PR accordingly, but it's as easy for maintainers to do so directly probably.
This PR also already contains too much noise.

hugovk:
> I generally think, "update pip" is a reasonable thing to ask at this point, given that as more people drop compatibility for Python 2, `Requires-Python` is going to be critical.

Agreed, see https://github.com/boto/botocore/pull/1415 or https://github.com/boto/botocore/pull/1434 to add `Requires-Python` support in botocore.

PR #1409, opened 2018-03-16
Make redirects for S3 head bucket and object more strict

### LINKED ISSUE #1400: botocore 1.9.x doesn't raise ExpiredToken error
s3client.Object's `upload_file` method does no longer raise `ExpiredToken` error if the given token expires in botocore 1.9.x. I have this code and catch `ClientError` such that the upper stream can renew the token. This works fine before 1.9.0:
```
try
...
            akey = meta[u'client'].Object(s3location.bucket_name, s3path)
            akey.upload_file(
                data_file,
                Callback=meta[u'put_callback'](
                    data_file,
                    os.path.getsize(data_file),
                    output_stream=meta[u'put_callback_output_stream']) if
                        meta[u'put_callback'] else None,
                ExtraArgs={
                    u'Metadata': s3_metadata,
                    u'ContentEncoding': u'gzip',
                },
                Config=TransferConfig(
                    multipart_threshold=SnowflakeS3Util.DATA_SIZE_THRESHOLD,
                    max_concurrency=max_concurrency,
                    num_download_attempts=10,
                )
            )
...
        except botocore.exceptions.ClientError as err:
            if err.response[u'Error'][u'Code'] == 'ExpiredToken':
                logger.debug(u"AWS Token expired. Renew and retry")
                meta[u'result_status'] = ResultStatus.RENEW_TOKEN
                return
```
But the new botocore 1.9.0+ returns HTTP 400 but doesn't raise `ExpiredToken` error code.

Is there any way to detect the token expiration?

joguSD:
Attempting to redirect 400s when performing a head bucket or object is too generic and can cause infinite retries in some circumstances (bad credentials, incorrect SSE-C parameters). This updates the logic to only retry a 400 in the case that the request was previously redirected by `requests` upon receiving a `307` from the global s3 endpoint.

Closes #1400, boto/boto3#1482

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -891,13 +891,21 @@ class S3RegionRedirector(object):
         error = response[1].get('Error', {})
         error_code = error.get('Code')
 
+        is_special_head_object = (
+            error_code == '301' and
+            operation.name in ['HeadObject', 'HeadBucket']
+        )
         # We have to account for 400 responses because
         # if we sign a Head* request with the wrong region,
+        # and it gets redirected by a temporary redirect
         # we'll get a 400 Bad Request but we won't get a
         # body saying it's an "AuthorizationHeaderMalformed".
-        is_special_head_object = (
-            error_code in ['301', '400'] and
-            operation.name in ['HeadObject', 'HeadBucket']
+        is_special_400_head_object = (
+            error_code == '400' and
+            operation.name in ['HeadObject', 'HeadBucket'] and
+            response[0].history and
+            response[0].history[-1].status_code == 307 and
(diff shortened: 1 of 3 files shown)


botocore/utils.py
```diff
@@ -932,6 +940,9 @@ def redirect_from_error(self, request_dict, response, operation, **kwargs):
             'bucket': bucket,
             'endpoint': endpoint
         }
+        # if the signing context doesn't change, don't retry
+        if request_dict['context']['signing'] == signing_context:
```
jamesls: Can you add a debug log here so we know when this happens?

PR #1419, opened 2018-03-23
Prevent infinite loops when handling S3 region redirects

jamesls:
This adds an alternate fix for https://github.com/boto/botocore/pull/1409 .  I've pulled in @joguSD's patch and added additional unit/integration tests.

The fix is to set a flag when we redirect the first time and verify we haven't previously redirected on subsequent requests.  I also added an integration test for the specific case mentioned in https://github.com/boto/botocore/issues/1400.

As for the regression caused from #1409, we need buckets with DNS propagated in order to actually test this, but I did write a one off integration test to verify this works as expected.  I don't think it makes sense to include in our main integration test suite, I think the unit/functional tests suffice.  If you're curious though, here's the test: https://gist.github.com/jamesls/9e1f0eb8e6b632f62ea5cb38131de2ae

### DIFF
--- a/.changes/next-release/bugfix-s3-85294.json
+++ b/.changes/next-release/bugfix-s3-85294.json
@@ -1,0 +1,5 @@
+{
+  "category": "``s3``", 
+  "type": "bugfix", 
+  "description": "Fix bug where invalidate head_object requests would cause an infinite loop (alternate fix to `#1400 <https://github.com/boto/botocore/issues/1400>`__)"
+}
--- a/.changes/next-release/bugfix-s3-85294.json
+++ b/.changes/next-release/bugfix-s3-85294.json
@@ -1,5 +1,5 @@
 {
   "category": "``s3``", 
   "type": "bugfix", 
-  "description": "Fix bug where invalidate head_object requests would cause an infinite loop (alternate fix to `#1400 <https://github.com/boto/botocore/issues/1400>`__)"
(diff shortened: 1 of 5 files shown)


.changes/next-release/bugfix-s3-85294.json
```diff
@@ -0,0 +1,5 @@
+{
+  "category": "``s3``", 
+  "type": "bugfix", 
+  "description": "Fix bug where invalidate head_object requests would cause an infinite loop (alternate fix to `#1400 <https://github.com/boto/botocore/issues/1400>`__)"
```
joguSD: invalid*

tests/functional/test_s3.py
```diff
@@ -339,6 +339,51 @@ def test_resign_request_with_region_when_needed(self):
                      '?encoding-type=url')
         self.assertEqual(calls[1].url, fixed_url)
 
+    def test_resign_request_in_us_east_1(self):
```
kyleknap: It might be good to have the other case when the initial `bad_request_response` should resolve to a 400 on the user side (i.e. bad SSE-C key case). It looks like there is a integration test for this, but it would be nice if there was a functional or unit test as well.

jamesls: Added a test for this case in 68b3997a2b9867ad51a5a590ba511d36f231e149

PR #1424, opened 2018-04-05
Add support for cliArgumentName in the member metadata

dstufft:
Adds a member value ``cliArgumentName`` that can be used to override what the automatically generated name on the CLI is.

This includes validation that the property is of the right type (although I decided not to try to validate that the ``cliArgumentName`` doesn't exist anywhere else, because I didn't see a great way of doing that besides validating the entire schema of a service model (and even then, it would block that name from being used as a member name itself, or a model name).

The tests also include validation that the cliArgumentName given does not collide with any other members on the same model, it contains the original argument name, and that within a single service, the same cliArgumentName is used for the same original name consistently (so that you don't have ``Region`` being renamed to ``--foo-region`` in one command, and ``--bar-region`` in another command for the same service).

### DIFF
--- a/tests/functional/test_cliargumentname_config.py
+++ b/tests/functional/test_cliargumentname_config.py
@@ -1,0 +1,94 @@
+# Copyright 2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from jsonschema import Draft4Validator
+
+import botocore.session
+from botocore import xform_name
+
+
+SERVICE_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "shapes": {
+            "type": "object",
+            "patternProperties": {
+                ".+": {
+                    "type": "object",
+                    "properties": {
+                        "members": {
+                            "type": "object",
+                            "patternProperties": {
+                                ".+": {
+                                    "type": "object",
+                                    "properties": {
+                                        "cliArgumentName": {"type": "string"},
+                                    }
+                                }
+                            }
+                        }
+                    }
+                },
+            },
+        }
+    },
+}
+
+
+def test_lint_cliargmentname_configs():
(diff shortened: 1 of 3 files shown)


tests/functional/test_cliargumentname_config.py
```diff
@@ -0,0 +1,94 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from jsonschema import Draft4Validator
+
+import botocore.session
+from botocore import xform_name
+
+
+SERVICE_SCHEMA = {
```
kyleknap: Just realizing this, it may make sense to hoist this to the CLI functional tests (depending on what direction we go as it relates to the file the cliArgumentRenames). Mainly if the plan is to put these in the `cli-extras.json`, botocore will not automatically load the up because botocore will only look for `sdk-extras.json` and therefore from it's standpoint, it will never see the `cliArgumentName`. If it were in the `cli-extras.json` (assuming we add support for that extension in the CLI), the CLI will be able to pick up the cliArgumentRenames in a CLI driver's session.

However if the values are going to be part of the `service-2.json` and not in the `extras.json`, this will not be a problem and the tests can live in botocore or the CLI. But if we put it in the CLI functional tests, it will not matter if the final resting place is the `cli-extra.json` or `service-2.json`. It will be loaded no matter what for the tests.

Let me know what you think.

tests/functional/test_cliargumentname_config.py
```diff
@@ -0,0 +1,94 @@
...
+    assert original_cli_name in member_shape["cliArgumentName"]
+
+
+def _validate_schema(validator, waiter_json):
+    errors = list(e.message for e in validator.iter_errors(waiter_json))
+    if errors:
+        raise AssertionError('\n'.join(errors))
+
+
+def _lint_consistent_renames(renames):
```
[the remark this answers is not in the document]
dstufft: Yea, I was on the fence about this, and left it in because removing it is easier then adding it back in later since going from lax -> strict means you have to fix any errors that snuck in by accident.

tests/functional/test_cliargumentname_config.py
dstufft: I'm happy having these tests live in either place, and I agree that if we put them in ``cli-extras.json`` we probably don't want the SDK to be loading them by default. That does mean this is blocked on that decision, unless we commit to here for now, and if need be hoist it out later (but we may want to hold off on merging this until we have a decision on that anyways).

kyleknap: Yeah I think waiting to make changes/merge this PR until a decision is made makes sense to me.

PR #1425, opened 2018-04-06
Vpcendpoint paginator

rocky4570:
DescribeVpcEndpoints to be paginate-able. Add some form of tests, Tested against AWS to a degree as dont have lot of VPCs nor with endpoints. This fix is to satisfy https://github.com/boto/boto3/pull/1504 which is requiring client call to describe_endpoints to be paginate-able.

### DIFF
--- a/botocore/data/ec2/2016-11-15/paginators-1.json
+++ b/botocore/data/ec2/2016-11-15/paginators-1.json
@@ -77,6 +77,12 @@
       "limit_key": "MaxResults",
       "result_key": "Volumes"
     },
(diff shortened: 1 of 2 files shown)


copumpkin:
Ping? I'd love to see this get in, and it doesn't appear to have anything actually wrong with it

JordonPhillips:
We ended up going with @copumpkin 's PR. While I seriously admire the drive to add tests, they aren't actually needed in this case since we have generated tests for all paginators that pretty exhaustively cover them.

Issue #1429, opened 2018-04-10
Relax upper bound on python-dateutil


pganssle:
PR #1402 was over-zealous in tightening the upper bound on `python-dateutil`. As you can see in that thread and in #1406, this is causing problems for a lot of people, apparently.

As [my comment on 1406 indicates](https://github.com/boto/botocore/pull/1406#issuecomment-377535983), I think that it *should* be fine to just revert #1402 because `python-dateutil` declares `Requires-Python`. That said, the "belt and suspenders" approach for people stuck with a version of `pip` that does not support `Requires-Python` would be to add conditional requirements in [`requires`](https://github.com/boto/botocore/blob/develop/setup.py#L26):

```python
requires = ['jmespath>=0.7.1,<1.0.0',
            "python-dateutil>=2.1,<2.7.0; python_version == '2.6'",
            "python-dateutil>=2.1,<3.0.0; python_version >= '2.7'",
            'docutils>=0.10']
```

While you're at it, I think using an `if` statement on `sys.version_info` in `setup.py` is *also* the wrong thing to do here, because the version that executes `setup.py` is often but not always the same version that the package is being installed for:

```python
requires = ['jmespath>=0.7.1,<1.0.0',
            'docutils>=0.10',
            # python-dateutil dropped support for Python 2.6 in version 2.7.0
            'python-dateutil>=2.1,<2.7.0; python_version == "2.6"',
            'python-dateutil>=2.1,<3.0.0; python_version >= "2.7"',
            # For python2.6 we have a few other dependencies.
            # First we need an ordered dictionary so we use the
            # 2.6 backport.
            'ordereddict==1.1; python_version == "2.6"',
            # Then we need simplejson.  This is because we need
            # a json version that allows us to specify we want to
            # use an ordereddict instead of a normal dict for the
            # JSON objects.  The 2.7 json module has this.  For 2.6
            # we need simplejson.
            'simplejson==3.3.0; python_version== "2.6"'
            ]
```

prometheanfire:
The proposal looks fine, I'd just remove the spaces but that's an optional thing according to pep-508.

prometheanfire:
Is there an associated PR for this?

pganssle:
I'm just making general suggestions on how to solve the problem. Anyone is welcome to make a PR based on those suggestions.

prometheanfire:
ok, pr incoming

PR #1430, opened 2018-04-10
Add python_version specifiers to required packages

prometheanfire:
This patch adds version specifiers to packages requiring them for old
versions of python.  It also relaxes the version of python-dateutil
using the same method.

Closes: https://github.com/boto/botocore/issues/1429


setup.py
```diff
@@ -23,23 +23,14 @@ def find_version(*file_paths):
...
-    # Then we need simplejson.  This is because we need
-    # a json version that allows us to specify we want to
-    # use an ordereddict instead of a normal dict for the
-    # JSON objects.  The 2.7 json module has this.  For 2.6
-    # we need simplejson.
-    requires.append('simplejson==3.3.0')
-
+# The python_version specifiers are necessary for supporting older
+# versions of python.
+requires = 'docutils>=0.10'
```
pganssle: This needs to be an actual list, right now it's missing the `[` and `]`.

Also the first element needs a comma after it. if you added the square brackets right now, the first two entries would concatenate.

setup.py
```diff
@@ -23,23 +23,14 @@ def find_version(*file_paths):
...
+           'jmespath>=0.7.1,<1.0.0',
+           'ordereddict==1.1;python_version=="2.6"',
```
pganssle: It might be a good idea to preserve the comments.

setup.py
prometheanfire: ya, hand jamming doesn't always work the best (especially when distracted).

setup.py
prometheanfire: I can readd it, sure

pganssle:
Apparently whatever stack `boto` is using for Python 3.3 doesn't support this feature? I'm kinda surprised because `pip` on Python 3 evidently supports it, since `pip -r requirements.txt` works just fine in `dateutil`'s test suite.

This may be a specific problem with `wheel` on Python 3.3, possibly this feature was added after `wheel` dropped support for Python 3.3?

One albeit annoying way to handle this is to go back to using a `sys.version` conditional, but conditional only on Python 3.3. That way people will only have a problem if they use Python 2.6 to install a wheel created using Python 3.3, which is fairly unlikely.

```python
requires = [
    'docutils>=0.10',
    'jmespath>=0.7.1,<1.0.0',
]

# Python 3.3 was failing to build the wheels with python_version specifiers
if sys.version_info[:2] == (3, 3):
    requires += ['python-dateutil>=2.1,<3.0.0']
else:
    requires += [
        # the latest version that supports python2.6 is 2.6.*
        'python-dateutil>=2.1,<2.7.0;python_version=="2.6"',
        'python-dateutil>=2.1,<3.0.0;python_version>="2.7"',
        # required for ordered dictionary support in python2.6
        'ordereddict==1.1;python_version=="2.6"',
        # requrired for a json version that allows specification of
        # ordereddict use instead of a normal dict for json objects
        # in python2.6
        'simplejson==3.3.0;python_version=="2.6"'
    ]
```

pganssle:
Also you need to update [`requirements.txt`](https://github.com/boto/botocore/blob/develop/requirements.txt).

prometheanfire:
does botocore still support python3.3?  it was EOL'd  https://www.python.org/dev/peps/pep-0398/#x-end-of-life

pganssle:
I don't know anything about botocore, but Python 2.6 is also EOLed and both 2.6 and 3.3 are in the test suite, so I would guess that both are supported.

prometheanfire:
If I get confirmation on 3.3 I'll implement the if/else workaround, otherwise I'd like to avoid it.

pganssle:
How many places are these dependencies specified? Jesus, you also need to update [setup.cfg](https://github.com/boto/botocore/blob/develop/setup.cfg#L5).

Also, I do suspect you'll have an easier time getting this reviewed and merged if the test suite passes.

pganssle:
FWIW I can't replicate this failure on Python 3.3. It's likely that explicitly upgrading some package or another in the CI will fix this better than the if/else.

pganssle:
See #1432 for why this is failing on Python 3.6

prometheanfire:
@pganssle that should probably be part of a separate pull request (imo).

pganssle:
The other problem can be solved by adding `setuptools>=21.0` to `requirements.txt`. I think the default installed version on the Python 3.3 virtualenv is too old to handle these kinds of strings.

> @pganssle that should probably be part of a separate pull request (imo).

Possibly, but I assume this PR is blocked on that one.

prometheanfire:
Yep, I'd agree with that.  that setuptools version is only needed for py3.3 right?

pganssle:
It's needed for everything, but Python 3.3 is the only one where the installed version is lower than that.

setup.py
```diff
@@ -23,23 +23,18 @@ def find_version(*file_paths):
...
-    # 2.6 backport.
-    requires.append('ordereddict==1.1')
-    # Then we need simplejson.  This is because we need
-    # a json version that allows us to specify we want to
-    # use an ordereddict instead of a normal dict for the
-    # JSON objects.  The 2.7 json module has this.  For 2.6
-    # we need simplejson.
-    requires.append('simplejson==3.3.0')
-
+requires = [ 'setuptools>=21.0',
```
pganssle: This actually only needs to go in `requirements.txt` I believe. By the time you pass it to `install_requires` it's too late (since the old version of `setuptools` is already imported). I think the same goes for `setup.cfg`.

After PEP 518, this will go in `pyproject.toml`.

pganssle:
Ping @jamesls

Issue #1432, opened 2018-04-10
Test suite breaks on Python 3.6 with python-dateutil>=2.7.0


pganssle:
It is unclear why this is happening. An example is [here](https://travis-ci.org/boto/botocore/jobs/364800800).

I briefly took a look at it, I'm not really sure why it would make a difference whether that's on Python 3.5 or Python 3.6. The biggest difference in `dateutil` is that 3.6+ has the `datetime.fold` attribute, which means that `tz.enfold` has a slightly different behavior between the two. The failing test *does* have a time zone, but it is not an ambiguous time.

lbolla:
From the failure linked above, it looks like this should fail with `python-dateutil==2.7.2`, but it doesn't:
```
In [1]: import dateutil.parser
In [2]: dateutil.parser.parse('Wed May 20 13:51:10 PDT 2015')
   ...: 
/<redacted>/python2.7/site-packages/dateutil/parser/_parser.py:1165: UnknownTimezoneWarning: tzname PDT identified but not understood.  Pass `tzinfos` argument in order to correctly return a timezone-aware datetime.  In a future version, this raise an exception.
  category=UnknownTimezoneWarning)
Out[2]: datetime.datetime(2015, 5, 20, 13, 51, 10)
```

Instead, it raises a warning: maybe nosetests or something else fail with the above `maximum recursion exceeded` error because of that?

pganssle:
@lbolla That change isn't specific to Python 3.6, but I'll note that if the test suite is being run on a computer in Pacific time, that warning will not be raised because `PDT` will be recognized as a local time. It's also possible to call it like this:

```python
from dateutil.tz import gettz
from dateutil.parser import parse
parse('Wed May 20 13:51:10 PDT 2015', tzinfos={'PDT': gettz('America/Los_Angeles'), 'PST': gettz('America/Los_Angeles')})
```

Which will also succeed just fine.

pganssle:
Did some digging, this will only fail if you use `--with-xunit`.

Irritatingly:

`nosetests -x --pdb --with-xunit -s -v functional/`

Does not seem to stop on the failing test or set a breakpoint.

pganssle:
OK, so it's not 100% clear exactly *why* this is throwing an error and why only on Python 3.6, but I think that some combination of xunit and some Python 3.6 feature and the nested way that particular test is written is causing the warning that @lbolla mentioned to throw an error.

[...]

PR #1433, opened 2018-04-11
Allow newer versions of dateutil

### LINKED ISSUE #1429: Relax upper bound on python-dateutil

joguSD:
This updates dateutil to use version specific requirements and also updates nose to a version that fixes the stack limit bug on Python 3.6 (see [this issue](https://github.com/nose-devs/nose/issues/749)).

This also requires that we make changes to the CLI: https://github.com/aws/aws-cli/pull/3197
These must be merged at the same time.

Closes #1429, #1432, #1430

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 tox>=2.5.0,<3.0.0
 python-dateutil>=2.1,<2.7.0; python_version=="2.6"
 python-dateutil>=2.1,<3.0.0; python_version>="2.7"
-nose==1.3.0
+nose==1.3.7
 mock==1.3.0
 wheel==0.24.0
 docutils>=0.10
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 tox>=2.5.0,<3.0.0
-python-dateutil>=2.1,<2.7.0
+python-dateutil>=2.1,<2.7.0; python_version=="2.6"
+python-dateutil>=2.1,<3.0.0; python_version>="2.7"
 nose==1.3.0
 mock==1.3.0
 wheel==0.24.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,8 @@ universal = 1
 
 [metadata]
 requires-dist =
-	python-dateutil>=2.1,<2.7.0
+	python-dateutil>=2.1,<2.7.0; python_version=="2.6"
+	python-dateutil>=2.1,<3.0.0; python_version>="2.7"
 	jmespath>=0.7.1,<1.0.0
 	docutils>=0.10
 	ordereddict==1.1; python_version=="2.6"
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,6 @@ def find_version(*file_paths):
 
 
 requires = ['jmespath>=0.7.1,<1.0.0',
-            'python-dateutil>=2.1,<2.7.0',
             'docutils>=0.10']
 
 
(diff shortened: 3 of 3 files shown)


kyleknap:
Actually I just saw the linked PR. Can we just include that commit in this PR so we have all of these changes under a single PR?

dstufft:
@kyleknap They're different repositories.

kyleknap:
@dstufft Right. Nevermind about my comment then.

joguSD:
Not sure what's going on with the coverage. Might be related to upgrading nose.

lbolla:
Any chance this can be merged and released?

joguSD:
@lbolla We haven't forgotten about this. We're doing a little bit more verification on our side before we merge as this affects our release process.

prometheanfire:
I assume https://github.com/boto/botocore/pull/1430 can be closed now?

pganssle:
Why would you merge this one instead of #1430? This one fixes this the wrong way...

pganssle:
@joguSD Note that the change to the test in #1430 changes something that will eventually become an error, because the time zone abbreviation changed there is being silently elided on any executor without its time zone set to Pacific time.

joguSD:
@prometheanfire Yep. Unfortunately we can't merge changes to files in the data directory here directly, we get them from upstream. 

@pganssle This PR is what https://github.com/boto/botocore/pull/1402 looked like originally, minus the update to nose (which was causing the failures on 3.6). Switching `setup.py` to also use markers instead of the conditional can break customers using older versions of `pip`.

pganssle:
@joguSD Those customers always have the option of manually downgrading their version of `dateutil`. If they install `dateutil` beforehand, it will not be auto-upgraded by this. All versions of Python that `botocore` supports support a version of `pip` that respects `python_requires`. *And* this will break if `setup.py` is run with a different version of Python than the version for which this is installed.

Seems like very much the wrong way to solve this problem since the correct solution is to have people upgrade their version of `pip`.

joguSD:
@pganssle For some customers upgrading `pip` isn't always an option. We decided not to use env markers in the `setup.py` because it already currently uses a conditional. We're trying to introduce as little change as possible to support this.

joguSD:
Unfortunately, we did run into some issues with this PR in our release process. We're going to revert this temporarily until we get our release process updated to handle the new version of dateutil. We apologize for any inconvenience.

prometheanfire:
can you link us to the issue so we can track it?

joguSD:
@prometheanfire We put these changes back in and they should be going out in the next release.

prometheanfire:
good to hear :D

PR #1415, opened 2018-03-22
Add python_requires to help pip

hugovk:
When botocore drops old Python versions, this will help pip install the right version for people still running those old Python versions.
 
For more info on how this works:
* https://hackernoon.com/phasing-out-python-runtimes-gracefully-956f112f33c4
* https://github.com/pypa/python-packaging-user-guide/issues/450

PR #1434, opened 2018-04-12
Pass python_requires argument to setuptools

jdufresne:
Helps pip decide what version of the library to install.

https://packaging.python.org/tutorials/distributing-packages/#python-requires

> If your project only runs on certain Python versions, setting the
> python_requires argument to the appropriate PEP 440 version specifier
> string will prevent pip from installing the project on other Python
> versions.

https://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords

> python_requires
>
> A string corresponding to a version specifier (as defined in PEP 440)
> for the Python version, used to specify the Requires-Python defined in
> PEP 345.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,6 +54,7 @@ setup(
                   'botocore.vendored.requests': ['*.pem']},
     include_package_data=True,
     install_requires=requires,
+    python_requires=">=2.6, !=3.0.*, !=3.1.*, !=3.2.*",
     extras_require={
         ':python_version=="2.6"': [
             'ordereddict==1.1',


acranox:
From what I understand, python_requires is not supported in the version of setuptools that is available in the distro repos for Centos 7.x, so this will cause failures when trying to install botocore on that platform, if you're using the distro supplied setuptools.

jdufresne:
setuptools/distutils does not error on unknown arguments, it only warns. So older versions will continue to work, they just won't be able to take advantage of the `python_requires` feature. For example:

```
>>> from setuptools import setup
>>> setup(foo='bar')
/usr/lib64/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'foo'
  warnings.warn(msg)
```

acranox:
I think I figured it out: https://github.com/pypa/setuptools_scm/commit/96f68bbaa44e9853291ab356a202319fbd906cc4


three levels deep of dependency problems.  argh.

jdufresne:
I'm closing as this is a duplicate of #1415 which came first. Sorry, didn't see that originally.

PR #1441, opened 2018-04-19
define requirements using environment markers

### LINKED ISSUE #1440: universal wheel is built with python 2.6-only version pinning on python-dateutil
Someone built the universal wheel for 1.10.4 using Python 2.6. This has caused the universal wheel on PyPI to include the following `Requires-Dist: python-dateutil>=2.1,<2.7.0` instead of the more relaxed `Requires-Dist: python-dateutil>=2.1,<3.0`.

```
botocore 1.10.4 has requirement python-dateutil<2.7.0,>=2.1, but you'll have python-dateutil 2.7.2 which is incompatible.
```

The `setup.py` needs to have version pins using environment markers (similar to the already-defined and redundant ordereddict and simplejson deps in the extras) instead of imperatively defined in the code in order for the metadata to get into the wheel correctly.

mmerickel:
fixes #1440

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,6 @@
 import codecs
 import os.path
 import re
-import sys
 
 from setuptools import setup, find_packages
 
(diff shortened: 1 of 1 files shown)


dstufft:
Closing due to the comment in https://github.com/boto/botocore/issues/1440#issuecomment-382823580, thanks again!

PR #1442, opened 2018-04-19
Revert "Allow newer versions of dateutil"

joguSD:
Reverts boto/botocore#1433


joguSD:
@jhoelzl We ran into some release time issues that required us to back out on this change to get the release out. We've resolved those issues and put this change back in.

PR #1443, opened 2018-04-20
Revert "Revert "Allow newer versions of dateutil""

joguSD:
Reverts boto/botocore#1442


acranox:
python-dateutil 2.7.x uses python_requires, which requires setuptools > 24.3
Centos 7 apparently only includes setuptools 0.9.8, so this requirements change causes failures to install in Centos 7.

Going back to python-dateutil 2.6.1 allows botocore 1.10.6 to be installed on centos 7 with the system setuptools.

PR #1449, opened 2018-04-24
Add a history recorder scope

kyleknap:
The abstraction allows for the registering of handlers to a scope so that if a client is also registered to the same scope, the handler added to that scope will only receive events from all clients registered to that scope. In this commit, doc strings were added to the history module as well to clarify functionality. The functional tests  would be good to start with to help understand the in's and out's of the interface, but for a general overview of how you go about using this here is an example:
```python
import botocore.session
from botocore.history import get_global_history_recorder
from botocore.history import BaseHistoryHandler
from botocore.history import HistoryRecorderScope


class PrintHistoryHandler(BaseHistoryHandler):
    def emit(self, event_type, payload, source):
        if event_type == 'HTTP_REQUEST':
            payload = payload['url']
        elif event_type == 'HTTP_RESPONSE':
            payload = payload['headers']
        print('%s: %s' % (event_type, payload))


session = botocore.session.Session()
s3_client = session.create_client('s3')
ec2_client = session.create_client('ec2')


history_scope = HistoryRecorderScope()
scoped_handler = PrintHistoryHandler()
history_scope.add_handler(scoped_handler)
history_scope.register_client(ec2_client)

history_recorder = get_global_history_recorder()
history_recorder.add_scope(history_scope)
history_recorder.enable()

s3_client.list_buckets()
ec2_client.describe_regions()
```
That will print out (Note that there are no S3 events in the output):
```
API_CALL: {'operation': u'DescribeRegions', 'params': {}, 'service': 'ec2'}
HTTP_REQUEST: https://ec2.us-west-2.amazonaws.com/
HTTP_RESPONSE: {'date': 'Tue, 24 Apr 2018 22:44:42 GMT', 'vary': 'Accept-Encoding', 'content-length': '2720', 'content-type': 'text/xml;charset=UTF-8', 'server': 'AmazonEC2'}
PARSED_RESPONSE: {u'Regions': [{u'Endpoint': 'ec2.ap-south-1.amazonaws.com', u'RegionName': 'ap-south-1'}, {u'Endpoint': 'ec2.eu-west-3.amazonaws.com', u'RegionName': 'eu-west-3'}, {u'Endpoint': 'ec2.eu-west-2.amazonaws.com', u'RegionName': 'eu-west-2'}, {u'Endpoint': 'ec2.eu-west-1.amazonaws.com', u'RegionName': 'eu-west-1'}, {u'Endpoint': 'ec2.ap-northeast-3.amazonaws.com', u'RegionName': 'ap-northeast-3'}, {u'Endpoint': 'ec2.ap-northeast-2.amazonaws.com', u'RegionName': 'ap-northeast-2'}, {u'Endpoint': 'ec2.ap-northeast-1.amazonaws.com', u'RegionName': 'ap-northeast-1'}, {u'Endpoint': 'ec2.sa-east-1.amazonaws.com', u'RegionName': 'sa-east-1'}, {u'Endpoint': 'ec2.ca-central-1.amazonaws.com', u'RegionName': 'ca-central-1'}, {u'Endpoint': 'ec2.ap-southeast-1.amazonaws.com', u'RegionName': 'ap-southeast-1'}, {u'Endpoint': 'ec2.ap-southeast-2.amazonaws.com', u'RegionName': 'ap-southeast-2'}, {u'Endpoint': 'ec2.eu-central-1.amazonaws.com', u'RegionName': 'eu-central-1'}, {u'Endpoint': 'ec2.us-east-1.amazonaws.com', u'RegionName': 'us-east-1'}, {u'Endpoint': 'ec2.us-east-2.amazonaws.com', u'RegionName': 'us-east-2'}, {u'Endpoint': 'ec2.us-west-1.amazonaws.com', u'RegionName': 'us-west-1'}, {u'Endpoint': 'ec2.us-west-2.amazonaws.com', u'RegionName': 'us-west-2'}], 'ResponseMetadata': {'HTTPStatusCode': 200, 'RequestId': 'c2759f73-f7b9-414a-9c48-c2f12e8dd535', 'HTTPHeaders': {'date': 'Tue, 24 Apr 2018 22:44:42 GMT', 'vary': 'Accept-Encoding', 'content-length': '2720', 'content-type': 'text/xml;charset=UTF-8', 'server': 'AmazonEC2'}}}
```

The one thing that I feel like there may be a better interface for (but have not figured out and are open for ideas on) is the `scope_id` interface. Essentially the `scope_id` interface gives us the ability for `HistoryRecorder.record()` to emit identifiers to help the scope and its handlers determine if the recorded event is for them. Currently, I am using the `id(client)` to generate the `scope_id` that gets passed into the `HistoryRecorder.record()` call. This allows `botocore` to pass an identifier of the client to these globally emitted events without having to keep reference of the actual client objects, but it does open a couple of potential issues:

1) For each event that we want to be scopeable to a client, it must have access to the id of the client. To do this, the id is stored in the request context (which most abstractions by now have access to).

2) If the client gets deleted, there is the potential that another client gets created under the same id. This would cause the newly created client to start emitting events to that scope without actually have registering to it. A few ideas to fix this include:
    * Include the class name passed to the `scope_id` to better avoid collisions
    * Expose a `deregister_client` so that you can at least remove the client from the scope before it getting deleted.
    * Expose a `client_id` on `client.meta` that returns a uid. Thinking about this now this may be the best option to avoid collisions, but would like to get your opinions on it first.

Let me know what you all think.

### DIFF
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -568,11 +568,12 @@ class BaseClient(object):
     def _make_api_call(self, operation_name, api_params):
         operation_model = self._service_model.operation_model(operation_name)
         service_name = self._service_model.service_name
+        client_id = id(self)
         history_recorder.record('API_CALL', {
             'service': service_name,
             'operation': operation_name,
             'params': api_params,
-        })
+        }, scope_id=client_id)
         if operation_model.deprecated:
             logger.debug('Warning: %s.%s() is deprecated',
                          service_name, operation_name)
@@ -581,6 +582,7 @@ class BaseClient(object):
             'client_config': self.meta.config,
             'has_streaming_input': operation_model.has_streaming_input,
             'auth_type': operation_model.auth_type,
+            'client_id': client_id,
         }
         request_dict = self._convert_to_request_dict(
             api_params, operation_model, context=request_context)
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -203,6 +203,7 @@ class Endpoint(object):
         # (http_response, parsed_dict).
         # If an exception occurs then the success_response is None.
         # If no exception occurs then exception is None.
+        client_id = request.original.context.get('client_id')
         try:
             logger.debug("Sending http request: %s", request)
             history_recorder.record('HTTP_REQUEST', {
@@ -211,7 +212,7 @@ class Endpoint(object):
                 'streaming': operation_model.has_streaming_input,
                 'url': request.url,
                 'body': request.body
-            })
+            }, scope_id=client_id)
             streaming = any([
                 operation_model.has_streaming_output,
                 operation_model.has_event_stream_output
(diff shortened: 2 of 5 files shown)


kyleknap:
Closing this out. I have been playing around with and there are significant issues that boil down to the history recorder being global. Specifically a client's id can be reused if it gets deleted, this can cause a new client that is not registered to the scope to be called. Also since the history recorder object is global, there will always be some amount of memory leak (even if we just use client id's). Honestly purely because the history recorder is global, we should avoid having features rely on it.

Issue #82, opened 2013-07-19
Can't put contents to s3 bucket outside of US standard region with file object


jamesls:
When you use a file like object as the body param to s3.PutObject for a bucket outside of US standard, you'll get a 307 followed by a 400 bad request and eventually a socket timeout.  To repro:

1) Create a bucket outside of us standard.
2) Send a PutObject request with a file like object for the body: `op.call(endpoint, body=open('/some/file', 'rb'), ...)`

You'll see a 307 then a 400.  I think this is because requests is configured to follow redirects, but there's no location header and hence the bad request.

I think the fact that this is a file like object exposes the bug because we send two packets, one for the request+headers then one for the body.  When the body is a string (like what we use in the integration tests), then httplib will do a `msg += message_body` and send it as a single chunk which succeeds.

This also might be related to the virtual host handler we have.  I noticed this in the log messages:

```
botocore.handlers: DEBUG: Checking for DNS compatible bucket for: https://s3-us-west-2.amazonaws.com/botocoretest1374195914-317
botocore.handlers: DEBUG: URI updated to: https://botocoretest1374195914-317.s3.amazonaws.com
```

Would be good to get an integration test written for this also.

Issue #1122, opened 2017-01-06
Support for requests>=2.10.0?


shawnrusaw-wf:
Is there a timeline/plan to sync `botocore`'s vendored `requests` module to a more recent version? `2.10.0` is able to support making calls from Google App Engine via `requests_toolbelt` (https://toolbelt.readthedocs.io/en/latest/adapters.html#appengineadapter).

JordonPhillips:
Upgrading requests is generally a high risk, low reward task. So if we aren't directly benefiting from a new feature or bug fix, it generally isn't worth the risk. However, `2.10` did add the ability to use SOCKS proxies, which is an [open feature request](https://github.com/boto/botocore/issues/880).

Even then, we likely won't be able to update past `2.11.1` because they added a feature to attempt to rewind file-like objects on certain kinds of errors, which is something we implemented ourselves some time ago. Balancing those conflicting implementations will be very risky.

Deewiant:
I just ran into an issue with the outdated `requests` library. I was wondering why some `aws s3 cp` invocations were failing with the inscrutable `fatal error: [Errno 21] Is a directory`. Turns out that having set `REQUESTS_CA_BUNDLE=/etc/ca-certificates/extracted/cadir` broke `botocore` because support for setting `REQUESTS_CA_BUNDLE` to a directory came only in 2.9.0, and `botocore` is still using 2.7.0 though my installed version of `requests` is 2.18.1. This was a highly unexpected interaction to me, and it'd be great to get at least this issue fixed.

kirkhansen:
I've also recently run into the outdated requests library issue, but mine was for the socks5 proxy support I was expecting to work. It'd be great to get it updated to at least 2.10.

andrey-utkin:
> Upgrading requests is generally a high risk, low reward task.

> Balancing those conflicting implementations will be very risky.

This is exactly why it is rational to stick with upstream and work on merging your enhancements to upstream codebase :) Vendoring, you have to forward-port your patches yourself routinely anyway *and* listen to users' moaning about outdated version of vendored dependencies :)

As a person contributing to availability of botocore in Gentoo Linux, I am concerned with it bundling dependencies and not using system-provided packages. This is frowned upon in general as in such case user doesn't have same visibility of obsolete or vulnerable software being installed in their system.

With above rant I intend to gently ask you to consider un-vendoring vendored deps.

amilstead:
[...]

Issue #805, opened 2016-02-23
Clients are never GC'd


magcius:
```python
def __del__(obj):
    print "deleting", obj

class foo(object):
    pass

def main():
    import botocore.session
    session = botocore.session.get_session()
    session.__class__.__del__ = __del__
    client = session.create_client('s3', region_name='us-west-2')
    client.__class__.__del__ = __del__

    thing = foo()
    thing.__class__.__del__ = __del__

print "before"
main()
print "after"
```

Prints:

```
before
deleting <botocore.session.Session object at 0x7fd23c7ec550>
deleting <__main__.foo object at 0x7fd23c7a4110>
after
```

This is causing major memory leaks in our applications that create lots of clients.

jamesls:
Taking a look.  We have specific tests in place to ensure we don't leak memory when creating clients: https://github.com/boto/botocore/blob/develop/tests/functional/leak/test_resource_leaks.py#L39-L65

One of the things I've noticed from your code is that you're assigning a `__del__` to a client, which in this scenario will prevent the object from being gc'd because of the cycles created within a client.  This isn't specific to botocore, this applies to any python class:  For example:

```
import gc


class ObjectWithDel(object):
    def __init__(self):
        pass

    def __del__(self):
        # This method will never be called because we'll create a cyclic reference.
        print "__del__ called for:", id(self)


class ObjectWithNoDel(object):
    pass


def test():
    a = ObjectWithDel()
    a.a = a

    b = ObjectWithNoDel()
    b.b = b
    print "created a, id(a) ->", id(a), "type:", type(a)
    print "created b, id(b) ->", id(b), "type:", type(b)


test()
gc.collect()
print gc.garbage


$ python test.py
created a, id(a) -> 4353206544 type: <class '__main__.ObjectWithDel'>
created b, id(b) -> 4353206608 type: <class '__main__.ObjectWithNoDel'>
[<__main__.ObjectWithDel object at 0x10378a910>]
```

However, the real botocore client's don't actually implement a `__del__` so this behavior isn't something you'll actually see in real code.

Let me run a few more tests to confirm.

jamesls:
Here's another test I ran:

```
#!/usr/bin/env python

import gc
import botocore.session

session = botocore.session.get_session()

print "Normal case, using standard clients"
def test():
    clients = []
    for i in xrange(10):
        c = session.create_client('s3')
        clients.append(c)
test()
gc.collect()
print "Uncollected:", gc.garbage
```

Running this I get:

[...]

PR #1450, opened 2018-04-25
RFC: Avoid creation of circular references per request

kkopachev:
This PR aimed to reduce number of circular references remain after a request has being made. 
I am running a tornado app, which for incoming request talks with s3 to get/put an object. At some point, I noticed that default GC thresholds are way to big to successfully collect all the garbage, so app eats tons of memory after a long run. 
I could possibly play with thresholds and configure it to run frequently, but that's all feels more like a hack. 
I noticed there was an old issue https://github.com/boto/botocore/issues/805 saying that client objects are retained in memory to be GC'd later. So I figured that I can reuse client objects across multiple incoming web requests. 
However, digging dipper I found that even making request creates multiple of unreachable objects which are then waiting to be GC'd.

I put a simple test script in place to track down if it is possible to break circular references somewhere manually, so objects would be removed from memory without GC help (which does not seem to be guarranteed in CPYthon anyway)
```python
#!/usr/bin/env python

import gc
import botocore.session

print (gc.get_threshold())
gc.disable()

client = session.create_client('ec2', region_name='us-west-2')


def test(client):
    for i in range(0, 1):
        client.describe_instances(
            Filters=[{"Name": "instance-state-name", "Values": ["pending"]}])


gc.collect()
gc.set_debug(gc.DEBUG_STATS)
test(client)
gc.collect()
```
Example output would be:
```
(700, 10, 10)
gc: collecting generation 2...
gc: objects in each generation: 511 0 70118
gc: done, 67 unreachable, 0 uncollectable, 0.0324s elapsed.
gc: collecting generation 2...
gc: objects in each generation: 0 0 70196
gc: done, 0.0316s elapsed.
```
Note the 67 unreachable objects after call to `gc.collect()`. Now, when we increase number of iterations (effectively increasing number of requests made), number of unreachable objects will also grow. With 10 requests it outputs `gc: done, 382 unreachable, 0 uncollectable, 0.0309s elapsed.` 

I am proposing 2 changes in this PR. 
First one is to declare `_default_blob_parser` method as static, which gives a little bit of improvement.
1 iteration shows 64 unreachable objects and 10 iterations give 352 (vs 67 and 382). Not a lot, but helps a little.

Second proposed change is to use WeakMethod to attach response hook handler. And I would like to get some advice about it. 
The method, which attached as a hook on response event was added as a way to fix https://github.com/boto/botocore/issues/82 . I did remove this hook attachment and run s3 integration test to capture a failure, but it was passing. What happened is that in original issue virtual host-style URI was [using global service endpoint](https://github.com/boto/botocore/blob/e5610793c162e6df3551bd4f76545506eadf7a18/botocore/handlers.py#L124). In current version this is all changed and[ it uses region-specific endpoint](https://github.com/boto/botocore/blob/develop/botocore/utils.py#L755). So I am not sure if that redirect-hook is all that matters at all now. 
I see that removing the hook might cause other failures down the road, since in general rewinding body on post requests on some redirects are expected. I noticed that such core feature should have being done in `requests` library and it is. [Issue](https://github.com/requests/requests/issues/3079) [was fixed ](https://github.com/requests/requests/pull/3655). So maybe, instead of using `WeakMethod` it would be better to upgrade `requests` lib https://github.com/boto/botocore/issues/1122 ?
tldr: WeakMethod vs removing the hook vs upgrading requests? Which one to go with?

These 2 changes above, running test script would output 
`gc: done, 32 unreachable, 0 uncollectable, 0.0353s elapsed.`
**regardless** of number of iterations.
So now I can safely make aws requests from my long-living app without waiting for GC to kick in to clean circular garbage. Should be performance/memory win after all.

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -34,7 +34,7 @@ from botocore.vendored.requests.packages.urllib3.connectionpool import \
 from botocore.vendored.requests.packages.urllib3.connectionpool import \
     HTTPSConnectionPool
 
-if six.PY3:
+if six.PY34:
     from weakref import WeakMethod
 else:
     import weakref
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -34,6 +34,24 @@ from botocore.vendored.requests.packages.urllib3.connectionpool import \
 from botocore.vendored.requests.packages.urllib3.connectionpool import \
     HTTPSConnectionPool
 
+if six.PY3:
+    from weakref import WeakMethod
+else:
+    import weakref
+
+    class WeakMethod(object):
+        """A callable object. Takes one argument to init: 'object.method'.
+        Once created, call this object -- MyWeakMethod() --
+        and pass args/kwargs as you normally would.
+        """
+        def __init__(self, object_dot_method):
+            self.target = weakref.proxy(object_dot_method.__self__)
+            self.method = weakref.proxy(object_dot_method.__func__)
+
+        def __call__(self, *args, **kwargs):
+            """Call the method with args and kwargs as needed."""
+            return self.method(self.target, *args, **kwargs)
+
 
 logger = logging.getLogger(__name__)
 
@@ -401,7 +419,7 @@ class AWSPreparedRequest(models.PreparedRequest):
         self.original = original_request
         super(AWSPreparedRequest, self).__init__()
         self.hooks.setdefault('response', []).append(
-            self.reset_stream_on_redirect)
+            WeakMethod(self.reset_stream_on_redirect))
 
     def reset_stream_on_redirect(self, response, **kwargs):
         if response.status_code in REDIRECT_STATI and \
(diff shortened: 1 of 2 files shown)


botocore/awsrequest.py
```diff
@@ -401,7 +419,7 @@ def __init__(self, original_request):
         self.original = original_request
```
kkopachev: this one seems unused :open_mouth:

kkopachev:
anybody?

kkopachev:
Hi guys! 
I'd like to get some input on this. Is this merge-able/not?
Seems like a good thing to do in general. 
Thoughts?

nateprewitt:
It looks like this hook ended up getting removed in this [commit](https://github.com/boto/botocore/commit/818d23505620f5091e5a4114020b66b5e6d7aa7b) at the end of 2018. The weak reference should no longer be needed since the entire hook mechanism has been removed from modern versions of botocore. Please let us know if you have any other comments or concerns.

PR #1452, opened 2018-04-26
Add doc string to get_waiter

JordonPhillips:


botocore/client.py
```diff
@@ -745,6 +745,12 @@ def _get_waiter_config(self):
         return self._cache['waiter_config']
 
     def get_waiter(self, waiter_name):
+        """Returns an object that can wait for some condition.
+
+        :type waiter_name: str
+        :param waiter_name: The name of the waiter to get. See the waiters
+            section of the service docs for a list of available waiters.
+        """
```
kyleknap: Maybe we should include what it returns and the return type as well?

PR #1454, opened 2018-04-28
fix typo for Spot fleet

sharuzzaman:
fix typo for Spot fleet


stealthycoin:
Thanks for reporting this. This is a generated model we consume from upstream so we can't take this PR. I will submit this to the upstream owner though, thanks!

PR #1466, opened 2018-05-22
Stop vendoring requests and urllib3 while being backwards compatible

### LINKED ISSUE #756: filedescriptor out of range in select() while calling message.delete
s = boto3.session.Session(aws_access_key_id=aws_access_key_id, aws_secret_access_key=aws_secret_access_key, region_name=region_name)
    session = s.resource('sqs')
    
    queue = sqs.get_queue_by_name(QueueName=queue_name)
    while True:
        messages = queue.receive_messages(MaxNumberOfMessages=10, AttributeNames=['All'], MessageAttributeNames=['All'])
        if len(messages) == 0:
            break
        for message in messages:
            save_message(message)
            message.delete()
            

Stack Trace
---------------
    Exception in thread Thread-4:
      File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
        self.run()
      File "/usr/lib/python2.7/threading.py", line 763, in run
        self.__target(*self.__args, **self.__kwargs)
      File "/ws/amazon/bin/sqs", line 59, in each_thread
        message.delete()
      File "/usr/local/lib/python2.7/dist-packages/boto3/resources/factory.py", line 455, in do_action
        response = action(self, *args, **kwargs)
      File "/usr/local/lib/python2.7/dist-packages/boto3/resources/action.py", line 79, in __call__
        response = getattr(parent.meta.client, operation_name)(**params)
      File "/usr/local/lib/python2.7/dist-packages/botocore/client.py", line 310, in _api_call
      File "/usr/local/lib/python2.7/dist-packages/botocore/client.py", line 384, in _make_api_call
      File "/usr/local/lib/python2.7/dist-packages/botocore/endpoint.py", line 111, in make_request
      File "/usr/local/lib/python2.7/dist-packages/botocore/endpoint.py", line 140, in _send_request
        success_response, exception):
      File "/usr/local/lib/python2.7/dist-packages/botocore/endpoint.py", line 213, in _needs_retry
      File "/usr/local/lib/python2.7/dist-packages/botocore/hooks.py", line 226, in emit
      File "/usr/local/lib/python2.7/dist-packages/botocore/hooks.py", line 209, in _emit
        response = handler(**kwargs)
      File "/usr/local/lib/python2.7/dist-packages/botocore/retryhandler.py", line 183, in __call__
      File "/usr/local/lib/python2.7/dist-packages/botocore/retryhandler.py", line 250, in __call__
        caught_exception)
      File "/usr/local/lib/python2.7/dist-packages/botocore/retryhandler.py", line 265, in _should_retry
      File "/usr/local/lib/python2.7/dist-packages/botocore/retryhandler.py", line 313, in __call__
        caught_exception)
      File "/usr/local/lib/python2.7/dist-packages/botocore/retryhandler.py", line 222, in __call__
      File "/usr/local/lib/python2.7/dist-packages/botocore/retryhandler.py", line 355, in _check_caught_exception
        raise caught_exception
    ValueError: filedescriptor out of range in select()

### LINKED ISSUE #1248: A urllib3 connection is not put back in the pool after HEAD

### LINKED ISSUE #1258: 'AWSHTTPSConnection' object has no attribute 'ssl_context'
Hi,

I'm getting the above error when trying to use Route53 service. I have investigated and concluded that the fault is [the copying of functions from AWSHTTPConnection to AWSHTTPSConnection](https://github.com/boto/botocore/blob/37bbdc3ffc40e43500d8e044b418947dce4f7482/botocore/awsrequest.py#L256-L261):

```python
# Now we need to set the methods we overrode from AWSHTTPConnection
# onto AWSHTTPSConnection.  This is just a shortcut to avoid
# copy/pasting the same code into AWSHTTPSConnection.
for name, function in AWSHTTPConnection.__dict__.items():
    if inspect.isfunction(function):
        setattr(AWSHTTPSConnection, name, function)
```

The problem is that this ends up overriding the `__init__` function too, so the `VerifiedHTTPSConnection` constructor is never invoked. I don't know why this worked in the past, or why it has become a problem recently. I have locally patched that file to add `and function.__name__ != '__init__'` to the `if` above, and copied manually the constructor to `AWSHTTPSConnection` (to invoke `VerifiedHTTPSConnection` constructor). Things seem to work fine now.

### LINKED ISSUE #1325: urllib3 / botocore
Hello,
it seems that urllib version using by botocore does not manage ipv6 proxy correctly,
as https://github.com/shazow/urllib3/commits/master/urllib3/connectionpool.py 

can you correct this ?
Regards, 
Nicolas Prochazka

henriklindstrom:
Fixes #1258 
Fixes #1248 
Fixes #1325 
Fixes #756 
Fixes aws/aws-cli#2994

This updates awsrequest.py to be compatible with the latest urllib3 and rips out the vendored versions of requests and urllib3. In order to be compatible with other libraries that are using the vendored packages, requests and urllib3 are imported and added to sys.modules at the same path as before.

This also stops using select.select for checking sockets, in order to work in processes with more than 1024 open filedescriptors (this is fixed in urllib3 by https://github.com/urllib3/urllib3/pull/1001).


joguSD:
Thanks for taking the time to look into this. This is a pretty big PR that addresses a lot of issues. Would it be possible to break this up into smaller PRs that address issues more specifically? 

Right now, I think we would be receptive of fixes that address compatibility issues with newer versions of requests and urllib3 so long as they still work with our vendored version (hopefully that makes sense!). But as of right now I don't think we're quite ready to bite the bullet on unvendoring requests as there's likely to be edge cases that cause breaking behavior. For example, exception handling has changed in urllib3 since our vendored version and I know of at least [one place](https://github.com/boto/botocore/blob/f5cf249aaccc2b988b65176e6a5337086d060074/botocore/endpoint.py#L260-L261) that affects `botocore`.

henriklindstrom:
Yes this is a pretty big PR, but it is made in order to solve a single issue, #756 .
In order to do this, three parts is needed:
 - Fix the broken AWSConnection classes (in order to use a later version of urllib3)
 - Use a version of urllib3 that includes the urllib3/urllib3#1001 fix
 - Implement the same fix in the AWSConnection class

This is basically what this PR does, but instead of upgrading the vendored versions of request and urllib3, they are removed.
So, if you feel like its a too big leap to remove them, we would have to find a version of urllib3 that you feel comfortable with including, that also includes the needed fix.

I do however think that it would be worth investigating eventual edge cases from unvendoring the packages and taking the leap, since this situation will likely occur again.

Hernrup:
I think this PR is as small as It can get from the point of view that unvendoring in it itself is a much needed fix. We can clearly see from the number of open issues related to this that it is not possible to keep up with the fixes in upstream. Frustration aside, we should consider all the possible security patches that might be applied in upstream and will that will be hard to refer back while requests is vendored. This makes botocore unusually susceptible to 0-days. We can already see that the included CA-cert list is out of sync with those validated by https://github.com/certifi for example (later version of requests used certifi).

I get that is a high-risk change but I dont see a viable long-term solution other then biting the bullet.

joguSD:
I agree that the current state of things is not a viable long term solution. We've been playing around with a few ideas and have been thinking about potentially unvendoring requests and dropping it as a dependency entirely and taking urllib3 as a direct dependency. After auditing our usage of requests we really don't use most of the features it offers in botocore and it might be simpler for us to just use urllib3 directly in the long run. A large reason we vendor requests is to avoid version conflicts with other popular libraries that also use requests. Maybe the best way to avoid this is to simply not use it. Any thoughts on this?

joakimkarlsson:
I agree that in the long run, removing unnecessary dependencies is the way to go. But this needs to be fixed ASAP. We are currently forced to run a fork of botocore based on this PR in our production environment as the problem with exhausting file descriptors due to #756 makes the official release unviable to use. 

So I say we should do whatever is necessary to get this PR merged first, and once that is out of the way start looking at removing dependencies.

We'd be happy to help out in any way we can.

mwilck:
@joguSD, Linux distributions are very much against vendor-specific versions of official packages like `requests`. I've tried to push a package using your vendored `requests` as a fix for #1258 for openSUSE, and it was rejected because of the security threats. I'm pretty certain other Linux distros have the same preferences. By shipping a custom `requests`, you essentially take over responsibility for security fixes for that package, which has a significantly larger attack surface than botocore itself. Is that something you want to carry around in the long term?

That said, I, and I guess most users, don't care whether you solve the issue by taking the patches from here or by just ripping out `requests`. That's fully up to you.

Hernrup:
Have you decided on a way forward for this @jogusd ?
What can we do to help?

joguSD:
I've opened an issue to continue this discussion: https://github.com/boto/botocore/issues/1486
If you have any additional concerns post them on the proposal issue and give it a +1 to help show that we need to get this prioritized.

I'm working on a branch that implements what's described in the above issue that I will hopefully be able to share before too long.

joguSD:
We have un-vendored requests/urllib3 and are now using urllib3 directly as of `botocore` v1.11.0.

PR #1470, opened 2018-06-01
Add paginator for route53 ListHostedZonesByName

euank:
I've done minimal testing that this works as I expect by running:

```
AWS_DATA_PATH=$BOTOCORE_PATH/data aws route53 list-hosted-zones-by-name --page-size=1
```

With this change, it successfully lists all my zones and `--debug` shows it makes multiple calls to do so as I'd expect.
Without this change the `--page-size` pseudo-parameter doesn't exist so the command doesn't run at all.

### DIFF
--- a/botocore/data/route53/2013-04-01/paginators-1.json
+++ b/botocore/data/route53/2013-04-01/paginators-1.json
@@ -14,6 +14,19 @@
       "limit_key": "MaxItems",
       "result_key": "HostedZones"
     },
+    "ListHostedZonesByName": {
+      "more_results": "IsTruncated",
+      "limit_key": "MaxItems",
+      "result_key": "HostedZones",
+      "input_token": [
+        "DNSName",
+        "HostedZoneId"
+      ],
+      "output_token": [
+        "NextDNSName",
+        "NextHostedZoneId"
+      ]
+    },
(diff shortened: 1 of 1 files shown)


botocore/data/route53/2013-04-01/paginators-1.json
```diff
@@ -14,6 +14,19 @@
...
+    "ListHostedZonesByName": {
+      "more_results": "IsTruncated",
+      "limit_key": "MaxItems",
+      "result_key": "HostedZones",
+      "input_token": [
+        "DNSName",
+        "HostedZoneId"
+      ],
+      "output_token": [
+        "NextDNSName",
```
rocky4570: Is this a common standard name for tokens? - its a bit of a bug bare of mine that AWS already has NextToken nextToken and NextMarker for some other API calls - would be good if this were consistent.

euank:
Bump, I think this PR is still relevant.

kdaily:
Hi @euank,

Apologies for the long delay. I've requested this paginator be added by the Route53 team. Other AWS SDKs also use the same paginator definitions, so in this way they will all get the same implementation and behavior.

kdaily:
V374875885

PR #1473, opened 2018-06-05
Don't fetch creds for unsigned clients

### LINKED ISSUE #1472: Client creator tries to resolve credentials for UNSIGNED clients
Source issue is awslabs/awsprocesscreds#16

Basically the problem is that the Session object will try to resolve credentials even if you're creating an anonymous client. This is causing an infinite loop when you use the process provider since it needs to create an anonymous client to call assume role with saml.

JordonPhillips:
Fixes #1472 
Fixes awslabs/awsprocesscreds#16

### DIFF
--- a/.changes/next-release/bugfix-Credentials-65605.json
+++ b/.changes/next-release/bugfix-Credentials-65605.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Credentials",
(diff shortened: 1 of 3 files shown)


jamesls:
With this PR, what happens if you create an UNSIGNED client and try to invoke an API that needs signing?

JordonPhillips:
Nothing changes as far as I can tell. With or without the credentials present you'll never actually use them: https://github.com/boto/botocore/blob/develop/botocore/signers.py#L138

PR #1478, opened 2018-06-12
Only prepare AWSRequest once when signing it using SigV4Auth

kmichel-sereema:
AWSRequest.body is a property which is fairly slow to compute:
if part of the request data is not streamable, the whole request data is urlencoded,
which slowly iterates on all bytes in urllib.parse.quote_from_bytes .

When profiling a quite busy task worker, this reduced the CPU time spent in
AWSRequest.body from 21% to 5%.

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -264,20 +264,21 @@ class SigV4Auth(BaseSigner):
             # When payload signing is disabled, we use this static string in
             # place of the payload checksum.
             return UNSIGNED_PAYLOAD
-        if request.body and hasattr(request.body, 'seek'):
-            position = request.body.tell()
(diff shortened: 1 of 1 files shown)


medmunds:
Until this or something similar lands, a workaround is to monkey patch `AWSRequest.body` to cache the result: https://gist.github.com/medmunds/ad47a5001729040d7743ae20d32786e1

(Also, note that a change [introduced](https://github.com/boto/botocore/commit/818d23505620f5091e5a4114020b66b5e6d7aa7b#diff-cb00fc815f8f80ae15e7d0cecc74b395R379) in botocore 1.11 doubles the `urllib.parse.quote_from_bytes` calls during signature calculation, making this PR even more helpful.)

PR #1484, opened 2018-06-15
Update endpoint test to use service ids

jamesls:
This changes the two tests to more closely map to what
we wanted to test when we originally wrote these tests.

These tests were written before serviceId existed, which
is really want we want to use for testing.  These tests check
two things:

* Ensure that the directory names (i.e the client names) match
  the computed name based on the serviceId.  Existing services
  still need overrides but new services shouldn't as many overrides.
* Update the endpoints tests to be more direct about endpoint prefix
  matching.  We really just care that everything in the endpoints.json
  maps to something we can create a client for.  We now build the
  list of acceptable keys by looking up the endpoint prefix.

### DIFF
--- a/botocore/data/transcribe/2017-10-26/service-2.json
+++ b/botocore/data/transcribe/2017-10-26/service-2.json
@@ -6,7 +6,7 @@
     "jsonVersion":"1.1",
     "protocol":"json",
     "serviceFullName":"Amazon Transcribe Service",
-    "serviceId":"Transcribe Service",
+    "serviceId":"Transcribe",
     "signatureVersion":"v4",
     "signingName":"transcribe",
     "targetPrefix":"Transcribe",
--- a/botocore/data/transcribe/2017-10-26/service-2.json
+++ b/botocore/data/transcribe/2017-10-26/service-2.json
@@ -6,6 +6,7 @@
     "jsonVersion":"1.1",
     "protocol":"json",
     "serviceFullName":"Amazon Transcribe Service",
+    "serviceId":"Transcribe Service",
     "signatureVersion":"v4",
     "signingName":"transcribe",
     "targetPrefix":"Transcribe",
--- a/tests/functional/test_endpoints.py
+++ b/tests/functional/test_endpoints.py
@@ -16,70 +16,122 @@ from nose.tools import assert_equal
 from botocore.session import get_session
 
 
-# Several services have names that don't match for one reason or another.
 SERVICE_RENAMES = {
-    'application-autoscaling': 'autoscaling',
-    'appstream': 'appstream2',
-    'autoscaling-plans': 'autoscaling',
-    'dynamodbstreams': 'streams.dynamodb',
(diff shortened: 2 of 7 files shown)


tests/functional/test_endpoints.py
```diff
@@ -92,15 +144,22 @@ def test_service_name_matches_endpoint_prefix():
...
-        "prefix `%s`, actual: `%s`" % (
-            service_name, expected_endpoint_prefix, endpoint_prefix))
+        computed_name, actual_service_name,
+        "Actual service name `%s` does not match expected service name "
+        "we computed: `%s`" % (
+            actual_service_name, computed_name))
+
+
+def _get_computed_service_name(service_metadata):
+    if 'serviceId' not in service_metadata:
```
kyleknap: So it is unfortunate that we have not rolled all of the service id's out yet. However, I am a little hesitant to continue to use endpointPrefix because this test may all of a sudden fail if we pull in a service model that now has a service id. For example let's take `resourcegroupstaggingapi`. Right now its endpoint prefix is `tagging` and that override is hard coded. However when its service id gets added to the model, it will be `Resource Groups Tagging API` which translates to `resoruce-groups-tagging-api` which will cause the override to not happen and the test to fail. I'm not sure how many examples there are of this.

I think what may make sense is to try to compute the service id if it does not exist and update the override dictionary . To my knowledge, (@stealthycoin can confirm)  all of the service id's to be rolled are based off of the computed form so there should be no discrepancies.

kyleknap: Yep!

jamesls: Ok chatted offline and wanted to capture what we'd discuss.  In order to address this concern, I'll just update the PR to add all the missing service ids.  I'll also make a missing service id a a test failure.  That should address all the concerns here.

botocore/data/transcribe/2017-10-26/service-2.json
```diff
@@ -6,6 +6,7 @@
     "jsonVersion":"1.1",
     "protocol":"json",
     "serviceFullName":"Amazon Transcribe Service",
+    "serviceId":"Transcribe Service",
```
kyleknap: I think for this one it was going to be `Transcribe`.

tests/functional/test_endpoints.py
```diff
@@ -92,15 +152,22 @@ def test_service_name_matches_endpoint_prefix():
 
 
 def _assert_service_name_matches_endpoint_prefix(loader, service_name):
-    # Load the service model and grab its endpoint prefix
     service_model = loader.load_service_model(service_name, 'service-2')
-    endpoint_prefix = service_model['metadata']['endpointPrefix']
+    computed_name = _get_computed_service_name(
+        service_name, service_model['metadata'])
```
kyleknap: I added `service_id` property the other [day](https://github.com/boto/botocore/pull/1467). I think we can just use that. It will also raise errors if it is not present.

jamesls:

PR #1491, opened 2018-06-25
Add iter_chunks()/iter_lines() to StreamingBody

jamesls:
This pulls in https://github.com/boto/botocore/pull/1195 and incorporates review feedback [here](https://github.com/boto/botocore/pull/1195#discussion_r197903058).  The suggested changes make it similar to what requests exposes for its streaming responses.

I also added a few additional test cases.

### DIFF
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -80,6 +80,53 @@ class StreamingBody(object):
             self._verify_content_length()
         return chunk
 
+    def __iter__(self):
+        """Return an iterator to yield lines from the raw stream.
+        """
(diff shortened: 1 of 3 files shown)


botocore/response.py
```diff
@@ -80,6 +82,47 @@ def read(self, amt=None):
             self._verify_content_length()
         return chunk
 
+    def __iter__(self):
+        """Return an iterator to yield lines from the raw stream.
```
kyleknap: Returns chunks not lines now.

Issue #1486, opened 2018-06-19
[Proposal] Use urllib3 directly instead of requests


joguSD:
Botocore currently vendors a fairly dated version of `requests`. This particular version of `requests` also vendors the `urllib3` library. We originally vendored `requests` to give us more control over the specific version that was being used as we need to make changes to some of the behavior provided by both `requests` (redirect behavior, etc.) and `urllib3` (HTTP 100-continue support). By dropping `requests` and using `urllib3` directly we gain more control over our HTTP client behavior and simplify our dependencies.

Let's use this issue as an aggregate issue to contain discussion around unvendoring `requests` and switching to using `urllib3` directly.

Implementation: https://github.com/boto/botocore/pull/1495

Related Issues:
https://github.com/aws/aws-cli/issues/2994
https://github.com/boto/botocore/issues/1258
https://github.com/boto/botocore/issues/1248
https://github.com/boto/botocore/issues/756
https://github.com/boto/botocore/issues/1370
https://github.com/boto/botocore/pull/1466
https://github.com/boto/botocore/pull/1300
https://github.com/boto/botocore/pull/1385
https://github.com/boto/botocore/issues/1464

jamesls:
:+1:  yeah I've wanted to do this for a while.

joguSD:
Implemented in #1495

Issue #1464, opened 2018-05-18
botocore leak when using get + head object


thehesiod:
with the following modules installed:
```
bash-4.2# /valgrind/bin/python3 -m pip freeze
APScheduler==3.3.1
asynctest==0.12.0
aws-xray-sdk==1.0
botocore==1.10.21
cachetools==2.1.0
certifi==2018.4.16
chardet==3.0.4
cycler==0.10.0
datadog==0.20.0
decorator==4.3.0
dill==0.2.7.1
docutils==0.14
editdistance==0.3.1
future==0.16.0
gapic-google-cloud-pubsub-v1==0.15.4
google-api-python-client==1.6.7
google-auth==1.4.1
google-auth-httplib2==0.0.3
google-cloud-core==0.25.0
google-cloud-pubsub==0.26.0
google-gax==0.15.16
googleapis-common-protos==1.5.3
graphviz==0.8.3
grpc-google-iam-v1==0.11.4
grpcio==1.12.0
httplib2==0.11.3
idna==2.6
iso8601==0.1.12
jmespath==0.9.3
jsonpickle==0.9.6
kiwisolver==1.0.1
matplotlib==2.2.2
memory-profiler==0.52.0
nameparser==0.5.6
numpy==1.14.3
oauth2client==4.1.2
objgraph==3.4.0
ply==3.8
proto-google-cloud-pubsub-v1==0.15.4
protobuf==3.5.2.post1
psutil==5.4.5
pyasn1==0.4.2
pyasn1-modules==0.2.1
PyGithub==1.39
PyJWT==1.6.1
pyparsing==2.2.0
python-dateutil==2.7.3
pytz==2018.4
requests==2.18.4
rsa==3.4.2
setproctitle==1.1.10
simplejson==3.15.0
six==1.11.0
tzlocal==1.5.1
uritemplate==3.0.0
urllib3==1.22
wrapt==1.10.11
```

and the following test case: https://gist.github.com/thehesiod/dc04230a5cdac70f25905d3a1cad71ce

and run with memory_profiler like so in python 3.6.5:
`PYTHONMALLOC=malloc python3 /valgrind/bin/mprof run --interval=1 /tmp/test.py -test botocore_test`

yields the following graph:
![boto_head_object](https://user-images.githubusercontent.com/4001789/40210894-5fda02e2-59fc-11e8-8f2a-2482c88e19c9.png)

when run with:
[...]

PR #1385, opened 2018-02-13
Backport urllib3 upstream fix for reset connections to AWS servers, fixes #1248

ezyang:
This solves #1248 in botocore HEAD. The proper fix is to update urllib3 but I don't know why you are vendoring `requests` in the first place.

Upstream fix I backported was https://github.com/shazow/urllib3/pull/1235

Issue #1370, opened 2018-01-22
Closing a StreamingBody does not close the socket


timuralp:
The `botocore.response.StreamingBody` (https://github.com/boto/botocore/blob/develop/botocore/response.py#L29) class defines a close method. However, closing a `StreamingBody` does not result in closing the socket. I verified this behavior by looking at packet traces after making a request for a large object in S3. Here is a code snippet to reproduce the behavior:
```
import boto3
import time

client = boto3.client('s3')
resp = client.get_object(Bucket='test', Key='large')
resp['Body'].close()
time.sleep(10)
```

The `RST` packet should be sent after calling `close()`. Currently, I see it sent 10 seconds later, when the program exits (and a number of ZeroWindow messages in between).

It appears that to close the socket, boto should also call `release_conn` on the `_raw_stream`. After adding that change, the connection is closed immediately (`RST -> FIN -> FIN`).

timuralp:
Looking through the vendored urllib3 code, the right thing to do would be to close the connection when `close` is called on the `HTTPResponse` object and its `_connection` field is set. In the newer versions of urllib3 this was addressed through https://github.com/shazow/urllib3/pull/783.

Since current release of botocore vendors requests, which vendors urllib3, I think the right thing to do is to close the `_connection` field explicitly and call `release_conn`. Releasing it will force the `is_connection_dropped` function to detect that it's been closed and create a new connection on a subsequent request.

dstufft:
Looking through this, it looks to be correct to me. We should ideally fix this by pulling in a newer urllib3.

joguSD:
We have un-vendored requests/urllib3 and are now using urllib3 directly as of `botocore` v1.11.0.

PR #1495, opened 2018-07-04
Unvendor requests and use urllib3 directly

### LINKED ISSUE #756: filedescriptor out of range in select() while calling message.delete

joguSD:
This pull request begins to implement the changes discussed in https://github.com/boto/botocore/issues/1486

This builds on some of the work started in this PR: https://github.com/boto/botocore/pull/1466

The work is not yet done but this branch should largely be functional for the majority of use cases. In particular exception cases have been hard to verify, and there will be cases where different exceptions are now raised. This is an unavoidable consequence of moving away from our vendored dependencies. To continue building on this I would like to minimize dependency exception uses to single modules, and ensure that we don't leak exceptions from our dependencies to simplify both the exceptions generated from `botocore` and make it easier to verify `botocore`'s exception behavior.

Additionally, these changes impact our dependent packages `s3-transfer`, `boto3`, and `aws-cli` as all have references to our vendored dependencies (mostly exception classes). 

Downstream Updates:
Boto3: https://github.com/boto/boto3/pull/1668
CLI: https://github.com/aws/aws-cli/pull/3513

Closes #756, #1248, #1258, #1300, #1370, #1385, #1464, #1466, 
Closes https://github.com/aws/aws-cli/issues/2994


botocore/http_session.py
```diff
@@ -0,0 +1,172 @@
...
+            http_response = AWSResponse(
+                url=request.url,
+                raw=urllib_response,
+                status_code=urllib_response.status,
+                headers=urllib_response.headers,
+            )
+
+            if not request.stream_output:
+                # Cause the raw stream to be exhausted immediatly
+                http_response.content
```
jamesls: What's the advantage of doing this vs. using the `preload_content` param?

botocore/http_session.py
```diff
@@ -0,0 +1,172 @@
...
+
+            return http_response
+        except (NewConnectionError, socket.gaierror) as e:
+            raise EndpointConnectionError(endpoint_url=request.url, error=e)
+        except ProtocolError as e:
```
jamesls: Does the connection get release back to the pool manager in the case of exceptions like this?

botocore/http_session.py
```diff
@@ -0,0 +1,172 @@
...
+
+class Urllib3Session(object):
+    def __init__(self,
+                 verify=True,
+                 proxies=None,
+                 timeout=None,
+                 max_pool_connections=MAX_POOL_CONNECTIONS,
+    ):
+        self._verify = verify
+        self._proxies = proxies or {}
```
jamesls: `if proxies is not None`

tests/functional/test_lex.py
```diff
@@ -45,7 +45,7 @@ def test_unsigned_payload(self):
         # signature. The signature will be part of the authorization header.
         # Since we don't have direct access to the payload signature,
         # we compare the authorization instead.
-        authorization = request.headers.get('authorization')
+        authorization = request.headers.get('Authorization')
```
jamesls: Why did these change?  Are we not pulling in a case insensitive dict anymore?

botocore/endpoint.py
```diff
@@ -282,6 +211,9 @@ def _needs_retry(self, attempts, operation_model, request_dict,
             time.sleep(handler_response)
             return True
 
+    def send(self, request):
```
jamesls: I'd prefer not to add a new public method to endpoint.  Why does this need to be public.  AFAICT consuming code is still using `make_request`.

botocore/exceptions.py
```diff
@@ -60,21 +59,20 @@ class ApiVersionNotFoundError(BotoCoreError):
     fmt = 'Unable to load data {data_path} for: {api_version}'
 
 
-class EndpointConnectionError(BotoCoreError):
+class HTTPClientError(BotoCoreError):
+    fmt = ('Generic error when an HTTP Client raises an exception')
```
jamesls: You don't need parens here.

botocore/awsrequest.py
```diff
@@ -450,6 +471,92 @@ def prepare_body(self, data, files, json=None):
                 if 'Transfer-Encoding' in self.headers:
                     self.headers.pop('Transfer-Encoding')
 
+        if self.body and 'Content-Length' not in self.headers:
+            self.headers['Transfer-Encoding'] = 'chunked'
+
+
+class AWSResponse(object):
+
+    def __init__(self, url=None, status_code=None, headers=None, raw=None):
```
jamesls: Why are all of these defaulted to `None`?  Shouldn't these be required?

botocore/awsrequest.py
```diff
@@ -450,6 +471,92 @@ def prepare_body(self, data, files, json=None):
...
+
+    @property
+    def content(self):
+        """Content of the response, in bytes."""
+
+        if self._content is None:
+            # Read the contents.
+            try:
+                data_generator = self.raw.stream()
+            except AttributeError:
```
jamesls: When would `raw` not have a stream method on it?

botocore/awsrequest.py
```diff
@@ -450,6 +471,92 @@ def prepare_body(self, data, files, json=None):
...
+            self._content = bytes().join(data_generator) or bytes()
+
+        return self._content
+
+    @property
+    def text(self):
+        if not self.content:
+            return str('')
+
+        return self.content.encode('utf-8')
```
jamesls: Shouldn't this take into account encodings provided via response headers?

botocore/awsrequest.py
```diff
@@ -428,13 +425,37 @@ def reset_stream(self):
...
+            return
+
+        if isinstance(self.body, dict):
+            params = list(self.body.items())
+            self.body = urlencode(params, doseq=True)
+
+        try:
+            length = len(self.body)
+            self.headers['Content-Length'] = str(length)
+        except Exception as e:
```
jamesls: What would raise here?  The `len()` call?

botocore/awsrequest.py
```diff
@@ -54,7 +50,7 @@ def _read_status(self):
             return HTTPResponse._read_status(self)
 
 
-class AWSHTTPConnection(HTTPConnection):
+class AWSConnection(object):
     """HTTPConnection that supports Expect 100-continue.
```
jamesls: docstring is out of date.

botocore/awsrequest.py
```diff
@@ -450,6 +471,92 @@ def prepare_body(self, data, files, json=None):
...
+    @property
+    def content(self):
+        """Content of the response, in bytes."""
+
+        if self._content is None:
+            # Read the contents.
+            try:
+                data_generator = self.raw.stream()
+            except AttributeError:
+                def stream():
```
jamesls: Is there a reason not to just have this be a separate method and always use it (vs. using `raw.stream()` if it's available).

joguSD:
@jamesls What do you think would be the best format for tracking these types of comparisons?  I don't necessarily think it makes sense to put these in-code (as the difference are mostly removals). My workflow on this has been less about "this is what requests does" and more about "this is what botocore expects" (passing all the tests). In general my answer to "Why is this different from requests?" will be "because botocore doesn't use it, afaik".

At a really high level we're basically taking two things from requests and re-implementing the subset that botocore excepts. These two pieces are:
* requests model objects that represent a request & response (`AWSRequest`, `AWSPreparedRequest`, `AWSResponse`)
* Requests adapter class for `urllib3`

As for the `Urllib3Session` which parallels to the `HTTPAdapter` class in requests there shouldn't be much difference. The big difference is:
* Our vendored version of requests implemented chunked encoding for request bodies. Newer versions of urllib3 support this by setting `chunked=True` when calling `urlopen`, so we don't need it in our http session class. Additionally, I've not enabled this flag because afaik we never use chunked encoding when making requests.

The model objects do have a lot of changes as I basically gutted them to have next to no logic and only implemented the things that we relied on requests for. This turns out to be not a whole lot of functionality as botocore is already producing / preparing generally well structured requests before handing them off to requests. The behavior that I've found we needed is:
* Converting the modeled params dictionary to the query string
* Determining content length of the body
* Converting the headers to a case-insensitive dictionary
The biggest differences here are url preparation and determining the content length of the body. Requests parses the url does some validations and conversions on it and reconstructs it. From what I've gathered this process basically just fixes any issues the url may have and I've never ever seen it actually change a url after it was done. I verified this by running all of the tests asserting that the url before and after are the same. As for determining content-length of the body I tried to keep it simple and implement the logic that we document (it's `bytes` or a `file-like`).

botocore/http_session.py
```diff
@@ -0,0 +1,172 @@
...
+    pyopenssl.extract_from_urllib3()
+except ImportError:
+    pass
+
+filter_ssl_warnings()
+logger = logging.getLogger(__name__)
+DEFAULT_TIMEOUT = 60
+MAX_POOL_CONNECTIONS = 10
+# TODO: move the vendored cacert when we drop requests
+DEFAULT_CA_BUNDLE = os.path.join(
```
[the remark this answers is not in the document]
joguSD: I didn't copy it over as the CA_BUNDLE kinda muddies the diff. I was planning on moving this at the same time we remove the vendored requests directory.

botocore/http_session.py
```diff
@@ -0,0 +1,172 @@
...
+
+def get_auth_from_url(url):
+    parsed_url = urlparse(url)
+    try:
+        return unquote(parsed_url.username), unquote(parsed_url.password)
+    except (AttributeError, TypeError):
+        return None, None
+
+
+def get_proxy_headers(proxy_url):
```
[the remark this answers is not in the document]
joguSD: I'll see what I can do. This logic is more or less directly ported from what requests does.

botocore/http_session.py
```diff
@@ -0,0 +1,172 @@
...
+    headers = {}
+    username, password = get_auth_from_url(proxy_url)
+    if username and password:
+        basic_auth = construct_basic_auth(username, password)
+        headers['Proxy-Authorization'] = basic_auth
+    return headers
+
+
+def fix_proxy_url(proxy_url):
```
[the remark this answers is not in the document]
joguSD: Yeah, I haven't touched docstrings yet as I didn't know how much churn there would be. I'll try adding some documentation for the things that I think are little more solid.

botocore/http_session.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def get_cert_path(verify):
+    if verify is not True:
+        return verify
+
+    return where()
+
+
+class Urllib3Session(object):
```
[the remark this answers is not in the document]
joguSD: 1) All of the proxy logic inside the class directly relates to taking a properly formed proxy url and turning that into a urllib3 proxied connection. The logic outside the class don't relate to anything urllib3 specific.
2) `URLLib3`?
3) This correlates to the adapter. It's less of consolidating the two classes and more-so removing the part we didn't use. We effectively passed every parameter every time we made a request so the session didn't provide much. The parts it did give us: retries & redirects we either didn't rely (retries) on or are now our handling in our own retry handler (redirects).

botocore/http_session.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def _path_url(self, parsed_url):
+        path = parsed_url.path
+        if not path:
+            path = '/'
+        if parsed_url.query:
+            path = path + '?' + parsed_url.query
+        return path
+
+    def _verify_cert(self, conn, url, verify):
```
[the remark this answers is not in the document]
joguSD: This sets it up so that Urllib3 will do the verification.

botocore/http_session.py
joguSD: Not sure if there is an advantage. This is just how requests did it.

botocore/http_session.py
joguSD: Not sure on this. I'll do some testing to see if urllib3 will close the connection in error cases.

tests/functional/test_lex.py
joguSD: We weren't originally, but I've since added a case insensitive implementation. I'll revert these changes to reduce noise.

botocore/endpoint.py
joguSD: Yeah, I think private is probably better as well. I'm using `Endpoint.send` as a replacement for places we were patching out requests methods. In the long run I'd like to remove these patches and simply pass a mock/fake http_session into the endpoint though.

botocore/awsrequest.py
joguSD: Yeah, perhaps these just should be positional arguments.

botocore/awsrequest.py
joguSD: When it's not a raw response  from urllib3. 

Currently, this will never happen in practice but we relied on this functionality from requests in some tests where we mock the raw response with a file-like object.

botocore/awsrequest.py
joguSD: It could, requests had some fairly complex logic for trying to determine the content encoding. The only place the `text` property is used is in the `ContainerMetadataFetcher`. I think I might just remove this have do the decoding there.

botocore/awsrequest.py
joguSD: Yep.

botocore/awsrequest.py
joguSD: Kind of? The functionality and inheritance described is still true, it just doesn't actually get inherited until it's used as a mixin.

botocore/awsrequest.py
joguSD: This is another case of 'it's what requests does'. I think the motivation here was that the Response object is generic and the underlying response could come from a different library other than urllib3 that may or may not have a stream method.

botocore/http_session.py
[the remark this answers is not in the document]
jamesls: >  kinda muddies the diff

Given we won't likely won't remove the vendored requests directory for a while, I'd prefer if this new stuff had minimal dependencies on anything in `vendored/` unless necessary.  That way we can just remove `vendored/` when appropriate.  It will also give customers something to switch to if for whatever reason they're depending on our cacert.pem.

botocore/http_session.py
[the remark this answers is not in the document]
jamesls: > This logic is more or less directly ported from what requests does.

It's still confusing to follow and I'd prefer we come up with a better abstraction for proxies if we're going to be maintaining/supporting this.

botocore/http_session.py
```diff
@@ -0,0 +1,172 @@
...
+        self._verify = verify
+        self._proxies = proxies or {}
+        if timeout is None:
+            timeout=DEFAULT_TIMEOUT
+        if not isinstance(timeout, (int, float)):
+            timeout = Timeout(connect=timeout[0], read=timeout[1])
+        self._timeout = timeout
+        self._max_pool_connections = max_pool_connections
+        self._proxy_managers = {}
+        self._http_pool = PoolManager(
```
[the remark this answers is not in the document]
jamesls: What's interesting is that the `max_pool_connections` is really "max persistent connections", and in the case of s3 with virtual hosted addressing, this number is potentially 100 max persistent connections by default (10 per pool, 10 pool total, 1 pool used per bucket because it's a different hostname).  A connection pool would make that closer to what you'd expect `max_pool_connections` to mean, but could potentially result in worse performance.  I suppose we should keep this as is for now to minimize risky changes in the switchover.

botocore/http_session.py
jamesls: Yes I'm aware requests does it like this.  I always found it odd because it's relying on attribute access to trigger external side effects, and if `preload_content` is identical it seems like that's a more explicit (and preferable) way of doing this.

botocore/endpoint.py
jamesls: Another option is to add short circuiting behavior at the endpoint layer via events, similar to how you can short circuit at the client layer via `before-call`.  We could add a `sending-http-request` event.  The main benefit there is that it should be less susceptible to churn as internals change (registering an event on a client and behavior based on returned values from events is a public API that can't change).  We could in theory make these changes to the develop branch and not have to make any (or fewer) changes to the tests in this branch.

botocore/awsrequest.py
jamesls: Yeah I'd prefer to simplify and remove this fallback if our tests are the only thing relying on it.

botocore/awsrequest.py
jamesls: Why's that?  We should scope this to something more specific than `Exception`.

botocore/awsrequest.py
jamesls: In that case, I'd prefer to remove this if nothing's relying on it.  If we are going to keep it at least extract it out to a method instead of a nested function.

jamesls:
> What do you think would be the best format for tracking these types of comparisons?

If you're asking where's the best place to document where we deviate from what requests does, I think it's worth putting in the code somewhere as either a comment or a docstring.

A lot of your previous comment would have been really helpful to read as a comment/docstring while reviewing the code.

botocore/http_session.py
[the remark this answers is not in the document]
joguSD: Copied the vendored CA Cert and updated this.

botocore/http_session.py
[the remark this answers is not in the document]
joguSD: Looking into refactoring the proxy code.

botocore/http_session.py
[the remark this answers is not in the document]
joguSD: Yep, that was the big difference I noticed as well.

botocore/awsrequest.py
joguSD: I removed it and just updated the tests that were dependent on it.

botocore/awsrequest.py
joguSD: Switched this to just use content and decode.

botocore/http_session.py
joguSD: Looks like urllib3 will never preload the content if the response is chunked. Considering this we'll have to do it externally.

botocore/endpoint.py
joguSD: I don't think the event idea will work as we're not just using the mocks to fake the response, we're also using it to assert things about the request.

botocore/http_session.py
joguSD: From what I can tell urllib3 will return the connection if we hit an exception when trying to urlopen or read from a response, see:
https://github.com/urllib3/urllib3/blob/master/src/urllib3/connectionpool.py#L644-L657
https://github.com/urllib3/urllib3/blob/master/src/urllib3/response.py#L354-L373

In my testing I also found this to be the case (the connection pool would have 10 connections again in the except block).

botocore/awsrequest.py
```diff
@@ -338,18 +332,38 @@ def _urljoin(endpoint_url, url_path):
...
+        self.method = method
+        self.url = url
+        self.headers = HTTPHeaders()
+        self.data = data
+        self.params = params
+        self.auth_path = auth_path
+        self.stream_output = stream_output
+
+        if headers is not None:
+            for key, value in headers.items():
```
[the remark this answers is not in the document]
joguSD: I'll check if this works but `self.headers` isn't a `dict` it's a `HTTPHeaders` object which has some weird behavior and this is how we were doing this previously so I just left it.

botocore/cacert.pem
```diff
@@ -0,0 +1,4433 @@
+
```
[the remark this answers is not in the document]
joguSD: I'm trying to keep changes to a minimum for now. I think this is something we could come back to.

botocore/awsrequest.py
```diff
@@ -428,27 +444,125 @@ def reset_stream(self):
...
+class _HeaderKey(object):
+    def __init__(self, key):
+        self._key = key
+        self._lower = key.lower()
+
+    def __hash__(self):
+        return hash(self._lower)
+
+    def __eq__(self, other):
+        return self._lower == other._lower
```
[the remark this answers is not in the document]
joguSD: I can do an `isinstance` check I suppose.

botocore/awsrequest.py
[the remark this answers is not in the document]
joguSD: Yeah this doesn't work.

tests/integration/test_client_http.py
```diff
@@ -0,0 +1,203 @@
...
+class BackgroundTaskFailed(Exception):
+    pass
+
+
+@contextmanager
+def background(target, args=(), timeout=10):
+    thread = threading.Thread(target=target, args=args)
+    thread.daemon = True
+    thread.start()
+    yield target
```
jamesls: You'll need a try/finally here.  For the tests that are expecting a failure being raised, the code after the yield won't be called.  It doesn't technically matter because they are background threads, but it's a pretty small change to fix this.

tests/integration/test_client_http.py
```diff
@@ -0,0 +1,203 @@
...
+                client.describe_regions()
+        except ConnectionClosedError:
+            pass
+        except BackgroundTaskFailed:
+            self.fail('Fake EC2 service was not called.')
+        else:
+            self.fail('Excepted exception was not thrown')
+
+
+def random_port():
```
jamesls: I would suggest doing what we did for chalice where we have the socket module find an unused port for us.  This function has a small chance of failing, but it'd be nice to have something that we know will always work: https://github.com/aws/chalice/blob/master/tests/functional/test_local.py#L95

tests/integration/test_client_http.py
```diff
@@ -0,0 +1,203 @@
...
+        try:
+            with background(run_server, args=(FakeEC2, self.port)):
+                client.describe_regions()
+        except ReadTimeoutError:
+            # Note: This was vendored requests ConnectionError & ReadTimeout
+            pass
+        except BackgroundTaskFailed:
+            self.fail('Fake EC2 service was not called.')
+        else:
+            self.fail('Excepted exception was not thrown')
```
jamesls: Expected.

tests/integration/test_client_http.py
```diff
@@ -0,0 +1,203 @@
...
+)
+
+
+class TestClientHTTPBehavior(unittest.TestCase):
+    def setUp(self):
+        self.port = random_port()
+        self.localhost = 'http://localhost:%s/' % self.port
+        self.session = botocore.session.get_session()
+
+    def test_can_proxy_https_request_with_auth(self):
```
jamesls: Do you think it's worth adding a test for plain http proxies?

tests/integration/test_client_http.py
```diff
@@ -0,0 +1,203 @@
...
+    def test_connect_timeout_exception(self):
+        config = Config(connect_timeout=0.2, retries={'max_attempts': 0})
+        client = self.session.create_client('ec2', endpoint_url=self.localhost,
+                                            config=config)
+
+        def no_accept_server():
+            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            sock.bind(('', self.port))
+            time.sleep(4)
```
jamesls: Can we use `threading.Events` instead?  We had similar sleeps in chalice and it ended up being flaky.  I think threading events will make these tests more robust.

botocore/http_session.py
```diff
@@ -0,0 +1,216 @@
...
+        self._max_pool_connections = max_pool_connections
+        self._proxy_managers = {}
+        self._manager = PoolManager(
+            strict=True,
+            timeout=self._timeout,
+            maxsize=self._max_pool_connections,
+        )
+
+    def _get_proxy_manager(self, proxy_url):
+        if proxy_url not in self._proxy_managers:
```
jamesls: I don't think there's a test for the case where the proxy url is already in the proxy managers dict.

botocore/http_session.py
```diff
@@ -0,0 +1,216 @@
...
+                proxy_headers=proxy_headers,
+                maxsize=self._max_pool_connections
+            )
+
+        return self._proxy_managers[proxy_url]
+
+    def _path_url(self, url):
+        parsed_url = urlparse(url)
+        path = parsed_url.path
+        if not path:
```
jamesls: Can you remind me again when `path` would be false?

botocore/http_session.py
```diff
@@ -0,0 +1,216 @@
...
+        if url.lower().startswith('https') and verify:
+            conn.cert_reqs = 'CERT_REQUIRED'
+            conn.ca_certs = get_cert_path(verify)
+        else:
+            conn.cert_reqs = 'CERT_NONE'
+            conn.ca_certs = None
+
+    def send(self, request):
+        try:
+            request_target = self._path_url(request.url)
```
jamesls: Could you break this method up?  It's fairly lengthy.

I also don't quite follow the logic where `request_target` is set to `self._path_url(request.url)`, but for the proxy case `request_target` is set to just `request.url` (line 170).  Maybe update the comment to explain why we need to use the  full url in the proxy case.

botocore/http_session.py
```diff
@@ -0,0 +1,216 @@
...
+            raise ConnectTimeoutError(endpoint_url=request.url, error=e)
+        except URLLib3ReadTimeoutError as e:
+            raise ReadTimeoutError(endpoint_url=request.url, error=e)
+        except ProtocolError as e:
+            raise ConnectionClosedError(
+                error=e,
+                request=request,
+                endpoint_url=request.url
+            )
+        except Exception as e:
```
jamesls: Are we sure we want to catch `Exception` here?  Is is possible to restrict this to a list of exceptions we know about?  This will consider even trivial bugs/typos as an `HTTPClientError`.

botocore/retryhandler.py
```diff
@@ -30,8 +30,8 @@
 # this mapping with more specific exceptions.
 EXCEPTION_MAP = {
     'GENERAL_CONNECTION_ERROR': [
-        ConnectionError, ClosedPoolError, Timeout,
-        EndpointConnectionError
+        ReadTimeoutError, ProxyConnectionError, EndpointConnectionError,
```
jamesls: Can we just use `HTTPClientError` here?

tests/integration/test_s3.py
```diff
@@ -813,18 +813,17 @@ def test_can_get_bucket_location(self):
     def test_request_retried_for_sigv4(self):
         body = six.BytesIO(b"Hello world!")
 
-        original_send = adapters.HTTPAdapter.send
+        original_send = Endpoint._send
```
jamesls: What other options do we have besides patching out a private method?

tests/unit/test_awsrequest.py
```diff
@@ -97,83 +97,6 @@ def tell(self):
...
-    def setUp(self):
-        self.tempdir = tempfile.mkdtemp()
-        self.request = AWSRequest(url='http://example.com')
-        self.prepared_request = self.request.prepare()
-        self.filename = os.path.join(self.tempdir, 'foo')
-
-    def tearDown(self):
-        shutil.rmtree(self.tempdir)
-
-    def test_should_reset_stream(self):
```
jamesls: Why is this test being removed?  It's still a method on the aws request object so it seems like we should still be testing it.

tests/unit/test_awsrequest.py
```diff
@@ -97,83 +97,6 @@ def tell(self):
...
-            fake_response = Mock()
-            fake_response.status_code = 307
-
-            # Then requests calls our reset_stream hook.
-            self.prepared_request.reset_stream_on_redirect(fake_response)
-
-            # The stream should now be reset.
-            self.assertEqual(body.tell(), 0)
-
-    def test_cannot_reset_stream_raises_error(self):
```
jamesls: In general, it seems like the coverage for `awsrequest` via `tests/unit/test_awsrequest.py` dropped with this PR.  It seems like we should at least keep the coverage we had before, and preferably raise the coverage given we're making changes to that file.

tests/integration/test_smoke.py
```diff
@@ -285,15 +285,14 @@ def test_client_can_retry_request_properly():
 
 def _make_client_call_with_errors(client, operation_name, kwargs):
     operation = getattr(client, xform_name(operation_name))
-    original_send = adapters.HTTPAdapter.send
-    def mock_http_adapter_send(self, *args, **kwargs):
+    original_send = Endpoint._send
+    def mock_endpoint_send(self, *args, **kwargs):
         if not getattr(self, '_integ_test_error_raised', False):
             self._integ_test_error_raised = True
-            raise ConnectionError("Simulated ConnectionError raised.")
```
jamesls: While I get that we're updating this to use our new exceptions, I do think it's useful to keep a few integration tests around that use the old exceptions to ensure that behavior isn't changed.

I would also prefer we leave the existing integration tests alone and add entirely new ones that verify we can catch the new exceptions.

This would serve as a nice safety check to ensure the same integration tests in develop can run against this PR.

tests/integration/test_smoke.py
```diff
@@ -285,15 +285,14 @@ def test_client_can_retry_request_properly():
 
 def _make_client_call_with_errors(client, operation_name, kwargs):
     operation = getattr(client, xform_name(operation_name))
-    original_send = adapters.HTTPAdapter.send
-    def mock_http_adapter_send(self, *args, **kwargs):
+    original_send = Endpoint._send
```
jamesls: Same comment as before, but we *really* shouldn't be touching internal methods in an integration test.

If we need to add better hooks into the endpoint to make it more testable via some public API, I'm fine with that.

tests/integration/test_client_http.py
joguSD: Done

tests/integration/test_client_http.py
joguSD: Done

tests/integration/test_client_http.py
joguSD: We could, but we would need a proxy server that understands it. I think at that point we're mostly testing urllib3, but it still could be worth it. I'll see what I can do.

tests/integration/test_client_http.py
joguSD: Yeah, this sounds good. Done.

botocore/http_session.py
joguSD: Broke it up a little bit more. I've updated the comment to note that proxying an HTTP requires the absolute URI so the host to connect to is part of the request target.

botocore/http_session.py
joguSD: I think so, it ensures we never directly leak any exceptions from our dependency or a mistake in our implementation. What's your concern with an exception from this layer being wrapped as an `HTTPClientError`?

botocore/retryhandler.py
joguSD: Possibly, I was thinking it might be nice to just have some of the `HTTPClientError` exception types extend from a `RetryableHTTPClientError` and then use that instead.

tests/unit/test_awsrequest.py
joguSD: These tests were for `reset_stream_on_redirect` method on the `AWSPreparedRequest` class that's been removed (since we aren't relying on / hooking into redirects from our client). I'll see if I can rework some of these tests to test `reset_stream` directly without relying on `reset_stream_on_redirect` calling it.

botocore/http_session.py
joguSD: Added a test for this.

botocore/http_session.py
jamesls: Primarily that actual bugs that raise things like `TypeError`/`ValueError` etc will be misrepresented as a `HTTPClientError`.  Ideally we'd want these things to fail fast and propagate all the way up the stack vs. possibly being retried and surfaced as a (somewhat confusing) `HTTPClientError`.

Note a huge deal, but I'd prefer to avoid it if possible.

tests/unit/test_awsrequest.py
joguSD: Added a bunch of tests and fixed some of these tests.

tests/integration/test_smoke.py
joguSD: Added an integration test for one of the exception types (Read timeout) to ensure in a real case we're catching the old exceptions. I also created some unit tests that verify a new raised exception will be caught by the old exception type.

tests/integration/test_smoke.py
joguSD: Yeah, I think this is another case for a before_send and after_send type of event.

kyleknap:
Also it would be good to have a change log entry along with upgrading notes to point at for gray areas when upgrading.

botocore/utils.py
```diff
@@ -1030,9 +1048,9 @@ class ContainerMetadataFetcher(object):
 
     def __init__(self, session=None, sleep=time.sleep):
         if session is None:
-            session = requests.Session()
-            session.trust_env = False
```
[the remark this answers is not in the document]
joguSD: Correct, URLLib3Session will not pull proxy settings from the environment.

requirements.txt
```diff
@@ -8,3 +8,4 @@ docutils>=0.10
 behave==1.2.5
 -e git://github.com/boto/jmespath.git@develop#egg=jmespath
 jsonschema==2.5.1
+urllib3>=1.20,<1.24
```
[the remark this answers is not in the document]
joguSD: 1.20 added a utility function to support things other than `select` which fixes some issues with file descriptors that are large. We use this function in our 100-Continue implementation so we require at least 1.20. 1.24 locks us to the current major version. 1.23 actually dropped python 3.3 support so you end up just getting 1.22 instead.

tests/integration/test_client_http.py
```diff
@@ -0,0 +1,205 @@
...
+        class AuthProxyHandler(ProxyHandler):
+            def validate_auth(self):
+                proxy_auth = self.headers.get('Proxy-Authorization')
+                return proxy_auth == 'Basic dXNlcjpwYXNz'
+
+        try:
+            with background(run_server, args=(AuthProxyHandler, self.port)):
+                client.describe_regions()
+        except BackgroundTaskFailed:
+            self.fail('Background task did not exit, proxy was not used.')
```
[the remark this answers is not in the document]
joguSD: You get a `botocore.exceptions.ProxyConnectionError` because of `OSError: Tunnel connection failed: 401 Unauthorized`

tests/unit/test_endpoint.py
```diff
@@ -307,98 +260,75 @@ def test_creates_endpoint_with_configured_url(self):
         self.assertEqual(endpoint.host, 'https://endpoint.url')
 
     def test_create_endpoint_with_default_timeout(self):
+        mock_session = Mock()
```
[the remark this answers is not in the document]
joguSD: Done

tests/unit/test_s3_addressing.py
```diff
@@ -40,9 +40,8 @@ def get_prepared_request(self, operation, params,
                              force_hmacv1=False):
         if force_hmacv1:
             self.session.register('choose-signer', self.enable_hmacv1)
-        with patch('botocore.endpoint.BotocoreHTTPSession') as \
-                mock_http_session:
-            mock_send = mock_http_session.return_value.send
+        # TODO: fix with stubber / before send event
+        with patch('botocore.endpoint.Endpoint._send') as mock_send:
```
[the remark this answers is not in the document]
joguSD: Agreed.

botocore/awsrequest.py
```diff
@@ -428,27 +444,133 @@ def reset_stream(self):
...
+            params = urlencode(list(params.items()), doseq=True)
+            self.url = '%s?%s' % (url, params)
+        else:
+            self.url = url
+
+    def prepare_headers(self, headers):
+        headers = headers or {}
+        self.headers = HeadersDict(headers.items())
+
+    def prepare_body(self, data):
```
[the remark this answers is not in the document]
joguSD: Not exactly. `requests` will do a lot more to try and determine the content-length. I kept it simple in ours to do more explicitly what is documented.

botocore/endpoint.py
```diff
@@ -212,32 +170,9 @@ def _get_response(self, request, operation_model, attempts):
...
-                return (None, better_exception)
-            elif self._looks_like_bad_status_line(e):
-                better_exception = ConnectionClosedError(
-                    endpoint_url=e.request.url, request=e.request)
-                return (None, better_exception)
-            else:
-                return (None, e)
+            http_response = self._send(request)
+        except HTTPClientError as e:
+            return (None, e)
```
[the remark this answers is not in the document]
joguSD: I've added logging to the URLLib3Session class.

joguSD:
@kyleknap  Added the appropriate notes and changelog.

botocore/parsers.py
```diff
@@ -250,7 +250,13 @@ def parse(self, response, shape):
         if isinstance(parsed, dict):
             response_metadata = parsed.get('ResponseMetadata', {})
             response_metadata['HTTPStatusCode'] = response['status_code']
-            response_metadata['HTTPHeaders'] = dict(response['headers'])
+            http_headers = {}
+            # Ensure that the http header keys are all lower cased. Older
+            # versions of urllib3 (< 1.11) would unintentionally do this for us
+            # (see urllib3#633). We need to do this conversion manually now.
+            for header in response['headers']:
```
jamesls: I'm not a fan of this.  It doesn't seem like it's the parsers responsibility to handle the lower casing, which I think is further hinted at by the fact that we're changing our parser because our http client changed (which I would not expect).

What was the reason for not fixing this in the custom headers dictionary class we made?

docs/source/index.rst
```diff
@@ -25,6 +25,19 @@ Contents:
...
+* The vendored versions of ``requests`` and ``urllib3`` are no longer being
+  used and have been replaced with a direct dependency on upstream ``urllib3``
+  and ``requests`` is no longer a dependency of ``botocore``.  While these
+  vendored dependencies are still in the ``botocore`` package they should not
+  be used as they will be removed in the future. Any code that imports from
+  ``botocore.vendored.requests.*`` should be updated accordingly. Specifically,
+  the use of ``botocore.vendored.requests.exceptions.*`` or
+  ``botocore.vendored.requests.packages.urllib3.exceptions.*`` must be updated
+  to the corresponding exception classes in ``botocore.exceptions``.
+* The version of ``urllib3`` used to make HTTP requests has been updated.
```
jamesls: Would be nice to put the urllib3 version in the upgrade notes so people don't have to track down the setup.py file, along with the version of requests/urllib3 we have vendored.

tests/unit/test_http_session.py
```diff
@@ -0,0 +1,196 @@
...
+            request_url=ANY,
+        )
+
+    def test_forwards_max_pool_size(self):
+        URLLib3Session(max_pool_connections=22)
+        self.pool_manager_cls.assert_called_with(
+            maxsize=22,
+            timeout=ANY,
+            strict=True,
+            ssl_context=ANY,
```
jamesls: We don't seem to be actually testing we're passing the correct ssl_context to the pool manager in any of our tests.  We could pass `None` and these tests would still pass.

I think we need at least one test that ensures it's a sane value.

botocore/httpsession.py
```diff
@@ -0,0 +1,279 @@
...
+    def __init__(self,
+                 verify=True,
+                 proxies=None,
+                 timeout=None,
+                 max_pool_connections=MAX_POOL_CONNECTIONS,
+    ):
+        self._verify = verify
+        self._proxy_config = ProxyConfiguration(proxies=proxies)
+        self._pool_classes_by_scheme = {
+            'http': botocore.awsrequest.AWSHTTPConnectionPool,
```
jamesls: It doesn't seem like this change is tested anywhere.  We should test either that we don't mess with the global namespace somehow.

botocore/parsers.py
joguSD: We're not fixing this in the custom headers dictionary because the previous headers object wasn't responsible for doing this either. This change in behavior would have happened even if we just upgraded requests. This behavior was never intentional on our part or urllib3s part, but is now part of our public contract for what we return. The parsers are what converts the response's header object into a `dict` and is what ultimately dictates the final format and structure of what we return from client operations.

docs/source/index.rst
joguSD: Do you mean the version range?

tests/unit/test_http_session.py
joguSD: Strictly speaking, `None` is a sane value (urllib3 will default it for us). I'm not even sure I could assert the type of object passed either... I'll see what I can do.

docs/source/index.rst
jamesls: Yeah.

botocore/httpsession.py
joguSD: Added a test to ensure that the global urllib3 defaults are not set to ours.

botocore/parsers.py
jamesls: Ok, I think that sounds reasonable.  Could you have this off to a utils function?  Some of the other dict manipulation functions are already in utils (e.g `merge_dicts`).

joguSD: Sure, sounds reasonable.

mikegrima:
Hello everyone! This PR broke ttps://github.com/spulec/moto really hard, and many projects depend on it for unit testing (See: https://github.com/spulec/moto/issues/1793).

Can some guidance be provided on how to fix? Pinning to an older version of boto3 can only be exercised in the short term.

joguSD:
@mikegrima Seems like the crux of it is these [lines](https://github.com/spulec/moto/blob/9c6e7f2ee797d1b878f26c0c795f9d72c46c1579/moto/core/models.py#L174-L175). Moto was monkey patching the vendored version of requests which is no longer used. I imagine a similar patch could be applied to our [`URLLib3Session` class](https://github.com/boto/botocore/blob/0670e4f241240ae3b4baee6767c3b658c5ae97f3/botocore/httpsession.py#L136) but I would highly recommend *not* doing that as we don't consider these as part of our public interface as they're undocumented and may change. 

We have a few [TODOs](https://github.com/boto/botocore/blob/f8365c3a5b2074f3999c918a84e76869e58311c3/tests/functional/test_apigateway.py#L35-L38) similar to this where we have a less than optimal solution to patching out the HTTP layer. We're looking into potentially implementing something via the event system to ignore the HTTP layer, but I'm still not 100% sure what will come out of that.

This is more of an open question to the moto team:
What functionality do you require so that you can avoid applying monkey patches?

mikegrima:
Hi @joguSD . I contribute to moto as many of our projects (internally and externally) make heavy use of it, but don't really work on the super low level components of it. That's more of a question for @spulec or @jackdanger.

From my perspective: I just want my tests to work and not be locked out of new features because we're stuck on an old version of boto.

mikegrima:
@joguSD : Can you comment on https://github.com/spulec/moto/issues/1793#issuecomment-416942466 and suggest the best approach for @spulec to take?

joguSD:
@mikegrima I think the 'best' approach is going to come down to how we expose the appropriate functionality to stub out the HTTP layer in `botocore`. I gave my opinion on various resolutions and proposed some changes we could possibly make in `botocore` to see if it would allow `moto` what they need to fix it last week, but it doesn't seem that anyone on the `moto` side has had any feedback on what I proposed.

Bartascnk (commented):
[Bartascnk@gmail.com]()

PR #1499, opened 2018-07-06
Added endpoint destructor to close http_session

dpitch40:
To resolve ResourceWarnings from requests lib about unclosed SSLSockets - see requests/requests#1882


dpitch40:
I'm not at the job where this came up anymore, so I can't easily tell if it's still relevant. I'll close it.

PR #1505, opened 2018-07-19
Ensure unicode client errors are handled correctly

joguSD:
If a service returns unicode in their response (for example, if  unicode was in the request) Python will attempt to represent the exception using ascii which fails for obvious reasons.

Fixes: https://github.com/boto/boto3/issues/1628

### DIFF
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -357,6 +357,11 @@ class ClientError(Exception):
             operation_name=operation_name,
             retry_info=retry_info,
         )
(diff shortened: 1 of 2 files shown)


ztane:
This applies to Python 2 only, naturally.

Just found out that Secrets Manager gives the following message when an invalid secret name is given: "Secrets Manager *can’t* find the specified secret.".. it has the \u2019 for an apostrophe.

nateprewitt:
@joguSD I don't think we need this anymore. Any objections to closing?

PR #1507, opened 2018-07-23
Timestamp format

stealthycoin:
Add support for more timestamp formats.

### DIFF
--- a/botocore/serialize.py
+++ b/botocore/serialize.py
@@ -39,8 +39,9 @@ and if a str/unicode type is passed in, it will be encoded as utf-8.
 """
 import re
 import base64
-from xml.etree import ElementTree
 import calendar
+import datetime
+from xml.etree import ElementTree
 
 from botocore.compat import six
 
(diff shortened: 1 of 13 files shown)


botocore/serialize.py
```diff
@@ -486,6 +495,15 @@ def _partition_parameters(self, partitioned, param_name,
             elif isinstance(param_value, bool):
                 partitioned['query_string_kwargs'][
                     key_name] = str(param_value).lower()
+            elif member.type_name == 'timestamp':
+                timestamp_format = member.serialization.get('timestampFormat')
```
joguSD: Any reason you chose not to do this?
```python
timestamp_format = member.serialization.get('timestampFormat', 'iso8601')
```
Dunno if we need a class level variable for this as this seems to be a pretty specific edge case in serialization.

botocore/serialize.py
```diff
@@ -486,6 +495,15 @@ def _partition_parameters(self, partitioned, param_name,
...
+                if timestamp_format is None:
+                    timestamp_format = self.QUERY_STRING_TIMESTAMP_FORMAT
+                timestamp_format = timestamp_format.lower()
```
joguSD: Don't think we need this, you made `_convert_timestamp_to_str` take care of lower-casing.

botocore/serialize.py
```diff
@@ -518,7 +536,12 @@ def _convert_header_value(self, shape, value):
         if shape.type_name == 'timestamp':
             datetime_obj = parse_to_aware_datetime(value)
             timestamp = calendar.timegm(datetime_obj.utctimetuple())
-            return self._timestamp_rfc822(timestamp)
+            timestamp_format = shape.serialization.get('timestampFormat')
```
joguSD: Similar feedback as above, any reason not to do:
```python
timestamp_format = shape.serialization.get('timestampFormat', 'rfc822')
return self._convert_timestamp_to_str(timestamp, timestamp_format)
```
I think I would prefer this as it seems the precedent is to use `_convert_timestamp_to_str` instead of calling the `_timestamp_<format>` methods directly.

At the very least we should be consistent in the current diff and have this be a similar style class variable like `HEADER_TIMESTAMP_FORMAT = 'rfc822'` or just set the default for the `get` call for both of these edge cases.

botocore/serialize.py
stealthycoin: Being consistent with everything else in this file.

PR #1512, opened 2018-07-31
Only auto-redirect 400s for s3 head operations if region header is present

joguSD:
We attempt to automatically redirect 400s for `Head*` operations to catch the case that a redirect was automatically followed by requests but fails authorization because it was signed with the wrong region. This can lead to infinite recursion in the case that we get a 400 but the redirection header is not present (e.g. the bucket doesn't exist). In this case we would fallback to attempting another `HeadBucket` to get the region header which leads to infinite recursion.

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -911,6 +911,7 @@ class S3RegionRedirector(object):
 
         error = response[1].get('Error', {})
         error_code = error.get('Code')
+        response_metadata = response[1].get('ResponseMetadata', {})
 
         # We have to account for 400 responses because
         # if we sign a Head* request with the wrong region,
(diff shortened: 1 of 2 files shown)


botocore/utils.py
```diff
@@ -911,22 +911,28 @@ def redirect_from_error(self, request_dict, response, operation, **kwargs):
...
         # body saying it's an "AuthorizationHeaderMalformed".
         is_special_head_object = (
             error_code in ['301', '400'] and
-            operation.name in ['HeadObject', 'HeadBucket']
+            operation.name == 'HeadObject'
+        )
+        is_special_head_bucket = (
+            error_code in ['301', '400'] and
+            operation.name == 'HeadBucket' and
+            'x-amz-bucket-region' in response_metadata.get('HTTPHeaders', {})
```
JordonPhillips: Why not check this for both HeadBucket and HeadObject?

PR #1515, opened 2018-08-10
New waiter definition for CloudFormation change set execution.

etanol:
With the introduction of CloudFormation change sets, the process of stack update can be instrumented to anticipate certain number of operations.

While there is a waiter defined for changeset creation, there is no waiter for changeset execution.  Adding a similar definition for symmetry.


etanol:
Travis build failing for Python 2.6, which should be fixed when merging #1514

Issue #1510, opened 2018-07-30
Cannot presign url for s3 delete_objects operation


ExProbitasFiducia:
So far as I can tell, the endpoint requires a query parameter, `delete` in the url.
See https://docs.aws.amazon.com/AmazonS3/latest/API/multiobjectdeleteapi.html for documentation.
`generate_presigned_post` has no way to provide this query parameter, and so cannot correctly sign a url to be used with this api.

`generate_presigned_url` does not work either, as it has no way to add what the hash of the body will be when signing the url.

Hopefully I am missing something here.

birdbrained:
think this is the relevant code: https://github.com/boto/botocore/blob/939fbebcae0e0519f040063d61f7dc354cd09dd8/botocore/signers.py#L457

JordonPhillips:
`generate_presigned_post` is specifically for the `put_objects` operation, what you want to use is `generate_presigned_url` with the POST method. The difference is the way that the signatures are generated, afaik S3 only supports PutObject with the `presigned_post` way of signing.

birdbrained:
That's the case I found, and what I've ended up using. There is still a problem here however, `generate_presigned_url` doesn't have a way to pass headers in like `generate_presigned_post` does.

birdbrained:
We were forced to write a modified version of this function in order to sign delete operations.
https://github.com/CenterForOpenScience/waterbutler/pull/351/files#diff-74bbac431e282383494483a4d0af3031R48

birdbrained:
See https://github.com/boto/botocore/pull/1516 for a pr that allows headers to be passed into the url signing process.

birdbrained:
Bump for a response

swetashre:
@ExProbitasFiducia - There is no parameter for custom headers in boto3's ```generate_presigned_url```. The headers get added based on the parameters that you would normally pass to the client's method.  

Please let me know if you have any questions.

ExProbitasFiducia:
https://github.com/CenterForOpenScience/waterbutler/pull/351/files#diff-e343fe1cb1d426f8d1f366e29e565554bb2d55cd77055b69e92e94b776d1216aR48-R114 for a link to the line numbers with the monkeypatch i'd written

ExProbitasFiducia:
@felliott @mfraezz might be interested in any progress here.

PR #1516, opened 2018-08-14
Pass headers into `generate_presigned_url`

ExProbitasFiducia:
Some operations that require a body must have a signature of that body
with which to sign the url; otherwise the body may be changed after the
url has been signed. Passing headers into `generate_presigned_url`
allows the signature to be correctly generated with the header that
provides a guarantee that the body of the request has not been changed.

See https://github.com/boto/botocore/issues/1510 for the relevant issue.

### DIFF
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -531,7 +531,7 @@ def add_generate_presigned_url(class_attributes, **kwargs):
 
 
 def generate_presigned_url(self, ClientMethod, Params=None, ExpiresIn=3600,
-                           HttpMethod=None):
+                           HttpMethod=None, Headers=None):
     """Generate a presigned url given a client, its method, and arguments
 
     :type ClientMethod: string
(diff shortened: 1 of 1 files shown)


birdbrained:
Example usage:
```
client.generate_presigned_url(
    'delete_objects',
    Params={
        'Bucket': bucket_name,
        'Delete': {
            'Objects': [{'Key': object['Key']} for object in objects_to_delete]
        }
    },
    Headers={
        'Content-Length': str(len(delete_payload)),
        'Content-MD5': compute_md5(BytesIO(delete_payload))[1],
        'Content-Type': 'text/xml'
    }
)
```

ExProbitasFiducia:
I think this is still relevant.

tim-finnigan:
Checking in — thanks for creating this PR and for your patience here. I brought this PR up with the team and the consensus was that more information was required regarding use cases and what this proposed change is trying to address. I reopened the related issue (https://github.com/boto/botocore/issues/1510) for further discussion going forward and we can revisit this PR as needed.

PR #1522, opened 2018-08-21
Create redirect branch

JordonPhillips:
This branch removes as much of the docs as possible to allow rtd to
redirect for us. Whatever is left is redirected by javascript to the
new doc site's index.


JordonPhillips:
Holding off on this since the rtd folks said they're going to try making an nginx level change over in rtfd/readthedocs.org#4546

PR #1524, opened 2018-08-22
Migrate to service ids for event names

JordonPhillips:
This moves events from using some combination of service name and
enpoint prefix to always using the new service id. This maintains
compatiblity with the previous state, but nonetheless updates all
internal usage.

It also migrates the aliasing strategy to using a wrapper class so custom
emitters are also aliased.


botocore/utils.py
```diff
@@ -897,6 +973,11 @@ def deep_merge(base, extra):
         base[key] = extra[key]
 
 
+def translate_service_id(service_id):
```
kyleknap: I would prefer we call this something like `get_service_event_name` as `translate_service_id` is a little vague and there are other ways to translate service id's. I see this as being similar to the `get_service_module_name` function.

botocore/utils.py
```diff
@@ -53,6 +52,83 @@
     ConnectTimeoutError,
 )
 S3_ACCELERATE_WHITELIST = ['dualstack']
+# In switching events from using service name / endpoint prefix to service
+# id, we have to preserve compatibility. This maps the instances where either
+# is different than the transformed service id.
+EVENT_ALIASES = {
```
kyleknap: What's your thoughts on adding some functional tests for these aliases? I was thinking it would be similar to what we did for `sagemaker` but be more programmatic. Specifically, we would have two tests where one emits events for service names that do not match the service id and the other emits events for endpoint prefixes that do not match the service id. Then in each yielded case we would register a handler to an event that uses the appropriate alias and make sure it gets fired. The list for the cases to run would also be statically generated. I think this can give us a lot of confidence in these aliasings.

botocore/hooks.py
```diff
@@ -343,48 +343,86 @@ def __copy__(self):
...
+            if '.' not in old_part:
+                event_parts = event_name.split('.')
+                try:
+                    # Theoretically a given event name could have the same part
+                    # repeated, but in practice this doesn't happen
+                    event_parts[event_parts.index(old_part)] = new_part
+                except ValueError:
+                    continue
+                new_name = '.'.join(event_parts)
+            elif old_part in event_name:
```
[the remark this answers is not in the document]
JordonPhillips: I ended up just fixing it to be strictly correct, and also added tests

botocore/hooks.py
```diff
@@ -343,48 +343,86 @@ def __copy__(self):
...
             aliased_event_name, handler, unique_id, unique_id_uses_count
         )
 
     def _alias_event_name(self, event_name):
         for old_part, new_part in self._event_aliases.items():
-            if old_part in event_name:
+
+            # We can't simply do a string replace for everything, otherwise we
+            # might end up tranlating substrings that we never intended to
+            # translate. So for most cases we try to break the event up and
```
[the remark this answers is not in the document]
JordonPhillips: done

botocore/signers.py
```diff
@@ -58,14 +58,22 @@ class fires two events scoped to a service and operation name:
     :type event_emitter: :py:class:`~botocore.hooks.BaseEventHooks`
     :param event_emitter: Extension mechanism to fire events.
 
+    :type translated_service_id: str
```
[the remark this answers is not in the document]
JordonPhillips: done

botocore/signers.py
```diff
@@ -58,14 +58,22 @@ class fires two events scoped to a service and operation name:
...
+    :param translated_service_id: The service id for the service. This
```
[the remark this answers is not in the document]
JordonPhillips: done

botocore/utils.py
JordonPhillips: done

tests/functional/test_event_alias.py
```diff
@@ -0,0 +1,612 @@
...
+
+
+def _assert_handler_called(client_name, event_part):
+    hook_calls = []
+
+    def _hook(**kwargs):
+        hook_calls.append(kwargs['event_name'])
+
+    session = _get_session()
+    session.register('creating-client-class.%s' % event_part, _hook)
```
kyleknap: So I was thinking we would have two separate tests:

* `test_event_alias_service_name`: This would use the `create-client-class` event and we would do the `_assert_handler_called` for both `service_name` and `service_id`

* `test_event_alias_endpoint_prefix`: This would use one of the events that was based on endpoint prefexies and we would do the `_assert_handler_called` for both `endpoint_prefix` and `service_id`

The main reasons that I suggest this is that:
1) We are not testing any of the aliasing for service names
2) It is not representative to register on the `endpoint_prefix` for the `creating-client-class` since it used to rely on the `service_name` before.

botocore/utils.py
```diff
@@ -129,6 +129,29 @@
     "streams.dynamodb": "dynamodb-streams",
     "tagging": "resource-groups-tagging-api"
 }
+# These endpoint prefixes represent multiple services. Prior to the service
+# id change, you had to register the endpoint prefix so there was bleed
+# over. We need to preserve that behavior for backwards compatibility.
+DUPLICATE_EVENTS = {
```
kyleknap: I think we should be injecting these into the EventAliaser and also add unit tests for this functionality with your own custom duplicate events dict.

botocore/hooks.py
```diff
@@ -352,79 +352,86 @@ def __init__(self, event_emitter, event_aliases=None):
...
+            )
+            if location is not None:
+                for service_id in service_ids:
+                    new_parts = copy.copy(event_parts)
+                    new_parts[location[0]:location[1]] = [service_id]
+                    # No need to alias here because we're already using a
+                    # mapping of endpoint prefix to service id.
+                    new_event_name = '.'.join(new_parts)
+                    logger.debug(
+                        'Registering %s for backwards compatibility' %
```
kyleknap: It might be good to put the original event name in this log message as well

tests/functional/test_event_alias.py
JordonPhillips: We talked offline about this. I've added another yield to test registering with the client name, and registering against an event that used to be endpoint_prefix is covered by the sagemaker tests (since there's no generic place to test that).

botocore/hooks.py
```diff
@@ -378,8 +416,12 @@ def unregister(self, event_name, *args, **kwargs):
...
 
     def _register(self, func, event_name, *args, **kwargs):
-        for endpoint_prefix, service_ids in DUPLICATE_EVENTS.items():
-            event_parts = event_name.split('.')
+        event_parts = event_name.split('.')
+        if not event_parts[0] in self._duplicate_event_whitelist:
+            self._alias(func, event_name, *args, **kwargs)
+            return
+
+        for endpoint_prefix, service_ids in self._duplicate_events.items():
```
kyleknap: We may want to rename this not to be specific to `endpoint_prefix` and `service_id` as the `duplicate_events` may not necessarily be related to endpoint prefixes and service ID's.

botocore/hooks.py
```diff
@@ -345,10 +346,47 @@ def __copy__(self):
...
+        :param event_aliases: A mapping event subsection to event
+            subsection, where the key is a subsection to be replaced by the
+            value. These can cover multiple event parts, but must be entire
+            event parts. So if you had the event ``foo.bar.baz`` you could
+            replace ``bar`` or ``foo.bar`` but not ``ba``. By default,
+            all services whose service name or endpoint prefix were
+            different at the time of the service id migration are aliased.
+
+        :type duplicate_events: dict
+        :param duplicate_events: A mapping of endpoint prefix to a list of
```
kyleknap: I think we should reword the docstring to not be specific toward endpoint prefixes and serviceId's. I think it is fine to mention it as the default but still should be described as a general mapping sort of like what you did for `event_aliases` and `duplicate_events`.

docs/source/development/upgrading.rst
```diff
@@ -0,0 +1,98 @@
...
+
+To get the proper event name, you will need to get the service's event name,
+like so, substituting in whatever service you are using::
+
+    from botocore import Session
+    from botocore.utils import get_service_event_name
+
+    session = Session()
+    client = session.create_client('elbv2')
+    service_id = client.meta.service_model.service_id
```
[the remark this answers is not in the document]
JordonPhillips: That would be an absolutely massive table that  would need to be updated every time a service releases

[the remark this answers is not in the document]
JordonPhillips: Also that makes it really hard to use if there's no blessed programmatic way to get it.

docs/source/development/upgrading.rst
```diff
@@ -0,0 +1,98 @@
+Upgrading Notes
+===============
+
+Notes to refer to when upgrading ``botocore`` versions.
+
+1.12.0
```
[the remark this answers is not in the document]
JordonPhillips: 👍

docs/source/development/upgrading.rst
```diff
@@ -0,0 +1,98 @@
...
+The botocore event system was changed to emit events based on the service id
+rather than the endpoint prefix or service name.
+
+Why Was The Change Was Made
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This was done to handle several issues that were becoming increasingly
+problematic:
+
+* Services changing their endpoint prefix would cause some registered events to
```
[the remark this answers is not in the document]
JordonPhillips: 👍

docs/source/development/upgrading.rst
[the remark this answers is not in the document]
kyleknap: I was a little ambiguous here. I meant a table for the services that shared an endpoint prefix and will require an update to the event name. Not every singe service. So something like:

Service | Old event name | New event name
------- | ---------------- | -----------------
`neptune` | `before-call.rds` | `before-call.neptune`


We do not have do every event either (i.e. `after-call`, `request-created`, etc.). We can just stick to a single event when doing this table and list out the public-facing events that are affected.

botocore/args.py
```diff
@@ -23,6 +23,7 @@
 from botocore.signers import RequestSigner
 from botocore.config import Config
 from botocore.endpoint import EndpointCreator
+from botocore.utils import get_service_event_name
```
joguSD: This is unsued.

docs/source/development/upgrading.rst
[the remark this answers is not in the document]
kyleknap: >> Also that makes it really hard to use if there's no blessed programmatic way to get it.

I don't disagree there should be a programmatic way to get the event name, but I do not think it should be this helper function. I think there are better ways to expose this. So I would prefer we do not make this public by documenting it and suggesting people to use it, and instead explore the interfaces we could expose and commit to one we like.

botocore/utils.py
```diff
@@ -897,6 +973,14 @@ def deep_merge(base, extra):
         base[key] = extra[key]
 
 
+def get_service_event_name(service_id):
```
joguSD: -1 on this being an external method.
+1 on this being on the service model more directly, perhaps even making a `ServiceID ` object that has blessed transformations like this. I could see this being transformed in other ways too.

botocore/handlers.py
```diff
@@ -273,6 +274,8 @@ def _needs_s3_sse_customization(params, sse_member_prefix):
 def register_retries_for_service(service_data, session,
                                  service_name, **kwargs):
     loader = session.get_component('data_loader')
+    service_id = service_data.get('metadata', {}).get('serviceId')
+    service_event_name = get_service_event_name(service_id)
     endpoint_prefix = service_data.get('metadata', {}).get('endpointPrefix')
```
joguSD: Is this relevant anymore?

botocore/handlers.py
```diff
@@ -273,6 +274,8 @@ def _needs_s3_sse_customization(params, sse_member_prefix):
 def register_retries_for_service(service_data, session,
                                  service_name, **kwargs):
     loader = session.get_component('data_loader')
+    service_id = service_data.get('metadata', {}).get('serviceId')
```
joguSD: Wonder if there's anything we can do to make this more inline with how all of the other events are setup.

joguSD:
Also, does this have any CLI implications?

botocore/handlers.py
JordonPhillips: yes

docs/source/index.rst
```diff
@@ -25,6 +25,105 @@ Contents:
...
+
+* Services changing their endpoint prefix would cause some registered events to
+  no longer fire (but not all).
+* New services that launch using an endpoint that another service is using
+  won't be able to be uniquely selected. There are a number of cases of this
+  already.
+* Services whose client name and endpoint prefix differed would require two
+  different strings if you want to register against all events.
+
+How Do I Know If I'm Impacted
```
jamesls: What about cases where the service id changed?  Not technically impacted but it might be good to document which event names have changed with the switchover to service ids so customers can start migrating their code and not rely on the aliasing (which will eventually go away).

botocore/client.py
```diff
@@ -105,6 +108,8 @@ def _load_service_model(self, service_name, api_version=None):
 
     def _register_retries(self, client):
         endpoint_prefix = client.meta.service_model.endpoint_prefix
+        service_id = client.meta.service_model.service_id
+        service_event_name = get_service_event_name(service_id)
```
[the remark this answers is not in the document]
JordonPhillips: The problem with putting it on the service model is that it really isn't a direct representation of the service model. Now I'd like to put it in the client meta, but that would still require us to use this function in places where we haven't created that.

botocore/client.py
```diff
@@ -585,9 +598,10 @@ def _make_api_call(self, operation_name, api_params):
         request_dict = self._convert_to_request_dict(
             api_params, operation_model, context=request_context)
 
+        service_id = get_service_event_name(self._service_model.service_id)
```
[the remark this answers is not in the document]
JordonPhillips: It's the name of the service used in events.

docs/source/index.rst
JordonPhillips: I do mention that briefly later on, but I can make it more explicit.

botocore/client.py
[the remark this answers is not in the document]
kyleknap: Talked to @jamesls and @joguSD offline about this. I think it is fine to put it on the service model under the argument that valid transformations of the service id is an extension of the service model. However, the name of these transformations should not be client specific. We talked about this idea as well, but I would be fine with setting the `ServiceModel.service_id` to a class like this that subclasses from `str`:
```python
class ServiceID(str):
    def hyphenize(self):
        return self.replace(' ', '-').lower()
```
The nice part about making a separate ServiceID object is that we will not have to bloat the `ServiceModel` class with properties like `hyphened_service_id` as right now there are five different accepted ways the service id can be represented. So if we ever needed a new one, we can just add a method to this object. Also we will not have to pollute a lot of the modules with the current `get_service_event_name` import from the `utils` module. Thoughts?

kyleknap:
The updates to the upgrading notes look fine. For me, there is just remaining conversation related to how we consume service id for the event emission that needs to get resolved.

botocore/client.py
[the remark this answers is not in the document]
JordonPhillips: Seems fine to me. I talked about something like this before @jamesls and @joguSD chimed in.

JordonPhillips:
updated

botocore/signers.py
```diff
@@ -40,7 +39,7 @@ class fires two events scoped to a service and operation name:
     and disabling signing per operation.
 
 
-    :type service_id: str
+    :type service_id: botocore.model._ServiceId
```
kyleknap: I think this is not an underscore here

PR #1527, opened 2018-08-23
Modified lines-from-chunks generator to ensure support for "\r\n"

nlw0:
The current code produces empty strings sometimes if the input file follows this unconventional and bizarre "\r\n" newline standard, that nonetheless is still found in some legacy systems. The proposed code tries to make sure whatever `splitlines()` understands as newline is supported.


nlw0:
Created an up-to-date rebased version: #1533

PR #1533, opened 2018-08-24
Modified lines-from-chunks generator to ensure support for "\r\n"

nlw0:
The current code produces empty strings sometimes if the input file follows this unconventional and bizarre "\r\n" newline standard, that nonetheless is still found in some legacy systems. The proposed code tries to make sure whatever `splitlines()` understands as newline is supported.

This PR is the same as #1527, but rebasing after lots of changes upstream.

### DIFF
--- a/tests/unit/test_response.py
+++ b/tests/unit/test_response.py
@@ -149,15 +149,6 @@ class TestStreamWrapper(unittest.TestCase):
         with self.assertRaises(ReadTimeoutError):
             stream.read()
 
-
-class FakeRawResponse(six.BytesIO):
-    def stream(self, amt=1024, decode_content=None):
-        while True:
-            chunk = self.read(amt)
-            if not chunk:
-                break
-            yield chunk
-
     def test_streaming_line_abstruse_newline_standard(self):
         for chunk_size in range(1, 30):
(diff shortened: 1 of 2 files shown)


tests/unit/test_response.py
```diff
@@ -158,6 +158,15 @@ def stream(self, amt=1024, decode_content=None):
                 break
             yield chunk
 
+    def test_streaming_line_abstruse_newline_standard(self):
```
jamesls: It looks like the rebase got messed up here.  This test is now part of the `FakeRawResponse` class so it won't be run.  I think this was meant to be part of `TestStreamWrapper`.

nlw0: Oops, sorry for the messy rebase. Just made a new patch

nlw0:
@jamesls @joguSD  I moved back the test. Is it OK now?

nlw0:
Still waiting for news, maybe it's because I didn't "resolve the conversation"?... Sorry!

nlw0:
So, is there anything else I might do to land this?

jamesls:
Sorry for the delay, looking at this now.

PR #1539, opened 2018-08-29
Ensure the SigV4 signer signs the host as all lowercase

joguSD:
Succeeds: https://github.com/boto/botocore/pull/1538

Basically, if the endpoint has mixed casing we sign using the mixed case but it's actually sent across the wire as all lowercase (this if the `Host` header is not explicitly set).

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -180,7 +180,11 @@ class SigV4Auth(BaseSigner):
             if lname not in SIGNED_HEADERS_BLACKLIST:
                 header_map[lname] = value
         if 'host' not in header_map:
-            header_map['host'] = self._canonical_host(request.url)
+            # Ensure we sign the lowercased version of the host, as that
+            # is what will ultimately be sent on the wire.
+            # TODO: We should set the host ourselves, instead of relying on our
+            # HTTP client to set it for us.
+            header_map['host'] = self._canonical_host(request.url).lower()
         return header_map
 
     def _canonical_host(self, url):
--- a/tests/integration/test_smoke.py
+++ b/tests/integration/test_smoke.py
@@ -91,6 +91,7 @@ SMOKE_TESTS = {
  's3': {'ListBuckets': {}},
  'sdb': {'ListDomains': {}},
  'ses': {'ListIdentities': {}},
+ 'shield': {'GetSubscriptionState': {}},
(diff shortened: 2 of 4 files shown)


tests/integration/test_smoke.py
```diff
@@ -91,6 +91,7 @@
  's3': {'ListBuckets': {}},
  'sdb': {'ListDomains': {}},
  'ses': {'ListIdentities': {}},
+ 'shield': {'GetSubscriptionState': {}},
```
kyleknap: I think it may be good to put an explicit integration test where we upper case parts of the endpoint just in-case if `shield` changes their endpoint in the `endpoint.json` to be lowercase.

tests/functional/test_auth_sigv4.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kyleknap: Is there a reason we are not putting this in the unit test suite under `auth/test_sigv4.py`?

botocore/auth.py
```diff
@@ -180,7 +180,7 @@ def headers_to_sign(self, request):
             if lname not in SIGNED_HEADERS_BLACKLIST:
                 header_map[lname] = value
         if 'host' not in header_map:
-            header_map['host'] = self._canonical_host(request.url)
+            header_map['host'] = self._canonical_host(request.url).lower()
```
kyleknap: It would be good to add a comment on why we are doing this.

tests/integration/test_smoke.py
joguSD: Do you have a suggestion for where to put that?

tests/functional/test_auth_sigv4.py
joguSD: I added a new file because the other test file seems specific to the tests generated from the aws4_testsuite directory. I could go either way on this.

kyleknap: I think we should be put it in the`unit` test directory. Not too picky on which file, I just suggested that one because it looked similar to some of the lines that you had.

tests/integration/test_smoke.py
joguSD: Added a test to `integration/test_client.py`

PR #1545, opened 2018-08-31
docfix: "cals" -> "calls"

tedder:
fix spelng.


joguSD:
Unfortunately we don't accept changes to the service models as we pull these documents from upstream.

Issue #1462, opened 2018-05-12
Operations/methods missing paginators 


kapilt:
I brought this up at the openspace at pycon, there are lots of client methods missing paginator metadata. I went ahead and coded up simple script to identify all the missing paginators in botocore json sdk metadata.

```python
import boto3
import yaml

def main():
    session = boto3.Session()
    services = session.get_available_services()
    op_missing_paginator = {}
    for s in services:
        c = session.client(s)
        op_to_py = {v: k for k, v in c._PY_TO_OP_NAME.items()}
        for op in c.meta.service_model.operation_names:
            opm = c.meta.service_model.operation_model(op)
            if opm.output_shape is None:
                continue
            if 'NextToken' not in opm.output_shape.members:
                continue
            py_name = op_to_py[opm.name]
            if not c.can_paginate(py_name):
                op_missing_paginator.setdefault(s, []).append(py_name)
    print(yaml.safe_dump(op_missing_paginator, default_flow_style=False))
```

Which results in the following output

```yaml
acm-pca:
- list_certificate_authorities
- list_tags
alexaforbusiness:
- list_device_events
- search_address_books
- search_contacts
application-autoscaling:
- describe_scheduled_actions
appstream:
- describe_directory_configs
- describe_fleets
- describe_image_builders
- describe_sessions
- describe_stacks
- list_associated_fleets
- list_associated_stacks
autoscaling:
- describe_load_balancer_target_groups
- describe_load_balancers
autoscaling-plans:
- describe_scaling_plan_resources
- describe_scaling_plans
budgets:
- describe_budgets
- describe_notifications_for_budget
- describe_subscribers_for_notification
clouddirectory:
- list_incoming_typed_links
- list_object_children
- list_object_parents
- list_outgoing_typed_links
cloudformation:
- describe_account_limits
- describe_change_set
- list_change_sets
- list_stack_instances
- list_stack_set_operation_results
- list_stack_set_operations
- list_stack_sets
cloudhsm:
- list_hapgs
- list_hsms
- list_luna_clients
cloudtrail:
- list_public_keys
- list_tags
cloudwatch:
- get_metric_data
codepipeline:
- list_webhooks
cognito-identity:
- list_identities
- list_identity_pools
- lookup_developer_identity
cognito-idp:
- admin_list_groups_for_user
- admin_list_user_auth_events
- list_groups
- list_identity_providers
- list_resource_servers
- list_user_pool_clients
- list_user_pools
- list_users_in_group
[...]

PR #1548, opened 2018-09-04
Add missing RDS paginators

brandond:
Looks like someone added a paginator for `DescribeDBClusters` but didn't add paginators for any of the other Cluster actions.

### DIFF
--- a/botocore/data/rds/2014-10-31/paginators-1.json
+++ b/botocore/data/rds/2014-10-31/paginators-1.json
@@ -1,16 +1,52 @@
 {
   "pagination": {
+    "DescribeCertificates": {
+      "input_token": "Marker",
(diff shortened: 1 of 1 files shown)


brandond:
@stealthycoin @JordonPhillips seems like #1462 is a better long-term approach, as opposed to asking folks to manually maintain the paginators separate from the service definitions.

JordonPhillips:
Auto-generating isn't foolproof - if the response contains anything but a list and the pagination token then it isn't safe to guess the result key. I know a while ago we actually did go through and auto-gen a bunch of them, but we haven't done so since. It would probably be worth doing that. In the meantime, this pr looks good so I'll go ahead and merge it.

PR #1549, opened 2018-09-06
Code for adding payload hash in canonical headers for non S3 requests.

pritha-srivastava:
Signed-off-by: Pritha Srivastava <prsrivas@redhat.com>


JordonPhillips:
What is the reason for adding this in?

swetashre:
@pritha-srivastava - Thank you for the contribution. Following up on old PR. Could you please provide some more context on the reason for adding the changes ?

PR #1552, opened 2018-09-10
Adding s3_list_objects_encoding_type_url handler to ListObjectsV2

drocamor:
This is to support https://github.com/aws/aws-cli/pull/3549 which has the AWSCLI use ListObjectsV2 for operations that list buckets.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -730,41 +730,37 @@ def decode_list_object(parsed, context, **kwargs):
     # Amazon S3 includes this element in the response, and returns encoded key
     # name values in the following response elements:
     # Delimiter, Marker, Prefix, NextMarker, Key.
-    if parsed.get('EncodingType') == 'url' and \
-                    context.get('encoding_type_auto_set'):
-        # URL decode top-level keys in the response if present.
-        top_level_keys = ['Delimiter', 'Marker', 'NextMarker']
-        for key in top_level_keys:
-            if key in parsed:
-                parsed[key] = unquote_str(parsed[key])
-        # URL decode nested keys from the response if present.
-        nested_keys = [('Contents', 'Key'), ('CommonPrefixes', 'Prefix')]
-        for (top_key, child_key) in nested_keys:
-            if top_key in parsed:
-                for member in parsed[top_key]:
-                    member[child_key] = unquote_str(member[child_key])
-
+    _decode_list_object(
+        top_level_keys=['Delimiter', 'Marker', 'NextMarker'],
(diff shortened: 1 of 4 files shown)


botocore/handlers.py
```diff
@@ -880,6 +880,8 @@ def remove_subscribe_to_shard(class_attributes, **kwargs):
 
     ('before-parameter-build.s3.ListObjects',
      set_list_objects_encoding_type_url),
+    ('before-parameter-build.s3.ListObjectsV2',
+     set_list_objects_encoding_type_url),
```
jamesls: You also need to register the decoding after the response is received (line 945).

One handler injects the encoding type (`set_list_objects_encoding_type_url`), and the other handler (`decode_list_object`) decodes objects if we injected the encoding type param.  We need both.

drocamor: Ah got it.

botocore/handlers.py
```diff
@@ -943,6 +945,7 @@ def remove_subscribe_to_shard(class_attributes, **kwargs):
     ('before-parameter-build.route53', fix_route53_ids),
     ('before-parameter-build.glacier', inject_account_id),
     ('after-call.s3.ListObjects', decode_list_object),
+    ('after-call.s3.ListObjectsV2', decode_list_object),
```
kyleknap: So unfortunately. We cannot purely reuse the `decode_list_objects`. There are different keys that need to be decoded according to the [docs](https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html). Specifically, we need to decode: `Delimiter`, `Prefix`, `ContinuationToken`, `Key`, and `StartAfter`. I would recommend trying to see if we can consolidate the logic and make a new `decode_list_objects_v2` handler that decodes it for the V2 response shape and use that instead.

botocore/handlers.py
```diff
@@ -745,6 +745,27 @@ def decode_list_object(parsed, context, **kwargs):
                     member[child_key] = unquote_str(member[child_key])
 
 
+def decode_list_object_v2(parsed, context, **kwargs):
```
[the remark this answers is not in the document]
drocamor: Got it!

tests/unit/test_handlers.py
```diff
@@ -873,6 +873,69 @@ def test_decode_list_objects_with_delimiter(self):
         handlers.decode_list_object(parsed, context=context)
         self.assertEqual(parsed['Delimiter'], u'\xe7\xf6s% asd\x08 c')
 
+    def test_decode_list_objects_v2(self):
```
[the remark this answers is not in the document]
drocamor: Added an integration test.

PR #1556, opened 2018-09-14
Add the send-request event

joguSD:
This adds a `send-request` event which will allow an event handler to return a response and short circuit botocore's default HTTP stack.

This also adds a test utility to stub the HTTP layer and replaces quite a few places where we were patching out internal methods.

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -340,6 +340,97 @@ def _urljoin(endpoint_url, url_path):
     return reconstructed
 
 
+class AWSRequestPreparer(object):
+    """
+    This class performs preparation on AWSRequest objects similar to that of
+    the PreparedRequest class does in the requests library. However, the logic
+    has been boiled down to meet the specific use cases in botocore. Of note
+    there are the following differences:
+        This class does not heavily prepare the URL. Requests performed many
+        validations and corrections to ensure the URL is properly formatted.
+        Botocore either performs these validations elsewhere or otherwise
+        consistently provides well formatted URLs.
+
+        This class does not heavily prepare the body. Body preperation is
+        simple and supports only the cases that we document: bytes and
+        file-like objects to determine the content-length. This will also
+        additionally prepare a body that is a dict to be url encoded params
+        string as some signers rely on this. Finally, this class does not
+        support multipart file uploads.
+
+        This class does not prepare the method, auth or cookies.
+    """
+    def prepare(self, original):
+        method = original.method
+        url = self._prepare_url(original)
+        body = self._prepare_body(original)
+        headers = self._prepare_headers(original, body)
+        stream_output = original.stream_output
+
+        return AWSPreparedRequest(method, url, headers, body, stream_output)
+
+    def _prepare_url(self, original):
+        url = original.url
+        if original.params:
+            params = urlencode(list(original.params.items()), doseq=True)
+            url = '%s?%s' % (url, params)
+        return url
+
+    def _prepare_headers(self, original, prepared_body):
+        headers = HeadersDict(original.headers.items())
+
+        # If the transfer encoding or content length is already set, use that
+        if 'Transfer-Encoding' in headers or 'Content-Length' in headers:
+            return headers
+
+        # Ensure we set the content length when it is expected
+        if original.method not in ('GET', 'HEAD', 'OPTIONS'):
+            length = self._determine_content_length(prepared_body)
+            if length is not None:
+                headers['Content-Length'] = str(length)
+            else:
+                # Failed to determine content length, using chunked
+                # NOTE: This shouldn't ever happen in practice
+                headers['Transfer-Encoding'] = 'chunked'
+
+        return headers
+
+    def _prepare_body(self, original):
+        """Prepares the given HTTP body data."""
+        body = original.data
+        if body == b'':
+            body = None
+
+        if isinstance(body, dict):
+            params = list(body.items())
+            body = urlencode(params, doseq=True)
+
+        return body
+
+    def _determine_content_length(self, body):
+        # No body, content length of 0
+        if not body:
+            return 0
+
+        # Try asking the body for it's length
+        try:
+            length = len(body)
+            return length
+        except (AttributeError, TypeError) as e:
+            pass
+
+        # Try getting the length from a seekable stream
+        if hasattr(body, 'seek') and hasattr(body, 'tell'):
+            orig_pos = body.tell()
+            body.seek(0, 2)
+            end_file_pos = body.tell()
+            body.seek(orig_pos)
+            return end_file_pos - orig_pos
+
+        # Failed to determine the length
+        return None
+
+
 class AWSRequest(object):
     """Represents the elements of an HTTP request.
 
@@ -347,6 +438,8 @@ class AWSRequest(object):
     boiled down to meet the specific use cases in botocore. That being said this
     class (even in requests) is effectively a named-tuple.
     """
+
+    _REQUEST_PREPARER = AWSRequestPreparer()
 
     def __init__(self,
                  method=None,
@@ -383,51 +476,34 @@ class AWSRequest(object):
 
     def prepare(self):
         """Constructs a :class:`AWSPreparedRequest <AWSPreparedRequest>`."""
-        return AWSPreparedRequest(self)
+        return self._REQUEST_PREPARER.prepare(self)
 
     @property
     def body(self):
-        p = AWSPreparedRequest(self)
-        p.prepare_body(self.data)
-        if isinstance(p.body, six.text_type):
-            p.body = p.body.encode('utf-8')
-        return p.body
+        body = self.prepare().body
+        if isinstance(body, six.text_type):
+            body = body.encode('utf-8')
+        return body
 
 
 class AWSPreparedRequest(object):
-    """Represents a prepared request.
+    """Represents a finalized prepared request ready to be sent on the wire.
 
-    This class is originally inspired by requests.models.PreparedRequest, but
-    has been boiled down to meet the specific use cases in botocore. Of note
-    there are the following differences:
-        This class does not heavily prepare the URL. Requests performed many
-        validations and corrections to ensure the URL is properly formatted.
-        Botocore either performs these validations elsewhere or otherwise
-        consistently provides well formatted URLs.
+    Requests at this stage should be treated as final, and the properties of
+    the request should not be modified.
 
-        This class does not heavily prepare the body. Body preperation is
-        simple and supports only the cases that we document: bytes and
-        file-like objects to determine the content-length. This will also
-        additionally prepare a body that is a dict to be url encoded params
-        string as some signers rely on this. Finally, this class does not
-        support multipart file uploads.
-
-        This class does not prepare the method, auth or cookies.
-
-    :ivar method: HTTP Method
+    :ivar method: The HTTP Method
     :ivar url: The full url
     :ivar headers: The HTTP headers to send.
     :ivar body: The HTTP body.
     :ivar stream_output: If the response for this request should be streamed.
-    :ivar original: The original AWSRequest
     """
-    def __init__(self, original):
-        self.method = original.method
-        self.prepare_url(original.url, original.params)
-        self.prepare_headers(original.headers)
-        self.prepare_body(original.data)
-        self.stream_output = original.stream_output
-        self.original = original
+    def __init__(self, method, url, headers, body, stream_output):
+        self.method = method
+        self.url = url
+        self.headers = headers
+        self.body = body
+        self.stream_output = stream_output
 
     def __repr__(self):
         fmt = (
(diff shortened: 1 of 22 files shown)


botocore/awsrequest.py
```diff
@@ -340,6 +340,97 @@ def _urljoin(endpoint_url, url_path):
...
+
+    def _prepare_body(self, original):
+        """Prepares the given HTTP body data."""
+        body = original.data
+        if body == b'':
+            body = None
+
+        if isinstance(body, dict):
+            params = list(body.items())
+            body = urlencode(params, doseq=True)
```
JordonPhillips: Wait really? urllb3 takes a list of tuples instead of a dict?

botocore/awsrequest.py
```diff
@@ -340,6 +340,97 @@ def _urljoin(endpoint_url, url_path):
...
+            return headers
+
+        # Ensure we set the content length when it is expected
+        if original.method not in ('GET', 'HEAD', 'OPTIONS'):
+            length = self._determine_content_length(prepared_body)
+            if length is not None:
+                headers['Content-Length'] = str(length)
+            else:
+                # Failed to determine content length, using chunked
+                # NOTE: This shouldn't ever happen in practice
```
JordonPhillips: Maybe log something here to help debugging in case it ever does

botocore/awsrequest.py
```diff
@@ -340,6 +340,97 @@ def _urljoin(endpoint_url, url_path):
...
+
+    def _determine_content_length(self, body):
+        # No body, content length of 0
+        if not body:
+            return 0
+
+        # Try asking the body for it's length
+        try:
+            length = len(body)
+            return length
```
JordonPhillips: This could be one line

botocore/awsrequest.py
```diff
@@ -340,6 +340,97 @@ def _urljoin(endpoint_url, url_path):
...
+        # Try getting the length from a seekable stream
+        if hasattr(body, 'seek') and hasattr(body, 'tell'):
+            orig_pos = body.tell()
+            body.seek(0, 2)
+            end_file_pos = body.tell()
+            body.seek(orig_pos)
+            return end_file_pos - orig_pos
+
+        # Failed to determine the length
+        return None
```
JordonPhillips: You should probably log something here too, or throw an error. Will this happen in practice?

botocore/awsrequest.py
```diff
@@ -383,51 +476,34 @@ def __init__(self,
...
+        return self._REQUEST_PREPARER.prepare(self)
 
     @property
     def body(self):
-        p = AWSPreparedRequest(self)
-        p.prepare_body(self.data)
-        if isinstance(p.body, six.text_type):
-            p.body = p.body.encode('utf-8')
-        return p.body
+        body = self.prepare().body
```
JordonPhillips: So why do we bother with any of the other prep if we only use the body

tests/unit/test_endpoint.py
```diff
@@ -149,33 +155,41 @@ def test_retry_on_socket_errors(self):
...
+                         'needs-retry.ec2.DescribeInstances')
         self.assertEqual(call_args[3][0][0],
+                         'request-created.ec2.DescribeInstances')
+        self.assertEqual(call_args[4][0][0],
+                         'send-request.ec2.DescribeInstances')
+        self.assertEqual(call_args[5][0][0],
                          'needs-retry.ec2.DescribeInstances')
 
-    def test_retry_attempts_added_to_response_metadata(self):
+    def est_retry_attempts_added_to_response_metadata(self):
```
JordonPhillips: `est`?

botocore/endpoint.py
```diff
@@ -171,7 +171,12 @@ def _get_response(self, request, operation_model, attempts):
                 'url': request.url,
                 'body': request.body
             })
-            http_response = self._send(request)
+            service_id = operation_model.service_model.service_id.hyphenize()
+            event_name = 'send-request.%s.%s' % (service_id, operation_model.name)
```
JordonPhillips: How about calling this `before-send` to be more descriptive and more consistent with our other event names

botocore/awsrequest.py
joguSD: This is `urlencode`, so yes?

botocore/awsrequest.py
joguSD: This should never happen in practice as we only accept `bytes`/`text` and `file-like` objects, it would only have if we got something else for a body, which means they disabled validation.

tests/__init__.py
```diff
@@ -359,3 +362,66 @@ def assert_url_equal(url1, url2):
...
+        self.reset()
+
+    def reset(self):
+        self._requests = []
+        self._responses = []
+
+    def add_response(self, response):
+        self._responses.append(response)
+
+    def create_response(self, url=None, status=200, headers=None, body=None):
```
kyleknap: What's your thoughts on just having a single `add_response()` method? It would optionally take the `url`, `status`, `headers`, `body`, and `response` (the `response` would be mutually exclusive with the `url`, `status`, etc.). Then we would also expose an `add_error()` method that accepts an exception. The nice part about this is that it more closely follows the `Stubber` interface and it is very explicit how to provide responses/exceptions.

tests/__init__.py
```diff
@@ -359,3 +362,66 @@ def assert_url_equal(url1, url2):
...
+        raw = RawResponse(body)
+        response = AWSResponse(url, status, headers, raw)
+        self._responses.append(response)
+
+    @property
+    def requests(self):
+        return self._requests
+
+    @property
+    def request_count(self):
```
kyleknap: While this is not a public interface, I think it would be fine to just expose the `requests` property to keep the interface surface area small in case we ever want to make this abstraction public.

tests/__init__.py
```diff
@@ -359,3 +362,66 @@ def assert_url_equal(url1, url2):
...
+            yield client
+        finally:
+            events.unregister('send-request', self)
+
+    def __call__(self, **kwargs):
+        request = kwargs.get('request')
+        self._requests.append(request)
+        if self._responses:
+            response = self._responses.pop(0)
+            if isinstance(response, Exception):
```
kyleknap: Internally, it may be worth exposing the items of the list as a dictionary/namedtuple with the following format (or something like this):

```
{
    "type": "<error|response>",
    "payload": response
}
```
So instead of having to check what the instance is, you can just look at the `type`. Furthermore, it should be pretty easy to do this if we had explicit methods for adding responses and errors.

tests/__init__.py
```diff
@@ -359,3 +362,66 @@ def assert_url_equal(url1, url2):
...
+                raise response
+            else:
+                return response
+        else:
+            raise Exception('Need to set responses')
```
kyleknap: We should probably throw our own custom exception for this code path (like we do for the `UnStubbedResponseError` with the `Stubber` class).

tests/__init__.py
```diff
@@ -359,3 +362,66 @@ def assert_url_equal(url1, url2):
...
+    @property
+    def requests(self):
+        return self._requests
+
+    @property
+    def request_count(self):
+        return len(self._requests)
+
+    @contextlib.contextmanager
+    def wrap_client(self, client):
```
kyleknap: What are your thoughts about exposing an `__enter__` and `__exit__` on the `HTTPStubber` that does this and have the `client` be passed into the `__init__()`?  It is more consistent to what is done in the `Stubber` class. Then you do not have to worry about passing in the client for every `with` statement:

```python
class TestSomething(unittest.TestCase):
    def setUp(self):
        self.client = session.create_client(...)
        self.http_stubber = HTTPStubber(self.client)


    def test_with_stubbed_layer(self):
        with self.http_stubber:
            self.http_stubber.add_response()
            self.client.make_call()
            # assert something
```

botocore/awsrequest.py
joguSD: Looks like we got an issue related to this: https://github.com/boto/botocore/issues/1561

tests/__init__.py
kyleknap: I am fine with how you have it now. This comment was more related to how to generalize it a bit more if there was an `add_error` to the stubber.

tests/__init__.py
```diff
@@ -359,3 +362,66 @@ def assert_url_equal(url1, url2):
...
+
+class RawResponse(BytesIO):
+    def stream(self, **kwargs):
+        contents = self.read()
+        while contents:
+            yield contents
+            contents = self.read()
+
+
+class BotocoreHTTPStubber(object):
```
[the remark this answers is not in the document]
kyleknap: Hmm maybe we should call it `ClientHTTPStubber` then? I think `Botocore` is still a little ambiguous because you can still make HTTP requests from the botocore layer via other mechanisms like credential resolving with IMDS and container metadata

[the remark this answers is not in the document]
joguSD: Yeah, I think that makes sense as this method does not stub those.

tests/__init__.py
```diff
@@ -359,3 +361,64 @@ def assert_url_equal(url1, url2):
...
+class ClientHTTPStubber(object):
+    def __init__(self, client):
+        self.reset()
+        self._client = client
+
+    def reset(self):
+        self.requests = []
+        self.responses = []
+
+    def create_response(self, url=None, status=200, headers=None, body=None):
```
jamesls: Why not use the verbs from the original stubber (`add_response()`)?  It seems confusing to have something called `create_response` without actually returning anything.

tests/__init__.py
```diff
@@ -359,3 +361,64 @@ def assert_url_equal(url1, url2):
...
+        if url is None:
```
jamesls: These strings can just be default values,

tests/__init__.py
```diff
@@ -359,3 +361,64 @@ def assert_url_equal(url1, url2):
...
+
+    def __enter__(self):
+        self.start()
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.stop()
+
+    def __call__(self, **kwargs):
+        request = kwargs.get('request')
```
jamesls: I don't follow when `request` wouldn't be in the kwargs.  Why not use `__call__(self, request, **kwargs)`.

tests/__init__.py
joguSD: I'm fine with this, I originally had an `add_response` and `create_response`.

tests/__init__.py
joguSD: It will always be there, not sure why I'm doing the get to be honest.

tests/__init__.py
```diff
@@ -359,3 +361,58 @@ def assert_url_equal(url1, url2):
...
+            contents = self.read()
+
+
+class ClientHTTPStubber(object):
+    def __init__(self, client):
+        self.reset()
+        self._client = client
+
+    def reset(self):
+        self.requests = []
```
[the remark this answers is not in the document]
joguSD: As a general rule, I would agree. Though in this case, I want `reset` to basically do what `__init__` does so it makes sense to put all of the initialization logic in `reset` and just call it from `__init__`. I don't feel too strongly on this, mostly did it to remove duplication and ensure `reset` gets you back to a clean slate in case this class ever got more complicated.

tests/__init__.py
```diff
@@ -359,3 +361,58 @@ def assert_url_equal(url1, url2):
...
+
+class HTTPStubberException(Exception):
+    pass
+
+
+class RawResponse(BytesIO):
+    def stream(self, **kwargs):
+        contents = self.read()
+        while contents:
+            yield contents
```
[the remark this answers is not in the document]
joguSD: I think you're right, but from a general correctness perspective I would rather the stream method work like this just in case that assumption ever changed.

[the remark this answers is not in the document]
JordonPhillips: Right now a customer doesn't really have any way to make it behave differently since the interface asks for bytes. What about letting it take an object? That would also let customers do things like emulate a slow connection.

In any case, if you want this behavior you should have a test for it.

tests/__init__.py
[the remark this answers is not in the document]
JordonPhillips: I'm -0 on keeping it since it doesn't actually violate a PEP

tests/__init__.py
```diff
@@ -359,3 +361,58 @@ def assert_url_equal(url1, url2):
...
+            contents = self.read()
+
+
+class ClientHTTPStubber(object):
```
JordonPhillips: Wait, why is this in `tests` and not `stub`? Customers won't actually be able to use it here. Also, docs / examples should be added.

kyleknap: We did not want to make it a public interface quite yet.

tests/__init__.py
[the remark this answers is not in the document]
joguSD: Seeing as there's no real way to currently test it and the fact that this isn't a public interface yet I think it's okay to leave it as is.

tests/__init__.py
joguSD: This PR is really for the `before-send` event. This class is currently something we're using to exercise the event and ensure that it works for what we need it to do. Making this HTTP stubber an actual public stubber is a feature we can come back to later.

botocore/awsrequest.py
joguSD: I've opened another PR to address this: https://github.com/boto/botocore/pull/1566

docs/source/topics/events.rst
```diff
@@ -18,33 +18,90 @@ to events.
...
+  ``'before-call.<service>.<operation>'``
+
+:Description:
+  This event is emitted when an operation is called and provides access to the
+  parameters that will be passed when calling the operation.
+
+:Keyword Arguments Emitted:
+
+  :type params: dict
+  :param params: A dictionary representing the keyword arguments that will be
```
kyleknap: Is this referring to the client parameters or the request dict? I'm pretty sure this will be the request dict.

docs/source/topics/events.rst
```diff
@@ -18,33 +18,90 @@ to events.
...
+                 used to invoke the operation call.
+
+  :type context: dict
+  :param context: A dictionary representing the operation context. Additional
+                  information can be stored in this context to share state
+                  between events for the operation call.
+
+:Expected Return Value: A tuple of (http_response, parsed_response) where the
```
kyleknap: We can return `None` here as well. We may want to follow the wording from `before-send` on how returning will affect the behavior.

docs/source/topics/events.rst
```diff
@@ -18,33 +18,90 @@ to events.
...
-     - When an operation is being called (``Operation.call``).
-     - ``operation`` - The newly created :class:`botocore.operation.Operation`
-       object.
-     - Ignored.
-   * - **after-call.<service>.<operation>**
-     - After an operation has been called, but before the response is parsed.
-     - ``http_response`` - The HTTP response, ``parsed`` - The parsed data.
-     - Ignored.
+The list below shows all of the events emitted by botocore.  In some cases, the
+events are listed as ``event-name.<service>.<operations>``, in which
```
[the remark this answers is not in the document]
kyleknap: Agreed we should be documenting to use service id

PR #1560, opened 2018-09-18
More specific service id error

JordonPhillips:
This gives a better error when the service id is missing from a
model. For example, if you called from the CLI you would currently
see an error like this:

```
$ aws secretsmanager help
'NoneType' object has no attribute 'replace'
```

Now you would see something like this:

```
$ aws secretsmanager help
The model being used for the service secretsmanager is missing the
serviceId metadata property, which is required.
```

### DIFF
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -506,8 +506,13 @@ class UndefinedModelAttributeError(Exception):
     pass
 
 
-class MissingServiceIdError(UndefinedModelAttributeError, BotoCoreError):
+class MissingServiceIdError(UndefinedModelAttributeError):
     fmt = (
         "The model being used for the service {service_name} is missing the "
         "serviceId metadata property, which is required."
     )
+
+    def __init__(self, **kwargs):
+        msg = self.fmt.format(**kwargs)
+        Exception.__init__(self, msg)
+        self.kwargs = kwargs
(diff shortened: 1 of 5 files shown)


botocore/exceptions.py
```diff
@@ -475,3 +475,15 @@ class MD5UnavailableError(BotoCoreError):
 
 class MetadataRetrievalError(BotoCoreError):
     fmt = "Error retrieving metadata: {error_msg}"
+
+
+class UndefinedModelAttributeError(Exception):
+    pass
+
+
+class MissingServiceIdError(UndefinedModelAttributeError, BotoCoreError):
```
jamesls: I'm a little hesitant to also subclass from BotoCoreError as well.  It can technically change which `except` class you hit, though I realize how unlikely it is that code is catching `UndefinedModelAttributeError`.

I think the safer change is just to subclass `UndefinedModelAttributeError` for now.

JordonPhillips: Well that's why I had a test that specifically catches `UndefinedModelAttributeError`.

JordonPhillips:
man what is up with codecov

botocore/exceptions.py
jamesls: It's not about a test, it's about code that's using this.  If I have an except clause that catches both `BotoCoreError` and `UndefinedModelAttributeError`, this PR can potentially change which except clause I hit.

If we're going to do that I think we should have a good reason to do this.  Why does it need to also subclass from `BotoCoreError`.   If it's just so we can utilize it's `__init__` method which does the `self.fmt.format(**kwargs)` I don't feel that's a strong enough rationale.

JordonPhillips: I did end up changing it since I didn't have strong feelings one way or another

Issue #1561, opened 2018-09-18
Very slow (>60s) API calls with large params


medmunds:
Problem: `boto3.client("ses").send_raw_email(..., RawMessage=b"message w/ 7MB attachment")` takes over a minute in AWS Lambda—and only the last few seconds are actually posting to the API endpoint. Most of the time is spent in `urlencode` while _repeatedly_ re-running `AWSPreparedRequest.prepare_body()`.

This likely affects any botocore client API using sigv4 auth, though it'll only be noticeable with larger params like SendRawEmail's RawMessage.

### Issue

I'm implementing an email forwarder using AWS SES and Lambda, and couldn't figure out why my Lambda function was timing out on messages with attachments. X-Ray showed the problem was in SendRawEmail:

<img width="1095" alt="aws x-ray ses sendrawemail" src="https://user-images.githubusercontent.com/639984/45712410-21d9f280-bb41-11e8-85fb-2dc7cc5d869c.png">

A little local profiling identified two problems:

1. Python's `urlencode` is not particularly fast. Addressing that seems out of scope for boto.

2. botocore's `AWSPreparedRequest.prepare_body()` is called *9 times* on the same params dict, re-urlencoding those large params every time. This seems excessive.

[Python 3.6; botocore 1.12.3 local dev <strike>or 1.10.74</strike> and Lambda; 256MB Lambda memory allocation]

### Fix

Eight of the nine `prepare_body()` calls are due to accessing `request.body` four times in `SigV4Auth`—and the `AWSRequest.body` property implementation unintentionally calls `prepare_body()` twice each time. The ninth is the one you'd expect via `AWSRequest.prepare()` as the request is sent.

I'm not sure the best way to proceed:

* No matter what, the [extraneous `prepare_body()` call](https://github.com/boto/botocore/blob/1.12.5/botocore/awsrequest.py#L391) should be removed from `AWSRequest.body`. (The body is already prepared by `AWSPreparedRequest()`'s init, on the preceding line.)

* A simple improvement for this particular case is to cache `request.body` in auth (e.g., #1478). But `AWSRequest.body` is still an expensive property to compute, which could easily trip up other code the same way. (We could add caching into the body prop implementation itself, but it's not clear to me how to reliably invalidate it.)

[...]

PR #1566, opened 2018-09-24
Improve prepare_body performance

joguSD:
When performing a SigV4 calculation we access the `AWSRequest.body` property many times which can potentially be expensive depending on the type of body being sent. This removes multiple calls to the body property to avoid calculating it many times, as the value doesn't change within the signer. 

This also removes an additional call to `prepare_body` that is no longer needed after some refactoring to how the body property works. The request is prepared in the `AWSPreparedRequest` constructor now meaning that calculating the `prepare_body` again is no longer needed.

Combined this reduces the number of times we prepare the body from 9 to 2, once when we calculate the signature and once when we actually send it. There's more profiling information in the linked issues.

Closes: #1478 
Fixes: #1561

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -268,20 +268,21 @@ class SigV4Auth(BaseSigner):
             # When payload signing is disabled, we use this static string in
             # place of the payload checksum.
             return UNSIGNED_PAYLOAD
-        if request.body and hasattr(request.body, 'seek'):
-            position = request.body.tell()
-            read_chunksize = functools.partial(request.body.read,
+        request_body = request.body
+        if request_body and hasattr(request_body, 'seek'):
+            position = request_body.tell()
+            read_chunksize = functools.partial(request_body.read,
                                                PAYLOAD_BUFFER)
             checksum = sha256()
             for chunk in iter(read_chunksize, b''):
                 checksum.update(chunk)
             hex_checksum = checksum.hexdigest()
-            request.body.seek(position)
+            request_body.seek(position)
             return hex_checksum
-        elif request.body:
+        elif request_body:
             # The request serialization has ensured that
             # request.body is a bytes() type.
-            return sha256(request.body).hexdigest()
+            return sha256(request_body).hexdigest()
         else:
             return EMPTY_SHA256_HASH
 


joguSD:
@jamesls 

1) I think caching the body property of the `AWSPreparedRequest` makes sense, and we more or less do that now. However, this is the `AWSRequest` class and I don't think caching it makes sense as the underlying `data` that produces the `body` *could* change. I agree the burden to known if a property is expensive to compute or not isn't optimal, but I don't think we can change that at this point.

2) [This](https://github.com/boto/botocore/blob/b9da4b4f52ac1556f33fc6cb4418b225bd107419/botocore/awsrequest.py#L490-L491) is the potentially expensive part. For the query serializers we ultimately produce a `dict(str, str)` that is later turned into bytes when preparing the request. I think ideally we would actually just have the query serializers produce the bytes directly rather than relying on request preparation to do this conversion. This way we treat the query format much more like json or xml. However, that's a much larger refactoring.

In my testing, I couldn't reproduce individual `prepare_body` calls taking more than half a second for 8MB body when run locally, but on lambda I was able to reproduce the `prepare_body` results in #1561.

So I decided to do some digging into what was actually taking so long in the `urlencode` and found that [this line](https://github.com/python/cpython/blob/fad6af2744c0b022568f7f4a8afc93fed056d4db/Lib/urllib/parse.py#L835) is the culprit. For some reason the `join` takes **significantly** longer on lambda than it does when ran locally. As pointed out in #1561, this scales with the memory setting in lambda.

For anyone curious, this is the lambda function I used to test:
```python
import time
import contextlib

@contextlib.contextmanager
def measure_time(msg):
    start_perf = time.perf_counter()
    start_proc = time.process_time()
    try:
        yield
    finally:
        end_perf = time.perf_counter()
        end_proc = time.process_time()
        print(msg, 'perf', end_perf - start_perf)
        print(msg, 'proc', end_proc - start_proc)

def run_test(event, context):
    with measure_time('join'):
        body = ''.join(byte for byte in ('A' * 8 * 1024 * 1024))

if __name__ == '__main__':
    run_test({}, {})

```

joguSD:
After a rebase this is now just the original commit in #1478

PR #1568, opened 2018-09-26
Properly handle error responses with null or not present body

theredcat:
When a HTTP response contain no body or the body field is None, botocore crashes with this error :

```
'NoneType' object has no attribute 'strip'
[ ...]
  File "/[...]/dist-packages/botocore/parsers.py", line 237, in parse
    if self._is_generic_error_response(response):
  File "/[...]/dist-packages/botocore/parsers.py", line 270, in _is_generic_error_response
    body = response['body'].strip()
```

The typical trigger for this error is a timeout that generate an internal http 599

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -271,8 +271,11 @@ class ResponseParser(object):
         # To prevent this case from happening we first need to check
         # whether or not this response looks like the generic response.
         if response['status_code'] >= 500:
-            body = response['body'].strip()
-            return body.startswith(b'<html>') or not body
+            if 'body' not in response or response['body'] is None:
+                return True
+            else:
+                body = response['body'].strip()
+                return body.startswith(b'<html>') or not body
 
     def _do_generic_error_parse(self, response):
         # There's not really much we can do when we get a generic
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -273,9 +273,9 @@ class ResponseParser(object):
         if response['status_code'] >= 500:
             if 'body' not in response or response['body'] is None:
                 return True
-            else:
-                body = response['body'].strip()
-                return body.startswith(b'<html>') or not body
+
+            body = response['body'].strip()
+            return body.startswith(b'<html>') or not body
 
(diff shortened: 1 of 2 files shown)


botocore/parsers.py
```diff
@@ -271,8 +271,11 @@ def _is_generic_error_response(self, response):
         # To prevent this case from happening we first need to check
         # whether or not this response looks like the generic response.
         if response['status_code'] >= 500:
-            body = response['body'].strip()
-            return body.startswith(b'<html>') or not body
+            if 'body' not in response or response['body'] is None:
+                return True
+            else:
```
SeleniumK: Unnecessary else :)

SeleniumK:
https://github.com/boto/botocore/issues/1570

I opened an issue yesterday for this, thank you for the pr! :D

theredcat:
@SeleniumK You're welcome ! I've removed the useless else:)

joguSD:
@theredcat Do you have any more context on when this happens?
What service is returning this? 
Does the API being called have a large request or is otherwise complex?

Just trying to understand the underlying cause here a little more.
The PR looks good to me though.

theredcat:
@joguSD I use https://github.com/thumbor/thumbor with an S3 backend and few weeks ago, AWS had trouble on the eu-west-1 region that impacted my thumbor instances. S3 api calls were hiting timeouts and the error in my first message start popping up in my logs.

I have no proxy.

joguSD:
@theredcat After some discussion in the other thread I've tracked this down to the `tornado-botocore` package. The unfortunate thing is the way that's implemented is via patching out `botocore` internals to use an async HTTP client and thus has very strict `botocore` requirements. To pick up this patch you'll need to get the following repos also updated to use newer versions of `botocore` that include this patch.
See:
https://github.com/thumbor-community/aws/blob/7b40255be6c846c6ed27a2ab037ed74d6c8d2c84/setup.py#L60-L61
https://github.com/nanvel/tornado-botocore/blob/07542ddc032c6fcf49e819e48c543d5a3ad953b2/requirements.txt#L1

Issue #1403, opened 2018-03-13
NoCredentialsError when reading creds from instance metadata


shuoh:
We have a script running on EC2 that keeps making DynamoDB API calls. Since recently the script always fails with the error below after successfully running for 1-2 hours :

```
  File "/home/ec2-user/buddyai_chatbot_deployed_app/app/venv/local/lib/python2.7/site-packages/boto3/resources/factory.py", line 520, in do_action
    response = action(self, *args, **kwargs)
  File "/home/ec2-user/buddyai_chatbot_deployed_app/app/venv/local/lib/python2.7/site-packages/boto3/resources/action.py", line 83, in __call__
    response = getattr(parent.meta.client, operation_name)(**params)
  File "/home/ec2-user/buddyai_chatbot_deployed_app/app/venv/local/lib/python2.7/site-packages/botocore/client.py", line 324, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/ec2-user/buddyai_chatbot_deployed_app/app/venv/local/lib/python2.7/site-packages/botocore/client.py", line 609, in _make_api_call
    operation_model, request_dict)
  File "/home/ec2-user/buddyai_chatbot_deployed_app/app/venv/local/lib/python2.7/site-packages/botocore/endpoint.py", line 143, in make_request
    return self._send_request(request_dict, operation_model)
  File "/home/ec2-user/buddyai_chatbot_deployed_app/app/venv/local/lib/python2.7/site-packages/botocore/endpoint.py", line 168, in _send_request
    request = self.create_request(request_dict, operation_model)
  File "/home/ec2-user/buddyai_chatbot_deployed_app/app/venv/local/lib/python2.7/site-packages/botocore/endpoint.py", line 152, in create_request
    operation_name=operation_model.name)
  File "/home/ec2-user/buddyai_chatbot_deployed_app/app/venv/local/lib/python2.7/site-packages/botocore/hooks.py", line 227, in emit
    return self._emit(event_name, kwargs)
  File "/home/ec2-user/buddyai_chatbot_deployed_app/app/venv/local/lib/python2.7/site-packages/botocore/hooks.py", line 210, in _emit
    response = handler(**kwargs)
  File "/home/ec2-user/buddyai_chatbot_deployed_app/app/venv/local/lib/python2.7/site-packages/botocore/signers.py", line 90, in handler
    return self.sign(operation_name, request)
[...]

Issue #1049, opened 2016-10-12
Handle errors/garbage coming back for metadata server creds


kapilt:
We've noticed that the ec2 metadata server isn't particularly stable with any concurrency, which in turn can cause exceptions in botocore, and in our app lockups.

metadata server concurrency issues
- https://ahmetalpbalkan.com/blog/comparison-of-instance-metadata-services/
- https://github.com/hashicorp/terraform/issues/6222

botocore traceback on retrieval

```
  File "/usr/local/custodian/local/lib/python2.7/site-packages/boto3/session.py", line 256, in client
    aws_session_token=aws_session_token, config=config)
  File "/usr/local/custodian/local/lib/python2.7/site-packages/botocore/session.py", line 815, in create_client
    credentials = self.get_credentials()
  File "/usr/local/custodian/local/lib/python2.7/site-packages/botocore/session.py", line 443, in get_credentials
    'credential_provider').load_credentials()
  File "/usr/local/custodian/local/lib/python2.7/site-packages/botocore/credentials.py", line 1072, in load_credentials
    creds = provider.load()
  File "/usr/local/custodian/local/lib/python2.7/site-packages/botocore/credentials.py", line 477, in load
    metadata = fetcher.retrieve_iam_role_credentials()
  File "/usr/local/custodian/local/lib/python2.7/site-packages/botocore/utils.py", line 187, in retrieve_iam_role_credentials
    val = json.loads(val)
  File "/usr/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 369, in decode
    raise ValueError(errmsg("Extra data", s, end, len(s)))
ValueError: Extra data: line 9 column 2 - line 9 column 14 (char 846 - 858)
```

Ideally botocore should catch this on relevant error codes and retry with delay.

Source issue: https://github.com/capitalone/cloud-custodian/issues/506

JordonPhillips:
Is the metadata service sending back an error response or just garbled data?

kapilt:
its unclear, from the other links it looks like its sending back a non 200 error code and non json.

joshuadfranklin:
Specifically a  HTTP/1.1 429 Too Many Requests 
"@phinze I was able to confirm that the metadata service does start sending 429s when it's being hit heavily. In my concurrency test it seems to break down after 25 concurrent connections are made."

jamesls:
[...]

PR #1569, opened 2018-09-26
Update IMDS to fix some issues

kyleknap:
Fixes: https://github.com/boto/botocore/issues/1403
Fixes: https://github.com/boto/botocore/issues/1049

Specifically with this we add the ability to retry on empty and invalid json credential responses. I also did quite a bit of refactoring on the `InstanceMetadataFetcher` to simplify it down. The behavior (except for the additional retry cases) should be identical in terms of how it currently got used in botocore. To help with the review, here are the places where I changed a code path:

* Eliminated [branch](https://github.com/boto/botocore/blob/17aa8d5674ab2fa429d9783983f6774118921717/botocore/utils.py#L295-L297) for if the first IMDS response returns something ending in a `\`. This was essentially dead code. If it were hit, it would have been a runtime error because it did not match the signature.
* Eliminated the [checking](https://github.com/boto/botocore/blob/develop/botocore/utils.py#L304-L306) if it starts with a `{`. We now always `json.loads` it. If we hit the code path if it did not get JSON loaded, it would have failed in trying to do the [dictionary access](https://github.com/boto/botocore/blob/develop/botocore/utils.py#L318-L325).
* Removed [looping](https://github.com/boto/botocore/blob/develop/botocore/utils.py#L293-L294) through the fields for the initial IMDS call. There will ever only be one role attached to an instance. Therefore, we would never hit cases where we had to make more than one subsequent call to IMDS. In the case where there is no role attached to the instance, we get a 4xx response which raises a retries exceeded error.
* Flattened the [data](https://github.com/boto/botocore/blob/17aa8d5674ab2fa429d9783983f6774118921717/botocore/utils.py#L318) dictionary. Since we are no longer iterating through fields, there is no reason to have a field key.

I also ran all of the botocore integration tests (using IMDS as the credential provider) and those test pass (there were a couple intermittent failures, but it was not because of IMDS). I also manually tested the following cases:

* No role attached to the instance. Botocore appropriately raises an error that it cannot find credentials.
* A role attached but does not have EC2 as a trusted entity. We get a key error on this (we probably make a better error in the future) because IMDS returns an error response that is formatted as JSON and the HTTP status code is 200. I did not change this to avoid changing too much as this same issue existed prior to refactoring.

Let me know what you all think.

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -263,15 +263,14 @@ class InstanceMetadataFetcher(object):
     def retrieve_iam_role_credentials(self):
         try:
             role_name = self._get_iam_role()
-            if role_name:
-                credentials = self._get_credentials(role_name)
-                return {
-                    'role_name': role_name,
-                    'access_key': credentials['AccessKeyId'],
-                    'secret_key': credentials['SecretAccessKey'],
-                    'token': credentials['Token'],
-                    'expiry_time': credentials['Expiration'],
-                }
+            credentials = self._get_credentials(role_name)
+            return {
+                'role_name': role_name,
+                'access_key': credentials['AccessKeyId'],
+                'secret_key': credentials['SecretAccessKey'],
+                'token': credentials['Token'],
+                'expiry_time': credentials['Expiration'],
+            }
         except _RetriesExceededError:
             logger.debug("Max number of attempts exceeded (%s) when "
                          "attempting to retrieve data from metadata service.",
@@ -282,7 +281,7 @@ class InstanceMetadataFetcher(object):
         return self._get_request(
             url=self._url,
             num_attempts=self._num_attempts,
-            needs_retry=self._is_non_ok_response
+            needs_retry=self._does_not_contain_role_name
         ).text
 
     def _get_credentials(self, role_name):
@@ -320,15 +319,20 @@ class InstanceMetadataFetcher(object):
                     "service request to %s: %s", url, e, exc_info=True)
         raise _RetriesExceededError()
 
-    def _is_non_ok_response(self, response):
-        return response.status_code != 200
+    def _does_not_contain_role_name(self, response):
+        return (
+            self._is_non_ok_response(response) or
+            self._is_empty(response)
+        )
 
     def _does_not_contain_credentials(self, response):
-        if (self._is_non_ok_response(response) or
+        return (self._is_non_ok_response(response) or
                 self._is_empty(response) or
-                self._is_invalid_json(response)):
-            return True
-        return False
+                self._is_invalid_json(response)
+        )
+
+    def _is_non_ok_response(self, response):
+        return response.status_code != 200
 
     def _is_empty(self, response):
         return not response.content
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -260,70 +260,83 @@ class InstanceMetadataFetcher(object):
             proxies=get_environ_proxies(self._url),
         )
 
-    def _get_request(self, url, timeout, num_attempts=1):
+    def retrieve_iam_role_credentials(self):
+        try:
+            role_name = self._get_iam_role()
+            if role_name:
+                credentials = self._get_credentials(role_name)
+                return {
+                    'role_name': role_name,
+                    'access_key': credentials['AccessKeyId'],
+                    'secret_key': credentials['SecretAccessKey'],
+                    'token': credentials['Token'],
+                    'expiry_time': credentials['Expiration'],
+                }
+        except _RetriesExceededError:
+            logger.debug("Max number of attempts exceeded (%s) when "
+                         "attempting to retrieve data from metadata service.",
+                         self._num_attempts)
+        return {}
+
+    def _get_iam_role(self):
+        return self._get_request(
+            url=self._url,
+            num_attempts=self._num_attempts,
+            needs_retry=self._is_non_ok_response
+        ).text
+
+    def _get_credentials(self, role_name):
+        r = self._get_request(
+            url=self._url + role_name,
+            num_attempts=self._num_attempts,
+            needs_retry=self._does_not_contain_credentials
+        )
+        return json.loads(r.text)
+
+    def _get_request(self, url, num_attempts=1, needs_retry=None):
         if self._disabled:
             logger.debug("Access to EC2 metadata has been disabled.")
             raise _RetriesExceededError()
-
         headers = {}
         if self._user_agent is not None:
             headers['User-Agent'] = self._user_agent
 
         for i in range(num_attempts):
             try:
-                AWSRequest = botocore.awsrequest.AWSRequest
-                request = AWSRequest(method='GET', url=url, headers=headers)
+                request = botocore.awsrequest.AWSRequest(
+                    method='GET', url=url, headers=headers)
                 response = self._session.send(request.prepare())
+                if needs_retry is not None and needs_retry(response):
+                    logger.debug(
+                        "Metadata service returned retryable response with "
+                        "status code of %s for url: %s, content body: %s",
+                        response.status_code, url, response.content
+                    )
+                else:
+                    return response
             except RETRYABLE_HTTP_ERRORS as e:
                 logger.debug("Caught exception while trying to retrieve "
                              "credentials: %s", e, exc_info=True)
-            else:
-                if response.status_code == 200:
-                    return response
         raise _RetriesExceededError()
 
-    def retrieve_iam_role_credentials(self):
-        data = {}
-        url = self._url
-        timeout = self._timeout
-        num_attempts = self._num_attempts
+    def _is_non_ok_response(self, response):
+        return response.status_code != 200
+
+    def _does_not_contain_credentials(self, response):
+        if (self._is_non_ok_response(response) or
+                self._is_empty(response) or
+                self._is_invalid_json(response)):
+            return True
+
+    def _is_empty(self, response):
+        return not response.content
+
+    def _is_invalid_json(self, response):
         try:
-            r = self._get_request(url, timeout, num_attempts)
-            if r.content:
-                fields = r.content.decode('utf-8').split('\n')
-                for field in fields:
-                    if field.endswith('/'):
-                        data[field[0:-1]] = self.retrieve_iam_role_credentials(
-                            url + field, timeout, num_attempts)
-                    else:
-                        val = self._get_request(
-                            url + field,
-                            timeout=timeout,
-                            num_attempts=num_attempts,
-                        ).content.decode('utf-8')
-                        if val[0] == '{':
-                            val = json.loads(val)
-                        data[field] = val
-            else:
-                logger.debug("Metadata service returned non 200 status code "
-                             "of %s for url: %s, content body: %s",
-                             r.status_code, url, r.content)
-        except _RetriesExceededError:
-            logger.debug("Max number of attempts exceeded (%s) when "
-                         "attempting to retrieve data from metadata service.",
-                         num_attempts)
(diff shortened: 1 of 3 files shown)


botocore/utils.py
```diff
@@ -260,70 +260,83 @@ def __init__(self, timeout=DEFAULT_METADATA_SERVICE_TIMEOUT,
...
+                if needs_retry is not None and needs_retry(response):
+                    logger.debug(
+                        "Metadata service returned retryable response with "
+                        "status code of %s for url: %s, content body: %s",
+                        response.status_code, url, response.content
+                    )
+                else:
+                    return response
             except RETRYABLE_HTTP_ERRORS as e:
                 logger.debug("Caught exception while trying to retrieve "
```
JordonPhillips: While we're at it, might as well update this message to be more clear and consistent with the other log message

kyleknap: So something like? 
```python
logger.debug("Caught HTTP exception while making metadata service request to %s: %s', url, e, exc_info=True)
```

JordonPhillips: Toss 'retryable' in there, it's not clear that it was retried

botocore/utils.py
```diff
@@ -260,70 +260,83 @@ def __init__(self, timeout=DEFAULT_METADATA_SERVICE_TIMEOUT,
...
 
-    def retrieve_iam_role_credentials(self):
-        data = {}
-        url = self._url
-        timeout = self._timeout
-        num_attempts = self._num_attempts
+    def _is_non_ok_response(self, response):
+        return response.status_code != 200
+
+    def _does_not_contain_credentials(self, response):
```
JordonPhillips: Why not just return the condition? Something like:

```python
return any(
    self._is_non_ok_response(response),
    self._is_empty(response),
    self._is_invalid_json(response),
)
```

kyleknap: Mainly being paranoid about changing behavior. I only wanted those additional checks to happen if I get back a 200 response back. Using `any` like that would cause all the checks to run even if say it was a non-200 response and I am not too sure what all of the non-200 response necessarily will look like from IMDS. Don't want to be making any assumptions.

JordonPhillips: Ah yeah, for some reason I thought it would lazily evaluate those. Regardless doing `if (multiline condition) return True` is a bit weird, especially since you never return False.

tests/unit/test_utils.py
```diff
@@ -1748,15 +1750,54 @@ def test_deepcopy_returns_same_object(self):
         self.assertIs(botocore.UNSIGNED, copy.deepcopy(botocore.UNSIGNED))
 
 
+class RawResponse(BytesIO):
```
joguSD: I think we have like 5 of these mock raw response classes. After we merge this let's look into consolidating these. I've added a TODO for this in my before-send PR.

botocore/utils.py
```diff
@@ -260,70 +260,83 @@ def __init__(self, timeout=DEFAULT_METADATA_SERVICE_TIMEOUT,
...
-
         headers = {}
         if self._user_agent is not None:
             headers['User-Agent'] = self._user_agent
 
         for i in range(num_attempts):
             try:
-                AWSRequest = botocore.awsrequest.AWSRequest
-                request = AWSRequest(method='GET', url=url, headers=headers)
+                request = botocore.awsrequest.AWSRequest(
```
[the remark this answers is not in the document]
JordonPhillips: eh, I like it this way better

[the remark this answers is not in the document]
kyleknap: It was more of a style thing. Setting a CamelCase variable in a function is a little weird. I rather keep it or change the import to `from botocore.awsrequest import AWSRequest` if that is better?

tests/unit/test_utils.py
kyleknap: Yep. I was hoping your `before-send` PR would get merged before I sent this up. So I pretty much copied and pasted what you had. I plan on using it once your PR gets merged.

botocore/utils.py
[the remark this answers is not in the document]
joguSD: Looks like I'm outnumbered then.

botocore/utils.py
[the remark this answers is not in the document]
kyleknap: I'm not sure if the IMDS response will ever have an empty `{}` and I'm not sure what it means if it does as if it did, it would most likely be intentional. The empty body and incomplete JSON are more for cases where IMDS does not send over the complete response. So I'm not sure if I want to be try to make assumptions about this case.

kyleknap: I guess that would work, but seems a little overkill for three checks. Might make more sense if there are ever any more checks. Also depending on more tests, I may be adding the `_is_empty` check to the role name checks so I may be able to flatten this to two checks.

botocore/utils.py
```diff
@@ -320,15 +319,20 @@ def _get_request(self, url, num_attempts=1, needs_retry=None):
...
-        return response.status_code != 200
+    def _does_not_contain_role_name(self, response):
+        return (
+            self._is_non_ok_response(response) or
+            self._is_empty(response)
+        )
 
     def _does_not_contain_credentials(self, response):
-        if (self._is_non_ok_response(response) or
+        return (self._is_non_ok_response(response) or
```
joguSD: Can we make the formatting on this consistent with the other function like this:
```python
return (
    self._is_...
)
```

botocore/utils.py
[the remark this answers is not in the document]
joguSD: If this were to ever happen we'd end up getting a KeyError though, we're already making the assumption that those 4 credential keys will be there, so why not validate it.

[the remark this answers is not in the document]
kyleknap: Talked offline about this. We are going to check for the fields we need and if they are not there, we will log what we got back and return an empty dictionary. We elected to not throw an error because someone may be relying on the IMDS credential provider to not throw an error (say they have a custom credential chain).

kyleknap:
Ok updated based on feedback. It should be good to look at. The latest change improves cases where we run into key errors (e.g. IMDs responds back with a 200 status code but JSON error message)

botocore/utils.py
```diff
@@ -260,70 +265,108 @@ def __init__(self, timeout=DEFAULT_METADATA_SERVICE_TIMEOUT,
...
+
+    def _get_credentials(self, role_name):
+        r = self._get_request(
+            url=self._url + role_name,
+            num_attempts=self._num_attempts,
+            needs_retry=self._does_not_contain_credentials
+        )
+        return json.loads(r.text)
+
+    def _get_request(self, url, num_attempts=1, needs_retry=None):
```
jamesls: Do we need to default to 1 here?  Seems like it's safer to just require you to provide `self._num_attempts` when calling this method.

botocore/utils.py
```diff
@@ -260,70 +265,108 @@ def __init__(self, timeout=DEFAULT_METADATA_SERVICE_TIMEOUT,
...
+            else:
+                # IMDS can return a 200 response that has a JSON formatted
+                # error message (i.e. if ec2 is not trusted entity for the
+                # attached role). We do not necessarily want to retry for
+                # these and we also do not necessarily want to raise a key
+                # error. So at least log the problematic response and return
+                # an empty dictionary to signal that it was not able to
+                # retrieve credentials.
+                logger.debug('Response received %s is missing '
+                             'credentials.', credentials)
```
jamesls: I don't think we should be logging the body here.  We can't guarantee what's in the body.  It seems safer to call out specific cases we know about, e.g check for the specific schema for when ec2 is not a trusted entity for the attached role and print the response in that scenario.

Otherwise maybe print the key names but not the values or something similar.

botocore/utils.py
```diff
@@ -260,70 +265,108 @@ def __init__(self, timeout=DEFAULT_METADATA_SERVICE_TIMEOUT,
             proxies=get_environ_proxies(self._url),
         )
 
-    def _get_request(self, url, timeout, num_attempts=1):
+    def retrieve_iam_role_credentials(self):
+        try:
+            role_name = self._get_iam_role()
+            credentials = self._get_credentials(role_name)
```
jamesls: I think the retry should trigger a restart of the *entire* process, including looking up the role name.  Right now we assume that if we get the role_name, we only need to retry if we aren't able to get credentials.  This won't handle the case where we get a bad role name and then try to retrieve credentials (which will never work).

botocore/utils.py
```diff
@@ -260,70 +265,108 @@ def __init__(self, timeout=DEFAULT_METADATA_SERVICE_TIMEOUT,
...
+            if self._contains_all_credential_fields(credentials):
```
jamesls: Why isn't `_contains_all_credential_fields` not part of the `self._does_not_contain_credentials` retry check?  It seems like part of checking if the response does not contain credentials is to also check that all the credential keys are present in the response.

botocore/utils.py
kyleknap: No we do not need a default. I'm fine with removing it.

botocore/utils.py
kyleknap: >> Check for the specific schema for when ec2 is not a trusted entity for the attached role and print the response in that scenario. Otherwise maybe print the key names but not the values or something similar.

I will probably do both: one log message for JSON errors and another that just prints out the missing keys.

botocore/utils.py
kyleknap: I put it in the [comment](https://github.com/boto/botocore/pull/1569/files/f5611ae5bfe0b374d3ce9ba85a67cb9590fcdcb1#diff-6c5fe2c0e8325d002e7c164f10a30dbcR281). I don't think we should be retrying these cases. I think I'm going to rename these methods to make it a bit clearer. I can see how the names make it confusing on why the `_contains_all_credential_fields` is not in the `_does_not_contain_credentials` method

botocore/utils.py
kyleknap: I was thinking about that as well, but I decided not to do that in order to keep parity with the existing behavior where we only do retries on a per IMDS request basis. I do not think we want to make the general assumption to refetch the role name on retryable credential grabs. For example when you either swap a role or disassociate a role, it is possible to get a 404 response back for a credential request. For the case where the role was completely disassociated, it is not appropriate to retry the entire process as there are no more credentials available. For role swaps, it would be appropriate to retry the entire process. The big problem is that it is impossible to disambiguate the 404's between both scenarios.

jamesls: Following up from offline discussion:  we should keep this logic as you have it.  It keeps the logic closer to the previous behavior.  While it's possible that we miss a case where we have a role name that is completely invalid, this realistically only affects the initial request to fetch credentials.  Once botocore has credentials, any refresh attempts during the advisory refresh is not a hard failure.  This means in order for this to be a hard failure we'd need to repeatedly see this failure which would require the role's repeatedly being switched while botocore is fetching credentials.

kyleknap:
@jamesls I updated based on feedback/offline discussions. Should be good to look at again.

botocore/utils.py
```diff
@@ -260,70 +265,125 @@ def __init__(self, timeout=DEFAULT_METADATA_SERVICE_TIMEOUT,
...
-        return final_data
+            json.loads(response.text)
+            return False
+        except ValueError:
+            self._log_imds_response(response, 'invalid json')
+            return True
+
+    def _log_imds_response(self, response, reason_to_log, log_body=False):
+        statement = (
+            "Metadata service returned " + reason_to_log + " response "
```
stealthycoin: How come `reason_to_log` is the only one not being formatted in at log time?

kyleknap: That was a refactoring artifact. I had `reason_to_log` optional at some point. I can update it.

PR #1572, opened 2018-10-02
Docs: describe-trusted-advisor-checks

istairbn:
describe-trusted-advisor-checks currently only works when the region is set to us-east-1. Adding this to the documentation

### DIFF
--- a/botocore/data/support/2013-04-15/service-2.json
+++ b/botocore/data/support/2013-04-15/service-2.json
@@ -180,7 +180,7 @@
       "errors":[
         {"shape":"InternalServerError"}
       ],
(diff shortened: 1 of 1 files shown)


jschwarzwalder:
@istairbn Thank you for your feedback. Your request has been shared with the Trusted Advisor Team. 

Updating the description here will not fix the issue, and it is possible that Trusted Advisor will work in other regions in the future. 

For more information about Trusted Advisor, you can checkout the [AWS Support User Guide](https://docs.aws.amazon.com/awssupport/latest/user/trustedadvisor.html?id=docs_gateway).

stealthycoin:
Closing as there is nothing to do here. This file is consumed from the service team as they are responsible for their own docs.

Issue #1537, opened 2018-08-29
DeprecationWarning in botocore/parsers.py:434 with Python 3.7


stig:
Hello! I'm using Sceptre with Python 3.7 on OS X and it uses boto3 and botocore internally. I am getting loads of these deprecation warnings:

> /Users/stig/.virtualenvs/sceptre/lib/python3.7/site-packages/botocore/parsers.py:434: DeprecationWarning: This method will be removed in future versions.  Use 'list(elem)' or iteration over elem instead.
>   if value.getchildren():

-----
Version information:

```
$ python --version 
Python 3.7.0
```

```
>>> botocore.__version__
 '1.11.2' 
```

OS: macOS High Sierra (10.13.6)

```
$ sceptre --version
Sceptre, version 1.4.1
```

joguSD:
Looks like these warnings are new in 3.7 but the referenced features have been documented as deprecated for a while, see: https://bugs.python.org/issue29209
It seems like the current plan is completely remove these deprecated `xml` features in 3.9+

Specifically, `xml.etree.cElementTree` has been deprecated since 3.3 and the [getchildren](https://github.com/boto/botocore/blob/f3f8454d7eb873984a0df71031a9fa3892e6c4b0/botocore/parsers.py#L434-L436) function has been deprecated since 2.7/3.2.

DiegoPomares:
https://github.com/boto/botocore/pull/1577

kyleknap:
Thanks @DiegoPomares  for the PR. I just merged it. You all should no longer be seeing that warning in the next release of `botocore`. Closing issue.

PR #1577, opened 2018-10-10
Fix annoying DeprecationWarning in parsers.py

DiegoPomares:
Starting with Python3.7, botocore issues the following warning:
```
/parsers.py:434: DeprecationWarning: This method will be removed in future versions.  Use 'list(elem)' or iteration over elem instead.
  if value.getchildren():
```
Which is super annoying when working with CloudFormation as it pops up a lot.
Addresses: https://github.com/boto/botocore/issues/1537


partounian:
Would love for someone to merge this @jamesls

PR #1585, opened 2018-10-22
Add specifications for available regions for missing services

j-kroll:
Add specifications for available regions for missing services, as raised in the boto3 issue https://github.com/boto/boto3/issues/1662. Below is the source used to identify each service's regions, with the list of services taken from the original issue:

1. appsync (A)
1. cloudsearchdomain (B: under CloudSearch)
1. connect (A)
1. dlm (B: under EC2)
1. eks (A)
1. iot-jobs-data (B: under IoT Device Management)
1. iotanalytics (n/a - already existed in the file)
1. macie (A)
1. mobile (A)
1. mq (A)
1. pi (B: under RDS)
1. transcribe (A)

A: [AWS region table](https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services/)
B: [AWS console](https://console.aws.amazon.com/console)

### DIFF
--- a/botocore/data/endpoints.json
+++ b/botocore/data/endpoints.json
@@ -1233,7 +1233,7 @@
           "us-west-2" : { }
         }
       },
-      "iot-jobs-data" : {
+      "data.jobs.iot" : {
         "endpoints" : {
           "ap-northeast-1" : { },
           "ap-northeast-2" : { },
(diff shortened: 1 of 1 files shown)


stealthycoin:
Hi thanks for the contribution, however; this file is given to us from an upstream source and we cannot modify it. These changes would be clobbered in our next release.

PR #1590, opened 2018-10-25
Add test to ensure we can load old models

jamesls:
This copies over the existing lambda service model
as it exists today and adds a test that ensures we can
create a client and make a successful stubbed request.
This ensures that botocore is able to use older service
models.


tests/functional/test_model_backcompat.py
```diff
@@ -0,0 +1,56 @@
...
+    loader.search_paths.insert(0, FIXED_MODELS_DIR)
+
+    # The model dir we copied was renamed to 'custom-lambda'
+    # to ensure we're loading our version of the model and not
+    # the built in one.
+    client = session.create_client(
+        'custom-lambda', region_name='us-west-2',
+        aws_access_key_id='foo', aws_secret_access_key='bar',
+    )
+    with Stubber(client) as stubber:
```
kyleknap: While I agree I don't think we need to be making an actual API request, what is your thoughts on using the [ClientHTTPStubber](https://github.com/boto/botocore/blob/334a179078e71760cfc8487b129ae894f886d6a0/tests/__init__.py#L380) instead? The nice part about this is we will cover a bit more in our stack and still will not be making an API call over the wire.

tests/functional/models/custom-lambda/2015-03-31/paginators-1.json
```diff
@@ -0,0 +1,22 @@
+{
```
kyleknap: What was the reasoning for pulling in the `paginators-1.json` and `examples-1.json` as well? It looks like we are not doing any assertions around there functionality.

tests/functional/test_model_backcompat.py
jamesls: In general I'm not opposed but I'd prefer to stick with the stubber for now because it's a publicly documented API that will remain backwards compatible.  Ideally we're not changing this test so I tried to stick with APIs that should never require modifications.

If we move the `ClientHTTPStubber` over to the stub module, then perhaps, but it sounds like there's some more vetting we'd like to do with the `ClientHTTPStubber` before making it public.

tests/functional/models/custom-lambda/2015-03-31/paginators-1.json
jamesls: Primarily just to ensure we're using a consistent snapshot of the all the models.  I'd not want us to load the service model from this directory and the paginators from the built in directory.

If you want, I an add some sort of assertion on getting paginators from the clients to ensure we exercise that code.

The examples-1.json however, I don't feel that strongly about.

kyleknap: I think it would make sense to add a test for paginators. I do not care to strongly about the `examples-1.json`. We can keep it there and not put any assertions on it. Do we want to add an empty `waiters-2.json` file? We may run into the same problem of loading from the builtin directory if waiters are added in the future for `Lambda`.

tests/functional/test_model_backcompat.py
kyleknap: Yeah I agree we would want to do more vetting before making it public. While I do not think we should be touching this test very much, I can see us updating it if we add support for a new model and if we did update the `ClientHTTPStubber`, I'm not sure if it would be a significant update to the interface. I think the main thing I'm concerned about is that the stubber is registered near the beginning of the `before-call` event (via `before-call.*.*`), which is a popular event we use. So if we registered a handler after the stubber handler and it caused issues for the legacy models, we may be blind to it because the stubber causes all subsequently registered handlers to not get ran.

jamesls:

PR #1591, opened 2018-10-25
Refactor session config loading customization

stealthycoin:
Adds a new session component to give more fine grained control over config variable loading. This PR maintains backwards compatibility with the existing `session_instance_vars` system while adding a new provider based component to the session. 

All of the classes located in `botocore/configprovider.py` are intended to be a part of the public API of a session and will need to maintain backwards compatibility.

cc @kyleknap @jamesls @JordonPhillips @dstufft @joguSD

### DIFF
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -95,10 +95,10 @@ class DefaultConfigChainBuilder(object):
         return ChainProvider(providers=providers)
 
 
-class ConfigProvider(object):
-    """The ConfigProvider object loads configuration values lazily."""
+class ConfigProviderComponent(object):
+    """The ConfigProviderComponent object loads configuration values lazily."""
     def __init__(self, mapping=None):
-        """Initialize a ConfigProvider.
+        """Initialize a ConfigProviderComponent.
 
         :type mapping: dict
         :param mapping: The mapping parameter is a map of string to a subclass
@@ -158,6 +158,16 @@ class ConfigProvider(object):
 
         """
         self._cache[logical_name] = value
+
+    def update_mapping(self, new_mapping):
+        """Update the config mapping.
+
+        :type new_mapping: dict
+        :param new_mapping: The new mapping of logical names to config
+            providers. Each name in this map will be added to or override an
+            existing mapping.
+        """
+        self._mapping.update(new_mapping)
 
 
 class BaseConfigValueProvider(object):
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -20,47 +20,60 @@ import copy
 def create_botocore_default_config_mapping(chain_builder):
     return {
         'profile': chain_builder.build_config_chain(
+            instance_var='profile',
             env_vars=['AWS_DEFAULT_PROFILE', 'AWS_PROFILE'],
         ),
         'region': chain_builder.build_config_chain(
+            instance_var='region',
             env_vars='AWS_DEFAULT_REGION',
             config_property='region',
             default=None,
         ),
         'data_path': chain_builder.build_config_chain(
+            instance_var='data_path',
             env_vars='AWS_DATA_PATH',
             config_property='data_path',
             default=None
         ),
         'config_file': chain_builder.build_config_chain(
+            instance_var='config_file',
             env_vars='AWS_CONFIG_FILE',
             default='~/.aws/config',
         ),
         'ca_bundle': chain_builder.build_config_chain(
+            instance_var='ca_bundle',
             env_vars='AWS_CA_BUNDLE',
             config_property='ca_bundle',
         ),
         'api_versions': chain_builder.build_config_chain(
+            instance_var='api_versions',
             config_property='api_versions',
             default={},
         ),
         'credentials_file': chain_builder.build_config_chain(
+            instance_var='credentials_file',
             env_vars='AWS_SHARED_CREDENTIALS_FILE',
             default='~/.aws/credentials',
         ),
         'metadata_service_timeout': chain_builder.build_config_chain(
+            instance_var='metadata_service_timeout',
             env_vars='AWS_METADATA_SERVICE_TIMEOUT',
             config_property='metadata_service_timeout',
-            default=1
+            default=1,
+            cast=int,
         ),
         'metadata_service_num_attempts': chain_builder.build_config_chain(
+            instance_var='metadata_service_num_attempts',
             env_vars='AWS_METADATA_SERVICE_NUM_ATTEMPTS',
             config_property='metadata_service_num_attempts',
-            default=1
+            default=1,
+            cast=int,
         ),
         'parameter_validation': chain_builder.build_config_chain(
+            instance_var='parameter_validation',
             config_property='parameter_validation',
             default=True,
+            cast=bool,
         ),
     }
 
@@ -88,8 +101,8 @@ class DefaultConfigChainBuilder(object):
             environ = copy.copy(os.environ)
         self._environ = environ
 
-    def build_config_chain(self, env_vars=None, config_property=None,
-                         default=None):
+    def build_config_chain(self, instance_var=None, env_vars=None,
+                           config_property=None, default=None, cast=None):
         """Build a config chain following the standard botocore pattern.
 
         In botocore most of our config chains follow the the precendence:
@@ -97,6 +110,11 @@ class DefaultConfigChainBuilder(object):
 
         This is a convenience function for creating a chain that follows
         that precendence.
+
+        :type instance_var: str
+        :param instance_var: The insance variable to associate with this key.
+            Instance variables are looked up in the session instance variable
+            mapping.
 
         :type env_vars: str or list of str or None
         :param env_vars: One or more environment variable names to search for
@@ -113,12 +131,24 @@ class DefaultConfigChainBuilder(object):
             callable it will be treated as a lazy value, and the callable will
             be called when the value is needed.
 
+        :type cast: None or callable
+        :param cast: If this value is None then it has no affect on the return
+            type. Otherwise, it is treated as a function that will cast our
+            provided type.
+
         :rvalue: ConfigChain
         :returns: A ConfigChain that resolves in the order env_vars ->
             config_property -> default. Any values that were none are
             omitted form the chain.
         """
         providers = []
+        if instance_var is not None:
+            providers.append(
+                ScopedConfigValueProvider(
+                    name=instance_var,
+                    scoped_config_method=self._session.instance_variables,
+                )
+            )
         if env_vars is not None:
             providers.append(
                 DictConfigValueProvider(
@@ -140,7 +170,7 @@ class DefaultConfigChainBuilder(object):
         if default is not None:
             providers.append(default)
 
-        return ChainProvider(providers=providers)
+        return ChainProvider(providers=providers, cast=cast)
 
 
 class ConfigProviderComponent(object):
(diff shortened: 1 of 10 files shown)


botocore/configprovider.py
```diff
@@ -0,0 +1,419 @@
...
+            providers.append(
+                ClosedDictProvider(
+                    keys=env_vars,
+                    source=self._environ,
+                )
+            )
+        if config_property is not None:
+            providers.append(
+                LazyProvider(
+                    provider_function=lambda: ClosedDictProvider(
```
kyleknap: It might make sense to rename `provider_function` to something like `provide_factory`. First reading this, I would have assumed we would have been providing a lazily called function (similar to `provide()`) that provides the value.

botocore/configprovider.py
```diff
@@ -0,0 +1,419 @@
...
+        if env_vars is not None:
+            providers.append(
+                ClosedDictProvider(
+                    keys=env_vars,
+                    source=self._environ,
+                )
+            )
+        if config_property is not None:
+            providers.append(
+                LazyProvider(
```
kyleknap: I wonder if it makes sense to have a generalized `LazyProvider`? It took me a while to follow the laziness with the use of `lambda`. I could also see us implement specialized providers such as `SessionInstanceProvider` or `ScopedConfigProvider` that just calls the slow method when `provide()` gets called. Another nice part about this approach is that we would be able to cache the responses back if we plan to call the slow method a lot and it is slow.

botocore/utils.py
```diff
@@ -245,25 +240,106 @@ def set_value_from_jmespath(source, expression, value, is_first=True):
 
 class InstanceMetadataFetcher(object):
 
-    _REQUIRED_CREDENTIAL_FIELDS = [
-        'AccessKeyId', 'SecretAccessKey', 'Token', 'Expiration'
-    ]
+    class _RetriesExceededError(Exception):
```
kyleknap: What's the reasoning for moving this error class to inside this class?

botocore/utils.py
```diff
@@ -245,25 +240,106 @@ def set_value_from_jmespath(source, expression, value, is_first=True):
...
         self._disabled = env.get('AWS_EC2_METADATA_DISABLED', 'false').lower()
         self._disabled = self._disabled == 'true'
         self._user_agent = user_agent
         self._session = botocore.httpsession.URLLib3Session(
             timeout=self._timeout,
-            proxies=get_environ_proxies(self._url),
+            proxies=get_environ_proxies(self._base_url),
+        )
+
+    def get_request(self, url_path, custom_retry=None):
```
kyleknap: I think for the signature here we just expose a `needs_retry` that overrides whatever the general `_needs_retry` method. It is a little weird that no matter what a user of this interface cannot override the default empty/non-200 retry logic.

botocore/session.py
```diff
@@ -972,6 +892,52 @@ def lazy_register_component(self, name, no_arg_factory):
...
+    def __delitem__(self, key):
+        del self._store[self.__keytransform__(key)]
+
+    def __iter__(self):
+        return iter(self._store)
+
+    def __len__(self):
+        return len(self._store)
+
+    def __keytransform__(self, key):
```
kyleknap: Can you elaborate on why we need this method?

botocore/session.py
```diff
@@ -972,6 +892,52 @@ def lazy_register_component(self, name, no_arg_factory):
             pass
 
 
+class SessionVarsShim(collections.MutableMapping):
```
kyleknap: I think it would be fine to just call this `SessionVarsMap`. The `Shim` part is a little vague for me.

botocore/session.py
```diff
@@ -972,6 +892,52 @@ def lazy_register_component(self, name, no_arg_factory):
...
+    def __init__(self, session, *args, **kwargs):
+        self._session = session
+        self._store = dict()
+        self.update(dict(*args, **kwargs))
```
kyleknap: This is a little hard to follow what is being passed in here. Is there a reason we do not just accept the original `session_vars_map` dictionary as an optional argument? The nice part about that is we would no longer need an `if` statement when instantiating this.

botocore/configprovider.py
```diff
@@ -0,0 +1,419 @@
...
+            default=None,
+        ),
+        'data_path': chain_builder.build_config_chain(
+            env_vars='AWS_DATA_PATH',
+            config_property='data_path',
+            default=None,
+        ),
+        'config_file': chain_builder.build_config_chain(
+            env_vars='AWS_CONFIG_FILE',
+            default='~/.aws/config',
```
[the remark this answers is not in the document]
stealthycoin: Not sure, this is unchanged from before.

botocore/utils.py
```diff
@@ -243,28 +238,108 @@ def set_value_from_jmespath(source, expression, value, is_first=True):
...
+    """Internal exception used when the number of retries are exceeded."""
+    pass
+
+
 class InstanceMetadataFetcher(object):
 
-    _REQUIRED_CREDENTIAL_FIELDS = [
-        'AccessKeyId', 'SecretAccessKey', 'Token', 'Expiration'
-    ]
+    _RETRIES_EXCEEDED_ERROR = _RetriesExceededError
```
kyleknap: We should rename this to `_RETRIES_EXCEEDED_ERROR_CLS` and raise instances of the class. Right now I think we are just raising the exception class.

tests/unit/test_config_provider.py
```diff
@@ -0,0 +1,418 @@
...
+        self.assertEqual(value, expected_value)
+
+    def test_chain_builder_can_provide_instance(self):
+        self.assert_chain_does_provide(
+            instance_map={'foo': 'bar'},
+            environ_map={},
+            config_map={},
+            build_config_chain_args={},
+            logical_name_to_provide='foo',
+            expected_value='bar',
```
kyleknap: So for these tests to make it even more readable, it would be nice to put the name of where it came from as the value so instead of using `bar` make the value `instance-val` to indicate it came from the instance values. You did it a little bit later on for the keys, but it would be nice to do it for the values as well.

tests/unit/test_config_provider.py
```diff
@@ -0,0 +1,418 @@
...
+            **build_config_chain_args
+        )
+        value = chain.provide()
+        self.assertEqual(value, expected_value)
+
+    def test_chain_builder_can_provide_instance(self):
+        self.assert_chain_does_provide(
+            instance_map={'foo': 'bar'},
+            environ_map={},
+            config_map={},
```
kyleknap: Maybe call this the `scoped_config_map`? It is a little more specific on what config is being referenced

botocore/configprovider.py
```diff
@@ -0,0 +1,416 @@
...
+        :type logical_name: str
+        :param logical_name: The logical name of the config value that this
+            chain is responsible for loading.
+
+        :type instance: bool
+        :param instance: This indicates whether or not session instance
+            variable lookup should be included in the config chain. By default
+            it is True.
+
+        :type env_vars: str or list of str or None
```
kyleknap: Just a little more feedback on the names after looking at the tests... Maybe call this `env_varnames` or something else to indicate it is a name? Typically, `env_vars` or `env` is associated as a map and it threw me off when I was reading through the tests. The same thing goes for `config_property`, but not as much. Maybe try to find a parameter name that better indicates it expects the `name`?

tests/unit/test_config_provider.py
```diff
@@ -0,0 +1,418 @@
...
+def assert_chain_does_provide(providers, expected_value):
+    provider = ChainProvider(
+        providers=providers,
+    )
+    value = provider.provide()
+    assert_equal(value, expected_value)
+
+
+def test_chain_provider():
+    # Each case is a tuple with the first element being the expected return
```
kyleknap: Not too picky about this, but if you wanted you could copy a bit of what we do in [testing s3 addressing](https://github.com/boto/botocore/blob/7052b0f91fca61bf5206ec73240a99094fd2207a/tests/functional/test_s3.py#L855) where we have a class to specify cases. The nice part about this is that you can have parameter names so we would not have to explain the parameters in the comments and they would be self-documenting.

tests/unit/test_session.py
```diff
@@ -97,49 +138,28 @@ def test_profile_when_pulled_from_env(self):
...
-        }
-        session = create_session(session_vars=env_vars)
+        self.update_session_config_mapping(
+            'profile', env_vars=['BAR_DEFAULT_PROFILE', 'BAR_PROFILE']
+        )
         self.environ.pop('BAR_DEFAULT_PROFILE', None)
         self.environ['BAR_PROFILE'] = 'second'
-        self.assertEqual(session.get_config_variable('profile'), 'second')
-
-    def test_profile(self):
```
kyleknap: Why did this test get removed?

stealthycoin: Wasn't doing anything useful.

tests/unit/test_config_provider.py
stealthycoin: Personally I like the way I did it better for this case since its super easy to see the chain and the result it should give with no extra verbosity.

botocore/configprovider.py
```diff
@@ -0,0 +1,418 @@
...
+        ),
+        'parameter_validation': chain_builder.create_config_chain(
+            instance_name='parameter_validation',
+            config_property_name='parameter_validation',
+            default=True,
+        ),
+    }
+
+
+class ConfigChainFactory(object):
```
jamesls: Would you consider this part of the public API?  Seems like we really only want the `ConfigValueStore` to be considered public right?  It's worth documenting what we consider to be public here.

botocore/credentials.py
```diff
@@ -67,7 +67,7 @@ def create_credential_resolver(session, cache=None):
     env_provider = EnvProvider()
     container_provider = ContainerProvider()
     instance_metadata_provider = InstanceMetadataProvider(
-        iam_role_fetcher=InstanceMetadataFetcher(
+        iam_role_fetcher=InstanceMetadataCredentialFetcher(
```
jamesls: I'm a little concerned about this impacting people.  I'd prefer we keep the name the same, and just create something new like `IMDSClient` for making arbitrary requests to IMDS.  We can revisit the names in a major version bump.

botocore/utils.py
```diff
@@ -243,28 +238,108 @@ def set_value_from_jmespath(source, expression, value, is_first=True):
...
         self._disabled = env.get('AWS_EC2_METADATA_DISABLED', 'false').lower()
         self._disabled = self._disabled == 'true'
         self._user_agent = user_agent
         self._session = botocore.httpsession.URLLib3Session(
             timeout=self._timeout,
-            proxies=get_environ_proxies(self._url),
+            proxies=get_environ_proxies(self._base_url),
         )
 
+    def get_request(self, url_path, retry_func=None):
```
jamesls: So I didn't mind this as an internal method on the original imds fetcher class, but this just seems like a weird public interface to me (specifically the `retry_func` part).  Is there any way we can just keep the retry stuff in the original class that fetches iam role credentials?  This class could just return the response and the class that's retrieving the iam role could still decide if it needed to retry.

botocore/configprovider.py
stealthycoin: Ah I hadn't thought about that specifically. I don't think there is any reason for it to be public since its a convenience class for us. It would only be useful for people that want to make the same configuration chain as we do, which I don't think is a big use case. Mainly I was considering the `ConfigStore` as the public class since it gets exposed through the `get_component('config_store')` on the session. And thats the main way to interact with overriding config providers. From there the `BaseProvider` is the interface they would use so that is public as well.

Should we add this to the botocore docs somewhere or document it in code?

botocore/credentials.py
stealthycoin: Ah good point, I can change the names around.

botocore/utils.py
stealthycoin: This was to make it easier to write more clients since they all share that functionality. Making it public was to indicate that its a publicly facing function since it will have a consumer in a different project (CLI).

PR #1594, opened 2018-10-31
Add missing paginator definitions for CloudFormation

slai:
This adds the missing pagination definitions for the CloudFormation APIs, as described in #1462.

I've tested DescribeChangeSet and ListChangeSets and both work as expected after this change,

```
client = boto3.client('cloudformation', region_name='eu-west-1')

paginator = client.get_paginator('list_change_sets')
page_iterator = paginator.paginate(StackName='MyStack')

for page in page_iterator:
    print(page)
```

### DIFF
--- a/botocore/data/cloudformation/2010-05-15/paginators-1.json
+++ b/botocore/data/cloudformation/2010-05-15/paginators-1.json
@@ -8,7 +8,23 @@
     "DescribeChangeSet": {
       "input_token": "NextToken",
       "output_token": "NextToken",
-      "result_key": "Changes"
+      "result_key": "Changes",
+      "non_aggregate_keys": [
+          "ChangeSetName",
+          "ChangeSetId",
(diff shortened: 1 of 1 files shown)


joguSD:
All of these look good except for `DescribeChangeSet`. That operation has quite a few fields in the output in addition to the `Changes` `result_key`. I'm pretty sure they could all be added as `non_aggregate_keys`, but a couple of them are lists which is making me a little hesitant right now. I'd need to first confirm that the auxiliary lists are also not paginated.

We would need to ensure that all output keys are accounted for, otherwise they'll be dropped from `build_full_results` output, which in turn effects the CLI in a backwards incompatible way.

joguSD:
Merged in https://github.com/boto/botocore/commit/3b2a60b963bc48a41ecfc9d9f010a976c592f0bc

PR #1596, opened 2018-10-31
Add support for pickling exceptions

kyleknap:
There has been a few requests/PRs for this:

https://github.com/boto/botocore/issues/834
https://github.com/boto/botocore/pull/1191
https://github.com/boto/botocore/pull/835

### DIFF
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -16,6 +16,18 @@ from botocore.vendored import requests
 from botocore.vendored.requests.packages import urllib3
 
 
+def _exception_from_packed_args(exception_cls, args=None, kwargs=None):
+    # This is helpful for reducing Exceptions that only accept kwargs as
+    # only positional arguments can be provided for __reduce__
+    # Ideally, this would also be a class method on the BotoCoreError
+    # but instance methods cannot be pickled.
+    if args is None:
+        args = ()
+    if kwargs is None:
+        kwargs = {}
+    return exception_cls(*args, **kwargs)
+
+
 class BotoCoreError(Exception):
     """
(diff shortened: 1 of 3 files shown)


botocore/exceptions.py
```diff
@@ -395,6 +414,12 @@ def _get_retry_info(self, response):
                                   metadata['RetryAttempts'])
         return retry_info
 
+    def __reduce__(self):
+        # Subclasses of ClientError's are dynamically generated and
+        # cannot be pickled unless they are attributes of a
+        # module. So at the very least return a ClientError back.
+        return ClientError, (self.response, self.operation_name)
```
joguSD: This is a sub class of `BotoCoreError` and not `ClientError`.

JordonPhillips: This ***is*** `ClientError`

kyleknap:
@JordonPhillips Yeah it is unfortunate. I would not say there is nothing we can do about generated exceptions. We can potentially still add it in the future, but it may be tricky. For example, this PR: https://github.com/boto/botocore/pull/1191 has an example of how you could do it. The only hesitation that I had in looking it over was that PR was it set dynamic exception modules as attributes to the global `exceptions` module when these dynamically generated exceptions are only scoped to client instances.

However just making the exceptions pickleable goes a long way in using `botocore` in processes. The fact that most of the errors we raise in botocore are not pickleable is pretty annoying.

botocore/exceptions.py
joguSD: Oh, for some reason I though this was in `ConnectionError` and was really confused.

PR #1603, opened 2018-11-06
Add the SESSION_VARIABLES back to the session object

stealthycoin:
The session.session_vars_map was a publicly accessible map that used to
have tuples that configured how variables were loaded, this map was
removed in favor of a new system. The new system is still in place but
the tuples have been put back so that they can be read from the session.

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -30,6 +30,7 @@ import botocore.client
 from botocore.configprovider import ConfigValueStore
 from botocore.configprovider import ConfigChainFactory
 from botocore.configprovider import create_botocore_default_config_mapping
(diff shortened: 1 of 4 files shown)


botocore/session.py
```diff
@@ -168,10 +225,7 @@ def _register_builtin_handlers(self, events):
                     self._events.register_last(event_name, handler)
 
     def _register_config_store(self):
-        chain_builder = ConfigChainFactory(session=self)
-        config_store_component = ConfigValueStore(
-            mapping=create_botocore_default_config_mapping(chain_builder)
-        )
+        config_store_component = ConfigValueStore()
```
jamesls: Why did this change?

PR #1619, opened 2018-11-29
Add Content-MD5 header to new operations that require it

stealthycoin:


jweinst1:
@stealthycoin , any reason why these were added with `calculate_md5` and not `conditionally_calculate_md5` ? I am experiencing issues now using boto3 copy in fips environment.

PR #1620, opened 2018-12-01
Fix EnvironmentID requirement

joelparkerhenderson:


joguSD:
As noted in our [contributing guide](https://github.com/boto/botocore/blob/develop/CONTRIBUTING.rst) we can't merge this as we get the service definitions from upstream.

PR #1625, opened 2018-12-11
Add PendingDeprecationWarning to internal requests methods

jamesls:
This produces no output yet unless you explicitly turn on
warnings.  In a future release we'll have this emit a warning
by default.

I opted to put this in the high level `botocore.vendored.requests.*`
methods because that's the more common usage of requests and I wanted
better error messages.  Moving it to the `Session.request` method
would mean we couldn't give as helpful of an error message.

## Testing

```
$ cat driver.py
from botocore.vendored import requests


print(requests.get('https://httpbin.org'))
print(requests.head('https://httpbin.org'))

$ python driver.py
<Response [200]>
<Response [200]>

$ python -Wd driver.py
/botocore/botocore/vendored/requests/api.py:60: PendingDeprecationWarning: You are using the get() function from 'botocore.vendored.requests'.  This is not a public API in botocore and will be removed in the future. Additionally, this version of requests is out of date.  We recommend you install the requests package, 'import requests' directly, and use the requests.get() function instead.
  PendingDeprecationWarning
<Response [200]>
/botocore/botocore/vendored/requests/api.py:60: PendingDeprecationWarning: You are using the head() function from 'botocore.vendored.requests'.  This is not a public API in botocore and will be removed in the future. Additionally, this version of requests is out of date.  We recommend you install the requests package, 'import requests' directly, and use the requests.head() function instead.
  PendingDeprecationWarning
<Response [200]>
```

### DIFF
--- a/botocore/vendored/requests/api.py
+++ b/botocore/vendored/requests/api.py
@@ -10,8 +10,18 @@ This module implements the Requests API.
 :license: Apache2, see LICENSE for more details.
 
 """
+import warnings
 
 from . import sessions
+
+
+_WARNING_MSG = (
+    "You are using the {name}() function from 'botocore.vendored.requests'.  "
+    "This is not a public API in botocore and will be removed in the future. "
+    "Additionally, this version of requests is out of date.  We recommend "
+    "you install the requests package, 'import requests' directly, and use "
+    "the requests.{name}() function instead."
+)
 
 
 def request(method, url, **kwargs):
@@ -45,6 +55,10 @@ def request(method, url, **kwargs):
       >>> req = requests.request('GET', 'http://httpbin.org/get')
       <Response [200]>
     """
+    warnings.warn(
+        _WARNING_MSG.format(name=method),
+        PendingDeprecationWarning
+    )
 
     session = sessions.Session()
     response = session.request(method=method, url=url, **kwargs)


botocore/vendored/requests/api.py
```diff
@@ -45,6 +55,10 @@ def request(method, url, **kwargs):
       >>> req = requests.request('GET', 'http://httpbin.org/get')
       <Response [200]>
     """
+    warnings.warn(
+        _WARNING_MSG.format(name=method),
+        PendingDeprecationWarning
```
JordonPhillips: This should be `DeprecationWarning` since it's already deprecated. Though I'm not sure deprecated is right at all since this was never meant to be used in the first place. Might be better to stick with `UserWarning` since that will be shown by default.

joguSD: Hmm, I think either `PendingDeprecationWarning` or `DeprecationWarning` is appropriate.
I'm -1 on labeling this as a `UserWarning` just to get it to show up by default, as I think it would go against what users would expect of a deprecation warning.

jamesls: Let's switch it to `DeprecationWarning` for now.  I think we'll eventually convert this to a warning that shows up by default, but I want to put something in place now and let it sit for a while before we start printing warnings to stderr.

jamesls:
@joguSD so my main concern was with trying to get a decent error message.  Putting in the `Session.request` would catch cases where you're using a session directly at the cost of a more generic error message.  Perhaps once we start printing warnings by default we can also add it to the `Session.request` method?

PR #1626, opened 2018-12-14
Add support for future SDK telemetry

kyleknap:


botocore/awsrequest.py
```diff
@@ -315,7 +315,7 @@ def create_request_object(request_dict):
     r = request_dict
     request_object = AWSRequest(
         method=r['method'], url=r['url'], data=r['body'], headers=r['headers'])
-    request_object.context.update(r['context'])
+    request_object.context = r['context']
```
jamesls: Why did this change?

botocore/utils.py
```diff
@@ -130,6 +130,22 @@
 }
 
 
+class _RetriesExceededError(Exception):
```
jamesls: Why is this needed?

kyleknap: Its not. I think it is the result of fixing merge conflicts where the `_RetriesExceededError` got moved in the `utils` module and now I have duplicates of it. I'll fix it.

botocore/awsrequest.py
kyleknap: I added it so that a handler mutating the `context` in the `request-created` event can have the changes be available in subsequent events (e.g. `response-received`). The other option would be to update the `request-created` event to include the `context` but I was trying to avoid expanding the event, especially because the `AWSRequest` object already includes the `context` as a property.

Issue #1460, opened 2018-05-08
Behavior for empty environment variables


posener:
In most programs, empty environment variable value results in a behavior as it was not defined.
In AWS CLI, defining an empty AWS_PROFILE environment variable results in the exception: `botocore.exceptions.ProfileNotFound: The config profile () could not be found`.
Botocore searches for a profile equal to an empty string.

To my opinion it should have the same behavior as AWS_PROFILE would not have been defined: using the `[default]` profile.

* `botocore.__version__ = '1.5.75'`
* OS: GNU/Linux

tim-finnigan:
Thanks for your patience on this issue. I brought this up with the team and commented [here](https://github.com/boto/botocore/pull/1927#issuecomment-1530249482) on the related PR. As mentioned there, changes regarding environment variables now must be considered at a cross-SDK level in order to help ensure more consistency. Please consider creating an issue in our cross-SDK repository (https://github.com/aws/aws-sdk) for further tracking if this is still something that is needed.

PR #1628, opened 2018-12-16
check for empty strings for Region (and others) and provide sensible fall-back

### LINKED ISSUE #1627: 'region' value never checked for empty string, nor any sane default applied
The api makes various calls to resolve service endpoints but at no stage does it actually check for empty strings nor does it provide the sensible default which per AWS docs is supposed to be 'us-east-1'. As a result of never checking for inputs programs throw ValueError on malformed endpoints like 'https://s3..amazonaws.com' - notice the double dot.

tb3088:
fixes #1627


tb3088:
addresses #1460 to some extent. The codebase relies extensively on testing for just None which is sloppy.

JordonPhillips:
Validating an empty region would be a breaking change in the case where you provide an endpoint url or where you are making unsigned requests with a service where region does not impact the url. We explicitly don't want a default region - it is not safe to assume where somebody should run a bit of code.

tb3088:
> it is not safe to assume where somebody should run a bit of code.

ok, I'll stipulate to that. However the rest of the diffs are valid since now the Boto API intercepts and treats as errors values that are empty string whereas previously it just blithely continued even though that value was clearly wrong.

tb3088:
What's with the rampant failure to check for 'empty' objects? There must be a zillion IF statements that should only proceed if the objects has a value that is non-empty. 'is [not] None' is an insufficient test.

JordonPhillips:
It's generally better to be explicit, unless you have a reason not to. When `None` is the default value we check for that explicitly,  because otherwise you start implicitly supporting any falsey values.

swetashre:
> Validating an empty region would be a breaking change in the case where you provide an endpoint url or where you are making unsigned requests with a service where region does not impact the url. We explicitly don't want a default region - it is not safe to assume where somebody should run a bit of code.

I am closing this PR based upon this comment.

PR #1632, opened 2018-12-20
Add support for endpoint discovery traits

jamesls:
This commit adds support for using endpoint discovery traits to determine the appropriate endpoint to use when sending a request.


botocore/serialize.py
```diff
@@ -460,6 +470,8 @@ def serialize_to_request(self, parameters, operation_model):
         if host_prefix is not None:
             serialized['host_prefix'] = host_prefix
 
+        self._serialize_api_version(serialized, operation_model)
```
stealthycoin: Are we really supposed to add this to every request?

jamesls: No, I'll update.  It should only be for the `DescribeEndpoints` call.

jamesls: Just to clarify, the serializer code handles this properly, but I think this should probably move to a separate handler.  Ideally we're not touching the serializer for this change.

PR #1633, opened 2018-12-21
Add generated paginators

JordonPhillips:
This adds generated paginators for a number of services. This only adds
the simple cases where the response contains only two elements: a list
which is the result key and the next token.

The script that generated these will look for the standard values
('NextToken', 'MaxResults') but also look for any values that are common
within a given service.

### DIFF
--- a/.changes/next-release/enhancement-Paginator-44918.json
+++ b/.changes/next-release/enhancement-Paginator-44918.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Paginator",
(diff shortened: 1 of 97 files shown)


copumpkin:
This looks awesome, thanks!

JordonPhillips:
I got a request to break this up into 98 PRs, one for each service. Thank god for `hub`

JordonPhillips:
After breaking out 20ish prs we decided it would be better to just add more generated tests so we really don't have to look at these.

Issue #1634, opened 2018-12-21
UnicodeDecodeError during logging of response body


reversefold:
While using secretsmanager if a secret doesn't exist the logging in parsers.py fails with a UnicodeDecodeError when trying to log the body of the message. This is happening to me even though debug logging is off:
```
Traceback (most recent call last):
  File "/var/cache/salt/master/git_pillar/54cafa49bdab016ddc25e48d69219c60ab0d4d22219733cf1ec652020b07cdc9/secrets/init.sls", line 45, in run
    get_secret_value_response = client.get_secret_value(SecretId=secret_key)
  File "/usr/local/lib/python2.7/dist-packages/botocore/client.py", line 320, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python2.7/dist-packages/botocore/client.py", line 611, in _make_api_call
    operation_model, request_dict, request_context)
  File "/usr/local/lib/python2.7/dist-packages/botocore/client.py", line 630, in _make_request
    return self._endpoint.make_request(operation_model, request_dict)
  File "/usr/local/lib/python2.7/dist-packages/botocore/endpoint.py", line 102, in make_request
    return self._send_request(request_dict, operation_model)
  File "/usr/local/lib/python2.7/dist-packages/botocore/endpoint.py", line 135, in _send_request
    request, operation_model, context)
  File "/usr/local/lib/python2.7/dist-packages/botocore/endpoint.py", line 167, in _get_response
    request, operation_model)
  File "/usr/local/lib/python2.7/dist-packages/botocore/endpoint.py", line 218, in _do_get_response
    response_dict, operation_model.output_shape)
  File "/usr/local/lib/python2.7/dist-packages/botocore/parsers.py", line 235, in parse
    LOG.debug('Response body:\n%s', response['body'])
  File "/usr/lib/python2.7/logging/__init__.py", line 1155, in debug
    self._log(DEBUG, msg, args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/salt/log/setup.py", line 310, in _log
    self, level, msg, args, exc_info=exc_info, extra=extra
  File "/usr/lib/python2.7/logging/__init__.py", line 1285, in _log
    record = self.makeRecord(self.name, level, fn, lno, msg, args, exc_info, func, extra)
  File "/usr/lib/python2.7/dist-packages/salt/log/setup.py", line 344, in makeRecord
    exc_info, func)
[...]

PR #1635, opened 2018-12-21
#1634: Use repr instead of str as the response body can have non-ascii chars

reversefold:
See issue #1634

---
This change is [Reviewable](https://reviewable.io/reviews/boto/botocore/1635)


JordonPhillips:
Could we get a test to verify this?

swetashre:
Following up on old PR. I am not able to reproduce the issue. Could we get a test for this ?

PR #1660, opened 2018-12-24
Use github instead of git_hub in operation names

JordonPhillips:
Our camel case to snake case code doesn't understand that while GitHub
is cased like that, it's actually one word. This fixes the specific
cases where GitHub is used in an operation name and aliases the old
names so nobody breaks.


kyleknap:
It may be nice to add a test for the aliasing of the methods similar to the ones we have for MTurk. If we want to go a little farther we could even generalize the test so it is easy to add test cases in the future.

PR #1663, opened 2019-01-01
Remove use of unnecessary compat shim six.binary_type

jdufresne:
On Python 2, starting with Python 2.6, the type bytes exists as an alias
of of str, same as the previously used six.binary_type. By using bytes,
the code has less indirection and is more forward compatible with modern
Python idioms.


nateprewitt:
Thanks for the PR, @jdufresne! It looks like this was unfortunately left unattended for quite some time.

We agree the shim should go and I believe there's a separate PR from Hugo removing these references rebased against the current codebase. We're planning to integrate that after we run pyupgrade, so I'll resolve this in favor of the other PR.

PR #1664, opened 2019-01-04
Add more fields to the ApiCall event 

kyleknap:
Specifically added the following fields to the ApiCall event:

* UserAgent
* FinalHttpStatusCode
* FinalAwsException
* FinalAwsExceptionMessage
* FinalSdkException
* FinalSdkExceptionMessage

### DIFF
--- a/botocore/monitoring.py
+++ b/botocore/monitoring.py
@@ -388,12 +388,25 @@ class CSMSerializer(object):
     def _serialize_attempts(self, attempts, event_dict, **kwargs):
         event_dict['AttemptCount'] = len(attempts)
         if attempts:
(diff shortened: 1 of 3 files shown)


botocore/monitoring.py
```diff
@@ -459,12 +472,20 @@ def _get_access_key(self, request_headers):
         return auth_match.group('access_key')
 
     def _get_region(self, request_headers):
+        if not self._is_signed(request_headers):
```
JordonPhillips: Why return no region if the request is unsigned?

kyleknap: It is part of the specification. The region used in the event must come from the region used when signing.

PR #1668, opened 2019-01-11
Add event stream support to the JSON protocol

joguSD:
This adds support for event streams to the JSON RPC protocol and support for parsing event stream exceptions.


botocore/parsers.py
```diff
@@ -658,16 +649,29 @@ def _do_parse(self, response, shape):
...
+                    'Code': response['headers'].get(':error-code', ''),
+                    'Message': response['headers'].get(':error-message', ''),
+                }
             }
-        }
         return error
 
     def _parse_payload(self, response, shape, member_shapes, final_parsed):
-        if shape.serialization.get('event'):
+        if shape.serialization.get('event') or shape.metadata.get('exception'):
```
jamesls: It doesn't seem like the `event` and `exception` code path should be consolidated:

* The `exception` shapes would never have an `eventpayload` so that code path will never be reached.
* We don't actually need to parse the error body according to its shape because `_do_error_parse` only checks for a `message` key.  Anything else it parses it just throws away.

Seems like it's simpler to just call `self._initial_body_parse` in `_do_error_parse`, and if you really wanted to be safe, a subsequent call to`self._parse_shape`.

botocore/parsers.py
```diff
@@ -722,6 +726,39 @@ def _initial_body_parse(self, xml_string):
...
 
+class JSONParser(BaseJSONParser):
+
+    EVENT_STREAM_PARSER_CLS = EventStreamJSONParser
+
+    """Response parser for the "json" protocol."""
+    def _do_parse(self, response, shape):
+        parsed = {}
+        if shape is not None:
+            event_name, event_shape = self._get_event_stream_member(shape)
```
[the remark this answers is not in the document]
joguSD: Yep, this is unfortunate. For now I've moved the logic to find this into the models which gives us caching at the client level. To avoid additional logic in the parsers I've added it to both the base `Shape` and the `Structure`, so short of a modelling change I think the impact of this should be minimal now.

botocore/parsers.py
joguSD: To be safe I'll still parse the body according to the shape.

joguSD:
The CLI side should be good already:
https://github.com/aws/aws-cli/blob/bae9a900501856fcee3389fb320fc831718b721d/awscli/customizations/removals.py#L43-L44

PR #1674, opened 2019-01-27
fix deprecation warnings found during running tests

mikedlr:
Running some test cases I got a bunch of deprecation warnings.

This is a very simple patch to eliminate those.  Mostly trivial docstring changes but one set of collections abstract base class changes which will cause compatibility problems with python 3.8.


mikedlr:
I think that https://github.com/boto/botocore/pull/1675 fixes the reason why my tests were calling in to the vendored code.

Jamim:
Hello @jamesls,

Could you please take a look at this PR?

Thanks!

nateprewitt:
This file ended up getting removed when we unvendored urllib3. It should now be compliant when installing a current version of urllib3. Thanks for the contribution!

PR #1675, opened 2019-01-27
unvendor urllib in exceptions class

mikedlr:
According to 

- https://botocore.amazonaws.com/v1/documentation/api/latest/index.html#upgrading-to-1-11-0
- https://github.com/boto/botocore/issues/1248

urllib has been unvendored and should not be used any more.  In the exceptions class however there's still a reference to it I found with 

    grep -r import .| grep urllib3  | grep botocore.vendored

this patch removes the urllib vendoring in the exceptions class.  Maybe rules like the grep above should also be added to the CI to ensure that all deprecated vendored classes stay removed?

### DIFF
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -13,7 +13,7 @@
 # language governing permissions and limitations under the License.
 from __future__ import unicode_literals
 from botocore.vendored import requests
-from botocore.vendored.requests.packages import urllib3
+import urllib3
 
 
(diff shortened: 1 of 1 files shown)


nateprewitt:
Thanks for the PR, @mikedlr. While we did remove the majority of the vendored code, we still provide exceptions from this folder. In order to maintain consistency in existing code, we're going to leave this reference for now. In a future version, we may revisit this change if our stance on the current vendoring changes.

PR #1679, opened 2019-02-06
Conform to vpc endpoint standard of {service}.{region}.{dnsSuffix}.

buroa:
SQS VPC endpoints by default use the {service}.{region}.{dnsSuffix} standard. The ones that are listed in the defaults.go use the old queue.amazonaws.com standard. This should be fixed.


buroa:
https://github.com/boto/botocore/issues/1681

PR #1694, opened 2019-03-06
Fix typo in RoboMaker documentation

ryanewel:


stealthycoin:
Thanks for the find. These files are generated upstream from us and we consume them as is. Any changes we made here would just be clobbered in the next release. I will make sure upstream is aware of this typo so it gets fixed at the source.

PR #1699, opened 2019-03-14
Don't modify arguments to generate_presigned_post

RasmusWL:
This is motivated by a real-world example, where I did

```
my_conditions = <...>

generate_presigned_post(Bucket=.., key='foo', Conditions=my_conditions)
generate_presigned_post(Bucket=.., key='bar', Conditions=my_conditions)
```

and performing the second presigned post failed because `my_conditions` was
altered to include `{'key': 'foo'}` after the first call


RasmusWL:
Great to hear you got that fixed @swetashre :tada:

PR #1705, opened 2019-03-31
Clarify Batch documentation for describe_jobs

adelevie:
This change clarifies that the input to `describe_jobs` should be a list of strings rather than a space-separated string.


jamesls:
Thanks for reporting.  Per our [contributing guidelines](https://github.com/boto/botocore/blob/develop/CONTRIBUTING.rst#contributing-code) the service-2.json files come from an upstream source so we can't accept PRs to those files.  I've cut a ticket internally to get our doc team to update this.  Thanks again.

HLeeHart:
The update is now live.

PR #1715, opened 2019-04-13
Find the partition of a region

jkinred:
This commit implements the core functionality for a feature request
against `boto3`.

See https://github.com/boto/boto3/issues/1868.

I implemented the method with a more verbose name. Happy to adjust if needed.

### DIFF
--- a/botocore/regions.py
+++ b/botocore/regions.py
@@ -115,6 +115,16 @@ class EndpointResolver(BaseEndpointResolver):
                     result.append(endpoint_name)
         return result
 
+    def get_partition_for_region(self, region_name):
+        for partition in self._endpoint_data['partitions']:
+            if region_name in partition['regions']:
+                return partition['partition']
+            if 'regionRegex' not in partition:
+                continue
+            if re.compile(partition['regionRegex']).match(region_name):
+                return partition['partition']
+        raise ValueError('Invalid region name: {0}'.format(region_name))
+
     def construct_endpoint(self, service_name, region_name=None):
         # Iterate over each partition until a match is found.
         for partition in self._endpoint_data['partitions']:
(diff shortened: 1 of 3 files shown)


shorea:
@joguSD is this something that can be merged? This would be a pretty useful feature to have.

jkinred:
If this is a useful feature that may be merged, I will invest some time in cleaning up the checks and rebasing.

jeking3:
I would make use of this if it was merged as well.  It seems pretty useful for ARN generation.

aph3rson:
+1ing this PR as well, this would be helpful for instances where the partition is necessary for certain functionality (e.g. determining console and signin URLs, which [my tool](https://github.com/netspi/aws_consoler) depends on).

jonrau1:
Throwing another +1 into the mix. My tool, ElectricEye, relies on the partition a lot for creation resource ARNs and creating Security Hub findings. This would allow me to not have to hard-code partitions and create separate implementations per partition.

jfuss:
This would be really useful for SAM as well. Is there a way we can get this merged @kyleknap or @stealthycoin?

jfuss:
@nateprewitt or @joguSD might be better people to ask merge-ablity of the feature.

wong-a:
Another +1! This is a useful method. What's blocking it?

yuan-bwn:
+1 for visibility!
It would be really great customer can fetch this info directly from boto instead of implementing all different sorts of helper methods.

tests/unit/test_session.py
```diff
@@ -379,6 +379,13 @@ def test_provides_empty_list_for_unknown_service_regions(self):
         regions = self.session.get_available_regions('__foo__')
         self.assertEqual([], regions)
 
+    def test_provides_correct_partition_for_region(self):
```
nateprewitt, line 382: Can we get a couple more tests. I think we want to make sure we're validating the regex with something like `af-south-99` (doesn't exist but matches regex), and a partition other than `aws` (`cn-north-1` or `us-gov-east-1` would work).

PR #1717, opened 2019-04-16
Ignore empty strings for env var creds

jamesls:
This pulls in https://github.com/boto/botocore/pull/1687 and simplifies the code and adds a changelog entry.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -921,7 +921,10 @@ class EnvProvider(CredentialProvider):
         """
         Search for credentials in explicit environment variables.
         """
-        if self._mapping['access_key'] in self.environ:
+
+        access_key_var = self._mapping['access_key']
+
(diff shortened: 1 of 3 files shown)


botocore/credentials.py
```diff
@@ -950,30 +953,32 @@ def _create_credentials_fetcher(self):
...
+                if token:
+                    credentials['token'] = token
                     break
-            credentials['token'] = token
 
-            expiry_time = environ.get(mapping['expiry_time'])
-            if require_expiry and expiry_time is None:
+            credentials['expiry_time'] = None
+            expiry_time = environ.get(mapping['expiry_time'], '')
+            if expiry_time:
```
kyleknap: Are there tests for the empty string expiry time? My guess is that they should be similar to the ones that got added for when access and secret keys were set to an empty string.

jamesls:

PR #1718, opened 2019-04-17
Header X-Amz-Content-SHA256

hendawy:
Enabling signature v4 header *X-Amz-Content-SHA256*
Example:
```
import boto3

config = Config(signature_version='v4')
client = boto3.client('s3', region_name='eu-central-1', config=config)
filename = '/my/file.test' 

with open(filename,"rb") as f:
    client.put_object(
        Key="file.test", Body=f,
        Bucket="bucket-name",
        ContentSHA256=sha256_hash_hexdigest)
```

### DIFF
--- a/botocore/data/s3/2006-03-01/service-2.json
+++ b/botocore/data/s3/2006-03-01/service-2.json
@@ -1601,6 +1601,7 @@
     "ContentLanguage":{"type":"string"},
     "ContentLength":{"type":"long"},
     "ContentMD5":{"type":"string"},
+    "ContentSHA256":{"type":"string"},
     "ContentRange":{"type":"string"},
     "ContentType":{"type":"string"},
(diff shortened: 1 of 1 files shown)


jamesls:
Hi, thanks for the pull request.

Unfortunately we aren't able to accept contributions to our service-2.json file.  These are generated from an upstream source so we don't modify them directly (more info in our contributing guide here: https://github.com/boto/botocore/blob/develop/CONTRIBUTING.rst#contributing-code).

If you can share your use case we can try and come up with an alternate approach.  Botocore will automatically compute and inject this header for you when signing requests.  Is there a reason you want to explicitly add this value yourself?

nateprewitt:
Resolving since we haven't heard back. Please feel free to reopen if you have more info. Thanks!

PR #1723, opened 2019-05-06
Pluggable HMAC signers

### LINKED ISSUE #1689: Refactoring HMAC signers in botocore/auth.py to support Yubikey-based signing
I have an initial implementation of a package for storing AWS access keys on Yubikey devices (https://github.com/pyauth/exile) and having the Yubikey sign AWS requests. This allows me to store the secret key for any given key ID on a Yubikey instead of in the home directory, protecting it in the YubiKey (which acts as an HSM, so theft of home directory contents does not compromise the key), and optionally further requiring user interaction (pressing a button on the key) to sign the request.

To integrate this into botocore, I currently have to monkey-patch signers in `botocore/auth.py`. The first step to not needing this monkey-patching is to factor out HMAC signers so instead of taking the literal secret HMAC key and string to sign, they can take a `(botocore.credentials.Credentials, string_to_sign)` tuple.

After that I can make a `botocore.credentials.CredentialProvider` subclass that would indicate that the credential is actually in the Yubikey, and think of how to structure the rest of the plumbing.

Does this sound reasonable? If so, I can get started on the first PR for factoring out the HMAC signers.

kislyuk:
To accommodate sequestration of keys into HSMs, signers should not assume that a key is always available verbatim from `self.credentials.secret_key`.

This PR introduces a new class, botocore.credentials.SigningCredentials, and utility methods in botocore.auth to delegate HMAC signing to the credentials class if it's a subclass of SigningCredentials. This allows a custom credentials subclass to be passed in to botocore at init time, that signs the message using a key whose value is only known to a processor inside a hardware security key (such as a YubiKey).

Closes #1689

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -24,6 +24,8 @@ import time
 import calendar
 import json
 
+from botocore.credentials import (Credentials, SigningCredentials,
+                                  ReadOnlyCredentials)
 from botocore.exceptions import NoCredentialsError
 from botocore.utils import normalize_url_path, percent_encode_sequence
 from botocore.compat import HTTPHeaders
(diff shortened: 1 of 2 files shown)


kislyuk:
@JordonPhillips may I please get a review on this?

kislyuk:
Any updates on this?

kislyuk:
Still waiting for feedback on this.

kislyuk:
It would be great if this PR received a review.

indrora:
Can I get

* Tests that show there aren't any regressions from the old behavior
* Tests that document the feature.
* Documentation on how to use this feature
* An example of using this change in action.

kellertk:
Hi there! Apologies for the possible confusion; I deleted a comment on this PR that I think might have been meant for a different PR.

Thank you for your contribution, but the team has decided not to move forward with this PR at this time.

PR #1724, opened 2019-05-06
Prefer AWS_PROFILE over AWS_DEFAULT_PROFILE

samanthaq:
If one has both `AWS_DEFAULT_PROFILE` and `AWS_PROFILE` environment variables set in their environment, botocore will prefer the default profile. I believe the intended behaviour is to use `AWS_PROFILE` when set (similar to preferring a `profile` set in the config over the environment variables).

This can be confirmed by running the introduced `functional/test_session.py:test_default_profile_precedence` test, or by setting both environment variables to valid aws profiles and calling, e.g., `aws sts get-caller-identity`.

See https://github.com/boto/botocore/issues/1725

### DIFF
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -43,7 +43,7 @@ from botocore import utils
 #: found.
 BOTOCORE_DEFAUT_SESSION_VARIABLES = {
     # logical:  config_file, env_var,        default_value, conversion_func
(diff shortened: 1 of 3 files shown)


joguSD:
Thanks for submitting this PR, it's well put together. Unfortunately, we can't merge this change as it's not backwards compatible and would break customers relying on the reverse resolution order currently.

PR #1727, opened 2019-05-09
Retry by default for Kinesis endpoints

mjschultz:
Several Kinesis endpoints respond with LimitExceededException, this makes the default behavior for all of them retry. At least the following endpoints have small TPS limits that this will benefit: ListStreams, CreateStream, DeleteStream, DescribeStream, GetRecords

### DIFF
--- a/botocore/data/_retry.json
+++ b/botocore/data/_retry.json
@@ -180,7 +180,7 @@
       }
     },
     "kinesis": {
-      "DescribeStream": {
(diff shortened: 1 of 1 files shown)


jamesls:
Thanks for the PR, this change ended up being needed for https://github.com/boto/botocore/pull/1719/files as well, which was just merged.  Botocore should now be retrying that exception for all Kinesis operations.

PR #1750, opened 2019-06-05
Move S3 bucket versioning test to test_s3.py

joguSD:
This also adds a simple waiter class for use in our tests to help in cases where we need to wait for API to reach eventual consistency.

### DIFF
--- a/tests/integration/test_s3.py
+++ b/tests/integration/test_s3.py
@@ -11,7 +11,10 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
-from tests import unittest, temporary_file, random_chars, ClientHTTPStubber
+from tests import (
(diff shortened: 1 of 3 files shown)


tests/integration/test_s3.py
```diff
@@ -154,6 +157,16 @@ def _wait_for_key(self, bucket_name, key_name, extra_params=None,
...
             waiter.wait(**params)
 
+    def _check_bucket_versioning(self, bucket, enabled=True):
+        response = self.client.get_bucket_versioning(Bucket=bucket)
+        status = response.get('Status')
+        return status == 'Enabled' if enabled else status != 'Enabled'
+
+    def wait_until_versioning_enabled(self, bucket, min_successes=3):
+        waiter = ConsistencyWaiter()
+        for _ in range(min_successes):
```
kyleknap: What's your thoughts on including the min successes in the logic of the `wait` command as well? I suspect we will be relying on min successes quite often.

joguSD: I've been going back and forth on this. I'll incorporate the min success logic into consistency waiter as well.

PR #1752, opened 2019-06-06
Raise NoCredentialsError when no credentials are found

tehtechguy:
Certain credential providers, such as the `InstanceMetadataProvider`  have the chance to fail to load credentials on start, even when the credentials exist and valid. Typically, such failures would get handled by the `RefreshableCredentials` class; however, that refresh only occurs if the provider was able to find credentials (thereby being the chosen credential provider).

The above situation results in a scenario where it's possible to create a session that does not have credentials. This will result in the client raising a `NoCredentialsError` upon the first usage of the credentials. That error will not have the ability to auto-resolve.

This PR moves the failure farther up the pipe, such that if credentials cannot be obtained, it will immediately bubble up to the user.

Having systematic credential failures like this are critical to be found upon session creation, rather than client use, as it enables services utilizing the session to know with certainty whether the session is usable or not.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -31,6 +31,7 @@ import botocore.configloader
 import botocore.compat
 from botocore.compat import total_seconds
 from botocore.compat import compat_shell_split
+from botocore.exceptions import NoCredentialsError
 from botocore.exceptions import UnknownCredentialError
 from botocore.exceptions import PartialCredentialsError
 from botocore.exceptions import ConfigNotFound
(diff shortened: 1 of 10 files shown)


tehtechguy:
Looking at the [Java SDK](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/auth/AWSCredentialsProviderChain.java#L136), it does appear that when no credentials are able to be loaded an exception occurs. This is therefore a mismatch between AWS SDKs.

James-TR:
Might want to look at changing the tests?

tehtechguy:
Thanks - I was too quick to look over the output from Travis. I just pushed up the test fixes. At the very least everything under `tests/unit` passes locally

tehtechguy:
I determined my local tests passed because I had valid AWS credentials, so it was finding those. I updated those tests (ones that created a client) to simply have a mock credentials object. Hopefully this one will pass 🤞

tehtechguy:
Tests fixed. All issues were around not mocking credentials, except the change to `test_misconfigured_source_profile`, which from what I could tell appeared to be a result of a missing parameter.

Krzysztof-Szyper-Epam:
@tehtechguy You might wanna have a look at merge conflicts here.

stobrien89:
Hi @tehtechguy,

Our apologies for the lack of attention to this.  We won't be able to merge this, as it would be a backwards incompatible change for some interfaces (not all client functionality requires credentials) to raise an error instead of returning `None`.  If you're still experiencing problems related to resolving credentials, we'd be more likely to accept something that makes credential existing credential providers more resilient.

PR #1757, opened 2019-06-10
Don't throw error for invalid old starting token

JordonPhillips:
Botocore has support for its own starting token formats, the older
of which simply splits up necessary data with triple underscores.
This format has been deprecated for a while due to various
shortcomings, but there's still input support for it for backwards
compatibility. The last element in that token format should be an
integer, which indicates how far to slice into the response. The
problem is that some services have tokens which include triple
underscores, so we think it's that format but malformed. This fixes
the problem by passing the full starting token along if parsing
fails.

Fixes boto/boto3#1984

### DIFF
--- a/.changes/next-release/bugfix-Paginator-26689.json
+++ b/.changes/next-release/bugfix-Paginator-26689.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Paginator",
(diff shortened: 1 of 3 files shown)


joguSD:
Not sure why codecov is freaking out, but this looks fine to me.

PR #1761, opened 2019-06-14
Use DEFAULT_URI_TEMPLATE if not exists hostname

shadchin:


shadchin:
Anyone?

shadchin:
Hello @jamesls,
Could you please take a look at this PR?
Thank you!

shadchin:
it seems to me that this is a mistake and needs to be fixed

nateprewitt:
Hi @shadchin,

It doesn't look like this PR was ever tied to an issue or provided context on the change. There _shouldn't_ be a case where `hostname` isn't populated in a normal region resolution flow. If that is the case, we'd need more information on how that state is being reached. The proposal here, seems reasonable in the context of the code at the time, but doesn't match the current state of botocore. Specifically, `hostname` as a variable no longer exists.

We're going to close this for now, but please feel free to open an issue with an explanation of how you're encountering this issue and we can talk through options. Thanks!

PR #1779, opened 2019-07-10
Ignore Reason Phrase when reading 100 (Continue) response

stuio:
HTTP Status message is optional.

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -237,9 +237,9 @@ class AWSConnection(object):
 
     def _is_100_continue_status(self, maybe_status_line):
         parts = maybe_status_line.split(None, 2)
-        # Check for HTTP/<version> 100 Continue\r\n
+        # Check for HTTP/<version> 100 *\r\n
         return (
-            len(parts) >= 3 and parts[0].startswith(b'HTTP/') and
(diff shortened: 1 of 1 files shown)


stuio:
Some extra context:

https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html

Specifically:

"The individual values of the numeric status codes defined for HTTP/1.1, and an example set of corresponding Reason-Phrase's, are presented below. The reason phrases listed here are only recommendations -- they MAY be replaced by local equivalents without affecting the protocol."

and

"The Reason-Phrase is intended to give a short textual description of the Status-Code. The Status-Code is intended for use by automata and the Reason-Phrase is intended for the human user. The client is not required to examine or display the Reason- Phrase."

We have a scenario where the Reason Phrase is omitted entirely for 100 (Continue) status, which although compliant, is causing issues with using the aws-cli.

This PR ensure boto-core processing of 100 (Continue) messages is compliant. We haven't examined other HTTP Status Line processing within the library so it's possible a broader PR covering all status line processing may be necessary which invalidates the need for this one.

nateprewitt:
It appears this PR is a duplicate of #1932. We're going to resolve this in favor of the other PR since it has tests around the change.

PR #1788, opened 2019-07-21
Blacklist docutils 0.15

dstufft:


asottile:
given upstream fixed this, should this be reverted now?

joguSD:
@asottile We intend to keep an upper bound going forward, all of our dependencies should have one and `docutils` was a bit of a miss since it releases so infrequently. We're going to do a little verification but I don't suspect there's much holding us back from bumping this to `<0.16`.

asottile:
here's hoping that doesn't take [3 months](https://github.com/aws/aws-cli/pull/4015) <3

glaubitz:
I'm packaging botocore and boto3 for openSUSE/SLE and we're therefore using the docutils version available in the various openSUSE distributions which will certainly be >= 0.15 for the openSUSE Factory distribution. We cannot pick arbitrary versions as the docutils package is shared with other python packages.

Thus, I was wondering what the reason is that newer versions of docutils are supposed to be blacklisted? Is there any incompatibility we need to be aware of

CC @rjschwei

asottile:
> I'm packaging botocore and boto3 for openSUSE/SLE and we're therefore using the docutils version available in the various openSUSE distributions which will certainly be >= 0.15 for the openSUSE Factory distribution. We cannot pick arbitrary versions as the docutils package is shared with other python packages.
> 
> Thus, I was wondering what the reason is that newer versions of docutils are supposed to be blacklisted? Is there any incompatibility we need to be aware of
> 
> CC @rjschwei

docutils 0.15 temporarily shipped a syntax error in python 2, it has since been corrected

Issue #1703, opened 2019-03-28
Use of urllib3 exposed to vulnerability CVE2018-20060


yoavrotems:
Hey,
I see that you're keeping urllib3 for legacy reasons https://github.com/boto/botocore/issues/1613.
We use your boto3 lib and found out that its opening our code (and anyone else who is including botocore) to a high severity vulnerability CVE2018-20060

Please could you reconsider removing (or updating) this library, to remove the vulnerability?

JordonPhillips:
It only opens your code to vulnerabilities if you are using it. If you are using the documented interface of botocore then you're fine. botocore does not and will not use it, so botocore / boto3 / cli usage does not expose you to those vulnerabilities.

We plan on removing that code in our next major version bump.

lizrice:
Do you know when that major version bump is planned? It does show up as a high-sev vulnerability in container image scans.

c22:
I don't consider myself an expert on vendored python modules, but I believe aws-cli may be unintentionally importing the vendored urllib3 from botocore via this line here:

https://github.com/aws/aws-cli/blob/develop/awscli/customizations/configure/__init__.py#L14

Probably the aws-cli issues is a better place for this but I thought I'd drop a note here, too.

I will file a report with the aws-cli issue tracker and hopefully someone can comment on whether there is an issue or not.

c22:
Also this issue seems to be essentially very similar to #1608.

cdubz:
@JordonPhillips,

According to @joguSD in aws/aws-cli#4082 --
> The only usage of this vendored package is extending the exception classes in case customers were using the vendored exceptions in try/excepts.

Since the usage is so basic here, is there any reason not to update urllib3 to an unaffected version, verify that the exception still works, and include the change in a maintenance release? I understand that the package is unaffected by the actual vulnerability, but it does raise flags and, e.g. with aquasecurity/kube-hunter#112, that can cause minor headaches.

countergram:
Right, this is currently our most common "high-severity" CVE reported across our containers. Having to make lots of exceptions & triple check that it really isn't used.

nateprewitt:
We don't vendor urllib3 directly anymore and allow up to the latest release. I'm going to resolve this since we shouldn't be exposing any of these issues in a current release.

struginskij:
Any update?

nateprewitt:
[...]

PR #1793, opened 2019-07-25
Upgrade requests to 2.22.0

countergram:
This addresses https://github.com/boto/botocore/issues/1703

The change consists of copying requests 2.22 files into the vendored/requests directory which appears to be what has done before, and adding the new dependencies that requests has since version 2.16 (which are loaded in __init__ and thus cannot be skipped even if only importing exceptions)

### DIFF
--- a/botocore/vendored/requests/__init__.py
+++ b/botocore/vendored/requests/__init__.py
@@ -6,7 +6,7 @@
 #          /
 
 """
-requests HTTP library
+Requests HTTP Library
 ~~~~~~~~~~~~~~~~~~~~~
 
 Requests is an HTTP library, written in Python, for human beings. Basic GET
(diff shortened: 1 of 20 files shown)


Jamim:
Conflicting PR: #1817

chrisdlangton:
@countergram for context https://github.com/boto/botocore/issues/1822
As i found out, not all files in vendored are actually "vendor file". The reason they exist embedded versus as a dependency is so boto3 maintainers can apply changes to the vendored libraries that couldn't/wouldn't be merged upstream.

A better approach for you, and this PR, would be to remove `vendored` requests altogether, and just reintroduce requests as a dependency. See what breaks?

Good luck with this PR, i fear you're breaking backwards compatibility or wiping out some boto3 added logic somewhere, but I am rooting for you!

joguSD:
We no longer use the vendored versions of requests. See this issue: #1829

PR #1795, opened 2019-07-26
Add AssumeRoleWithWebIdentity credential provider

joguSD:
The web identity provider uses the full config off of the session
instead of relying on the session's get_config. While the new config
system in session is nice it has two major probles that make it
unsuitable for this use case:

1) If a profile is statically configured we need to ignore the
environment variables for that configuration. While it is possible to
use the methods kwarg, it's extremely inefficient.
2) The larger blocker is that this provider needs to support loading
configuration from a profile that is not the currently configured
profile for a session (e.g. when used as a source_profile in an assume
role chain). This is not trivial to add to the session's get_config and
doesn't really fit into the way it is designed.

Using the full config is what the other profile based providers do,
and this makes this new provider consistent with that.

* Add retries to sts for IDPCommunicationError

* Backport the StubbedSession class from chalice

This backports the StubbedSession class from chalice and refactors some
of the assume role tests to use as it simplifies things. This does not
backport the improvements or wrapper class that chalice has for the
client level stubber.

### DIFF
--- a/tests/functional/test_credentials.py
+++ b/tests/functional/test_credentials.py
@@ -508,6 +508,58 @@ class TestAssumeRole(BaseAssumeRoleTest):
         actual_creds = session.get_credentials()
         self.assert_creds_equal(actual_creds, expected_creds)
         stubber.assert_no_pending_responses()
+        # Assert that the client was created with the credentials from the
+        # assume role with web identity call.
+        self.assertEqual(self.mock_client_creator.call_count, 1)
+        _, kwargs = self.mock_client_creator.call_args_list[0]
+        expected_kwargs = {
+            'aws_access_key_id': identity_creds.access_key,
+            'aws_secret_access_key': identity_creds.secret_key,
+            'aws_session_token': identity_creds.token,
+        }
+        self.assertEqual(kwargs, expected_kwargs)
+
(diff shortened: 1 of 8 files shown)


tests/functional/test_credentials.py
```diff
@@ -472,6 +478,37 @@ def test_process_source_profile(self):
...
+            'assume_role_with_web_identity',
+            identity_response,
+        )
+
+        expected_creds = self.create_random_credentials()
+        assume_role_response = self.create_assume_role_response(expected_creds)
+        stubber.add_response('assume_role', assume_role_response)
+
+        actual_creds = session.get_credentials()
+        self.assert_creds_equal(actual_creds, expected_creds)
```
kyleknap: We should make sure to assert that we are creating the sts client with the correct credentials similar to how we do it for the process provider.

tests/functional/test_credentials.py
```diff
@@ -472,6 +478,37 @@ def test_process_source_profile(self):
         }
         self.assertEqual(kwargs, expected_kwargs)
 
+    def test_web_identity_source_profile(self):
```
kyleknap: It's probably worth adding a functional test to make sure we do not use the env vars when chaining.

joguSD:
The force push was just to rebase off the latest develop.

PR #1801, opened 2019-08-02
Add Glue JobRunSucceeded waiter

jmsanders:
This adds a `JobRunSucceeded` waiter to Glue. Polling delay and max attempts are borrowed from the EMR `StepComplete` waiter.

### DIFF
--- a/botocore/data/glue/2017-03-31/waiters-2.json
+++ b/botocore/data/glue/2017-03-31/waiters-2.json
@@ -1,0 +1,42 @@
+{
+  "version": 2,
+  "waiters": {
+    "JobRunSucceeded": {
+      "delay": 30,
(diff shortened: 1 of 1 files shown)


jmsanders:
@kyleknap @JordonPhillips I noticed you've reviewed some of the more recent PRs that add waiters. Is there anything else you'd like to see here before this is ready to merge?

stobrien89:
Hi @jmsanders 

Thanks for the request. We are now submitting requests for paginators and waiters upstream so that all AWS SDKs benefit from a consistent implementation. I will pass on the request for this waiter to the service team. We will also update the guidance in the contributing guide to better reflect this policy. I've added an issue to track this here:

aws/aws-sdk#82

PR #1804, opened 2019-08-08
Add pagination support for `ListUsers`

delissonjunio:
Simple pagination support for Cognito's `ListUsers` call


swetashre:
@delissonjunio - Thank you for the PR. Paginator is already there for cognito-idp ListUsers 
https://github.com/boto/botocore/blob/develop/botocore/data/cognito-idp/2016-04-18/paginators-1.json#L51

Closing the PR as it has already been fixed.

PR #1806, opened 2019-08-09
Provide better error msg on config load errors on py3

jamesls:
In python3, it's possible that parse errors when reading the
cfg file will result in unicode decode errors instead of the
configparser errors.  As a result, we don't wrap that error
in our ConfigParseError which leads to an opaque error message:

```
ERROR: test_config_parse_error_bad_unicode (tests.unit.test_configloader.TestConfigLoader)
  File "botocore/tests/unit/test_configloader.py", line 73, in test_config_parse_error_bad_unicode
    raw_config_parse(filename)
  File "botocore/botocore/configloader.py", line 149, in raw_config_parse
    cp.read([path])
  File "/pythons/CPython-3.6.3/lib/python3.6/configparser.py", line 697, in read
    self._read(fp, filename)
  File "/pythons/CPython-3.6.3/lib/python3.6/configparser.py", line 1015, in _read
    for lineno, line in enumerate(fp, start=1):
  File ".virtualenvs/aws-cli36/lib/python3.6/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xad in position 2: invalid start byte

```


jamesls:
Looks like travis is passing, going to go ahead and merge.

Issue #454, opened 2015-02-10
Signature is absent in headers for 


Andrey-mp:
I have a set of tests. I wrote it for checking EC2 layer of OpenStack.
I run it against Amazon first and check that all tests are passed.

But botocore with version 0.86 has a strange behavior against Amazon.
First 5-8 tests are passed but all next tests are failed with message
'The request must contain the parameter AWSAccessKeyId (MissingParameter)'.
I checked and found that failed requests didn't have 'Authorization' header.

version 0.85 works well.

jamesls:
Are you still seeing this issue on the latest commit on the `develop` branch?

Andrey-mp:
yes. behavior is the same

jamesls:
Interesting.  We recently upgraded the version of requests we use, and we had to put a specific override in their to ensure the Auth header isn't being removed.  Can you share sample code that demonstrates this issue?

Andrey-mp:
All our tests are here - https://github.com/stackforge/ec2-api/tree/master/ec2api/tests/functional

botocoreclient.py - encapsulates access to botocore library by method name.
base.py - is a base class for all tests
api folder contains tests

tests run against OpenStack can be found here - http://logs.openstack.org/75/154075/11/check/check-functional-neutron-dsvm-ec2api/6d76d59/console.html

run against Amazon is the same except error message.

Andrey-mp:
This is my fail. In second run code worked with cached operation and new service/endpoint.
Fix for my code is here - https://review.openstack.org/#/c/156625/

PR #1810, opened 2019-08-16
Expose close() method in client instances

vtermanis:
Although the (now outdated) #1499 (add `__del__`) and #1231 (close/release connection back to pool) try to address https://github.com/boto/boto3/issues/454 , this PR suggests simply exposing a `close` method on `botocore.client` client instances. It should then be possible to e.g. use `contextlib.closing` or to manually close the client after usage:

```python
with closing(session.client(service_name='secretsmanager', ...)) as client:
    secret = client.get_secret_value('my_secret')
    # Make any additional calls for the duration of the client instance
```

~I have not added any tests since I wasn't sure what would be appropriate. This additional method however should not affect any existing behaviour.~ Have added test to exercise the new methods.


vtermanis:
> We encourage you to check if this is still an issue in the latest release

There still is no `close()` (or equivalent method).

vtermanis:
Rebased to be against current top of `develop` branch again.

akx:
👍 , would love to see this get merged – I currently have to rely on 

```
def close_client(client: BaseClient):
    # See https://github.com/boto/botocore/pull/1810 ...
    try:
        client._endpoint.http_session._manager.clear()
        return True
    except Exception:
        return False
```

akx:
Could we get e.g. @nateprewitt to take a look at this?

hannes-ucsc:
If I wanted the client to be long-lived in order to maximize the benefit of the connection pool, would I still get the warnings from #454 even with the changes from this PR?

nateprewitt:
Hi @vtermanis, I left a response to https://github.com/boto/boto3/issues/454 a bit earlier. I think we're going to try and get this added to an upcoming release. I've rebased your change onto our current `develop` branch to get CI running again.

I think we'll likely want to add some doc changes and potentially a couple more tests before merging. I'm still working through what those will be and will update once we have more. Please let me know if you'd like to continue this work or would rather I build on top of your existing PR. Thanks!

nateprewitt:
> If I wanted the client to be long-lived in order to maximize the benefit of the connection pool, would I still get the warnings from https://github.com/boto/botocore/issues/454 even with the changes from this PR?

@hannes-ucsc, to answer your question, you would likely still see these warnings for long-lived clients. The existence of the ConnectionPool is what's causing these warnings to arise. The outcome of this PR will only provide the ability to forcibly cleanup client ConnectionPools when you're finished with them.

vtermanis:
@nateprewitt - Thanks for considering this. It's been a while since I look at the codebase (and I don't have quite as much time as I used to - though I guess who has, eh?)  - so I think you building on top of this will be more efficient at this point.

nateprewitt (approved):
Alright, things are rebased and the tests are cleaned up. Let's ship it! Thanks again, @vtermanis

sparrowt:
https://github.com/boto/boto3/issues/454#issuecomment-1047277122 says the following regarding this PR:
> https://github.com/boto/botocore/pull/1810 has some surprising side effects that are going to catch people. When the user calls close, it invokes it on the PoolManager. That not only closes the connections for the client/resource in use, but ALL clients and resources created by the underlying Session. This may lead to unexpected failures in parallelized environments and introduces a new level of management required for client cleanup. You MUST ensure all users of the Session are ready for teardown before invoking close.
> ...
> We do need to make sure the implications are clearly documented for users though.

Is that still the case with the final code that was merged? (from an outsider's brief look it would appear so) I couldn't find any documentation either in this PR or generally on boto3/botocore docs so I'm wondering whether `close()` is safe to use or of the surprising side effects are still a risk.

mmerickel:
If that’s the case shouldn’t the close() method be on the session and not the client/resource? It’s where I looked for a method first before finding this ticket and using the version in the client as in the comments above.

PR #1812, opened 2019-08-19
Changes to boto3 and botocore to implement AWS rate limiting client side

museadmin:
Hello

This PR is for changes to both boto3 and botocore. The changes are explained in detail on the README of master project I've created to test them together here - https://github.com/museadmin/botorate/tree/master

Your contributing guidelines suggest that I should be supporting some deprecated versions of Python. I'm not clear if this is a real requirement or if your docs are just a little out of date. I raised an issue asking the question with boto3, but have decided to submit the PR to see what other specific issues you might want resolving.

If accepted, I think you would need to coordinate the merge with boto3, whom I'm also raising a PR with.

Bradley

boto3 PR - https://github.com/boto/boto3/pull/2086


j4v:
This is much needed as AWS's implementation of rate limiting is very hard to handle.

museadmin:
Would it be possible to get a little help with this PR from someone with some experience of managing the travis ci build process?

I'm still struggling with the build process for python2.x versions but will have a crack at the setup.py syntax tomorrow and see if I can work out why it doesn't install two packages I require.

The issue I need help with is for the 3.x builds which I think would now pass if you were pulling in my fork of boto3 instead of the current version. Looking at the failure it's complaining that I have an unexpected parameter being passed, (api_rate=0) which is the parameter I would be passing in from my fork of boto3:

```
AssertionError: Expected call: create_client(api_version=<ANY>, client_config=<botocore.config.Config object at 0x7f367a082438>, credentials=<ANY>, endpoint_url=<ANY>, is_secure=<ANY>, region_name=<ANY>, scoped_config=<ANY>, service_name=<ANY>, verify=<ANY>)
Actual call: create_client(api_rate=0, api_version=None, client_config=<botocore.config.Config object at 0x7f367a082438>, credentials=<botocore.credentials.Credentials object at 0x7f367a1a6e80>, endpoint_url=None, is_secure=True, region_name='us-west-11', scoped_config={'aws_access_key_id': 'fie', 'aws_secret_access_key': 'baz'}, service_name='sts', verify=None)
```

I don't know how to get your ci pipeline to use two forked PRs for boto3 and botocore at the same time....

museadmin:
Ok boto

Finally have a green build. 

My PR for boto3 won't pass until this is merged. 
Do I have to do anything else to get this in?

joguSD:
Thanks for putting this together, it looks like you've put quite a bit of work into it. Unfortunately, we feel that this is a little out of scope for what botocore should handling. I would suggest you pull this into it's own package and hook into one of the events the clients emit:

https://boto3.amazonaws.com/v1/documentation/api/latest/guide/events.html
https://botocore.amazonaws.com/v1/documentation/api/latest/topics/events.html

For what it's worth, we do agree that the current retry behavior isn't perfect and could handle being throttled much more gracefully. This is something we're actively researching to improve, though I cannot share much information on this now.

museadmin:
Thanks for the feedback @joguSD 

With regard to the scope of botocore, I'm not really asking it to handle anything beyond making the queue available as an option. It is entirely up to the consumer to instantiate, start and use it if they need it.

It would be a shame if you refuse it, because I was going to go on to submit PRs to all of the SDKs in each language to do the same. Then approach AWS to persuade them that leveraging client side limiting by exposing a limits API would put an end to this problem.

You comment that botocore, "could handle being throttled much more gracefully" suggests that you are unaware of the serious outages being caused by this issue. Presumably because those issues are being raised on other projects that are consuming yours, not your own.

May I suggest that you contact the ScoutSuite developers at NCCGroup to try and get the measure of this issue? Because it is certainly the case at present that we have conscientious AWS customers attempting to scan their production environments for vulnerabilities, and bringing down those environments as a result.

Something of an irony I am sure you will agree...

museadmin:
Closing for lack of interest from boto

j4v:
It's really a shame this has been closed. AWS API rate limiting is poorly implemented and very hard to handle. Having the official libraries help avoid hitting the limits would be tremendously helpful to the community and organizations that build upon them.

rams3sh:
@museadmin  @x4v13r64  I was looking for a similar solution for handling route53's throttling in the client side with time based rate limiting since it has only 5 calls allowed per second and I hit upon this issue.

By any chance , would you mind releasing a separate package that can patch the relevent modules / class of botocore or boto3 during runtime to achieve this. I know it's little hacky but it will definitely help others who are in same boat as you to achieve this without much modification to the underlying boto* package.  For reference, I had made something for caching over boto which uses unittest's patch method to achieve call based caching [here](https://github.com/rams3sh/botocache).

Issue #1822, opened 2019-09-10
Support custom SSLContext objects


chrisdlangton:
Basically there is this [boto3/issues/1976](https://github.com/boto/boto3/issues/1976) where I asked for there to be PFS (perfect forward secrecy) in boto3, the response is a feature request to add support for a custom `SSLContext` to the `requests.adapters.HTTPAdapter`.

It's been identified the best place to introduce this is in the botocore class `URLLib3Session` because end-users can provide custom botocore session to boto3, and boto3 itself calls `botocore.session.get_session()` when an end-user custom botocore session is not provided.

To see if i might contribute a PR i forked botocore and dug through the codebase to look for usage of `requests.adapters.HTTPAdapter` and i found a very very scary pattern... there are  forked vendored libraries hardcoded in botocore, frozen without patches (including security no patches!!!), particularly i found requests `adapters.py` file, but it has been hacked to pieces and doesn't even represent anything close to what requests actually looks like today...

Putting aside my ranting about some of the worst appsec i've seen in a while.. Just don't ever embed deps in your repos..

So, looking at botocore/httpsession.py class `URLLib3Session` i can tell that this is the enabler of `HTTPAdapter` usage, awesome! good start.

However there is no ability to provide a HTTPAdapter with a custom end-user provided SSLContext, because there is no existence of the ssl context representation in this HTTPAdapter at all... it's not a HTTPAdapter at all close to what urllib3 or requests uses despite the botocore comments...

Moving on. Next i looked for where botocore might call a `create_urllib3_context`, and again i find a method that is called this but is not even close the the _real_ urllib3 method, acording to comments, because _reasons_;

```python
def create_urllib3_context(ssl_version=None, cert_reqs=None,
                           options=None, ciphers=None):
    """ This function is a vendored version of the same function in urllib3

        We vendor this function to ensure that the SSL contexts we construct
        always use the std lib SSLContext instead of pyopenssl.
    """
```

But this isn't true either... Worse, it allows `context.set_ciphers(ciphers or DEFAULT_CIPHERS)` and `DEFAULT_CIPHERS` comes from `urllib3` (Okay i guess... might be better to use the std lib SSL, see below) and `ciphers` is an argument with default `None` as you can see above, but the method `create_urllib3_context` is called by;

```python
class URLLib3Session(object):
    def _get_ssl_context(self):
        return create_urllib3_context()
```

[...]

PR #1823, opened 2019-09-10
introduce PFS

chrisdlangton:
discussion https://github.com/boto/botocore/issues/1822

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -22,11 +22,12 @@ import socket
 
 import botocore.serialize
 import botocore.utils
-from botocore.utils import PERFECT_FORWARD_SECRECY_TUPLE
+from botocore.utils import PERFECT_FORWARD_SECRECY_CIPHERS
 from botocore.signers import RequestSigner
 from botocore.config import Config
 from botocore.endpoint import EndpointCreator
 from botocore.httpsession import URLLib3Session, create_urllib3_context
+from urllib3.util.ssl_ import ssl
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


chrisdlangton:
I've checked each file this PR touches in the codecov reports and all the LOC codecov says are not covered are not any LOC this PR changed...

chrisdlangton:
code review?
@jamesls 
@JordonPhillips 
@awstools 
@kyleknap

joguSD:
@chrisdlangton I think at this point trying to offer configuration for every knob in the SSLContext is too much and it's probably a better approach to just allow a custom SSLContext to be passed as a client configuration.

chrisdlangton:
@joguSD thank you for commenting.

Basically a custom SSLContext was impossible, i'll evaluate the new botocore state once the PR that removes the vendored `requests` is merged

chrisdlangton:
@joguSD it is curious you mention this, should I also refactor out the `use_ssl` and expect an end-user to know how to construct the primitives themselves?
no...
that is crazy fora framework/library.. it's nonsensical to suggest frameworks/libraries should not introduce a _feature flag_.

THis PR is introducing a _feature flag_ so an end-user can simply chose to _enforce_ the feature called _pfs_ (Perfect Forward Secrecy as it is known) which is actually just a select ciphersuite that (few) people actually know the correct options to use to do this (probably why not many people are doing this today, it's damn hard)

So i fundamentally disagree, we _should_ definitely be abstracticting away these extremely complex features behind a feature flag, botocore does this is almost every line of code.. the most analogous feature flag is the `use_ssl` flag...

chrisdlangton:
@joguSD it is curious you mention this, should I also refactor out the use_ssl and expect an end-user to know how to construct the primitives themselves?

no...

that is crazy..

For a framework/library, it is nonsensical to suggest frameworks/libraries should not introduce a feature flag.

This PR is introducing a _feature flag_ so an end-user can simply chose to enforce the feature called pfs (Perfect Forward Secrecy as it is known) which is actually just a select ciphersuite that (few) people actually know the correct options to use to do this (probably why not many people are doing this today, it's damn hard)

So i fundamentally disagree, we should definitely be abstracticting away these extremely complex features behind a feature flag, botocore does this is almost every line of code.. the most analogous feature flag is the use_ssl flag...

PR #1832, opened 2019-09-25
Changes for S3

byterock:
Playing with S3 which might require some changes here as well.


byterock:
I do not thing I pull this correctly will close

PR #1835, opened 2019-09-27
Allow stderr passthrough for credential_process

twooster:
Many organizations now require MFA for AWS login, or perhaps fetch
credentials from a command-line password manager. While `stdout` of
the `credential_process` is used to acquire the credential token,
`stderr` is captured for raising an exception back to the user.

In order enable user-interaction with the credential provider process,
it must have access to some stream the user can see: in this case,
`stderr`. This commit provides that change.

In order to maintain a relatively high level of backwards
compatibility, we will continue to capture `stderr` for raising
exceptions _unless_ `stderr` is reporting as being a TTY device, in
which case we assume that user-interaction is more favorable than
detailed exceptions.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -19,6 +19,7 @@ import getpass
 import threading
 import json
 import subprocess
+import sys
 from collections import namedtuple
 from copy import deepcopy
 from hashlib import sha1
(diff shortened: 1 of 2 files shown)


twooster:
Related PR, a while back: https://github.com/boto/botocore/pull/1349

ryan-gerstenkorn-sp:
Want to see if there is any plan's to merge this as well as mention this would be useful when using with aws-vault. Currently the unlock credentials prompt get's swallowed and it appears the process is stuck.

I spent some time trying to duplicate the stderr output so it can be raised in an error as well as passed through to the parent's stderr however something strange is going on and I was not able to get this to work properly. It seems like this is the next best option, was about to submit a PR for that before seeing this PR was already open.

tim-finnigan:
Thank you for creating this PR. This feature request is now being tracked here in our cross-SDK repository: https://github.com/aws/aws-sdk/issues/358. Implementations involving credential processing need to be considered across SDKs, so further review is required here. I will link this PR in that issue to reference this as a possible approach.

Issue #1710, opened 2019-04-08
ipv6 support in botocore without using dualstack


cpcunningham:
Is it possible to use IPV6  address directly in a url 
eg
 http://[999:1900:228f:20:172:19:9:35]/

Another user raised this issue and proposed a solution that was never merged.
https://github.com/boto/botocore/issues/1207

jamesls:
I'm not able to find a PR for this.  It looks like no one actually sent a PR.  Could you tell me more about how you're using ivp6 in URLs?  Is there a specific AWS endpoint you're using?  Not opposed to this, just would like to understand the use case a little better.

tpdanh:
Hi Jamesis, 
here is the use case:
client = boto3.client('s3', verify=False, endpoint_url='[2001:1900:228f:40:1:1:1:1]', aws_access_key_id=aws_access_key_id, aws_secret_access_key=aws_secret_access_key, config=Config(s3={'use_dualstack_endpoint': True}))

client.put_object(Bucket='vault-test', Key='Test-pd', Body='This is a test.')

tpdanh:
@jamesls  - do you have any feedback to the above query? thanks for any feedback..

tpdanh:
@jamesls - Hi Jamesis, we need this feedback on this rather quickly. Can you please help expedite this ? or get someone to help with this ?

jamesls:
Hi sorry for the delay, looks like the closing-soon-if-no-response label wasn't removed and this issue fell through the cracks.  Thanks for giving an example.  Out of curiosity, is there a reason you're specifying a hardcoded ipv6 address?  Just trying to understand this better so I can make a good case for prioritizing it.  If someone wants to send a PR for this issue, we should be able to review it quickly.

PR #1842, opened 2019-10-11
ipv6 endpoint support

joke-lee:


joke-lee:
@jamesls would you mind review it

https://github.com/boto/botocore/issues/1710

PR #1852, opened 2019-10-17
botocore documentation: Fix secretsmanager create-secret --secert-string example

scottrigby:
The current docs secretsmanager create-secret `--secret-string` JSON string example creates a secret value can't be converted to key name and value pairs:

Screen Shot 2019-10-17 at 6 18 33 PM

Whereas with this fix it can:

Screen Shot 2019-10-17 at 6 20 35 PM

I removed the link to the format a JSON parameter doc page because it's confusing for this command.


swetashre:
@scottrigby - Thank you for the PR. This file comes from upstream and is shared across different SDKs. So we can't make change to this file. I will contact the service team regarding the issue.

scottrigby:
@swetashre Thanks for the reply. Where can we make upstream pull requests for things like this?

PR #1853, opened 2019-10-20
Handle refreshing a shared credentials file

cknave:
This adds support for credential refreshing to `SharedCredentialProvider`.

If a new configuration key `aws_expiration` is present, `RefreshableCredentials` will be returned. Otherwise, standard `Credentials` will be returned. Credentials will be refreshed by re-parsing the file.

The key `aws_expiration` was chosen because it's already in use by https://github.com/zalando-incubator/kube-aws-iam-controller

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1156,6 +1156,7 @@ class SharedCredentialProvider(CredentialProvider):
 
     ACCESS_KEY = 'aws_access_key_id'
     SECRET_KEY = 'aws_secret_access_key'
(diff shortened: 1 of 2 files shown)


tim-finnigan:
Thank you for creating this PR and for your patience. I brought this up for discussion with the team, and the consensus was that this should be tracked as a feature request issue in our cross-SDK repository: https://github.com/aws/aws-sdk. The reason is because changes like this which involve credentials should be considered across SDKs in order to strive for consistency. Please consider creating an issue in that repository if this is still something you are interested in seeing.

PR #1857, opened 2019-10-24
Add sts regional endpoint configuration support

kyleknap:
To override the use of the global endpoint, set the `sts_regional_endpoints` config file variable or the `AWS_STS_REGIONAL_ENDPOINTS` environment variable to `regional`.

### DIFF
--- a/.changes/next-release/enhancement-sts-29751.json
+++ b/.changes/next-release/enhancement-sts-29751.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "``sts``",
+  "description": "Add support for configuring the use of regional STS endpoints."
+}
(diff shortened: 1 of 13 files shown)


ScottMaclure:
Question - I wanted to set this property in Config(), at run-time. I get the following error:

```
  File "aws_sts_regional_test.py", line 17, in <module>
    sts_regional_endpoints='regional'
  File "/path/to/botocore/config.py", line 172, in __init__
    args, kwargs)
  File "/path/to/botocore/config.py", line 199, in _record_user_provided_options
    'Got unexpected keyword argument \'%s\'' % key)
TypeError: Got unexpected keyword argument 'sts_regional_endpoints'
```

Looking at the code for config.py, you can see that before args.py code runs, validation is done on the Config inputs, and they must be present in OPTION_DEFAULTS or else it will raise an exception.

I assume this is a bug?

Background motivation - need to use the regional sts endpoint for non-default aws regions.

PR #1860, opened 2019-10-24
Add CloudFormation drift detection waiter

acbilimoria:
This pull request adds a CloudFormation drift detection waiter.  Polling delay and max attempts were borrowed from StackCreateComplete.

Resolves: https://github.com/boto/boto3/issues/1771

### DIFF
--- a/botocore/data/cloudformation/2010-05-15/waiters-2.json
+++ b/botocore/data/cloudformation/2010-05-15/waiters-2.json
@@ -116,6 +116,33 @@
         }
       ]
     },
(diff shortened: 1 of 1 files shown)


stobrien89:
Hi @acbilimoria,

Thanks for the request. We are now submitting requests for paginators and waiters upstream so that all AWS SDKs benefit from a consistent implementation. I will pass on the request for this waiter to the service team. We will also update the guidance in the contributing guide to better reflect this policy. I've added an issue to track this here:

aws/aws-sdk#81

PR #1861, opened 2019-10-24
Fix typo: DEFAUT -> DEFAULT

tipabu:


swetashre:
@tipabu - Thank you for the PR. The change looks good. Can you resolve the conflict so that we can merge the PR ?

tipabu:
> Can you resolve the conflict so that we can merge the PR ?

Thanks for letting me know about the conflict; done.

joguSD:
I think we're a little gun shy on making typo fixes like this in variable names. I know in the past it's caused issues for some people despite these variables not being public interfaces really. At the very least to feel comfortable with this change we would need an alias for the old spelling.

nateprewitt:
Going to resolve this as we haven't seen changes in the last several months and it no longer runs against CI. Please feel free to reopen if there are more changes to review. Thanks!

PR #1865, opened 2019-10-29
fix deprecation warnings for collections.abc (will break in Python 3.9)

terencehonles:


b4stien:
FYI the title seems misleading: it breaks with the newly released Python 3.8, not the next-year unreleased 3.9. We haven't tested yet, but the message seems straightforward about this:

```
/opt/api_venv/lib/python3.7/site-packages/botocore/awsrequest.py:624: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```

toolness:
Thanks, it would be great to see this PR merged!

@b4stien it looks like they actually pushed the removal to 3.9 when they released 3.8--see https://github.com/boto/botocore/issues/1615#issuecomment-548108554 or [this Twitter thread](https://twitter.com/hynek/status/1184845494759104512) for more details.

terencehonles:
@toolness I didn't realize it moved! I thought I just remembered it incorrectly.

I had checked the Python source and saw it in 3.8 when I was checking this and itypes, which I had on an ignore list for the DeprecationWarning and was preparing to start testing against 3.8.

Syntactical01:
Is this fixed yet? I feel like this should be a higher priority item. Lambda supports 3.8 yet we cannot safely use boto3 because of this deprecation.

tirkarthi:
Python 3.9 will raise an ImportError as per upstream PR : https://github.com/python/cpython/pull/10596

thompsjj:
Will anyone ever merge this?

terencehonles:
@tirkarthi thanks, I updated the test!

Now to get this merged :thinking:...

terencehonles:
This is no longer needed with #1922 which does the same thing.

PR #1867, opened 2019-10-30
Add endpoints for EKS, RDS Data, Textract, RDS PI

daisieh:
I don't know why EKS isn't listed now, but I added its endpoints as listed on https://docs.aws.amazon.com/general/latest/gr/rande.html#eks_region.


swetashre:
@daisieh - Thank you for the PR. We pull this file from upstream. So if you find any issue around this file then please open an issue in our repo and we will contact the service team to get this fixed.

PR #1874, opened 2019-11-06
Fix deprecation warning

labrys:
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -15,7 +15,7 @@ import sys
 import logging
 import functools
 import socket
-import collections
+import collections.abc
 
 import urllib3.util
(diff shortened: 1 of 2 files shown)


BoboTiG:
You should use that kind of code to fix the CI and Python 2 errors:
```python
try:
   from collections.abc import MutableMapping
except ImportError:
   from collections import MutableMapping
```
Ad then use `MutableMapping` in class declaration.

BoboTiG:
My bad, it did no see that this is a duplicate of #1865.

labrys:
Closing as duplicate

PR #1875, opened 2019-11-06
Update minor typo in Docs

Jarryd-rk:
Cals to InvokeEndpoint -> Calls to InvokeEndpoint


nateprewitt:
Thanks for the contribution, @Jarryd-rk! This request was passed on to the service team who owns these docs and they've been able to address it in more recent versions of the SDK. Going to close this now that the issue is resolved.

PR #1895, opened 2019-11-20
Imds metadata

stealthycoin:


botocore/utils.py
```diff
@@ -274,29 +281,56 @@ def __init__(self, timeout=DEFAULT_METADATA_SERVICE_TIMEOUT,
...
+        for i in range(self._num_attempts):
+            try:
+                response = self._session.send(request.prepare())
+                if response.status_code == 200:
+                    return response.text
+                elif response.status_code in (404, 403):
+                    return None
+                elif response.status_code in (400,):
+                    raise BadIMDSRequestError(request)
+            except ReadTimeoutError:
```
joguSD: Are we sure that this will always get raised as a read timeout? When I was working on the `urllib3` changes I remember some scenarios where it was ambiguous exactly what timeout error would get raised.

stealthycoin: In the repro case it appears to only be that error.

itajaja:
this change is really causing some faulty behaviors in kubernetes: https://github.com/boto/botocore/issues/2304 https://github.com/boto/botocore/issues/2421. any chance we can revisit?

PR #1900, opened 2019-11-22
WIP: Remove support for specifying api_version for models

garnaat:
Initial commit, removed support for multiple API versions.  Unit and functional tests pass.


tests/unit/test_waiters.py
```diff
@@ -671,11 +671,11 @@ def setUp(self):
         self.service = 'cloudfront'
         self.old_api_versions = ['2014-05-31']
```
kyleknap: We can probably remove that line as well.

tests/unit/test_waiters.py
```diff
@@ -697,46 +697,15 @@ def assert_invalidation_completed_call_count(self, api_version=None):
...
+            self.service)
         self.client.get_streaming_distribution.side_effect = [
             {'StreamingDistribution': {'Status': 'Deployed'}}
         ]
         waiter = create_waiter_with_client(waiter_name, waiter_model,
                                            self.client)
         waiter.wait()
         self.assertEqual(self.client.get_streaming_distribution.call_count, 1)
-
-    def test_distribution_deployed(self):
```
kyleknap: For these tests, we probably only want to remove the for loop around the different API versions. The first assertion still seems valuable to make sure that those cloudfront waiters are not broken.

botocore/loaders.py
```diff
@@ -340,7 +340,7 @@ def list_api_versions(self, service_name, type_name):
         return sorted(known_api_versions)
 
     @instance_cache
-    def load_service_model(self, service_name, type_name, api_version=None):
+    def load_service_model(self, service_name, type_name):
```
kyleknap: I'm wondering for now if it make sense to hold off on updating the `loaders` module for removing `api_version` right now? Given that we have not really addressed the structure of the data directory yet, it seems reasonable to keep the data loader interface in line with how the models are organized on disk.

botocore/session.py
```diff
@@ -726,12 +721,6 @@ def create_client(self, service_name, region_name=None, api_version=None,
         :param region_name: The name of the region associated with the client.
             A client is associated with a single region.
 
-        :type api_version: string
```
kyleknap: Don't think we need to do it in this PR, but it would be nice to get a follow up PR that removes the models of the old api versions in order to reduce botocore's footprint.

garnaat: Agreed.  I can do a separate PR after this one to remove old model data.

garnaat:
See https://github.com/aws/aws-cli/pull/4751

PR #1912, opened 2019-12-12
Removing all traces of old style s3 auth.

garnaat:
This PR removes sigv2 auth for S3.


tests/unit/test_s3_addressing.py
```diff
@@ -43,27 +40,24 @@ def get_prepared_request(self, operation, params, force_hmacv1=False):
             return http_stubber.requests[0]
 
     def enable_hmacv1(self, **kwargs):
-        return 's3'
```
stealthycoin: We can remove this function since it isn't used anymore.

kyleknap:
@garnaat could you possibly rebase this off of the HEAD of v2? I just merged a PR that synced v2 with the develop branch: https://github.com/boto/botocore/pull/1925 and there were some sigv2 test updates included in this resync. I figured it would be easier to resolve conflicts in these smaller PRs then the resync of v2 with develop.

botocore/client.py
```diff
@@ -276,45 +276,28 @@ def _set_s3_presign_signature_version(self, client_meta,
...
-        # the customer hasn't set a signature version so we default the
-        # signature version to sigv2.
+        # At this point, we want to default to sigV4 for all regions.
+        # SigV2 is no longer an option.
         client_meta.events.register(
-            'choose-signer.s3', self._default_s3_presign_to_sigv2)
+            'choose-signer.s3', self._default_s3_presign_to_sigv4)
 
-    def _default_s3_presign_to_sigv2(self, signature_version, **kwargs):
+    def _default_s3_presign_to_sigv4(self, signature_version, **kwargs):
```
kyleknap: I'm not sure if we still need this handler? It may be interesting to see if we can just remove the `_set_s3_presign_signature_version` method and see if botocore just defaults to sigv4. We really only needed this logic because the presigners would default to sigv2 while the API requests would default to sigv4.

tests/functional/test_s3.py
```diff
@@ -467,16 +459,16 @@ def test_correct_url_used_for_s3():
 
     # The default behavior for sigv2. DNS compatible buckets
```
kyleknap: We should probably update the comment to be about just DNS compatible bucket names and not mention sigv2. Also for the test cases, it may make sense to just remove the `signature_version` from the cases because we already have cases where we explicitly use `s3v4` in the next set of cases below this set of cases.

garnaat: Okay, I will rebase and incorporate feedback.  Thanks.

PR #1913, opened 2019-12-13
Remove STS legacy configration

stealthycoin:


botocore/args.py
```diff
@@ -224,42 +197,6 @@ def compute_s3_config(self, scoped_config, client_config):
...
-        )
-
-    def _get_sts_regional_endpoints_config(self):
-        sts_regional_endpoints_config = self._config_store.get_config_variable(
-            'sts_regional_endpoints')
-        if not sts_regional_endpoints_config:
-            sts_regional_endpoints_config = 'legacy'
-        if sts_regional_endpoints_config not in \
-                VALID_STS_REGIONAL_ENDPOINTS_CONFIG:
-            raise botocore.exceptions.InvalidSTSRegionalEndpointsConfigError(
```
kyleknap: We should remove `InvalidSTSRegionalEndpointsConfigError` from `exceptions.py` as well.

tests/functional/test_sts.py
```diff
@@ -99,16 +99,7 @@ def _get_signing_region(self, request):
...
-        sts = self.create_sts_client('us-west-2')
-        self.assert_request_sent(
-            sts,
-            expected_url='https://sts.amazonaws.com/',
-            expected_signing_region='us-east-1'
-        )
-
-    def test_legacy_region_with_regional_configured(self):
+    def test_sts_standard_endpoint(self):
         self.environ['AWS_STS_REGIONAL_ENDPOINTS'] = 'regional'
```
kyleknap: We probably want to remove setting this environment variable.

tests/functional/test_sts.py
```diff
@@ -44,12 +44,12 @@ def test_presigned_url_contains_no_content_type(self):
```
kyleknap: Looking over `tests/functional/test_credentials` from this PR: https://github.com/boto/botocore/pull/1857/files, it looks like we need to update the test to remove the `sts_regional_endpoints` config variable,

PR #1917, opened 2019-12-16
Update Description for GovCloud West to US-West

aivarson:
Looking to update the endpoints.json description for GovCloud West.


stealthycoin:
This file is consumed from an upstream source, and we do not own its contents so we cannot make this change.

PR #1918, opened 2019-12-16
Add proper description for GovCloud

sandeepvasani:
Added the description to distinguish between east and west regions,
https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html


sandeepvasani:
Looks like Alan already made this change.

PR #1921, opened 2019-12-16
[1915] Use delay function in waiter

### LINKED ISSUE #1915: Custom callable for waiter
https://github.com/boto/botocore/blob/develop/botocore/waiter.py#L331 has a hard-coded callable to pause the polling on the status checks.  It might help to add an option in the class init for a custom callable, e.g.

```python
def backoff_pause(tries):
    pause = backoff_delay(tries)
    time.sleep(pause)

def backoff_delay(tries):
    """An exponential back-off delay, with random jitter, and
    a maximum interval of 10 minutes.

    .. seealso::

            - https://docs.aws.amazon.com/general/latest/gr/api-retries.html
            - https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
    """
    max_interval = 600.0  # 10 minutes

    delay = int(1 + pow(tries * 0.3, 2))
    delay = min(max_interval, delay)
    return randint(int(delay / 2), delay)
```

The class signature to use a custom callable becomes something like:
```python
class Waiter(object):
    def __init__(self, name, config, operation_method, delay_method = None):
        """
        :type name: string
        :param name: The name of the waiter
        :type config: botocore.waiter.SingleWaiterConfig
        :param config: The configuration for the waiter.
        :type delay_method: callable
        :param delay_method: A callable that accepts `tries: int`
            and pauses execution (using `time.sleep(tries)` for example).
        :type operation_method: callable
        :param operation_method: A callable that accepts **kwargs
            and returns a response.  For example, this can be
            a method from a botocore client.
        """
        self._operation_method = operation_method
        self._delay_method = delay_method or time.sleep
        # The two attributes are exposed to allow for introspection
        # and documentation.
        self.name = name
        self.config = config
```

The it can be plugged in using something like:
```python
Waiter("backoff", config, batch_client.describe_jobs, delay_method=backoff_pause)
```

Suggestions for anything that could pause the waiter execution with async behavior would be great.  It would be interesting to have async waiters, if they don't exist already.

dazza-codes:
Fix #1915

- use a randomized exponential backoff delay
  for waiter retries by default
- allow a custom callable for waiter retries
- the motivation for this is to avoid API throttle limits for highly concurrent systems

### DIFF
--- a/botocore/waiter.py
+++ b/botocore/waiter.py
@@ -10,6 +10,8 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import random
+
 import jmespath
 import logging
 import time
@@ -263,7 +265,7 @@ class AcceptorConfig(object):
 
 
 class Waiter(object):
-    def __init__(self, name, config, operation_method):
+    def __init__(self, name, config, operation_method, delay_method=None):
         """
 
         :type name: string
@@ -277,8 +279,15 @@ class Waiter(object):
             and returns a response.  For example, this can be
             a method from a botocore client.
 
+        :type delay_method: Optional[callable]
+        :param delay_method: A callable with a function signature matching
+            `delay_method(attempt: int, delay: int, cap: int = 600)`; the
+            callable must call `time.sleep()` to pause execution between
+            status checks; the default uses a configured delay with
+            `Waiter.retry_delay(attempt, config.delay, cap=600)`.
         """
         self._operation_method = operation_method
+        self._delay_method = delay_method or self.retry_delay
         # The two attributes are exposed to allow for introspection
         # and documentation.
         self.name = name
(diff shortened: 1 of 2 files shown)


dazza-codes:
Does anyone ever look at community contributions?  I have no idea whether this will get reviewed or not and probably some bot will come along and mark it stale and close it.

dazza-codes:
Why did I ever bother to submit a PR on this project, just to have a github-bot close it?

kdaily:
Hi @dazza-codes,

I apologize for the delay in reviewing and responding to this PR. Unfortunately we will not be making these changes to the waiter in this package. Changes to how the waiter works would need to be implemented across all AWS SDKs that support them, as they all share a common implementation. This will require coordination with the rest of the AWS SDK teams as well as service teams.

The current waiter is not designed with these cases in mind. There's no pool for concurrency or mitigation strategy for these kinds of issues, and implementing this appropriately would require more abstraction.

It is an interesting feature request, but not one that has garnered much other attention. I would suggest that we leave the feature request #1915 open, but reframe it around the limitation of the waiter in the case of throttling and concurrency instead of this specific implementation.

PR #1922, opened 2019-12-17
Add support for python 3.8

stealthycoin:


botocore/awsrequest.py
```diff
@@ -29,6 +28,11 @@
      urlencode
 from botocore.exceptions import UnseekableStreamError
 
+try:
```
jamesls: This is what I was referring to when I mentioned moving the try/except into a compat module.  It's repeated 3 times and is a perfect fit for `compat.py`.

jamesls:
@joguSD Updated compat module, should be ready to look at again.  Just waiting for travis to pass.

PR #1927, opened 2019-12-30
Treat `AWS_PROFILE` with empty string same as if not defined

jstastny:
This is another take on https://github.com/boto/botocore/issues/1460

### Issue

I noticed the issues described in https://github.com/boto/botocore/issues/1460 when running tests on a fresh instance (continuous integration), which had no profiles and configuration in `~/.aws` directory and used `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. At the same time, I had the environment variable `AWS_PROFILE` set to empty string, hoping it would have the same meaning as if it was not defined. That caused the build failing with:
```
botocore.exceptions.ProfileNotFound: The config profile () could not be found
```

We use the `AWS_PROFILE` variable in the build scripts so that we can reuse the same build scripts for our local development, where we use profiles. The hope was that we could use the same scripts for continuous integration without any changes to the scripts (and without having to do `aws configure` to create profile on continuous integration or having to do some similar workarounds).

### Changes
Add conversion function to the `profile` entry in `configprovider.BOTOCORE_DEFAUT_SESSION_VARIABLES` that converts falsey values to `None`. This means empty string is converted to `None`.

### DIFF
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -270,7 +270,7 @@ class ConfigValueStore(object):
 
     def get_config_variable(self, logical_name):
         """
-        Retrieve the value associeated with the specified logical_name
+        Retrieve the value associated with the specified logical_name
         from the corresponding provider. If no value is found None will
         be returned.
 
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -48,7 +48,7 @@ logger = logging.getLogger(__name__)
 #: found.
 BOTOCORE_DEFAUT_SESSION_VARIABLES = {
     # logical:  config_file, env_var,        default_value, conversion_func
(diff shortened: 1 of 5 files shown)


tim-finnigan:
Thank you for creating this PR and for your patience. I brought this up for discussion with the team, and the consensus was that the related issue (https://github.com/boto/botocore/issues/1460) should be tracked in in the cross-SDK repository (https://github.com/aws/aws-sdk) if it is still something that is needed. The reason for this is that environment variables are shared across SDKs. (For more information on this please see: https://docs.aws.amazon.com/sdkref/latest/guide/settings-reference.html#EVarSettings.) Based on that feedback I'm going to close this PR but please let us know if you had any follow up questions.

benkehoe:
Are we sure this isn't fixed? The corresponding issue on the CLI is closed as fixed: https://github.com/aws/aws-cli/issues/5248

PR #1935, opened 2020-01-09
Update six.py

alanyee:
Using the constant over reusing a function


alanyee:
Where is the upstream `master`? I am making a PR to the `develop` since I suspected that was some commits ahead of `master` unless I am wrong about that.

hugovk:
This is upstream `master`:

* https://github.com/benjaminp/six/blob/master/six.py

"Vendoring" means that file is copied into this codebase every once in a while.

alanyee:
Oh okay, I did not realize that botocore was getting changes from another repo. I can close this PR if you want.

PR #1941, opened 2020-01-14
Fix crash when running under Python 3 due to incorrect string types

Tim020:


Tim020:
@moseb good shout, that's made things better! :)

nateprewitt:
Hey @Tim020, thanks for the PR! I didn't see info on the exact issue you're experiencing, and wanted to clarify that. Botocore should only be handling bodies in this context that are strictly bytes in the parser. Would you be able to provide a brief description of how you encountered this and a stack trace if one is still available?

nateprewitt:
Resolving since we haven't had any further responses. Please feel free to open an issue detailing the cases where you're encountering this problem and we can talk through potential options. Thanks!

PR #1948, opened 2020-01-22
Strictly use service id as the identifier in events

joguSD:
This removes event aliasing and cleans up any places that were relying on it to now use service id.


tests/functional/test_route53.py
```diff
@@ -28,6 +28,23 @@ def setUp(self):
             'MaxItems': '1'
         }
         self.operation_name = 'list_hosted_zones'
+        self.final_params = {}
+
+    def _store_final_params(self, params, **kwargs):
+        self.final_params = params
+
+    def test_route53_id_conversion(self):
+        # This conversion is handled in the before-parameter-build event, but
```
kyleknap: Thoughts on using the HTTP stubber instead?

botocore/handlers.py
```diff
@@ -975,11 +976,11 @@ def inject_api_version_header_if_needed(model, params, **kwargs):
 
     ('before-call.s3', add_expect_header),
     ('before-call.glacier', add_glacier_version),
-    ('before-call.apigateway', add_accept_header),
+    ('before-call.api-gateway', add_accept_header),
```
kyleknap: I'm assuming we don't have to apply this to api gateway v2 as well? I know for a while those two shared the same endpoint prefix so I don't know if we had an alias that would trigger the handler for api gateway v2 as well.

tests/functional/test_route53.py
joguSD: I thought about this but since this ends up as a query arg I'd have to parse the request URL and pull out the right arg to assert on. It felt more wrong than just hooking into an event that would have the correctly converted parameter.

botocore/handlers.py
joguSD: I checked but it doesn't seem like this event is triggering for API Gateway V2 on `v2` or the `develop` branch. Not sure if it ever has or should be hooked up to APIG V2.

tests/functional/test_route53.py
joguSD: I thought about this but since this ends up as a query arg I'd have to parse the request URL and pull out the right arg to assert on. It felt more wrong than just hooking into an event that would have the correctly converted parameter.

kyleknap: Fair enough.

PR #1956, opened 2020-02-03
Decorate EventAliaser._alias_event_name() with functools.lru_cache

fruch:
Since we pass via `_alias_event_name()` for each single event,
before this change it was iteratrating via all the known aliases of
all the AWS services, while only a small subset is need.

In my specific benchmark using only dynamodb api
(not directly with AWS, but with a local ScyllaDB),

I've managed to go from ~500-600rps to ~1000-1100rps

### DIFF
--- a/botocore/hooks.py
+++ b/botocore/hooks.py
@@ -12,9 +12,15 @@
 # language governing permissions and limitations under the License.
 import copy
 import logging
+
 from collections import defaultdict, deque, namedtuple
 from botocore.compat import accepts_kwargs, six
 from botocore.utils import EVENT_ALIASES
+
+try:
+    from functools import lru_cache
+except ImportError:
+    from backports.functools_lru_cache import lru_cache
 
 logger = logging.getLogger(__name__)
 
@@ -386,10 +392,9 @@ class EventAliaser(BaseEventHooks):
         return self._emitter.unregister(
             aliased_event_name, handler, unique_id, unique_id_uses_count
         )
-
+    @lru_cache(maxsize=1024)
     def _alias_event_name(self, event_name):
         for old_part, new_part in self._event_aliases.items():
-
             # We can't simply do a string replace for everything, otherwise we
             # might end up translating substrings that we never intended to
             # translate. When there aren't any dots in the old event name
@@ -413,9 +418,7 @@ class EventAliaser(BaseEventHooks):
                 continue
 
             new_name = '.'.join(event_parts)
-            logger.debug("Changing event name from %s to %s" % (
-                event_name, new_name
-            ))
+            logger.debug("Changing event name from %s to %s", event_name, new_name)
             return new_name
         return event_name
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ wheel==0.24.0
 docutils>=0.10,<0.16
 behave==1.2.5
 jsonschema==2.5.1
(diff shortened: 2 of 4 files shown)


fruch:
this is the code I've used to benchmark
https://github.com/fruch/hydra-locust/blob/master/benchmark.py

I'm still looking for more place, since with YCSB (i.e. java client) i'm getting ~12K rps.

Any other tricks to get more RPS out of the boto3/botocore ?

mschfh:
this patch improves boto3s `s3_client.generate_presigned_url` performance by about 2.5x (see also https://github.com/boto/boto3/issues/217)

before:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    40000    3.050    0.000    5.695    0.000 hooks.py:392(_alias_event_name)
  2360000    1.640    0.000    1.640    0.000 {method 'index' of 'list' objects}
  2940004    1.010    0.000    1.010    0.000 {method 'split' of 'str' objects}
    40000    0.201    0.000    0.769    0.000 hooks.py:178(_emit)
    80000    0.159    0.000    0.352    0.000 parse.py:853(quote_from_bytes)
```

after:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    40000    0.193    0.000    0.749    0.000 hooks.py:178(_emit)
    80000    0.164    0.000    0.356    0.000 parse.py:853(quote_from_bytes)
   710000    0.144    0.000    0.159    0.000 {built-in method builtins.isinstance}
    80000    0.133    0.000    0.640    0.000 utils.py:658(percent_encode)
    70000    0.117    0.000    0.138    0.000 parse.py:109(_coerce_args)
```

qk4l:
It looks like that LRU simply resolve problem with iteration over EVENT_ALIASES.
I got 25% boost with s3.download_file().

Is any change to see this improvement in upstream?

mschfh:
> Any other tricks to get more RPS out of the boto3/botocore ?
https://maori.geek.nz/make-pythons-dynamodb-client-faster-with-this-one-simple-trick-2eb2888269ce

nateprewitt:
Hey everyone,

We sat down this week to take a deeper look at this. Some of the initial concern with this change is the way `lru_cache` interacts with class methods when compared to stand alone functions. The cache keeps a reference to the method attached to a given instantiation of each client, which will keep the entire client in scope after it's been released elsewhere. This leads to memory leaks, which while capped at 1024 clients, makes this change problematic for memory constrained users.

functools introduced [`cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property) in Python 3.8 to help with this use case which would probably be the preferred approach in this scenario. Unfortunately, while we still support Python 3.6 & 3.7, we would need to bring on an additional dependency to `backports` which we'd prefer to avoid if possible.

I've done some basic benchmarking with a simplified change using a dictionary "cache" on the `EventAliaser` itself that shows equal or better performance to the lru_cache without the memory leak. I've posted an alternative proposal in #2453 with a rough outline of possible changes. We would encourage anyone still interested in seeing this issue improved to please test out the PR with your current use case. That will help us prioritize changes to this behavior going forward.

Thanks!

nateprewitt:
We've gone ahead and merged #2453 as the current solution to the caching issue. We're going to resolve this now as we believe the core issue is addressed. The fix should be available in the next Boto3/Botocore release.

fruch:
notice that this was solved, (only took me a year to notice it..)

thanks @nateprewitt

PR #1957, opened 2020-02-04
Only test and support py37+

jamesls:
Py37 because that's what we use to build cli v2,
Py38 so we can stay on top of keeping compatible with
new versions of python so it's easy for us to upgrade
python.


jamesls:
Based on team discussion, let's hold off on this for now.

PR #1960, opened 2020-02-05
Do not use sslCommonName from endpoint data

jamesls:
We originally introduced this to work around a py26 bug
where it couldn't read subjectAltNames for certs over a certain
size.  We now no longer need this due to py26 being dropped,
and v2 is a natural place to change these endpoints over.

As part of this change I had to update the functional test
for regions.  The main test generator was being shadowed by another
test name so as a result these tests were never actually running.

I had to update these tests to handle other endpoint related changes
we made (regional STS, S3 us-east-1, `s3.region` instead of
`s3-region`).

### DIFF
--- a/tests/functional/test_regions.py
+++ b/tests/functional/test_regions.py
@@ -17,6 +17,7 @@ from nose.tools import assert_equal, assert_raises
 
 from botocore.client import ClientEndpointBridge
 from botocore.exceptions import NoRegionError
+from botocore.compat import urlparse
 
 
 # NOTE: sqs endpoint updated to be the CN in the SSL cert because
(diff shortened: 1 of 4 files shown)


tests/functional/test_regions.py
```diff
@@ -465,8 +465,9 @@ def _test_single_service_region(service_name, region_name,
                                 expected_endpoint, resolver):
     bridge = ClientEndpointBridge(resolver, None, None)
     result = bridge.resolve(service_name, region_name)
-    expected = 'https://%s' % expected_endpoint
-    assert_equal(result['endpoint_url'], expected)
+    if not expected_endpoint.startswith('http'):
```
nateprewitt: Very minor, but we could use `urlparse` to ensure http is the scheme and not the start of the domain.

jamesls: Updated.

bpodgursky:
@jamesls Did this ever make it into a botocore 1.x release so that it could be used by boto3?  We're running into the issue with inconsistent sslCommonName data for services referenced by boto3. 

I'm a bit confused about the versioning but it doesn't seem like botocore v2 is possible to use from boto3 (https://github.com/boto/boto3/issues/2571) so I'm wondering if these changes ever get pulled back into the 1.x releases or if there's no plan for boto3 to ever pick this up.

PR #1962, opened 2020-02-05
Add branching requirement for urllib3

stealthycoin:
urllib3 dropped support for python 3.4 in version 1.25.8:
https://github.com/urllib3/urllib3/blob/master/CHANGES.rst#1258-2020-01-20

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,10 +23,19 @@ def find_version(*file_paths):
     raise RuntimeError("Unable to find version string.")
 
 
-requires = ['jmespath>=0.7.1,<1.0.0',
-            'docutils>=0.10,<0.16',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -23,10 +23,19 @@ def find_version(*file_paths):
...
-            'python-dateutil>=2.1,<3.0.0',
-            'urllib3>=1.20,<1.26']
+requires = [
+    'jmespath>=0.7.1,<1.0.0',
+    'docutils>=0.10,<0.16',
+    'python-dateutil>=2.1,<3.0.0',
+]
+
+
+if sys.get_version[:2] == (3, 4):
```
nateprewitt: I think this needs to be `sys.version_info`.

PR #1971, opened 2020-02-13
JSONFileLoader: dedent unnecessary code inside context manager

robgolding:
Apologies in advance for the fact that I can't explain exactly _why_ this change should be merged. It comes about because we were seeing intermittent `OSError: [Errno 9] Bad file descriptor` errors being raised by `botocore` in our CI suite under Python 3, which runs tests in parallel using `multiprocessing`:

```
   File "<censored>", line 71, in mock_s3
     s3 = botocore.session.get_session().create_client('s3')
   File "/usr/local/lib/python3.6/site-packages/botocore/session.py", line 828, in create_client
     endpoint_resolver = self._get_internal_component('endpoint_resolver')
   File "/usr/local/lib/python3.6/site-packages/botocore/session.py", line 695, in _get_internal_component
     return self._internal_components.get_component(name)
   File "/usr/local/lib/python3.6/site-packages/botocore/session.py", line 903, in get_component
     self._components[name] = factory()
   File "/usr/local/lib/python3.6/site-packages/botocore/session.py", line 155, in create_default_resolver
     endpoints = loader.load_data('endpoints')
   File "/usr/local/lib/python3.6/site-packages/botocore/loaders.py", line 132, in _wrapper
     data = func(self, *args, **kwargs)
   File "/usr/local/lib/python3.6/site-packages/botocore/loaders.py", line 420, in load_data
   File "/usr/local/lib/python3.6/site-packages/botocore/loaders.py", line 175, in load_file
     return json.loads(payload, object_pairs_hook=OrderedDict)
 OSError: [Errno 9] Bad file descriptor
```

I tried for days to pin this down, and eventually just tried the change I've made in this PR, so that the file descriptor is not left open for as long. Amazingly, that fixed the issue and we haven't seen another error of this type since.

The code is equivalent, so this has no effect on the correctness of the `load_file` method. I'm hoping this is an acceptable change so that we don't need to monkeypatch `botocore` anymore! And if anyone has an idea as to what could possibly be causing this, I would be extremely grateful 😄.

### DIFF
--- a/botocore/loaders.py
+++ b/botocore/loaders.py
@@ -171,8 +171,9 @@ class JSONFileLoader(object):
         # We specify "utf8" here to ensure the correct behavior.
         with open(full_path, 'rb') as fp:
             payload = fp.read().decode('utf-8')
-            logger.debug("Loading JSON file: %s", full_path)
-            return json.loads(payload, object_pairs_hook=OrderedDict)
+
+        logger.debug("Loading JSON file: %s", full_path)
+        return json.loads(payload, object_pairs_hook=OrderedDict)
 
 
 def create_loader(search_path_string=None):


nateprewitt (approved):
Hey @robgolding, thanks for the patch! Looking over the issue you're describing, I would say it's almost certainly a ulimits issue in your CI environment. The operating system is likely reaping file descriptors, causing this to fail when the context manager is kept open. While this isn't a complete protection against the issue, it seems like a reasonable improvement.

I've rebased your change onto the top of `develop` to rerun the tests, and the I think this should be ready to merge.

PR #1972, opened 2020-02-13
Add support for new retry modes

jamesls:
Add support for parsing retryable metadata in model

Add support for retries v2 standard mode

This adds the concept of "retry modes" which allows us to
introduce different retry behavior.

The default mode, `legacy`, is the existing retry behavior.
This means that new retry modes (in v1) require an opt-in.

The new `standard` mode moves away from the `_retry.json` and
contains a set of errors that are retried that is now consistent
with other AWS SDKs implementing standard mode.  This mode also
adds support for retry quotas which control how many unsuccessful
retries a client can make.

As part of this change, I've added a new total_max_attempts config
option.

This addresses the fact that there's already a max
attempts client config var in botocore v1, and that it
doesn't mean the same thing as the `AWS_MAX_ATTEMPTS`
env var.  To summarize:

* The client config's `max_attempts` remains unchanged.
* A new `total_max_attempts` var is added that maps to the
  `AWS_MAX_ATTEMPTS` and `max_attempts` env var and config file
  values.
* We internally convert client config `max_attempts` to
  `total_max_attempts` and ensure internally (in args.py and
  client.py) we only use this new value.

Add support for adaptive mode retries

This adds a new experimental retry mode, `adaptive`, that adds
support for client side rate limiting based on receiving
throttling responses.  `adaptive` mode includes all the behavior
on `standard` mode.


kyleknap:
We should also add a changelog entry for the new modes before merging.

PR #1973, opened 2020-02-13
[v2] Add support for adaptive mode retries

jamesls:
This adds a new experimental retry mode, `adaptive`, that adds
support for client side rate limiting based on receiving
throttling responses.  `adaptive` mode includes all the behavior
on `standard` mode.

### DIFF
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -40,6 +40,7 @@ from botocore.discovery import (
     block_endpoint_discovery_required_operations
 )
 from botocore.retries import standard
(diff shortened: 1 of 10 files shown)


OperationalFallacy:
Hi,
Thank you for this feature, it looks nice and advance.

I'm looking to do some testing for this new adaptive algorithm to see how better it is compare to various retry logic added through external packages (retries, retry or even custom code).

Is there a design document that outlines why it was implemented?

Roman

PR #1974, opened 2020-02-17
update list-stack-instances documentation

ejpir:
The StackInstance `documentation` for the response of the `Account` field is wrong. It does not return the name, it returns the ID.


ejpir:
Thanks Maurits!

ejpir:
change branch to develop

PR #1977, opened 2020-02-18
support error code attribute on exception classes

### LINKED ISSUE #1976: add error code to exception classes
currently given an exception instance/class there isn't a straightforward way to map it back to its exception code. the errorfactory should just annotate this on to the class.

```python
import boto3
client = boto3.clients('sqs')
err = client.exceptions.QueueDoesNotExist
# looking for error code hypothetically, this doesn't work, but currently looking for
err.code == 'AWS.SimpleQueueService.NonExistantQueue'
# the only way to map it back at the moment is the following requires poking at private implementation details.
err_map = {v: k for k, v in client.exceptions._code_to_exception.items()}
err_map[err] ==  'AWS.SimpleQueueService.NonExistantQueue'
```

kapilt:
closes #1976

### DIFF
--- a/botocore/errorfactory.py
+++ b/botocore/errorfactory.py
@@ -80,8 +80,8 @@ class ClientExceptionsFactory(object):
             shape = service_model.shape_for(shape_name)
             if shape.metadata.get('exception', False):
                 exception_name = str(shape.name)
-                exception_cls = type(exception_name, (ClientError,), {})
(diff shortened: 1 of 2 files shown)


kapilt:
the coverage drop here seems unrelated entirely to this pr which has 100% patch coverage, is the botocore coverage testing unstable (threaded coverage Heisenbergs?)

PR #1978, opened 2020-02-18
For SigV4 canonical query str, encode then sort.

### LINKED ISSUE #1951: SigV4Auth class does not construct canonical query string according to V4 spec
According to the docs at https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html, you're supposed to encode query params before sorting them:

> CanonicalQueryString specifies the URI-encoded query string parameters. You URI-encode name and values individually. You must also sort the parameters in the canonical query string alphabetically by key name. **The sorting occurs after encoding.**

However, the code here iterates over the sorted params and performs encoding **afterward**, inside of the loop for each pair:

https://github.com/boto/botocore/blob/526f1e0322bf87e74f69c894245ce8506b86e597/botocore/auth.py#L213-L220

As an FYI, I haven't checked other libraries, but the Java SDK one seems to not have this problem because it inserts encoded entries into a (inherently-sorted) `TreeMap` as it iterates over the originals.

houglum:
Fixes #1951, which describes that the library is currently sorting
before encoding. The S3 docs linked below explicitly say to encode
params before sorting them when creating the canonical query string:
https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -212,10 +212,15 @@ class SigV4Auth(BaseSigner):
 
     def _canonical_query_string_params(self, params):
         l = []
-        for param in sorted(params):
+        # Encode params before sorting.
+        encoded_params = {}
+        for param in params:
             value = str(params[param])
-            l.append('%s=%s' % (quote(param, safe='-_.~'),
-                                quote(value, safe='-_.~')))
+            encoded_params[quote(param, safe='-_.~')] = quote(value,
+                                                              safe='-_.~')
+        for param in sorted(encoded_params):
+            value = encoded_params[param]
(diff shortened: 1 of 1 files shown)


houglum:
*friendly ping*

The order of operations is still reversed in the code.  I've even offered up a PR to fix it.  Please take a look :)

kdaily:
This was fixed with the commit https://github.com/boto/botocore/commit/826e7dab3a01d04d4ec37dfeef4d0bd3556fb340, closing!

houglum:
Awesome! Thanks, @kdaily !

PR #1983, opened 2020-02-23
[Fix #1981] No "host" field in HTTP header in https session

### LINKED ISSUE #1981: No "host" field in HTTP header in https proxy session
With the assumption, the https connection with proxy has a session to use "CONNECT" method.
"CONNECT" method is available in http since version 1.1.
>(https://tools.ietf.org/html/rfc7231#section-4)

However, in botocore, the https connection with "CONNECT" use HTTP/1.0
This problem is caused by Python's standard library
(the pull and request are already sent to github: https://github.com/python/cpython/pull/8305)
In addion to that, there is no "Host" field in https session in Botocore in spite of the description in RFC7230 that "Host" field is required in HTTP/1.1.
>A client MUST send a Host header field in all HTTP/1.1 request messages.

It causes the failure of proxy routing.

siradori:
Fix for #1981


nateprewitt:
Hi @siradori,

I think this behavior ended up getting added in #2471. It's not explicitly supported at this time but the feature is available for testing while we collect more feedback. We'll resolve this in favor of the current approach.

Issue #1979, opened 2020-02-21
Can not PUT a 0 bytes object when Object Lock is enabled on s3 bucket


wimglenn:
When uploading an empty file to an s3 bucket which has object lock enabled:

    An error occurred (InvalidRequest) when calling the PutObject operation: Content-MD5 HTTP header is required for Put Object requests with Object Lock parameters

The API requires this header, but in botocore it does not inject it, I think problem is probably here within the [handler for `before-call.s3.PutObject`](https://github.com/boto/botocore/blame/1aac4d7c1f311826a1a4686e54444f6ded72ea45/botocore/handlers.py#L951):

https://github.com/boto/botocore/blame/1aac4d7c1f311826a1a4686e54444f6ded72ea45/botocore/handlers.py#L187-L196

Look at the if-statement:

    if request_dict['body'] and 'Content-MD5' not in params['headers']:
        # ... the params['headers']['Content-MD5'] gets set ...

It neglects the possibility that `request_dict["body"]` is empty.

For >= 1 bytes files, `content-md5` gets automatically set and upload succeeds.

swetashre:
Thank you for your post. I am not able to reproduce the issue with a Object Lock enabled s3 bucket. Can you please give me a code sample to reproduce the issue ? I have tested with this code:
```python
s3 = boto3.client('s3')
data = ''
res = s3.put_object(Bucket = 'bucketname', Body=data, Key='emptyfile.txt')
```
It would be helpful if you could provide me debug logs as well. You can enable logs by adding ```boto3.set_stream_logger('')``` to your code.

wimglenn:
Hi, it seems it doesn't occur with `put_object`.  Here is a reproducer:

```
import boto3

boto3.set_stream_logger('')
s3 = boto3.client('s3')
with open("./emptyfile.txt", "w") as f:
    pass
res = s3.upload_file(Bucket='jump-integration-testing', Filename="emptyfile.txt", Key='emptyfile.txt')
```

swetashre:
I am not able to reproduce the issue with ```upload_file``` api call. Can you please provide me debug log ? You can enable log by adding ```boto3.set_stream_logger('')``` to your code.

wimglenn:
Here is the log.  I have search/replaced the actual bucket name with `example-bucket` and replaced some strings which looked auth-related with XXXXX.

```
>>> botocore.__version__
'1.15.5'
>>> boto3.__version__
'1.12.5'
[...]

PR #1985, opened 2020-02-25
Include Content-MD5 header when body is empty

wimglenn:
Closes https://github.com/boto/botocore/issues/1979

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -25,7 +25,7 @@ import uuid
 
 from botocore.compat import (
     unquote, json, six, unquote_str, ensure_bytes, get_md5,
-    MD5_AVAILABLE, OrderedDict, urlsplit, urlunsplit, XMLParseError,
+    OrderedDict, urlsplit, urlunsplit, XMLParseError,
     ETree,
 )
 from botocore.docs.utils import AutoPopulatedParam
@@ -37,15 +37,11 @@ from botocore.signers import add_generate_db_auth_token
 from botocore.exceptions import ParamValidationError
 from botocore.exceptions import AliasConflictParameterError
 from botocore.exceptions import UnsupportedTLSVersionWarning
-from botocore.exceptions import MissingServiceIdError
 from botocore.utils import percent_encode, SAFE_CHARS
 from botocore.utils import switch_host_with_param
-from botocore.utils import hyphenize_service_id
 from botocore.utils import conditionally_calculate_md5
 
-from botocore import retryhandler
 from botocore import utils
-from botocore import translate
 import botocore
 import botocore.auth
 
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -2138,7 +2138,7 @@ def conditionally_calculate_md5(params, **kwargs):
     """Only add a Content-MD5 if the system supports it."""
     headers = params['headers']
     body = params['body']
-    if MD5_AVAILABLE and body and 'Content-MD5' not in headers:
+    if MD5_AVAILABLE and body is not None and 'Content-MD5' not in headers:
         md5_digest = calculate_md5(body, **kwargs)
         params['headers']['Content-MD5'] = md5_digest
 
--- a/tests/unit/test_handlers.py
+++ b/tests/unit/test_handlers.py
@@ -1223,7 +1223,7 @@ class TestAddMD5(BaseMD5Test):
             'headers': {}
         }
         self.md5_digest.return_value = b'8X\xf6"0\xac<\x91_0\x0cfC\x12\xc6?'
-        handlers.calculate_md5(request_dict)
+        conditionally_calculate_md5(request_dict)
         self.assertEqual(
             request_dict['headers']['Content-MD5'],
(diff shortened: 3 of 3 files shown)


swetashre:
@wimglenn - Can you write a test case to make sure the header is present ?

wimglenn:
@swetashre  Please check the CI failure.  It does not look related to my change at all.

wimglenn:
@swetashre  Is there anything more I can do to help get this bug fixed?

wimglenn:
@swetashre  Is it possible to get a review or get this patch merged? What's the hold-up?

joguSD:
@wimglenn Conceptually, these changes sound alright to me. We need to do some confirmation on our side around all the exact expectations around the content-md5 of empty bodies. We also have some changes in the works around this code path that will end up conflicting and might end up fixing this issue anyways. We'll follow up on this after those changes go out.

nateprewitt:
Hi @wimglenn, just wanted to follow up from our last comment. The changes referenced are now in place. Please feel free to rebase this on top of the master branch and we can take another look. Thanks!

jangrewe:
@nateprewitt Maybe you can have a look at https://github.com/boto/botocore/pull/2181 - this fix works fine for our issue.

wimglenn:
@nateprewitt  @joguSD  @swetashre  Rebased on develop, and updated after the refactor. I hope this doesn't languish again.

nateprewitt:
Thanks again for this PR, @wimglenn! I'm merging this into a feature branch for a couple more tweaks and we're going to try to get this merged into live asap.

radekwlsk:
@wimglenn @nateprewitt FYI: In Django (and especially `django-storages` with `boto3`) when using `ContentFile("")` as empty file stub to create S3 object, i.e. for presigned URL upload, this change makes it fail on save to S3.

```
TypeError: Unicode-objects must be encoded before hashing
  ...
  File "django/db/models/fields/files.py", line 87, in save
    self.name = self.storage.save(name, content, max_length=self.field.max_length)
  File "django/core/files/storage.py", line 52, in save
    return self._save(name, content)
  File "storages/backends/s3boto3.py", line 451, in _save
    obj.upload_fileobj(content, ExtraArgs=params)
  File "boto3/s3/inject.py", line 619, in object_upload_fileobj
    return self.meta.client.upload_fileobj(
  File "boto3/s3/inject.py", line 539, in upload_fileobj
    return future.result()
  File "s3transfer/futures.py", line 106, in result
    return self._coordinator.result()
  File "s3transfer/futures.py", line 265, in result
    raise self._exception
  File "s3transfer/tasks.py", line 126, in __call__
    return self._execute_main(kwargs)
  File "s3transfer/tasks.py", line 150, in _execute_main
    return_value = self._main(**kwargs)
  File "s3transfer/upload.py", line 692, in _main
    client.put_object(Bucket=bucket, Key=key, Body=body, **extra_args)
  File "botocore/client.py", line 357, in _api_call
  File "botocore/client.py", line 652, in _make_api_call
    handler, event_response = self.meta.events.emit_until_response(
  File "botocore/hooks.py", line 360, in emit_until_response
    return self._emitter.emit_until_response(aliased_event_name, **kwargs)
  File "botocore/hooks.py", line 243, in emit_until_response
    responses = self._emit(event_name, kwargs, stop_on_response=True)
  File "botocore/hooks.py", line 211, in _emit
    response = handler(**kwargs)
  File "botocore/utils.py", line 2219, in conditionally_calculate_md5
    md5_digest = calculate_md5(body, **kwargs)
  File "botocore/utils.py", line 2196, in calculate_md5
    binary_md5 = _calculate_md5_from_file(body)
  File "botocore/utils.py", line 2209, in _calculate_md5_from_file
    md5.update(chunk)
```

I suppose making it `ContentFile(b"")` will resolve that issue, as with string it uses `StringIO` instead of `BytesIO` buffer. But maybe text content could be handled here? To encode it first, before calculating md5.

wimglenn:
@radekwlsk  Yes I would think that's a bug in `django-storages`, that code should use BytesIO in the first place. Better to make sure you're sending bytes and not text, it's outside of the responsibility for botocore to guess any encoding to use.

nateprewitt:
Hi @radekwlsk, thanks for bringing this to our attention! It looks like `django-storages` has been violating the API constraint but happened to be getting away with it only for empty files. The [boto3 interface](https://github.com/boto/boto3/blob/develop/boto3/s3/inject.py#L588) they're using states that the fileobj supplied but be in binary mode to work correctly.

@wimglenn is correct that this will likely need to be fixed in that project since guessing around file encodings leads to a lot of unexpected failures.

radekwlsk:
> Hi @radekwlsk, thanks for bringing this to our attention! It looks like `django-storages` has been violating the API constraint but happened to be getting away with it only for "falsey" files. The [boto3 interface](https://github.com/boto/boto3/blob/develop/boto3/s3/inject.py#L588) they're using states that the fileobj supplied must be in binary mode to work correctly.

Thanks for quick response! I've already reported that isssue in `django-storages`. If there is no response I may create a PR myself then.

Issue #1939, opened 2020-01-13
OSError raised if called with 0.0 and different local timezone (UTC+1)


mbayrak78:
Hello,

we catched an OSError on following line if called with 0.0 and timezone UTC+1:

https://github.com/boto/botocore/blob/d6b91089832c5dd14c84d2f00a12204cf5a2b187/botocore/utils.py#L607

The bug is in the summation:
0.0 + (-3600) => can't be converted from OS (windows) to an epoch time earlier than zero point.

BR

swetashre:
@mbayrak78 - Thank you for your post. This looks to be an issue with an underlying library dateutil. I was able to find a related issue on their repository: https://github.com/dateutil/dateutil/issues/197. I am not sure if there is much we can do in boto3 to get around it.

These are some related github issue and stack overflow link which may help you to find a workaround : 
https://github.com/boto/boto3/issues/1238
https://stackoverflow.com/questions/38918668/dynamodb-create-table-calls-fails

mbayrak78:
@swetashre, thanks for your reply.
As writtenin one of the links you provided, boto3 can patch utils.py as:
https://github.com/boto/boto3/issues/1238#issuecomment-413023250

With that you ensure that at least no date earlier than minimum is calculated (otherwise an OSError occour).

Tobotimus:
> This looks to be an issue with an underlying library dateutil.

I disagree that this is a bug in dateutil. This is a bug in `botocore.utils.parse_timestamp`.

The `parse_timestamp` function naively makes use of the built-in `datetime.fromtimestamp()` classmethod despite the many warnings listed about the edge cases of that classmethod in its [documentation]((https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp)) (in particular, the warning about parsing timestamps outside of a platform-dependent range). I think the `parse_timestamp` function ought to use an alternative method to parse timestamps. This is especially necessary with regards to timestamps returned in API responses (which I gather are all relative to the Unix Epoch) whereas the `datetime.fromtimestamp` classmethod is intended for timestamps generated by the current platform, Unix or otherwise.

jwilges:
> > This looks to be an issue with an underlying library dateutil.
> 
> I disagree that this is a bug in dateutil. This is a bug in `botocore.utils.parse_timestamp`.

[...]

PR #1987, opened 2020-02-26
Prefer tzwinlocal over tzlocal in parse_timestamp to handle negative timestamps

jwilges:
Issue #1939 draws attention to an `OSError` that is raised on Windows in some timezones when `botocore.utils.parse_timestamp` relies on `tzlocal()` instead of using `tzwinlocal()` during this operation:
```python
datetime.datetime.fromtimestamp(value, tzlocal())
```

The **dateutil** maintainers are aware of this interoperability issue and have created issue [#848](https://github.com/dateutil/dateutil/issues/848) against **dateutil** to propose a change to make `tz.gettz()` conditionally return `tzwinlocal()` on Windows.

@pganssle discusses this motivation in [this comment](https://github.com/dateutil/dateutil/issues/197#issuecomment-448681838):

> I think my preferred solution to the very real UX challenge you have identified, @h-vetinari, is to make tz.gettz() return tz.tzwinlocal instead of tz.tzlocal. In general, I think that tz.gettz is intended to get you the "right time zone object" for the job, and it may make sense to make tz.tzwinlocal the default "local time" on platforms where it's available.

Since the proposed upstream changes for **dateutil** may involve them bumping their major version of the library (due to breaking changes), I am proposing a more targeted patch to resolve the Windows issue in **botocore** today. It may also be worth marking this change with a `TODO` to later use the new `tz.gettz()` call when **dateutil** eventually resolves their issue.

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -24,7 +24,7 @@ import socket
 import cgi
 
 import dateutil.parser
-from dateutil.tz import tzlocal, tzutc
+from dateutil.tz import tzlocal, tzutc, tzwinlocal
 
 import botocore
 import botocore.awsrequest
@@ -602,6 +602,8 @@ def parse_timestamp(value):
     This will return a ``datetime.datetime`` object.
 
     """
+    # Prefer tzwinlocal when available to handle negative timestamps.
+    tzlocal = tzwinlocal or tzlocal
     if isinstance(value, (int, float)):
         # Possibly an epoch time.
         return datetime.datetime.fromtimestamp(value, tzlocal())
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -24,13 +24,13 @@ import socket
 import cgi
 
 import dateutil.parser
-from dateutil.tz import tzlocal, tzutc, tzwinlocal
+from dateutil.tz import tzutc
 
 import botocore
 import botocore.awsrequest
 import botocore.httpsession
 from botocore.compat import json, quote, zip_longest, urlsplit, urlunsplit
-from botocore.compat import OrderedDict, six, urlparse
+from botocore.compat import OrderedDict, six, urlparse, get_tzinfo_options
 from botocore.vendored.six.moves.urllib.request import getproxies, proxy_bypass
 from botocore.exceptions import (
     InvalidExpressionError, ConfigNotFound, InvalidDNSNameError, ClientError,
@@ -590,6 +590,25 @@ def percent_encode(input_str, safe=SAFE_CHARS):
     return quote(input_str, safe=safe)
 
 
+def _parse_timestamp_with_tzinfo(value, tzinfo):
+    """Parse timestamp with pluggable tzinfo options."""
+    if isinstance(value, (int, float)):
+        # Possibly an epoch time.
+        return datetime.datetime.fromtimestamp(value, tzinfo())
+    else:
+        try:
+            return datetime.datetime.fromtimestamp(float(value), tzinfo())
+        except (TypeError, ValueError):
+            pass
+    try:
+        # In certain cases, a timestamp marked with GMT can be parsed into a
+        # different time zone, so here we provide a context which will
+        # enforce that GMT == UTC.
+        return dateutil.parser.parse(value, tzinfos={'GMT': tzutc()})
+    except (TypeError, ValueError) as e:
+        raise ValueError('Invalid timestamp "%s": %s' % (value, e))
+
+
 def parse_timestamp(value):
     """Parse a timestamp into a datetime object.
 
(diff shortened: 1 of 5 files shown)


jwilges:
Looks like `test_timestamp_example` fails (see below). I'll try to look into it tomorrow to see if the `tzwinlocal` idea can pan out okay.

```test_timestamp_example (tests.unit.docs.test_sharedexample.TestSharedExampleDocumenter) ... ERROR```

File: botocore/utils.py [hunk 1]
```diff
@@ -602,6 +602,8 @@ def parse_timestamp(value):
     This will return a ``datetime.datetime`` object.
 
     """
+    # Prefer tzwinlocal when available to handle negative timestamps.
+    tzlocal = tzwinlocal or tzlocal
```
nateprewitt: Rather than changing our default to `tzwinlocal`, how do you feel about using it as a best effort fallback? That will preserve the current default behavior for most users, but enable graceful recovery.

The rough outline of what I'm thinking would be moving this functions content into a private function that takes the value and a new optional parameter `tzinfo`. From there we add some try/except logic to this function to first attempt parsing with `tzlocal` and then again with `tzwinlocal` if the `OSError` is raised.

e.g.
```python
def _parse_timestamp(value, tzinfo=tzlocal):
     #<current logic of parse_timestamp>

def parse_timestamp(value):
    tzinfo_options = (tzlocal, tzwinlocal)
    for tzinfo in tzinfo_options:
        try:
            return _parse_timestamp(value, tzinfo)
        except OSError as e:
            # debug logging about fallback
    # We shouldn't reach this point

jwilges: Hey @nateprewitt, thanks for taking a look. Your suggestion is even more gentle/conservative than the one I quickly typed up in my initial changeset while maintaining a similar spirit: find the best `tzinfo` value for the current environment. Nice work.

I will try out a test suite run with your changes out of curiosity to see if your changes produce fewer noise in CI with respect to test failures. If I recall correctly when I ran my changeset against some of the tests that failed on `botocore`'s Travis CI Linux containers they passed for me on a local `python:3.7-slim` (Debian) container. If your changes run without any noise in CI, all the better!

The bright side is your changeset has faith that the system may resolve successfully with `tzlocal` in some cases before ever needing to fall back to `tzwinlocal`, but the trade off is of course that if a Windows system remains in a problematic timezone, timestamps within the range that fail `tzlocal` will consistently require a second attempt (and will log to debug or warn level accordingly). But only certain ranges of timestamps fail depending on the timezone, so perhaps it is fair to mention your change will not directly double the total number of attempts.

My original suggestion relied on a different angle of optimism: on Windows, regardless of the timezone, `tzwinlocal` (so far) provides us a `tzinfo` object that will yield the correct result for all timestamps in the underlying `time.localtime(...)` call.

There's no response yet upstream from the `dateutil` authors as to what they plan to do with respect to `tz.gettz()` for issue [#848](https://github.com/dateutil/dateutil/issues/848) beyond acknowledging that this is a known portability issue, but regardless of how that issue is resolved a change along the lines of what we have proposed here in `botocore` will be required to ensure the this part of the `botocore` library works on Windows as well as it does on POSIX systems.

jwilges:
> Hey @jwilges,
> 
> Thanks for contributing to botocore! I agree we can do a bit better here with the Windows experience and this looks like a good start.
> 
> I'm slightly concerned we're potentially changing the default behavior for all Windows users with this approach, including those not experiencing this issue. I left a quick brainstorm about an alternative approach if you have thoughts on it.
> 
> If we could also get some quick tests and a changelog entry in here, that would be great!

Hi @nateprewitt, agreed on all counts. I started this changeset/issue as a workable talking point to get us thinking at the very least.

I agree we should add tests. An integration test would best expose this by using a known-bad timezone (one outside of UTC) on a Windows container running the standard cpython distribution to trigger the underlying `OSError` from the [`_PyTime_localtime()` call in `time.localtime()`](https://github.com/python/cpython/blob/1b55b65638254aa78b005fbf0b71fb02499f1852/Modules/timemodule.c#L494). Unit tests although contrived are obviously feasible. It looks like `botocore` does not currently run Windows containers on Travis CI, so a proper integration test for this will involve a bit of scope creep probably.

botocore/utils.py
nateprewitt: Yep, I think simplifying to just using `tz.gettz()` is the long term goal if #848 gets resolved. We can probably make a note that as a todo.

I agree this is a slight performance hit with the retry and may be some additional noise. The thought process for the tradeoff is it's only for the edge case, and it avoids having to do more extensive testing around things like bad registry entries and any other potential anomalies on windows. We might look at a log filter at a later date if we find the debug noise is excessive.

nateprewitt:
> An integration test would best expose this by using a known-bad timezone (one outside of UTC) on a Windows container running the standard cpython distribution to trigger the underlying OSError from the _PyTime_localtime() call in time.localtime().

Yep, I think that's spot on.

>It looks like botocore does not currently run Windows containers on Travis CI, so a proper integration test for this will involve a bit of scope creep probably.

You're right Travis isn't currently running Windows so we'd miss this in PRs but we do have a separate testing process that runs for the daily releases. We're currently talking about our options for Windows testing on Github, so it'd be good to have this for that future case. I don't think it needs to be anything crazy, just a Windows only test that runs something like the example in the ticket.

If you're interested in implementing any or all of the above, we'd happily welcome it. Otherwise, I can look at trying to build on this PR later in the week. Just let us know what your preference would be and we'll go from there. Thanks again for starting the convo!

PR #1990, opened 2020-02-29
Not expect 100-continue with body size == 0.

fao89:
ref https://github.com/boto/botocore/pull/1328
HTTP RFC explicitly states that 100-continue should
not be set when there is no message body.

https://tools.ietf.org/html/rfc7231#section-5.1.1

> A client MUST NOT generate a 100-continue expectation in
> a request that does not include a message body.

This is a mandatory requirement, this PR fixes the current
implementation behavior by not sending `100-continue`
when content-length is '0' which means there is no message
body
fixes https://github.com/boto/boto3/issues/1341

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -282,9 +282,11 @@ def add_expect_header(model, params, **kwargs):
         return
     if 'body' in params:
         body = params['body']
-        if hasattr(body, 'read'):
+        size = getattr(body, "_size", None)
+        if size and size != "0":
             # Any file like object will use an expect 100-continue
-            # header regardless of size.
+            # except when size equals zero.
+            # https://tools.ietf.org/html/rfc7231#section-5.1.1
             logger.debug("Adding expect 100 continue header to request.")
             params['headers']['Expect'] = '100-continue'
 
--- a/tests/unit/test_awsrequest.py
+++ b/tests/unit/test_awsrequest.py
@@ -436,6 +436,14 @@ class TestAWSHTTPConnection(unittest.TestCase):
         response = conn.getresponse()
         self.assertEqual(response.status, 200)
 
+    def test_no_expect_header_set_no_body(self):
+        s = FakeSocket(b'HTTP/1.1 200 OK\r\n')
+        conn = AWSHTTPConnection('s3.amazonaws.com', 443)
+        conn.sock = s
+        conn.request('PUT', '/bucket/foo', b'')
+        response = conn.getresponse()
+        self.assertEqual(response.status, 200)
+
     def test_tunnel_readline_none_bugfix(self):
         # Tests whether ``_tunnel`` function is able to work around the
         # py26 bug of avoiding infinite while loop if nothing is returned.


daviddavis:
@kyleknap @stealthycoin @joguSD any chance we can get a review?

daviddavis:
@kyleknap @stealthycoin @joguSD hey any chance we can get a review?

sp-ricard-valverde (commented):
Is there any chance that this PR is considered for merging ? This is actually a very valid improvement for boto integrations to work properly.

Regarding the PR itself, I would consider changing:
````python
if getattr(body, "_size", None):
  ...
````
for:
````python
size = getattr(body, "_size", None)
if size and size != "0" :
...
````
If it makes sense.

**Comment by harshavardhana** [comment]
> @kyleknap @stealthycoin @joguSD hey any chance we can get a review?

Bump :+1:

File: botocore/handlers.py [hunk 1]
```diff
@@ -282,9 +282,11 @@ def add_expect_header(model, params, **kwargs):
         return
     if 'body' in params:
         body = params['body']
-        if hasattr(body, 'read'):
+        size = getattr(body, "_size", None)
+        if size and size != "0":
```
mdellweg, line 286:
```suggestion
        if size:
```

If `size` is zero, it evaluates to `false` automatically.

sp-ricard-valverde: That covers the case of when `size` is `"0"` as a string, so the value was not converted to `int`.

daviddavis:
@nateprewitt any chance we can get this merged?

botocore/handlers.py
```diff
@@ -282,9 +282,11 @@ def add_expect_header(model, params, **kwargs):
         return
     if 'body' in params:
         body = params['body']
-        if hasattr(body, 'read'):
+        size = getattr(body, "_size", None)
```
nateprewitt, line 285: This unfortunately won't work because it relies on a private variable that's not always present. This change is a fundamental rework of how we compute this which is breaking. Objects that currently implement `read` without `_size` will stop emitting a 100. This also introduces a forward compatibility risk, that if an object ever removes, renames, or changes data types of the private `_size` variable, this logic fails unexpectedly.

If we introduce a solution, we need it to rely solely on public interfaces. It also needs to operate without reading the body, because some inputs are not rewindable and the data will be lost. This is what introduces the complexity that have prevented previous fixes.

We agree with what the SDK _should_ do in this case, but given it's not a general HTTP client we've scoped it to accepted behavior within the AWS ecosystem. Otherwise, we wouldn't be able to support critical features in S3. Long term, we're looking at more flexible solutions for other platforms, but have nothing concrete at this time.

nateprewitt (commented):
Hi @fao89,

I wanted to quickly leave a note since this PR has some requests for updates. We appreciate the contribution but unfortunately can't accept it. We're willing to discuss alternatives in the tracking issue, but they need to be scoped to support Boto's current requirements.

I'll resolve this for now and we can track further discussion in the issue. Thanks!

PR #1996, opened 2020-03-10
Fixing stubber to handle client error for GetBucketLocation command

### LINKED ISSUE #1884: get_bucket_location returns NoneType error when client error is added
## Problem

When adding a client error through `get_bucket_location`, rather than raise a `ClientError` (or the specified Service Error Code, we instead receive;

 `AttributeError: 'NoneType' object has no attribute 'stream'`

## Working Example (with `get_bucket_logging`)

```
import boto3
from botocore.stub import Stubber

client = boto3.client('s3')
stubber = Stubber(client)
stubber.add_client_error('get_bucket_logging', 'NoSuchBucket')
stubber.activate()

# Raises botocore.errorfactory.NoSuchBucket
client.get_bucket_logging(Bucket="meh")
```
## Non-working Example (with `get_bucket_location`)

```
import boto3
from botocore.stub import Stubber

client = boto3.client('s3')
stubber = Stubber(client)
stubber.add_client_error('get_bucket_location', 'NoSuchBucket')
stubber.activate()

# Raises an AttributeError from within awsrequest.py
client.get_bucket_location(Bucket="meh")
```
The full stack response from this error is;
```
  File "untitled.py", line 24, in <module>
    client.get_bucket_location(Bucket="meh")
  File "/usr/local/lib/python3.7/site-packages/botocore/client.py", line 357, in _api_call
  File "/usr/local/lib/python3.7/site-packages/botocore/client.py", line 655, in _make_api_call
    model=operation_model, context=request_context
  File "/usr/local/lib/python3.7/site-packages/botocore/hooks.py", line 356, in emit
    return self._emitter.emit(aliased_event_name, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/botocore/hooks.py", line 228, in emit
  File "/usr/local/lib/python3.7/site-packages/botocore/hooks.py", line 211, in _emit
    response = handler(**kwargs)
  File "/usr/local/lib/python3.7/site-packages/botocore/handlers.py", line 485, in parse_get_bucket_location
    response_body = http_response.content
  File "/usr/local/lib/python3.7/site-packages/botocore/awsrequest.py", line 587, in content
    self._content = bytes().join(self.raw.stream()) or bytes()
AttributeError: 'NoneType' object has no attribute 'stream'
```
## Summary

I've tried searching open/closed issues but can't find something similar, nor a working example online using `get_bucket_location`. I fear this might be a bug within `botocore.stub` itself. Any help would be greatly appreciated! Thanks.

swetashre:
Fixes #1884

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -484,8 +484,7 @@ def parse_get_bucket_location(parsed, http_response, **kwargs):
     # The "parsed" passed in only has the ResponseMetadata
     # filled out.  This handler will fill in the LocationConstraint
     # value.
-    if 'LocationConstraint' in parsed or 'Error' in parsed:
-        # Response already set - a stub?
+    if http_response.raw is None:
         return
     response_body = http_response.content
     parser = xml.etree.cElementTree.XMLParser(
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -484,7 +484,7 @@ def parse_get_bucket_location(parsed, http_response, **kwargs):
     # The "parsed" passed in only has the ResponseMetadata
     # filled out.  This handler will fill in the LocationConstraint
     # value.
-    if 'LocationConstraint' in parsed:
+    if 'LocationConstraint' in parsed or 'Error' in parsed:
         # Response already set - a stub?
         return
     response_body = http_response.content
--- a/.changes/next-release/bugfix-Stubber-96340.json
+++ b/.changes/next-release/bugfix-Stubber-96340.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Stubber",
+  "description": "fixes `#1884 <https://github.com/boto/botocore/issues/1884>`__"
+}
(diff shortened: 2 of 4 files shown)


botocore/handlers.py
```diff
@@ -484,7 +484,7 @@ def parse_get_bucket_location(parsed, http_response, **kwargs):
     # The "parsed" passed in only has the ResponseMetadata
     # filled out.  This handler will fill in the LocationConstraint
     # value.
-    if 'LocationConstraint' in parsed:
+    if 'LocationConstraint' in parsed or 'Error' in parsed:
```
stealthycoin: What do people think about changing this if statement to 
`if http_response.raw is None:`?

The handler function is only needed to manually parse a raw response into a parsed response since it can't be modeled correctly. The precondition for this function is that this a real response we got from the service. And it has a raw_response, and an incorrect parsed_response that needs to be manually built  from the raw response. It should skip if this was a stubbed response as there is no raw response, only a preset parsed response.

I think it makes more sense to check this as directly as possible. `parse_get_bucket_location` depends on the raw response, so we should just check it. If it isnt there, there is nothing to do. The current if statements have the same effect of skipping the function in the correct cases, but they don't really do a good job of explaining why. They check for the side effects of the case where there is no raw response. We could then update the comment to say that there is no raw response to parse, this is likely a stubbed response.

swetashre: yeah, this makes sense to me. We can directly check for the raw response rather than checking for every side effects of it.

PR #2004, opened 2020-03-26
Added missing paginators for SecurityHub

jurajseffer:


jurajseffer:
Bump

kdaily:
Hi @jurajseffer, thanks for the request. We are now submitting requests for paginators and waiters upstream so that all AWS SDKs benefit from a consistent implementation. I will pass on this request for this paginator to the service team. I will also update the guidance in the contributing guide to better reflect this policy. I've added an issue to track this here:

https://github.com/aws/aws-sdk/issues/78

PR #2007, opened 2020-03-30
Move Kinesis tests to functional

nateprewitt:
Moving our kinesis tests to operate offline as a functional test to avoid erroneous failures due to throttling and other transient service issues.

Tests now validate we maintain our expected request structure going forward.

** Also fixing our tox script to only test our supported versions to save time validating.


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py26,py27,py33,py34,py35,py36,py37,py38
```
kyleknap, line 2: We still support py34, but I'm fine with dropping the other two.

tests/functional/test_kinesis.py
```diff
@@ -0,0 +1,92 @@
...
+                b64decode(record_foobar["Data"]).decode("utf-8"),
+                "foobar"
+            )
+            self.assertEqual(
+                b64decode(record_barfoo["Data"]).decode("utf-8"),
+                "barfoo"
+            )
+
+
+if __name__ == "__main__":
```
kyleknap, line 91: Not really a problem, but we don't necessarily need the `if` statement here as we really only run the tests with `nose`.

tests/functional/test_kinesis.py
```diff
@@ -0,0 +1,92 @@
...
+        with self.http_stubber as stub:
+            self.client.put_records(
+                StreamName=self.stream_name,
+                Records=[
+                    {"Data": "foobar", "PartitionKey": "foo"},
+                    {"Data": "barfoo", "PartitionKey": "foo"},
+                ],
+            )
+            self.assertEqual(len(stub.requests), 1)
+            # TODO: Decode for 3.4 and 3.5 (remove after deprecation)
```
kyleknap, line 75: I'm a little confused about the todo. Mainly I'm not sure what what we would need to remove? Is it specifying `utf-8` or the `decode` because I'd assume we would always need to decode it because it is binary.

tests/functional/test_kinesis.py
```diff
@@ -0,0 +1,92 @@
...
+            request = json.loads(stub.requests[0].body.decode("utf-8"))
+            self.assertEqual(len(request["Records"]), 2)
+
+            record_foobar = request["Records"][0]
+            record_barfoo = request["Records"][1]
+            self.assertEqual(
```
kyleknap, line 81: It might be worth adding a helper assertion method like `assert_base64encoded_record`, maybe we can come up with a better name, to help make the test case a little easier to read and also so we do not have duplicate the same logic for each assertion

PR #2020, opened 2020-04-14
Removed instruction from msk config to base64-encode

danielorf:
The doc for MSK (kafka) create_configuration [ServerProperties](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/kafka.html#Kafka.Client.create_configuration) param incorrectly instructs the user to base64-encode the config file.  Instead, the file should be read to a string or bytes variable as shown [here](https://docs.aws.amazon.com/msk/1.0/apireference/operations.html).

### DIFF
--- a/botocore/data/kafka/2018-11-14/service-2.json
+++ b/botocore/data/kafka/2018-11-14/service-2.json
@@ -1345,7 +1345,7 @@
         "ServerProperties": {
           "shape": "__blob",
           "locationName": "serverProperties",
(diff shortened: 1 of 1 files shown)


danielorf:
The [location](https://github.com/boto/botocore/blob/develop/tests/functional/retries/test_bucket.py#L104) of the failing test, combined with the seemingly unrelated change leads me to believe that the failure is a resource issue of the underlying travis instance or hardware.  Is it possible to re-run the failing test?

swetashre:
@danielorf - Thank you for the PR. We can't merge this as this file comes from upstream and shared across different sdk. I will reach out to the service team about the issue.

Issue #2026, opened 2020-04-24
Support querying codecommit of partition 'aws-cn' in session


zxkane:
**AWS CodeCommit** is GA in AWS China(for both `cn-north-1` and `cn-northwest-1`).

Above regions are not available when querying regions for service 'codecommit' for 'aws-cn'.

```python
>>> import botocore.session
>>> session = botocore.session.Session(profile = 'default')
>>> session.get_available_partitions()
['aws', 'aws-cn', 'aws-us-gov']
>>> session.get_available_regions('codecommit', 'aws-cn')
[]
>>> session.get_available_regions('codecommit', 'aws')
['ap-northeast-1', 'ap-northeast-2', 'ap-south-1', 'ap-southeast-1', 'ap-southeast-2', 'ca-central-1', 'eu-central-1', 'eu-west-1', 'eu-west-2', 'eu-west-3', 'sa-east-1', 'us-east-1', 'us-east-2', 'us-west-1', 'us-west-2']
```

swetashre:
@zxkane - Thank you for your post. This change was introduced yesterday here:
https://www.amazonaws.cn/en/new/2020/aws-codecommit-available-aws-china-beijing-sinnet-ningxia-nwcd/

As this is recent change, those region has not included in the endpoints.json file.  The ```endpoints.json``` file comes from upstream. So the service team will eventually include those regions as they are available now. I think it only matters of time.

PR #2027, opened 2020-04-24
Enable the codecommit support in China regions

henrysher:
Fix the issue #2026
Here attached the announcement: https://www.amazonaws.cn/en/new/2020/aws-codecommit-available-aws-china-beijing-sinnet-ningxia-nwcd/

---
This change is [Reviewable](https://reviewable.io/reviews/boto/botocore/2027)

### DIFF
--- a/botocore/data/endpoints.json
+++ b/botocore/data/endpoints.json
@@ -5724,6 +5724,12 @@
           "cn-northwest-1" : { }
         }
       },
+      "codecommit" : {
+        "endpoints" : {
+          "cn-north-1" : { },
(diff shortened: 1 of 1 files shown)


swetashre:
@henrysher - Thank you for putting up a PR for it. These china regions have already been added to the endpoint file 
https://github.com/boto/botocore/blob/develop/botocore/data/endpoints.json#L6066

In general endpoint.json file comes from upstream and unfortunately we can't accept change for this file. Closing the PR as the change has already been added.

PR #2033, opened 2020-04-27
Handling S3 connection drops

nateprewitt:
There are some operations where S3 streams data to us to keep a connection alive while waiting for a long running operation to complete. In cases where the client drops the connection or has other intermittent network connection issues, we need to be more resilient.

This patch will attempt to retry these cases where they're recoverable, and prompt the client application to retry the operation again if we're unable to recover. This will also avoid leaking exceptions from the underlying xml library which we'd been doing before in these rare cases.

### DIFF
--- a/tests/functional/test_s3.py
+++ b/tests/functional/test_s3.py
@@ -18,10 +18,10 @@ from nose.tools import assert_equal
 
 import botocore.session
 from botocore.config import Config
-from botocore.compat import urlsplit
-from botocore.compat import parse_qs
+from botocore.compat import datetime, urlsplit, parse_qs
 from botocore.exceptions import ParamValidationError, ClientError
 from botocore.exceptions import InvalidS3UsEast1RegionalEndpointConfigError
+from botocore.parsers import ResponseParserError
 from botocore import UNSIGNED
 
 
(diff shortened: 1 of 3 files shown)


botocore/compat.py
```diff
@@ -23,7 +23,7 @@
 
 from botocore.vendored import six
 from botocore.exceptions import MD5UnavailableError
-from dateutil.tz import tzlocal
+from dateutil.tz import tzlocal, tzutc
```
kyleknap, line 26: Was this import intentional?

tests/functional/test_s3.py
```diff
@@ -409,6 +409,74 @@ def test_client_region_remains_aws_global_for_regional(self):
         self.assertEqual(client.meta.region_name, 'aws-global')
 
 
+class TestS3Copy(BaseS3OperationTest):
+
+    def setUp(self):
```
kyleknap, line 414: We probably don't need the `setUp` definition here.

botocore/compat.py
nateprewitt: Yeah, it's being used in the tests file. I wanted to keep imports for different timezones consolidated if possible.

tests/functional/test_s3.py
nateprewitt: Good catch, I pulled the customizations out and this got missed in the clean up. Removed in latest revision.

botocore/compat.py
nateprewitt: Sure, I'll just yank that assert and pull the import.

PR #2036, opened 2020-04-28
Update issue templates

swetashre:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,20 @@
...
+A clear and concise description of what the bug is.
+
+**Steps to reproduce**
+If you have a runnable example, please include it as a snippet or link to a repository/gist for larger code examples.
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Debug logs**
+Full stack trace by adding `boto3.set_stream_logger('')` to your code.
```
kyleknap, line 20: We should probably use the botocore version of this, which boils down to:
```python
import botocore.session
botocore.session.Session().set_debug_logger('')
```

.github/ISSUE_TEMPLATE/guidance-issue.md
```diff
@@ -0,0 +1,18 @@
...
+
+Please fill out the sections below to help us address your issue.
+
+**What issue did you see ?**
+
+**Steps to reproduce**
+If you have a runnable example, please include it as a snippet or link to a repository/gist for larger code examples.
+
+**Debug logs**
+Full stack trace by adding `boto3.set_stream_logger('')` to your code.
```
kyleknap, line 18: Same comment here as in the `bug_report` template where we should be using the botocore version of `set_stream_logger`.

Issue #1814, opened 2019-08-23
How to use generated exceptions


takeda:
So I'm trying to use SSM and catch scenario where value doesn't exist. The call generates `botocore.errorfactory.ParameterNotFound: An error occurred (ParameterNotFound) when calling the GetParameter operation`

The thing is that ParameterNotFound does not exist the class with the error is dynamically generated and I can't reference it in the code (or at least not from `botocore.errorfactory.ParameterNotFound`). What's the point of having exceptions like this when they can't be caught?  Why bothering with dynamically generated class, when you could just have a generic error and inside of its properties I could extract the actual cause. Or is there a way to use it that I'm not aware of?

swetashre:
@takeda - Thank you for your post. Basically this error comes from the service when the requested parameter does not exist. You can see in the [documentation](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParameter.html) that what does the error means and what is the cause of this. You can find the root causes in the code for client errors only. 

Hope it helps and please let me know if you have any questions.

takeda:
The problem I'm having is that I can't use such exception in try-except clause, so it has a little of use as an actual exception.

swetashre:
@takeda - Modeled exceptions can be accessed through the client. Instead of ```botocore.errorfactory.ParameterNotFound``` you can do ```client.exceptions.ParameterNotFound```.
```python
client = boto3.client('ssm')

try:
   response = client.get_parameter(Name='sweta')
except client.exceptions.ParameterNotFound as e:
    print(e)
```
 I agree that this is confusing and the current documentation is insufficient. More information on this can be found at this pr: boto/botocore#1113 .

swetashre:
More information on how to catch exceptions with boto3/botocore
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html

Closing this issue as it has been fixed. Please reopen if you have any more concerns.

Issue #1489, opened 2018-06-22
error factory exceptions for generic errors


kapilt:
It would be nice to include generic exceptions, else users can only conditionally use exceptions classes on the basis of their existence without any clear guidance outside of trying each manually, and will accumulate as multiple error handling styles in a codebase.

```python
>>> import boto3
>>> client = boto3.client('s3')
>>> client.exceptions.AccessDenied
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/kapilt/projects/cloud-custodian/lib/python3.6/site-packages/botocore/errorfactory.py", line 53, in __getattr__
    self, name, ', '.join(exception_cls_names)))
AttributeError: <botocore.errorfactory.S3Exceptions object at 0x108f2efd0> object has no attribute 'AccessDenied'. Valid exceptions are: BucketAlreadyExists, BucketAlreadyOwnedByYou, NoSuchBucket, NoSuchKey, NoSuchUpload, ObjectAlreadyInActiveTierError, ObjectNotInActiveTierError
```

kapilt:
as an easy alternative, exposing error factory off the client and allowing users to create/retrieve exceptions by code would be nice.

kapilt:
something like

```python
except client.exceptions.for_code('AccessDenied'):
    pass
```

which would do a from_code for class cache lookup before constructing & caching.

JordonPhillips:
Yeah I think something like that would be great! The downfall of what we have now is it is based off of the service models, which don't always define every exception. A solution like `for_code` would be cool.

rollokb:
Question, if AccessDenied is returned by S3, why isn't it part of the service definition?

Issue #885, opened 2016-04-16
Consistent documentation for the ClientError


kevinburke:
The ClientError is now (I think?) the gateway to most AWS service exceptions. The documentation around it is not very clear.

[Here](http://botocore.readthedocs.org/en/latest/client_upgrades.html#error-handling), we're checking ['Error']['Code'] for the string `'InvalidInstanceID.NotFound'`:

``` python
from botocore.exceptions import ClientError

ec2 = session.get_client('ec2', 'us-west-2')
try:
    parsed = ec2.describe_instances(InstanceIds=['i-badid'])
except ClientError as e:
    logger.error("Received error: %s", e, exc_info=True)
    # Only worry about a specific service error code
    if e.response['Error']['Code'] == 'InvalidInstanceID.NotFound':
        raise
```

[Here](http://boto3.readthedocs.org/en/latest/guide/migrations3.html?highlight=clienterror#accessing-a-bucket) we're checking it for the string `"404"`. Note `int()` will throw if any other error is present in that field.

``` python
import botocore
bucket = s3.Bucket('mybucket')
exists = True
try:
    s3.meta.client.head_bucket(Bucket='mybucket')
except botocore.exceptions.ClientError as e:
    # If a client error is thrown, then check that it was a 404 error.
    # If it was a 404 error, then the bucket does not exist.
    error_code = int(e.response['Error']['Code'])
    if error_code == 404:
        exists = False
```
- Can you provide more guidance on what the field in Error.Code looks like, and how this matches up with exceptions mentioned (but [not linked](https://github.com/boto/boto3/issues/597) ) in the documentation like LimitExceededException?
- Is Error.Code guaranteed to always exist for a ClientError, or do I need another level of try/catch error wrapping around that?
- What other fields exist on `ClientError.response`, besides `Error` ?

jamesls:
* The error code will match the errors returned by the service, so the code in that example would be `LimitExceededException`.
* The error code will always exist.  If it doesn't, it's a bug.
* The other top level key is `ResponseMetadata`, which can have a `RequestId`.

I'll get the docs updated with more info on this, I agree this should be documented.

lenucksi:
👍 

Also a clear documentation of every possible exception dict content at best linked from or included at the location where the said content may occur would be very helpful.

I would also find it very appreciable if you could provide proper, exception classes per service and exception instead of the dict approach. That would allow to use try in a more precise, pythonic way without catching the ClientError blank and then wading around in Exception details.

[...]

PR #2037, opened 2020-04-29
Modeled exceptions field parsing and documentation

### LINKED ISSUE #1937: LexModelBuildingService Exceptions do not pull information from response body
# Issue
The order to delete all resources associated with a lex bot you must do the following:
* Delete all aliases associated with a bot in order to delete the bot itself
* Delete the bot in order to delete intents
* Delete intents in order to delete slot types
* Delete slot types

If you attempt to delete a lex resource which references another lex resource a `ResourceInUseException` is returned. The response body from AWS is in the following format:
```json
{
    "exampleReference": {
        "name": "mycoolbotalias",
        "version": null
    },
    "referenceType": "BotAlias"
}
```
The above AWS response was from attempting to delete a Lex bot when an alias existed for the bot. Botocore prints out the following when the `ResourceInUseException` is throw:
```
An error occurred (ResourceInUseException) when calling the DeleteBot operation:
```
As you can there's no information about the alias which must first be deleted to delete the bot, but the documentation states that if this exception is throw you will be given a reference to the resource that needs to be deleted. This can be seen here https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/lex-models.html#LexModelBuildingService.Client.delete_slot_type

I believe the issue is coming from some inconsistency in the Lex API, as botocore looks for any error messages in the body under a `message` or `Message` key, which in this case is not being returned. This can be seen in the BaseJsonParser class here https://github.com/boto/botocore/blob/develop/botocore/parsers.py#L602
```python
error['Error']['Message'] = body.get('message',
                                             body.get('Message',''))
```
As you can see if nothing is found the message is set to an empty string, a quick solution I mocked up locally to get around this, was to add the body as the default value if the `message` or `Message` key does not exist.
```python
error['Error']['Message'] = body.get('message',
                                             body.get('Message', body))
```
Running the code above results in the following being printed out:
```
An error occurred (ResourceInUseException) when calling the DeleteBot operation: {'exampleReference': {'name': 'mycoolbotalias', 'version': None}, 'referenceType': 'BotAlias'}
```

This issue is also in the aws cli (with current code)
image

With proposed code:
image

If you are happy with this solution I can open a pull request with this change, if there is another solution you would prefer, please let me know. 

Thanks!

joguSD:
This updates our modeled exception parsing logic to include parsing any additional modeled fields present on exceptions and adds documentation generation for the modeled exception classes.

Fixes #1937, #1814, #885, #1489


botocore/endpoint.py
```diff
@@ -216,9 +216,34 @@ def _do_get_response(self, request, operation_model):
...
+    ):
+        error_code = parsed_response.get("Error", {}).get("Code")
+        if error_code is None:
+            return
+        service_model = operation_model.service_model
+        error_shape = service_model.shape_for_error_code(error_code)
+        if error_shape is None:
+            return
+        modeled_parse = parser.parse(response_dict, error_shape)
+        # TODO: avoid naming conflicts with ResponseMetadata and Error
```
kyleknap, line 244: Did we at least add a test here to detect when the collisions occur?

joguSD: I've updated with some tests to validate this. It's unfortunately a little trickier than it seems at first glance to add this. Because `Error` and `ResponseMetadata` are injected at the parser level we need to do the renaming there, ideally doing the rename when we're parsing the members for a structure (in the `_handle_structure` methods of the parsers). The unfortunate thing is we don't have a great way to know when we're parsing a top-level structure vs a nested structure which we need for `ResponseMetadata` renames. We can work around this for `Error` as that only applies to structures with the trait `exception`. For `ResponseMetadata` we might have to just audit everywhere we might inject that key and ensure we aren't clobbering the field if it doesn't already exist.

PR #2038, opened 2020-04-30
Update shared lambda examples

kyleknap:


kyleknap:
We are going to have to hold on off on pulling this in. Some of the examples have invalid timestamps (e.g `${timestamp}` as a value.) I'm looking to get an updated `examples-1.json` file.

PR #2042, opened 2020-05-06
fix endpoint resolution with unknown regions

nateprewitt:
When a client is handed an unresolvable region, we create an assumed regionalized endpoint and send a request hoping it will work. In many cases, it won't and results in extra wait time for the connection, followed by an exception. 

This patch will change the behavior to instead validate if the service being called is regionalized. If not, we can fall back to the global endpoint and continue with the call. This should prevent unnecessary failures in cases where the region isn't relevant to the service request.

### DIFF
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -359,6 +359,15 @@ class ClientEndpointBridge(object):
         region_name = self._check_default_region(service_name, region_name)
         resolved = self.endpoint_resolver.construct_endpoint(
             service_name, region_name)
+
+        # If we can't resolve the region, we'll attempt to get a global
+        # endpoint for non-regionalized services (iam, route53, etc)
+        if not resolved:
+            # TODO: fallback partition_name should be configurable in the
+            # future for users to define as needed.
+            resolved = self.endpoint_resolver.construct_endpoint(
+                service_name, region_name, partition_name='aws')
+
         if resolved:
             return self._create_endpoint(
                 resolved, service_name, region_name, endpoint_url, is_secure)
--- a/.changes/next-release/enchancement-Endpoints-29513.json
+++ b/.changes/next-release/enchancement-Endpoints-29513.json
@@ -1,0 +1,5 @@
+{
+  "type": "enchancement",
+  "category": "Endpoints",
+  "description": "Improved endpoint resolution for clients with unknown regions"
+}
--- a/botocore/regions.py
+++ b/botocore/regions.py
@@ -115,7 +115,19 @@ class EndpointResolver(BaseEndpointResolver):
                     result.append(endpoint_name)
         return result
 
-    def construct_endpoint(self, service_name, region_name=None):
+    def construct_endpoint(self, service_name, region_name=None, partition_name=None):
+        if partition_name is not None:
+            valid_partition = None
+            for partition in self._endpoint_data['partitions']:
+                if partition['partition'] == partition_name:
+                    valid_partition = partition
+
+            if valid_partition is not None:
+                result = self._endpoint_for_partition(valid_partition, service_name,
+                             region_name, True)
+                return result
+            return None
+
         # Iterate over each partition until a match is found.
         for partition in self._endpoint_data['partitions']:
             result = self._endpoint_for_partition(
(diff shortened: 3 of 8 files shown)


tests/functional/test_endpoints.py
```diff
@@ -171,3 +173,45 @@ def _assert_service_name_matches_endpoint_prefix(session, service_name):
         "Actual service name `%s` does not match expected service name "
         "we computed: `%s`" % (
             actual_service_name, computed_name))
+
+
+class TestEndpointResolution(BaseSessionTest):
```
kyleknap, line 178: It might be worth if to see if we can move this over to `tests/functional/test_regions.py`. Mainly the tests there seem more specific to region/endpoint resolution. The tests in this file or more about making sure the information in our endpoints file matches up with the services we have.

tests/functional/test_endpoints.py
```diff
@@ -171,3 +173,45 @@ def _assert_service_name_matches_endpoint_prefix(session, service_name):
...
+
+    def test_regionalized_client_with_unknown_region(self):
+        client, stubber = self.create_stubbed_client('s3', 'not-real')
+        client.list_buckets()
+        # Validate we don't fall back to partition endpoint for
+        # regionalized services.
+        self.assertEquals(
+            stubber.requests[0].url,
+            'https://s3.not-real.amazonaws.com/'
+        )
```
kyleknap, line 203: For the `s3` tests, I think it would make sense to move them over to the `tests/functional/test_s3.py`. In addition, there are a fair amount of additional configurations such as using arns, use_us_east_1_regional, etc. We should add cases to make sure that the endpoint that we will try to hit is what we expect for these edge cases.

tests/functional/test_endpoints.py
```diff
@@ -171,3 +173,45 @@ def _assert_service_name_matches_endpoint_prefix(session, service_name):
...
+
```
kyleknap, line 204: On a similar note, we should add more tests for `sts` to make sure that unknown regions go to the regional endpoint and not the global one. Specifically, I was thinking we add the test cases in `tests/functional/test_sts.py` and make sure that the `AWS_STS_REGIONAL_ENDPOINTS` configuration always results in using the regional endpoint of an unknown sts region.

tests/functional/test_endpoints.py
```diff
@@ -171,3 +173,45 @@ def _assert_service_name_matches_endpoint_prefix(session, service_name):
...
+    def test_unregionalized_client_endpoint_resolution(self):
+        client, stubber = self.create_stubbed_client('route53', 'us-west-2')
+        client.list_geo_locations()
+        self.assertTrue(
+            stubber.requests[0].url.startswith('https://route53.amazonaws.com/')
+        )
+
+    def test_unregionalized_client_with_unknown_region(self):
```
kyleknap, line 212: It would be great if we can add the two test cases for IAM as well here. We definitely should have a check for that.

botocore/client.py
```diff
@@ -359,6 +359,12 @@ def resolve(self, service_name, region_name=None, endpoint_url=None,
         region_name = self._check_default_region(service_name, region_name)
         resolved = self.endpoint_resolver.construct_endpoint(
             service_name, region_name)
+
```
kyleknap, line 362: We should add a changelog entry for this as an `enhancement`.

botocore/client.py
```diff
@@ -373,6 +379,16 @@ def _check_default_region(self, service_name, region_name):
         if self.client_config and self.client_config.region_name is not None:
             return self.client_config.region_name
 
+    def _is_regionalized_endpoint(self, service_name, partition_name='aws'):
```
kyleknap, line 382: I'd prefer to not be accessing the `_endpoint_data`. Looking through the code, I'm not opposed to exposing a general `is_regionalized` method on the `EndpointResovler` class. Specifically:

1. The endpoint resolver is marked as an internal component on the session and that is the only location a user can hook into the resolver.
2. The purpose of the resolver is to be a utility for interpreting the endpoints file. This logic is really only interpreting the endpoints file. Plus it is actually useful functionality to have (i.e. I can see it being exposed on the session in the future like we do for `get_available_regions`).

PR #1969, opened 2020-02-08
Allow botocore loading from zip

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

https://github.com/boto/boto3/issues/1770

Issue #1950, opened 2020-01-28
alternate resource loaders for model files


mattsb42-aws:
I've been poking around trying to use PyOxidizer[1] to create a self-contained binary for a project of mine that uses boto3. Unfortunately, because PyOxidizer (and most other similar tools) load Python modules from memory rather than a filesystem, the model file loading that botocore and boto3 do for the built-in models breaks when you attempt to load anything.

I've been digging into this, and the simplest solution I've been able to come up with to create a branching path in the model loaders: if running in (iirc) CPython 3.7+, we can use a `ResoureLoader` to load data files contained in a Python module. On older versions of CPython and for custom models in `$AWS_DATA_PATH` we would still need the file loader, but this would remove the need for the relative filepath loading that botocore/boto3 currently do.

The way I've been thinking this would manifest is to take `Loader`[2] and turn it into a wrapper on top of two new loaders: a `FileLoader` that works everywhere but only looks at the extra search paths, and (depending on Python version) either a `FileLoader` that does what the current `Loader` does but only for the built-in model files or a `MemoryLoader` that loads the built-in models directly from memory via `ResourceLoader`s.

There is one caveat to this: doing this would require sprinkling `__init__.py` files throughout all directories in `data`, to turn them into Python modules from which data can be loaded.

boto3 would need similar changes for the loaders and data files that it uses.

Is this a change that you all would be interested in? Any additional concerns that I should keep in mind if I start hacking on this?

[1] https://pyoxidizer.readthedocs.io
[2] https://botocore.amazonaws.com/v1/documentation/api/latest/reference/loaders.html?highlight=model#botocore.loaders.Loader.load_service_model

swetashre:
Thank you for the post. Currently we are not willing to make these changes unless there is a strong need for it.  
For now it is possible to create a self contained binary with PyInstaller. Here is the link : 
https://github.com/aws/aws-cli/blob/v2/scripts/installers/make-exe

PR #2046, opened 2020-05-13
Eliminate __file__ for importlib.resources

philipkimmey:
# Background
We have an internal tool that we distribute to all developers and we're looking to improve how we distribute that package.

Everything works fine with a `.pex`-style approach, but eliminating the `__file__`-oriented assumptions would allow us to use [PyOxidizer](https://pyoxidizer.readthedocs.io/en/stable/) which we hope will reduce the extra ~400ms of overhead that the pex-based approach introduces. (400ms isn't the end of the world but isn't ideal for a CLI tool our team is using all day.)

Here's a somewhat random grab bag of relevant links:

- https://github.com/boto/boto3/issues/1770
- https://github.com/boto/botocore/pull/1969
- https://github.com/certifi/python-certifi/pull/116
- https://pyoxidizer.readthedocs.io/en/latest/oxidized_importer_behavior_and_compliance.html#no-file

# Code

This currently this just contains failing tests that demonstrate the issue. I recognize the test code is pretty rough but couldn't come up with a more clever way to demonstrate the issue (since `botocore` will usually already be imported in the Python interpreter when this test is run.). (It's possible that the proposed [Subinterpreter](https://www.python.org/dev/peps/pep-0554/) stuff would make that possible but that's theoretical and I'm not familiar enough to know if it would actually even help.)

### DIFF
--- a/botocore/__init__.py
+++ b/botocore/__init__.py
@@ -53,9 +53,6 @@ _xform_cache = {
 # individual case.
 ScalarTypes = ('string', 'integer', 'boolean', 'timestamp', 'float', 'double')
 
-BOTOCORE_ROOT = os.path.dirname(os.path.abspath(__file__))
-
-
 # Used to specify anonymous (unsigned) request signature
 class UNSIGNED(object):
     def __copy__(self):
--- a/botocore/data/__init__.py
+++ b/botocore/data/__init__.py
@@ -1,0 +1,1 @@
+__path__ = __import__('pkgutil').extend_path(__path__, __name__)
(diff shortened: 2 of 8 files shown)


mattsb42-aws:
Some historical conversation on this if it helps: https://github.com/boto/botocore/issues/1950

mattsb42-aws:
It is still an issue: https://github.com/boto/botocore/search?q=__file__

tcyrus:
I'm not sure if that search is the best example, as the only code that should cause issues with bundling is the code in the `botocore` directory. That being said, [it's still an issue](https://github.com/boto/botocore/search?q=__file__+path%3Abotocore).

aBurmeseDev:
Thank you all for chiming in. This was brought up to discuss with the team recently and per team discussion, eliminating _file_ at the base level module might be a breaking change and would have to find a better approach to accomplish it. Closing for now.

PR #2091, opened 2020-06-29
Remove redirection of stderr when calling credential_process

### LINKED ISSUE #2090: Stop redirecting stderr when calling credential_process
Currently, when resolving credentials using a `credential_process` [you read both stdout and stderr](https://github.com/boto/botocore/blob/v2/botocore/credentials.py#L969). This means that the `credential_process` itself has no way of communicating with the user, which would be very useful to e.g. request user input (passwords, MFA tokens etc) or simply to provide debug logging.

I would like remove the redirection of stderr and let the `credential_process` communicate directly to the user via the console, which is [exactly whats done in v1 and v2 of the AWS SDK for Go](https://github.com/aws/aws-sdk-go-v2/blob/master/aws/processcreds/provider.go#L301-L303).

itsdalmo:
This PR closes #2090 by removing the redirection of stderr when spawning a new `subprocess` with the `credential_process` command. Since stderr will now be output to the console, we are including the exit code in the error message from `botocore`.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -964,13 +964,11 @@ class ProcessProvider(CredentialProvider):
         # We're not using shell=True, so we need to pass the
         # command and all arguments as a list.
         process_list = compat_shell_split(credential_process)
-        p = self._popen(process_list,
-                        stdout=subprocess.PIPE,
-                        stderr=subprocess.PIPE)
-        stdout, stderr = p.communicate()
+        p = self._popen(process_list, stdout=subprocess.PIPE)
+        stdout, _ = p.communicate()
         if p.returncode != 0:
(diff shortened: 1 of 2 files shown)


itsdalmo:
Just noticed the following open PRs that address the same problem:
- https://github.com/boto/botocore/pull/1835
- https://github.com/boto/botocore/pull/1349

tim-finnigan:
Thank you for creating this PR. This feature request is now being tracked here in our cross-SDK repository: https://github.com/aws/aws-sdk/issues/358. Implementations involving credential processing need to be considered across SDKs, so further review is required here. I will link this PR in that issue to reference this as a possible approach.

Issue #761, opened 2016-01-18
First-class support for AssumeRole in sessions


copumpkin:
This is similar to my request for https://github.com/boto/boto/issues/3381, but for the botocore credentials/session system.

When I first looked for this, I got my hopes up because I saw the `AssumeRoleProvider` in [credentials.py](https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L589), but then it turned out to be fairly awkward to use programmatically with dynamically specified role metadata, as it seemed to assume fairly deeply that you wanted to use it the way `aws-cli` does, via a config file and static credentials.

What I'd really like to see is composable/fully programmatic solution to this. It would probably use much of the same logic that's already in `AssumeRoleProvider`, except with fewer assumptions about where the `AssumeRole` metadata information is coming from, and an API that makes it easy to create new assumed sessions from existing ones.

For example, here's an API I might enjoy using:

```python
# Gives me some default session using default credentials that have power 
# to AssumeRole into other accounts
session = botocore.session.get_session() 

# 
session1 = session.assume_role_session('arn:aws:iam::1234567890:role/JumpRole', role_session_name='hostile-takeover')
session2 = session.assume_role_session('arn:aws:iam::1111111111:role/SomeOtherJumpRole', role_session_name='hostile-takeover')

# arn:aws:iam::1234567890:role/JumpRole has the power to itself assume a role
session3 = session1.assume_role_session('arn:aws:iam::222222222:role/JumpRole', role_session_name='hostile-takeover')

# Now we wait for a few hours
time.sleep(3600 * 5)

client3 = session3.create_client('ec2', region_name='us-west-2')

# This should work, and transparently refresh credentials as needed up the stack (in this case, two credential refreshes would be needed since we're two AssumeRoles deep)
print client3.describe_instances()
```

I don't really care much about the API specifics, but I do want the entire `AssumeRole` session information to be (at least optionally) programmatic, and not implicitly loaded from some config file. The current `AssumeRoleProvider` also expects a `source_profile` which makes it hard to stack these things as I show above. Ideally, this would also work nicely with other `AssumeRole` variants, but that's far less pressing for me.

cc @jamesls who I think wrote (or at least ported) `AssumeRoleProvider` in botocore.

copumpkin:
[...]

PR #2096, opened 2020-07-03
Added ProgrammaticAssumeRoleProvider and Session.assume_role() for #761

benkehoe:
Issue #761 asks for first-class support for sts:AssumeRole, where there is a `Session.assume_role()` method that produces another session (`boto3.Session` would then also get an `assume_role()` method). To implement this, the existing `credentials.AssumeRoleProvider` seemed too intimately tied to loading credentials from config, in particular there did not seem to be a good way of injecting an arbitrary `Credentials` object, so I added a new, simpler `ProgrammaticAssumeRoleProvider` class that mainly just wraps the `AssumeRoleCredentialFetcher`. I've replicated the unit tests for `AssumeRoleProvider` except those that relate to configuration files.

I'm hoping to get feedback on whether this is a contribution that can feasibly be accepted before I add the functional tests and go to the effort of figuring out how the integration testing works.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1625,6 +1625,78 @@ class AssumeRoleProvider(CredentialProvider):
             )
         return credentials
 
+class ProgrammaticAssumeRoleProvider(CredentialProvider):
+    METHOD = 'programmatic-assume-role'
+
+    def __init__(self, client_creator, source_credentials, role_arn,
+            extra_args=None, prompter=getpass.getpass, cache=None, expiry_window_seconds=None):
+        """
+        :type client_creator: callable
+        :param client_creator: A factory function that will create
+            a client when called.  Has the same interface as
+            ``botocore.session.Session.create_client``.
+
+        :type source_credentials: Credentials
+        :param source_credentials: The credentials to use to create the
+            client for the call to AssumeRole.
+
+        :type role_arn: str
+        :param role_arn: The ARN of the role to be assumed.
+
+        :type extra_args: dict
+        :param extra_args: Any additional arguments to add to the assume
+            role request using the format of the botocore operation.
+            Possible keys include, but may not be limited to,
+            DurationSeconds, Policy, and RoleSessionName.
+
+        :type prompter: callable
+        :param prompter: A callable that returns input provided
+            by the user (i.e raw_input, getpass.getpass, etc.).
+
+        :type cache: dict
+        :param cache: An object that supports ``__getitem__``,
+            ``__setitem__``, and ``__contains__``.  An example of this is
+            the ``JSONFileCache`` class in aws-cli.
+
+        :type expiry_window_seconds: int
+        :param expiry_window_seconds: The amount of time, in seconds,
+        """
+
+        self._client_creator = client_creator
+        self._source_credentials = source_credentials
+        self._role_arn = role_arn
+        self._extra_args = extra_args
+        self._prompter = prompter
+        if cache is None:
+            cache = {}
+        self._cache = cache
+        self._expiry_window_seconds = expiry_window_seconds
+
+        self._fetcher = None
+
+
+    def _get_fetcher(self):
+        if not self._fetcher:
+            self._fetcher = AssumeRoleCredentialFetcher(
+                self._client_creator,
+                self._source_credentials,
+                self._role_arn,
+                extra_args=self._extra_args,
+                mfa_prompter=self._prompter,
+                cache=self._cache,
+                expiry_window_seconds=self._expiry_window_seconds
+            )
+        return self._fetcher
+
+    def load(self):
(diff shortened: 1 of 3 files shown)


benkehoe:
Pinging on this. Would love to make my first contribution to the Python SDK!

tim-finnigan:
Hi @benkehoe, thank you for the PR and sorry for the wait. I discussed this with the team and will share some points that came up. 

Currently we don’t provide public interfaces for any CredentialProviders in botocore. To support that would require a substantial refactoring of the codebase.

The implementation here doesn’t follow our existing CredentialProvider workflows and involves replacing the existing credential chain with an overwritten one. That makes the session inoperable for any credentials except the ARN used to assume the role. The current change also doesn't account for the use of profiles, which means the new session will be broken for common configuration cases.

For those reasons we think that currently it makes more sense for this to exist as a third-party plugin. But we may revisit this in a future refactoring or major version.

benkehoe:
Thanks for following up. It's true that botocore doesn't today provide programmatic credential definitions, but this is generally not true of other SDKs, e.g., [JavaScript (even in v2)](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/ChainableTemporaryCredentials.html). It's unfortunate that Python doesn't get the same capability.

For the record, though, I wanted to respond to this part:

> The implementation here doesn’t follow our existing CredentialProvider workflows and involves replacing the existing credential chain with an overwritten one. That makes the session inoperable for any credentials except the ARN used to assume the role. The current change also doesn't account for the use of profiles, which means the new session will be broken for common configuration cases.

This is intentional. The explicit purpose of this credential provider is to provide a programmatic way to get a *new* session that uses an assumed role based on a parent session. This new session should not be able to pick up any other credentials. Such a session can only be created through Python code, not through configuration (there's already a credential provider for that). For that reason it uses a `CredentialResolver` that only uses the programmatic assumed role credential provider. It is similar to how [a boto3 Session created with explicit credentials](https://github.com/boto/boto3/blob/5d44d5b2b47c88084507dbac7a40ae9ca529b559/boto3/session.py#L80) will not ever use the credential resolver. Would you be more amenable to this change if `Session.assume_role()` created the `DeferredRefreshableCredentials` directly and set them on the newly-created child session's `Session._credentials` field?

tim-finnigan:
Hi @benkehoe thanks for sharing those points. Unfortunately the team is still not considering the proposed changes at this time. We could make changes to more tightly integrate this with the existing interfaces, but this would still be a special case feature that we’d need to maintain outside of the normal credential flow.

Some SDKs may have some form of this feature as you mentioned, but generally speaking we are trying to move toward more standardization of feature implementation across SDKs. Adding another different approach here in botocore would take us further from alignment. 

But we will continue recommending your [aws-assume-role-lib](https://github.com/benkehoe/aws-assume-role-lib) library to those who could benefit from using it. Also this PR prompted a broader team discussion on how boto3 could better support/incorporate third-party libraries and plugins such as that one. If you have any thoughts to share regarding that please let us know.

benkehoe:
> we are trying to move toward more standardization of feature implementation across SDKs.

As part of that, please, *please* include programmatic configuration as part of that.

I have one last-ditch proposal. What if we added a new `CredentialProvider` subclass to the default chain, one that read from the session config and relied on the configuration values to present, e.g., `role_arn`, but instead of `source_profile`, it had to be `source_credentials_object` and be a `Credential` object? So creating an assumed role session would be like this:

```
parent_session = botocore.Session()
config = {
  "role_arn": "arn:aws:...",
  "source_credentials_object": parent_session.get_credentials()
}
assumed_role_session = botocore.Session(session_vars=config)
```
This credential provider would never get activated by existing config, and this config would never activate existing credential providers. Then that session creation would get wrapped up in a `Session.assume_role()` method.

My concern with AWS recommending 3rd party libraries like `aws-assume-role-lib` and [`requests-aws4auth`](https://github.com/tedder/requests-aws4auth) (because botocore doesn't expose request signing, see #1784) is that it encourages broader software supply chains for code that directly touches credentials. Like, in the absence of botocore stepping up to the plate, I guess it's the next best thing, but I'd much rather be able to recommend that people limit their dependencies for SDK credential handling to code coming directly from AWS. But in general, an AWS-curated list of plugin/extensions/etc. (with a healthy warning label) would be a good thing, I think.

benkehoe:
I've been a fool. This implementation is needlessly overcomplicated, because I missed how static credentials are used in a botocore Session. If a `boto3.Session` is allowed to set the `_credentials` field of a `botocore.Session`, I don't need any changes in botocore at all. If not, all I need is a change to the `set_credentials()` method to allow a `Credentials` object directly. I am going to close this PR, implement it that way on boto3, and open a PR there for further discussion.

PR #2115, opened 2020-07-31
Add some waiters for workspaces #2527

littlejo:
I need this feature:

https://github.com/boto/boto3/issues/2527


stobrien89:
Hi @littlejo,

Thanks again for the request. We are now submitting requests for paginators and waiters upstream so that all AWS SDKs benefit from a consistent implementation. I will pass on the request for these waiters to the service team. We will also update the guidance in the contributing guide to better reflect this policy. I've added an issue to track this here:

aws/aws-sdk#79

PR #2121, opened 2020-08-06
i#2120: Bump docutils to <0.17 due to conflict

dchichkov:


prometheanfire:
ping, any updates on this?

jurgenhaas:
I'd like to see this committed too.

prometheanfire:
more pings?

spackard:
I'd like to see the upper cap removed if it completely arbitrary. It should be set when there's a known backwards incompatibility issue. My 2 cents. Failing that, I'd like to see this merged as well.

prometheanfire:
> I'd like to see the upper cap removed if it is completely arbitrary. It should be set when there's a known backwards incompatibility issue. My 2 cents. Failing that, I'd like to see this merged as well.

I've tried to get them to do this, but upstream seems dead (all commits are from the bot only).

erdseb (approved):

erdseb:
@nateprewitt you seem to have write permission at `master`. We need your help here, do you have a minute?

dchichkov:
GitHub says - "Only those with write access to this repository can merge pull requests."
I assume you'll click merge (& no action items for me).

nateprewitt:
Hi everyone. I've already looked at this but it's is a breaking change due to [0.16](https://docutils.sourceforge.io/RELEASE-NOTES.html#release-0-16-2020-01-12) removing the `handles_io_errors` parameter from their API. Given docutils chose to make a breaking change on a non-major version, we have to treat this a strict dependency. We can start to get around this by moving our lower bound for Sphinx to 1.3.2 but we have a tertiary issue that `pip` will soon be releasing dependency resolution, breaking our Sphinx theme if we don't use Sphinx <1.3.

This is resolvable, but isn't as straight forward as the PR suggests. Pinning to a < 0.16 is a requirement for the project for the moment, and we're working on getting together some other work to get us to a state where things can upgrade. Thanks for your patience!

prometheanfire:
One option that should be explored is gating the docs stuff behind a flag (see all the flags used in cinder here).

https://github.com/openstack/cinder/blob/master/setup.cfg#L78-L118

prometheanfire:
imo, those users should have docutils in their requirements.  I'm surprised they have not broken already if they don't directly track their dependencies.

erdseb:
> Hi everyone. I've already looked at this but it's a breaking change due to [0.16](https://docutils.sourceforge.io/RELEASE-NOTES.html#release-0-16-2020-01-12) removing the `handles_io_errors` parameter from their API. Given docutils chose to make a breaking change on a non-major version, we have to treat this a strict dependency. We can start to get around this by moving our lower bound for Sphinx to 1.3.2 but we have a tertiary issue that `pip` will soon be releasing dependency resolution, breaking our Sphinx theme if we don't use Sphinx <1.3.
> 
> This is resolvable, but isn't as straight forward as the PR suggests. Pinning to a < 0.16 is a requirement for the project for the moment, and we're working on getting together some other work to get us to a state where things can upgrade. Thanks for your patience!

@nateprewitt thanks for these details! From a quick look I find no use of `handles_io_errors` in here or in sphinx. Could you elaborate how that removal affects the project and how the new dependency resolution breaks the theme with Sphinx >=1.3?

> @prometheanfire thanks for the suggestion! An `extra` would definitely be an option with a fresh project, but it gets complicated since some users are currently relying on the default botocore install to include docutils. Removing that will be a breaking change.

Interesting. I think the case can be argued but I get your point. Is bumping the mayor version from 0 to 1 an option here to communicate a move like that?

nateprewitt:
> @nateprewitt thanks for these details! From a quick look I find no use of handles_io_errors in here or in sphinx. Could you elaborate how that removal affects the project and how the new dependency resolution breaks the theme with Sphinx >=1.3?

It's Sphinx itself that used this prior to Sphinx 1.3.2 (sphinx-doc/sphinx#1834). Our theme requires <1.3, hence the problem with docutils 0.17. We could technically get around this now, but pip will start enforcing dependency resolution in the next major version which means we'll no longer be able to resolve.

> Interesting. I think the case can be argued but I get your point. Is bumping the mayor version from 0 to 1 an option here to communicate a move like that?

We don't have control over docutils unfortunately, so we're at the mercy of their versioning choices. It would take a *very* unique case for Botocore to get a major version because of a dependency. Like I said, I think we have a path forward, it may take a little time though to get everything in place. We are looking into getting this fixed though.

spackard:
My 2 cents on the topic is that if upstream dependencies are not expressing their direct dependencies explicitly and correctly and relying on this package to bring in a transitive dependency, then they deserve to be broken. I do not feel like removing a runtime dependency that does not actually exist and is only there for CI/development/documentation requires a major version bump, though an advisement in the change log wouldn't hurt.

nateprewitt:
This issue should now be resolved in Botocore 1.18.0.

PR #2123, opened 2020-08-08
Don't utf-8 encode header values on python3.

jalopezsilva:
With ``urllib3`` version 1.25.10 we're running into an issue as botocore is accidentally passing header values as bytes. The bytes are originating from ``_encode_headers`` in ``endpoint.py`` in which we are ``utf-8`` encoding ``str``s unnecessarily. ``str``s are already ``utf-8`` encoded in Python3 by default. 

The code that has the bug is meant to make sure python2 ``unicode`` type is transformed into str. We're naively checking ``six.text_type`` vs the value which will always be true. 

The test adds a repro of the problem. The stack trace I'm seeing in urllib3 is the following:

```
File "botocore/httpsession.py", line 283, in send
    chunked=self._chunked(request.headers),
  File "urllib3/connectionpool.py", line 687, in urlopen
    chunked=chunked,
  File "urllib3/connectionpool.py", line 400, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/urllib3/connection.py", line 215, in request
    elif headers["user-agent"] == SUPPRESS_USER_AGENT:
  File "urllib3/_collections.py", line 157, in __getitem__
    return ", ".join(val[1:])
TypeError: sequence item 0: expected str instance, bytes found
```

### DIFF
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -120,7 +120,7 @@ class Endpoint(object):
     def _encode_headers(self, headers):
         # In place encoding of headers to utf-8 if they are unicode.
         for key, value in headers.items():
-            if isinstance(value, six.text_type):
+            if not isinstance(value, str):
                 headers[key] = value.encode('utf-8')
 
     def prepare_request(self, request):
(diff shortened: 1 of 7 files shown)


jalopezsilva:
After syncing with @nateprewitt and the AWS team, we concluded we want to keep the byte representation in botocore. The urllib3 issue was properly fixed in https://github.com/urllib3/urllib3/issues/1930.

PR #2127, opened 2020-08-14
Add basic region_name validation

nateprewitt:
This PR will add some basic validation to the region_name parameter used to create new Session and BaseClient objects. The goal is to avoid accidentally setting up clients that will either never route, or lead to unexpected request routing. The change here will bubble up to boto3 in clients and resources, as well as the AWS CLI.

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -20,6 +20,7 @@ import copy
 import logging
 import os
 import platform
+import re
 import socket
 import warnings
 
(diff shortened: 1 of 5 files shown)


botocore/session.py
```diff
@@ -858,6 +863,15 @@ def _resolve_region_name(self, region_name, config):
         self._last_client_region_used = region_name
         return region_name
 
+    def _validate_region_name(self, region_name):
+        """Provided region_name must be a valid host label."""
+        if region_name is None:
+            return
+        valid_host_label = re.compile(r'^(?![0-9]+$)(?!-)[a-zA-Z0-9-]{,63}(?<!-)$')
```
joguSD, line 870: Do you think it makes sense to hoist this up to the class? Or maybe even create a utility function for doing host label validation?

joguSD:
Seems fine, just had one small comment.

botocore/session.py
nateprewitt: Yeah, I was originally thinking about having it live in utils but didn't know if we wanted it public. I'll move it over there really quick.

PR #2132, opened 2020-08-24
Update Endpoint Discovery to enable for required operations

nateprewitt:
When using services with endpoint discovery as a requirement, botocore currently throws an exception by default. This was originally intended to prompt users to opt in for this feature. As more services start to use this, we're switching the default to use endpoint_discovery automatically when required and opt-out when it's optional. This will maintain backwards compatibility, but allow users to not have to worry about configuration in the common case.

You can still explicitly disable this behavior with `endpoint_discovery_enabled=False` in your Config object, or with the `AWS_ENDPOINT_DISCOVERY_ENABLED` environment variable. Similarly, it can be used in all requests where endpoint discovery is available (optional or required) when explicitly set to `True`.

### DIFF
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -14,22 +14,22 @@ import logging
 import functools
 
 from botocore import waiter, xform_name
+from botocore.args import ClientArgsCreator
 from botocore.auth import AUTH_TYPE_MAPS
 from botocore.awsrequest import prepare_request_dict
 from botocore.docs.docstring import ClientMethodDocstring
 from botocore.docs.docstring import PaginatorDocstring
-from botocore.exceptions import ClientError, DataNotFoundError
-from botocore.exceptions import OperationNotPageableError
-from botocore.exceptions import UnknownSignatureVersionError
+from botocore.exceptions import (
+    ClientError, DataNotFoundError, OperationNotPageableError,
+    UnknownSignatureVersionError, InvalidEndpointDiscoveryConfigurationError
+)
 from botocore.hooks import first_non_none_response
 from botocore.model import ServiceModel
 from botocore.paginate import Paginator
-from botocore.utils import CachedProperty
-from botocore.utils import get_service_module_name
-from botocore.utils import S3RegionRedirector
-from botocore.utils import S3ArnParamHandler
-from botocore.utils import S3EndpointSetter
-from botocore.args import ClientArgsCreator
+from botocore.utils import (
+    CachedProperty, get_service_module_name, S3RegionRedirector,
+    S3ArnParamHandler, S3EndpointSetter, ensure_boolean
+)
 from botocore import UNSIGNED
 # Keep this imported.  There's pre-existing code that uses
 # "from botocore.client import Config".
@@ -197,13 +197,34 @@ class ClientCreator(object):
         elif self._config_store:
             enabled = self._config_store.get_config_variable(
                 'endpoint_discovery_enabled')
-        if enabled:
-            manager = EndpointDiscoveryManager(client)
+
+        enabled = self._normalize_endpoint_discovery_config(enabled)
+        if enabled and self._requires_endpoint_discovery(client, enabled):
+            discover = enabled is True
+            manager = EndpointDiscoveryManager(client, always_discover=discover)
             handler = EndpointDiscoveryHandler(manager)
             handler.register(events, service_id)
         else:
             events.register('before-parameter-build',
                             block_endpoint_discovery_required_operations)
+
+    def _normalize_endpoint_discovery_config(self, enabled):
+        """Config must either be a boolean-string or string-literal 'auto'"""
+        if isinstance(enabled, str):
+            enabled = enabled.lower().strip()
+            if enabled == 'auto':
+                return enabled
+            elif enabled in ('true', 'false'):
+                return ensure_boolean(enabled)
+        elif isinstance(enabled, bool):
+            return enabled
+
+        raise InvalidEndpointDiscoveryConfigurationError(config_value=enabled)
+
+    def _requires_endpoint_discovery(self, client, enabled):
+        if enabled == "auto":
+            return client.meta.service_model.endpoint_discovery_required
+        return enabled
 
     def _register_s3_events(self, client, endpoint_bridge, endpoint_url,
                             client_config, scoped_config):
--- a/botocore/discovery.py
+++ b/botocore/discovery.py
@@ -16,6 +16,7 @@ import weakref
 
 from botocore import xform_name
 from botocore.exceptions import BotoCoreError, HTTPClientError, ConnectionError
+from botocore.model import OperationNotFoundError
 from botocore.utils import CachedProperty
 
 logger = logging.getLogger(__name__)
@@ -59,8 +60,11 @@ class EndpointDiscoveryModel(object):
         return keys
 
     def discovery_required_for(self, operation_name):
-        operation_model = self._service_model.operation_model(operation_name)
-        return operation_model.endpoint_discovery.get('required', False)
+        try:
+            operation_model = self._service_model.operation_model(operation_name)
+            return operation_model.endpoint_discovery.get('required', False)
+        except OperationNotFoundError:
+            return False
 
     def discovery_operation_kwargs(self, **kwargs):
         input_keys = self.discovery_operation_keys
@@ -87,7 +91,7 @@ class EndpointDiscoveryModel(object):
 
 
 class EndpointDiscoveryManager(object):
-    def __init__(self, client, cache=None, current_time=None):
+    def __init__(self, client, cache=None, current_time=None, always_discover=True):
         if cache is None:
             cache = {}
         self._cache = cache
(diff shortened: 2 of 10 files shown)


tests/functional/models/awsfooservice/2020-08-20/service-2.json
```diff
@@ -0,0 +1,121 @@
...
+                "apiVersion": "2018-08-31",
+                "endpointPrefix": "fooendpoint",
+                "jsonVersion": "1.1",
+                "protocol": "json",
+                "serviceAbbreviation": "FooService",
+                "serviceId": "FooService",
+                "serviceFullName": "AwsFooService",
+                "signatureVersion": "v4",
+                "signingName": "awsfooservice",
+                "targetPrefix": "awsfooservice"
```
kyleknap, line 13: Let's rename this to test-discovery-endpoint or something similar to better indicate why we need the model.

tests/functional/test_discovery.py
```diff
@@ -0,0 +1,277 @@
...
+        with http_stubber as stubber:
+            stubber.add_response(status=200, body=json.dumps(response).encode())
+            stubber.add_response(status=200, body=b'{}')
+            client.test_discovery_required(Foo="bar")
+            self.assertEqual(len(stubber.requests), 2)
+            self.assertEqual(
+                stubber.requests[1].url, "https://discovered.domain"
+            )
+
+    def test_endpoint_discovery_enabled_with_random_string(self):
```
kyleknap, line 270: We should also add test cases where we enable/disable it using the shared config file or environment variables to make sure we handle value conversion of string to boolean correctly.

tests/functional/test_discovery.py
```diff
@@ -0,0 +1,277 @@
...
+            }]
+        }
+        with http_stubber as stubber:
+            stubber.add_response(status=200, body=json.dumps(response).encode())
+            stubber.add_response(status=200, body=b'{}')
+            client.test_discovery_optional(Foo="bar")
+            self.assertEqual(len(stubber.requests), 2)
+            self.assertEqual(
+                stubber.requests[1].url, "https://discovered.domain"
+            )
```
kyleknap, line 248: It might make sense to make helper assertions to make it a bit easier to understand the assertions (e.g `assert_endpoint_used`)

botocore/discovery.py
```diff
@@ -87,14 +91,15 @@ def _gather_ids(self, shape, params, ids=None):
...
-    def __init__(self, client, cache=None, current_time=None):
+    def __init__(self, client, cache=None, current_time=None, strict=True):
         if cache is None:
             cache = {}
         self._cache = cache
         self._failed_attempts = {}
         if current_time is None:
             current_time = time.time
         self._time = current_time
+        self._strict = strict
```
kyleknap, line 102: It might be worth renaming it to something else. Typically, I associate strict with causing validation errors. Maybe call it `always_discover`?

tests/functional/test_discovery.py
```diff
@@ -0,0 +1,371 @@
...
+        with http_stubber as stubber:
+            self.add_describe_endpoints_response(stubber, discovered_endpoint)
+            client.test_discovery_optional(Foo="bar")
+            self.assertEqual(
+                os.environ.get('AWS_ENDPOINT_DISCOVERY_ENABLED'), "True"
+            )
+            self.assert_endpoint_discovery_used(stubber, discovered_endpoint)
+
+    def test_endpoint_discovery_required_with_env_var_disabled(self):
+        self.environ['AWS_ENDPOINT_DISCOVERY_ENABLED'] = "False"
```
kyleknap, line 326: For one of these true/false cases, it would be interesting if we swiched the casing to lowercase `false` or `true` to make sure those hold up.

tests/functional/test_discovery.py
```diff
@@ -0,0 +1,371 @@
...
+            service_name="test-discovery-endpoint", config=config
+        )
+        with http_stubber as stubber:
+            self.add_describe_endpoints_response(stubber, discovered_endpoint)
+            client.test_discovery_optional(Foo="bar")
+            self.assert_endpoint_discovery_used(stubber, discovered_endpoint)
+
+    def test_endpoint_discovery_manual_auto_on_required_endpoint(self):
+        discovered_endpoint = 'https://discovered.domain'
+        config = Config(endpoint_discovery_enabled=" aUto  \n")
```
kyleknap, line 249: It seems weird that we would support this casing with whitespaces especially when we are providing it through the `Config` object. I think it may make sense to not try to do any `lower` or `strip` for the auto value. I do not feel too strongly about this though.

tests/functional/test_discovery.py
```diff
@@ -0,0 +1,371 @@
...
+            client.test_discovery_required(Foo="bar")
+            self.assert_endpoint_discovery_used(stubber, discovered_endpoint)
+
+    def test_endpoint_discovery_enabled_with_random_string(self):
+        config = Config(endpoint_discovery_enabled="bad value")
+        client, http_stubber = self.create_client(
+            service_name="test-discovery-endpoint", config=config
+        )
+        http_stubber.add_response(status=200, body=b'{}')
+        with self.assertRaises(EndpointDiscoveryRequired):
```
kyleknap, line 264: I feel like this should instead be a validation error for a bad configuration value instead.

botocore/client.py
```diff
@@ -197,14 +196,27 @@ def _register_endpoint_discovery(self, client, endpoint_url, config):
         elif self._config_store:
             enabled = self._config_store.get_config_variable(
                 'endpoint_discovery_enabled')
-        if enabled:
-            manager = EndpointDiscoveryManager(client)
+        if enabled and self._requires_endpoint_discovery(client, enabled):
+            discover = ensure_boolean(enabled)
```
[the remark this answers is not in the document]
nateprewitt: Auto is equivalent to false. `ensure_boolean` checks for an actual boolean value or if it matches a string "true". "auto" will never match that.

tests/functional/test_discovery.py
```diff
@@ -0,0 +1,371 @@
...
+        config = Config(endpoint_discovery_enabled=True)
+        client, http_stubber = self.create_client(
+            service_name="test-discovery-endpoint",
+            config=config,
+            endpoint_url=endpoint
+        )
+        with http_stubber as stubber:
+            stubber.add_response(status=200, body=b'{}')
+            client.test_discovery_required(Foo="bar")
+            self.assert_discovery_skipped(
```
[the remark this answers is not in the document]
nateprewitt: This is a requirement for the feature, if we specify an endpoint it needs to be used as the final endpoint.

tests/functional/test_discovery.py
```diff
@@ -0,0 +1,371 @@
...
+        client, http_stubber = self.create_client(
+            service_name="test-discovery-endpoint", config=config
+        )
+        http_stubber.add_response(status=200, body=b'{}')
+        with self.assertRaises(EndpointDiscoveryRequired):
+            client.test_discovery_required(Foo="bar")
+
+    def test_endpoint_discovery_required_with_env_var_default(self):
+        environ = os.environ.copy()
+        with mock.patch('os.environ', environ):
```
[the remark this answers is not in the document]
nateprewitt: Good catch, that's hold over from earlier.

tests/functional/test_discovery.py
```diff
@@ -0,0 +1,371 @@
...
+            discovered_endpoint = 'https://discovered.domain'
+            client, http_stubber = self.create_client(
+                service_name="test-discovery-endpoint", config=None
+            )
+            with http_stubber as stubber:
+                self.add_describe_endpoints_response(stubber, discovered_endpoint)
+                client.test_discovery_required(Foo="bar")
+                self.assertEqual(
+                    os.environ.get('AWS_ENDPOINT_DISCOVERY_ENABLED'), 'auto'
+                )
```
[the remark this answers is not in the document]
nateprewitt: Sure, I can pull these.

tests/functional/test_discovery.py
nateprewitt: So I could move this to the config file, but I think it's a pretty common case to get weird spaces or extra return character in config files. I just want to assert we're doing case normalization. If we can't do an enum for this, we should be doing out best to avoid silly failures.

tests/functional/test_discovery.py
nateprewitt: I agree this could be more direct, this was hold over from the previous behavior for bad values. Let me see if I can add some better handling.

tests/functional/test_discovery.py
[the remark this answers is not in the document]
kyleknap: That's fair. Thinking it over more, it would definitely be even more strange if the SDK actually tried endpoint discovery and ended up using a different endpoint than what was specified. Also there is no purpose to doing endpoint discovery if you already know what endpoint you want to use and explicitly specify it.

tests/functional/test_discovery.py
kyleknap: Ok I'm fine with keeping it as is then.

PR #2134, opened 2020-08-26
Remove dependency on nose.

mcepl:
Mostly independent of anything outside of the standard library, except I need pytest for marking tests as slow. If these marks can be sacrificed, there would be no new dependency.


mcepl:
OK, no idea, what’s wrong with ``functional/test_regions.py::TestRegions::test_known_endpoints``, but it doesn't seem like something caused by the switch from nose to pytest.

tests/functional/test_cognito_idp.py
```diff
@@ -114,7 +111,5 @@ def run(self):
             operation(**self._parameters)
             request = self._http_stubber.requests[0]
 
-        assert_false(
-            'authorization' in request.headers,
+        assert 'authorization' not in request.headers, \
             'authorization header found in unsigned operation'
```
nateprewitt, line 115: We can probably make this a `failure_msg` param or something similar to avoid the line wrap.

tests/functional/test_credentials.py
```diff
@@ -826,7 +826,7 @@ def test_credential_process_returns_error(self):
             # Finally `(?s)` at the beginning makes dots match newlines so
             # we can handle a multi-line string.
             reg = r"(?s)^((?!b').)*$"
-            with self.assertRaisesRegexp(CredentialRetrievalError, reg):
+            with six.assertRaisesRegex(self, CredentialRetrievalError, reg):
```
nateprewitt, line 829: Let's use [`pytest.raises`](https://docs.pytest.org/en/stable/reference.html#pytest-raises), and we can pass the regex as the `match` parameter. That'll avoid us needing to import six here.

tests/functional/test_history.py
```diff
@@ -87,10 +87,10 @@ def test_does_record_http_request(self):
         self.assertIsNone(body)
 
         streaming = payload['streaming']
-        self.assertEquals(streaming, False)
+        self.assertEqual(streaming, False)
```
nateprewitt, line 90: Let's just use plain `assert x == y` here.

tests/functional/test_paginate.py
```diff
@@ -217,22 +215,22 @@ def test_token_with_triple_underscores(self):
...
-
-
-def assert_token_encodes_and_decodes(token_dict):
-    encoded = TokenEncoder().encode(token_dict)
-    assert isinstance(encoded, six.string_types)
-    decoded = TokenDecoder().decode(encoded)
-    assert_equal(decoded, token_dict)
+class TestTokenEncoding(unittest.TestCase):
+    def test_token_encoding(self):
+        cases = [
```
nateprewitt, line 220: If we're gonna shift this, it's a good candidate for `pytest.parametrize` too.

tests/functional/test_paginate.py
```diff
@@ -217,22 +215,22 @@ def test_token_with_triple_underscores(self):
...
+        ]
+
+        for token_dict in cases:
+            self.assert_token_encodes_and_decodes(token_dict)
+
+    def assert_token_encodes_and_decodes(self, token_dict):
+        encoded = TokenEncoder().encode(token_dict)
+        assert isinstance(encoded, six.string_types)
+        decoded = TokenDecoder().decode(encoded)
+        self.assertEqual(decoded, token_dict)
```
nateprewitt, line 236: `assert x == y`

tests/unit/test_awsrequest.py
```diff
@@ -20,7 +20,10 @@
 import socket
 import sys
 
-from mock import Mock, patch
+try:
+    from mock import Mock, patch
```
nateprewitt, line 24: I think you've already added this in test. Let's consolidate this to `from test.mock import Mock, patch`.

tests/unit/test_eventstream.py
```diff
@@ -301,8 +297,13 @@ def test_all_positive_cases():
 def test_negative_cases():
     """Test that all negative cases raise the expected exception. """
     for (encoded, exception) in NEGATIVE_CASES:
-        test_function = raises(exception)(check_message_decodes)
-        yield test_function, encoded, None
+        try:
+            check_message_decodes(encoded, None)
```
nateprewitt, line 301: We may want to be using something like:
```python
with pytest.raises(exception):
    check_message_decodes(encoded, None)
```

That will be a bit easier to understand.

tests/unit/test_eventstream.py
```diff
@@ -455,14 +457,19 @@ def test_event_stream_wrapper_iteration():
...
 
-@raises(EventStreamError)
 def test_eventstream_wrapper_iteration_error():
-    raw_stream = create_mock_raw_stream(ERROR_EVENT_MESSAGE[0])
-    parser = Mock(spec=EventStreamXMLParser)
-    parser.parse.return_value = {}
-    output_shape = Mock()
-    event_stream = EventStream(raw_stream, output_shape, parser, '')
-    list(event_stream)
+    try:
```
nateprewitt, line 461: This should also done with a `with pytest.raises()` context manager.

tests/functional/test_client_class_names.py
```diff
@@ -69,13 +67,10 @@
...
-
-
-def _assert_class_name_matches_ref_class_name(client, ref_class_name):
-    assert_equal(client.__class__.__name__, ref_class_name)
+class TestClientClassNames(unittest.TestCase):
+    def test_client_has_correct_class_name(self):
+        session = botocore.session.get_session()
+        for service_name in SERVICE_TO_CLASS_NAME:
+            client = session.create_client(service_name, REGION)
+            self.assertEqual(client.__class__.__name__,
```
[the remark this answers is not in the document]
mcepl: Well, I didn’t want to change `unittest` based tests, pytest support them without problems anyway, and they don’t create any additional dependency, so if anybody wants to port those to pytest, they can do it in the next iteration.

scripts/ci/run-integ-tests
```diff
@@ -16,5 +16,4 @@ def run(command):
     return check_call(command, shell=True)
 
 
-run('nosetests --with-xunit --cover-erase --with-coverage '
-    '--cover-package botocore --cover-xml -v integration')
+run('pytest -v --junit-xml=test_results.xml --cov=botocore --cov-report=xml:coverage.xml -v integration/')
```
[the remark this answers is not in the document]
mcepl: Yup, that’s a bug. Fixed in 6abc5ac9.

tests/functional/csm/test_monitoring.py
```diff
@@ -50,7 +49,7 @@ class NonRetryableException(Exception):
 def test_client_monitoring():
     test_cases = _load_test_cases()
     for case in test_cases:
-        yield _run_test_case, case
+        _run_test_case(case)
```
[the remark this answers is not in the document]
mcepl: Again, switching to `parametrize` (and diving into deeper bowels of pytest generally) should be left as an exercise for future enhancement. Of course, I expect this PR to be thoroughly tested, but I believe it does exactly the same (albeit, not on background, so perhaps the speed of running the test suite will hurt slightly; I haven’t observed anything horrible when running it locally).

tests/functional/test_cognito_idp.py
```diff
@@ -95,8 +93,7 @@ def test_unsigned_operations():
         client = session.create_client('cognito-idp', 'us-west-2')
 
         for operation, params in operation_params.items():
-            test_case = UnsignedOperationTestCase(client, operation, params)
-            yield test_case.run
+            UnsignedOperationTestCase(client, operation, params).run()
```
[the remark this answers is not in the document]
mcepl: Same answer. Yes, I believe it does exactly the same (I don’t understand the point of that variable in the first place).

tests/functional/test_paginate.py
mcepl: I don’t think it matters, but changed in e8736768.

tests/unit/test_awsrequest.py
mcepl: Changed in commit 014a4657.

tests/unit/test_eventstream.py
mcepl: I didn’t want to add any more direct pytest dependency, than what’s absolutely necessary (I cannot mark tests as slow with unittest only), so that you won’t get to the same situation when pytest dies as nose just did.

tests/unit/test_eventstream.py
mcepl: Ditto.

tests/functional/test_paginate.py
mcepl: Ditto.

tests/functional/csm/test_monitoring.py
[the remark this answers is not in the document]
nateprewitt: So I did a quick check to verify I wasn't missing something. The implications of this change is we collapse 13 tests into one test. If any of those cases fail, we only get output for the first failure and all others are masked until we fix the first one. That will make debugging pretty painful, especially if a change introduces multiple issues.

tests/functional/test_client_class_names.py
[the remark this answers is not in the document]
nateprewitt: `assert` isn't specific to pytest, but would allow us to be consistent and avoid any dependency, including unittest. I agree this can be addressed later though.

tests/unit/test_eventstream.py
nateprewitt: I think if we're gonna rewrite the test, we should try to write the final version we want. We already have a hard dependency on pytest for the test suite now, so let's use it here if needed.

mcepl:
OK, this is getting really taking my time (which I don’t have). Could you just take this pull request (into separate branch I guess) as it is, and finish it according to your needs, please? Any rebasing is perfectly OK with me as long as I get at least one commit in your master branch in the end.

nateprewitt:
Alright, no worries. Thanks for you work @mcepl! I've changed to target branch to `pytest_migration` and we'll track the final deliverable there. We'll likely squash the commit list but maintain your original commit if that's OK. I'll leave this sitting until tomorrow and merge then if you don't have any other comments/questions.

glaubitz:
This patch unfortunately doesn't apply anymore - even with fuzz - meaning that I currently cannot update botocore to the latest upstream version in openSUSE.

mcepl:
@glaubitz Wasn’t this supposed to be merged on 2020-09-02 already?

glaubitz:
No, it was merged into boto:pytest_migration which is still a separate branch.

Your patch largely applies, but there are many new places where the old ```self.assertEqual```, ```self.assertTrue``` and so on are being used. I already started a manual rebase but noticed the many new instances of the old asserts so I gave up because it would take several hours to get these all sorted out.

mcepl:
@glaubitz Couldn’t you rebase `boto:pytest_migration` and apply it as a patch?

glaubitz:
I don't think it will rebase without issues. The problem is simply that there are new instances of ```self.assertEqual``` etc now.

mcepl:
There is nothing wrong with `self.assertEqual`, and when somebody doesn't bother to review and merge my patch in a year, I don't think it is my business to rebase it for them all the time.

Issue #2120, opened 2020-08-06
Cap on docutils < 0.16 prevents co-installability


dchichkov:
**Describe the bug**
I observe the following dependency conflict: 
  botocore 1.17.37 requires docutils<0.16,>=0.10, but you'll have docutils 0.16 which is incompatible.

Strangely, pipdeptree doesn't show constraints on docutils that prevent it from going to 0.15.  
But for some reason the constraints solver keeps docutils at 0.16.0 (current release on pip)

**Expected behavior**
Is constraint to <0.16 arbitrary?  It seems this constraint was placed there by:
https://github.com/boto/botocore/commit/a8bbf60b665efee53f5a5d2cbbeb6e5e9e0838f2

And constraining docutils was problematic before.

dchichkov:
Associated pull request - https://github.com/boto/botocore/pull/2121
No testing was performed.

erdseb:
This comment is related: https://github.com/boto/botocore/pull/2121#issuecomment-686762476

swetashre:
This issue is related to issue https://github.com/boto/botocore/issues/1942

nateprewitt:
This issue should now be resolved in Botocore 1.18.0.

PR #1944, opened 2020-01-17
setup.py: Allow use with docutils >=0.16 (fixes #1942)

moseb:
For #1942

Issue #1942, opened 2020-01-14
botocore 1.14.4 blocks users from updating docutils to 0.16


moseb:
Hi!

When botocore is used in a virtualenv next to other packages or in context of a whole GNU/Linux distribution, this line in `setup.py`
https://github.com/boto/botocore/blob/70c7767374c42bd86a896391dff55fc25ebfac27/setup.py#L27

blocks users from updating to docutils 0.16 which is not ideal. Can the upper boundary be dropped or be turned into semver-like `<1` at least to be more friendly to other packages and user freedom?

Thanks and best, Sebastian

PS: I have seen #900 and #933 (but I don't consider this the right direction while docutils is imported from public botocore API).

swetashre:
@moseb - Thank you for your post. Marking this as enhancement.

prometheanfire:
seems like we have to go through this issue every few weeks

neg3ntropy:
This is not an enhancement, It's a bug.

roll:
Hi, our stack users run into:

```
ERROR: botocore 1.15.36 has requirement docutils<0.16,>=0.10, but you'll have docutils 0.16 which is incompatible.
```

yan12125:
@swetashre: Could you have a look into https://github.com/boto/botocore/pull/2011? It's an updated fix for this issue.

edrozenberg:
Same issue here. Thanks to the very useful utility `pipdeptree` for making it easy to dig up issues (https://pypi.org/project/pipdeptree/)

```
Warning!!! Possibly conflicting dependencies found:
* botocore==1.16.20
 - docutils [required: >=0.10,<0.16, installed: 0.16]
```

katzdm:
Would be nice to have this constraint removed; `docutils==0.16.0` was released nearly six months ago.

yan12125:
@edrozenberg: since you are using system packages from Slackware, a workaround might be adding the patch at #2011 to the botocore package on Slackware, like what I did for Arch Linux [1].

[1] https://git.archlinux.org/svntogit/community.git/commit/trunk?h=packages/python-botocore&id=4f4eecd0bf7719bdc77856c1f40e3466e94ad8db

edrozenberg:
@yan12125 Thanks! That would certainly do it and I'll consider applying that sometime. For now I'll just ignore the issue since I don't care about building boto docs.

prometheanfire:
not sure that helps those that want to install from pip

yan12125:
[...]

PR #2154, opened 2020-09-16
Remove docutils dependency

### LINKED ISSUE #900: Consider removing docutils from dependencies
As far as I can see, `docutils` package is only required for developing `botocore` itself and can be left out for `botocore` users.

I'd suggest considering to remove `docutils` from package dependencies as it adds significant time overhead to the installation process.

nateprewitt:
Now that we've removed upstream dependencies on botocore to supply docutils, we're able to remove the dependency completely. This should not affect existing users of botocore since the dependency was primarily for help documentation in AWS CLI and local doc generation.

Doc generation should run now without this dependency and if extended usage of the doc suite is needed for some reason, requirements-docs.txt still supplies the original dependency tree. Please note `bcdoc` is not a public interface and is subject to further change without warning.

This should resolve #900, #933, #1942, #1944, #2120, #2121.


requirements.txt
```diff
@@ -2,6 +2,5 @@ tox>=2.5.0,<3.0.0
 nose==1.3.7
 mock==1.3.0
 wheel==0.24.0
-docutils>=0.10,<0.16
```
kyleknap, line 5: I think we should remove it from the `requirements-docs.txt` as well. There is not really a point to keep this around since the file is really only used to help with HTML doc generation and `docutils` was never actually ever used to generate botocore/boto3 docs.

kyleknap (commented):
Cool. Looks good to me. Just a small comment

requirements.txt
nateprewitt: We actually need this here because older versions of Sphinx have a relatively unbounded dependency on docutils. We'll need to install it until we upgrade Sphinx, so this identifier let's us keep control for now. I'll make a TODO to work on the Sphinx changes in a future PR.

kyleknap (approved):
Looks good! :ship:

Issue #879, opened 2016-04-13
Use io.IOBase as base class for StreamingBody


depp:
StreamingBody is a "file-like object", but file-like objects should probably inherit from IOBase or implement a similar suite of methods.  This will provide an easy fix for #767.

kyleknap:
Are there any methods in particular that you are looking for? It may be difficult to inherit from IOBase as not all of its methods may be possible to implement (given that we are reading from an unseekable stream from the server underneath).

akvadrako:
@kyleknap, IOBase can certainly work with streaming data. This is sufficient for my current use-case, but with a little more work I think it could be fully supported. My main interest was in line-by-line iteration:

```python
class StreamingBodyIO(RawIOBase):
    """Wrap a boto StreamingBody in the IOBase API."""
    def __init__(self, body):
        self.body = body

    def readable(self):
        return True

    def read(self, n=-1):
        n = None if n < 0 else n
        return self.body.read(n)
```

atrigent:
There are two other other open feature requests (#1197 and #767), which should be solved by inheriting from `IOBase` or `RawIOBase`, and one open PR (#1195) which should be closed because it does not use this method.

atrigent:
...actually, is there any point to keeping `StreamingBody` around at all? It looks like urllib3 implements content length checking itself, and it looks like that socket timeout thing might've been resolved as well.

GFernie:
When returning a file-like object, such as *StreamingBody*, *io.IOBase* is a very reasonable abstract to assume. From the *io.IOBase* docstring:

> This class provides empty abstract implementations for many methods that derived classes can override selectively; the default implementations represent a file that cannot be read, written or seeked.

E.g., the missing *io.IOBase* attributes make it very dificult to lazily read a compressed file from S3:

```python
from gzip import GzipFile
import boto3

body_stream_gz = boto3.resource('s3').Object('my_bucket', 'my_object.txt.gz').get()['Body']
with GzipFile(fileobj=body_stream_gz) as body_stream:
    content = body_stream.read('1000')

---------------------------------------------------------------------------
AttributeError: 'StreamingBody' object has no attribute 'tell'
```

Here's an accepted *cpython* PR for the same issue:
https://github.com/python/cpython/pull/3249

glyph:
[...]

PR #2156, opened 2020-09-17
implemented IOBase functions and included unit tests

zdutta:
Updated fix for https://github.com/boto/botocore/issues/879. Previous PR broke S3 integration tests. Object with underlying raw_stream was being destructed via GC. Overriding destructor to keep underlying stream.

### DIFF
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -47,6 +47,9 @@ class StreamingBody(IOBase):
         self._amount_read = 0
 
     def __del__(self):
+        #Extending destructor in order to preserve the underlying raw_stream.
+        #The ability to add custom cleanup logic introduced in Python3.4+.
+        #https://www.python.org/dev/peps/pep-0442/
         pass
 
     def set_socket_timeout(self, timeout):
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -47,9 +47,7 @@ class StreamingBody(IOBase):
         self._amount_read = 0
 
     def __del__(self):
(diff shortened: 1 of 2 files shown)


botocore/response.py
```diff
@@ -45,6 +46,9 @@ def __init__(self, raw_stream, content_length):
         self._content_length = content_length
         self._amount_read = 0
 
+    def __del__(self):
```
nateprewitt, line 49: This seems a bit scary, this will disable garbage collection for the object right? Is that what we're actually trying to do?

joguSD: Implementing `__del__` allows you to extend behavior at GC time, having a stubbed implementation doesn't disable GC. There's a very specific scenario around circular references where Python assumes the presence of a `__del__` indicates custom cleanup logic and thus does not clean up the circularly referenced objects, but I believe this was fixed in Python 3.5+.

botocore/response.py
```diff
@@ -45,6 +46,12 @@ def __init__(self, raw_stream, content_length):
...
+        #Extending destructor in order to preserve the underlying raw_stream.
```
nateprewitt, line 50: "Stubbing destructor in order to..." may be more accurate here but that's a pretty minor nit.

We can probably remove the info about __del__ after this sentence since it'll be safe to assume users digging this deep are familiar.

Last nit, it's best practice to have a space between the # and the first character for comments. After that I think we're good to go here.

nateprewitt (approved):
Cool, I think this looks good to me after JoguSD's explanation.

PR #2159, opened 2020-09-22
Add support for python 3.9

michael-k:
3.9.0 final is expected on Monday, 2020-10-05, see https://www.python.org/dev/peps/pep-0596/#schedule

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,9 @@ matrix:
     - python: 3.8
       dist: xenial
       sudo: true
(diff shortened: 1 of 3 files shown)


nateprewitt:
Hey @michael-k, apologies on the delay here. We've finished the required changes we needed to merge this today but the branch is significantly out of date. We've just merged #2498 which adds 3.9 and 3.10 support, so I'll resolve this as completed. Thanks for the contribution!

PR #2161, opened 2020-09-22
Fix pagination search issue of yielding None

### LINKED ISSUE #2160: Pagination's search incorrectly yielding None values
**Describe the bug**

`search()` incorrectly yields a value of `None` when there are no results when it should simply not yield anything.

**Steps to reproduce**
If you have a runnable example, please include it as a snippet or link to a repository/gist for larger code examples.

On an account with no CloudFront distributions:

```python
distributions = list(
    boto3.client("cloudfront")
    .get_paginator("list_distributions")
    .paginate()
    .search("DistributionList.Items[]")
)
# distributions ends up being [None]
print(f"Found {len(distributions)} distributions")
# prints "Found 1 distributions"
```

**Expected behavior**

The generator should not yield anything, resulting in an empty list.

ericfrederich:
Fixes #2160

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -324,8 +324,8 @@ class PageIterator(object):
             if isinstance(results, list):
                 for element in results:
                     yield element
-            else:
-                # Yield result directly if it is not a list.
+            elif results:
+                # Yield result directly if it is not a list and there are results.
                 yield results
 
     def _make_request(self, current_kwargs):


ericfrederich:
Who can review this?

botocore/paginate.py
```diff
@@ -324,8 +324,8 @@ def search(self, expression):
             if isinstance(results, list):
                 for element in results:
                     yield element
-            else:
-                # Yield result directly if it is not a list.
+            elif results:
+                # Yield result directly if it is not a list and there are results.
```
joseph-jones, line 328: What if the result evaluates to `False` but is not `None`?

```suggestion
            elif results is None:
                return
            else:
                # Yield result directly if it is not a list and there are results.
```

stobrien89:
Hey @ericfrederich,

Sorry again for the delay in contact.  I'm going to leave the issue open for now, but I'll be closing this because we're unable to take the change right now.  I'll also create an item in our backlog internally so this is top-of-mind for the next major version—  I'll see if there's a way we can take an external contribution when that time comes.

Issue #2165, opened 2020-09-23
Client-side parameter validation does not check 'max' value; displays erroneous error message that the max allowed value is infinity


emdelaney:
**Describe the bug**
The [parameter validation logic](https://github.com/boto/botocore/blame/develop/botocore/validate.py#L70-L86) does not validate the 'max' value specified in the AWS service model; it only checks the 'min' value. As a result, it displays incorrect error messages, suggesting that the max allowed value is always `inf`. 
 
**Steps to reproduce**
Using any recent version of the awscli: 
```
aws globalaccelerator create-listener --accelerator-arn irrelevant --protocol UDP --port-ranges FromPort=0,ToPort=65535

Parameter validation failed:
Invalid range for parameter PortRanges[0].FromPort, value: 0, valid range: 1-inf
```

The global accelerator service model [has a max value for this parameter of 65535](https://github.com/boto/botocore/blob/develop/botocore/data/globalaccelerator/2018-08-08/service-2.json#L1311) (and the API will reject requests beyond 65535), but the error message suggests `1-inf` is the valid range.

**Expected behavior**
The error message should display the model's maximum value, if it exists.

Additionally, should the validator validate 'max' constraints, in addition to 'min' ones?

**Debug logs**
```
Traceback (most recent call last):
  File "/usr/local/Cellar/awscli/1.17.0/libexec/lib/python3.8/site-packages/awscli/clidriver.py", line 217, in main
    return command_table[parsed_args.command](remaining, parsed_args)
  File "/usr/local/Cellar/awscli/1.17.0/libexec/lib/python3.8/site-packages/awscli/clidriver.py", line 358, in __call__
    return command_table[parsed_args.operation](remaining, parsed_globals)
  File "/usr/local/Cellar/awscli/1.17.0/libexec/lib/python3.8/site-packages/awscli/clidriver.py", line 527, in __call__
    return self._operation_caller.invoke(
  File "/usr/local/Cellar/awscli/1.17.0/libexec/lib/python3.8/site-packages/awscli/clidriver.py", line 649, in invoke
    response = self._make_client_call(
  File "/usr/local/Cellar/awscli/1.17.0/libexec/lib/python3.8/site-packages/awscli/clidriver.py", line 661, in _make_client_call
    response = getattr(client, xform_name(operation_name))(
  File "/usr/local/Cellar/awscli/1.17.0/libexec/lib/python3.8/site-packages/botocore/client.py", line 276, in _api_call
[...]

PR #2166, opened 2020-09-24
Show the correct max value when range validation fails.

emdelaney:
Implements a fix for the error-message part of #2165 , displaying the model's specified max value if the user gets an error when requesting a value smaller than the minimum.

There are tests in test_validate.py that suggest that not validating the maximum value specified by the model is intentional, so I have not changed that behavior; the only thing changing is the error message when a user provides a value that is too small. The maximum displayed will now be the model's maximum value, if it exists, instead of always being `inf`.

Before:
```
Invalid range for parameter X, value: -1, valid range: 1-inf
```

After (supposing our example model X has a max value of 1000):
```
Invalid range for parameter X, value: -1, valid range: 1-1000
```

### DIFF
--- a/botocore/validate.py
+++ b/botocore/validate.py
@@ -75,6 +75,9 @@ def range_check(name, value, shape, error_type, errors):
         min_allowed = shape.metadata['min']
         if value < min_allowed:
             failed = True
+    if 'max' in shape.metadata:
+        max_allowed = shape.metadata['max']
+        # Only validating the min, and not the max here, is intentional.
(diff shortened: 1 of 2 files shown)


RyanFitzSimmonsAK:
Hi @emdelaney, thanks for this PR and your patience on hearing back.  After bringing this up for further discussion with the team, we've decided to close this PR, and reopen https://github.com/boto/botocore/issues/2165 to track this issue.  Please follow up there if you have any questions.  Thanks!

PR #2179, opened 2020-10-13
Add logic to configure IPv6 env var

zdutta:
Enable user to set InstanceMetaDataFetcher endpoint to a custom IPv6 endpoint via environment variable or config file. Included additional functions to perform IPv6 URI validation.

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -23,9 +23,11 @@ import random
 import os
 import socket
 import cgi
+import warnings
 
 import dateutil.parser
 from dateutil.tz import tzutc
+from urllib3.util.url import IPV6_ADDRZ_RE
 
 import botocore
 import botocore.awsrequest
@@ -40,14 +42,16 @@ from botocore.exceptions import (
     MetadataRetrievalError, EndpointConnectionError, ReadTimeoutError,
     ConnectionClosedError, ConnectTimeoutError, UnsupportedS3ArnError,
     UnsupportedS3AccesspointConfigurationError, SSOTokenLoadError,
-    InvalidRegionError, UnsupportedOutpostResourceError,
+    InvalidRegionError, InvalidIMDSEndpointError, UnsupportedOutpostResourceError,
     UnsupportedS3ControlConfigurationError, UnsupportedS3ControlArnError,
-    InvalidHostLabelError,
+    InvalidHostLabelError, HTTPClientError
 )
+from urllib3.exceptions import LocationParseError
 
 logger = logging.getLogger(__name__)
 DEFAULT_METADATA_SERVICE_TIMEOUT = 1
 METADATA_BASE_URL = 'http://169.254.169.254/'
+METADATA_BASE_URL_IPv6 = 'http://[fe80:ec2::254%eth0]/'
 
 # These are chars that do not need to be urlencoded.
 # Based on rfc2986, section 2.3
@@ -274,10 +278,11 @@ class IMDSFetcher(object):
 
     def __init__(self, timeout=DEFAULT_METADATA_SERVICE_TIMEOUT,
                  num_attempts=1, base_url=METADATA_BASE_URL,
-                 env=None, user_agent=None):
+                 env=None, user_agent=None, config=None):
         self._timeout = timeout
         self._num_attempts = num_attempts
-        self._base_url = base_url
+        self._base_url = self._select_base_url(base_url, config)
+
         if env is None:
             env = os.environ.copy()
         self._disabled = env.get('AWS_EC2_METADATA_DISABLED', 'false').lower()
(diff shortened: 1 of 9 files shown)


botocore/utils.py
```diff
@@ -919,6 +956,8 @@ def is_valid_endpoint_url(endpoint_url):
         re.IGNORECASE)
     return allowed.match(hostname)
 
+def is_valid_ipv4_or_ipv6_endpoint(endpoint_url):
+    return is_valid_endpoint_url(endpoint_url) or is_valid_IPv6_endpoint_url(endpoint_url)
```
nateprewitt, line 960: I think we have an edge case here where unbracketed shorthand ipv6 addresses will pass the ipv4 check. Those aren't valid URIs, so we'll need to handle them appropriately. We either need to add additional constraints here to use the ipv6 check on an address with more than one `:` (we may need to evaluate without user information e.g. `url.netloc.split('@')[-1]`).

Alternatively, we could use urllib3's validation but need to scope it more appropriately as discussed above.

nateprewitt (changes requested):

tests/unit/test_utils.py
```diff
@@ -2316,20 +2332,65 @@ def test_disabling_env_var_not_true(self):
...
+
+        self.add_default_imds_responses()
+
+        fetcher = InstanceMetadataFetcher(config=config, base_url=url_arg)
+        result = fetcher.retrieve_iam_role_credentials()
+
+        self.assertEqual(result, self._expected_creds)
+        self.assertEqual(fetcher.get_base_url(), url_arg)
+
+    def test_IPv6_IMDS_not_allocated(self):
```
nateprewitt, line 2380: It looks like you don't have any tests around supplying `None` or an empty config. Let's make sure we get those in there.

botocore/utils.py
```diff
@@ -919,6 +961,8 @@ def is_valid_endpoint_url(endpoint_url):
         re.IGNORECASE)
     return allowed.match(hostname)
 
+def is_valid_ipv4_or_ipv6_endpoint(endpoint_url):
```
nateprewitt, line 964: We've had some back and forth on this, and I think this name still seems a bit wrong. The endpoint isn't necessarily an ip address and` is_valid_endpoint_url` isn't specific to IPv4.

Maybe we should either name this something like `is_valid_uri` which is a more complete set of rules compared to `is_valid_endpoint_url`. We could also update the definition to `is_valid_endpoint_url` to be this combination, but that may break expectations for users who are currently using it.

nateprewitt (changes requested):

tests/functional/test_ec2.py
```diff
@@ -133,3 +133,4 @@ def test_copy_snapshot_injects_presigned_url(self):
             'a94a6b52afdf3daa34c2e2a38a62b72c8dac129c9904c61aa1a5d86e38628537'
         )
         self.assertEqual(query_args['X-Amz-Signature'], [expected_signature])
+
```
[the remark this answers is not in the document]
nateprewitt: The easiest way to resolve this may to just reset the commit and remove your first change to this. Using `git add -p <filename or directory>` will let you choose which lines you commit and avoid issues like this.

botocore/utils.py
```diff
@@ -308,6 +343,11 @@ def _fetch_metadata_token(self):
                     raise BadIMDSRequestError(request)
             except ReadTimeoutError:
                 return None
+            except HTTPClientError as e:
+                if isinstance(e.kwargs.get('error'), LocationParseError):
+                    raise InvalidIMDSEndpointError(endpoint=url, error=e)
+                else:
+                    raise HTTPClientError(error=e.kwargs.get('error'))
```
nateprewitt, line 350: You should be able to just do `raise` or `raise e` here and it will rethrow the current exception being caught. That avoids having to try and recreate it again.

i.e.
```python
except HTTPClientError as e:
    if isinstance(e.kwargs.get('error'), LocationParseError):
        raise InvalidIMDSEndpointError(endpoint=url, error=e)
    raise
```

tests/unit/test_utils.py
```diff
@@ -2316,20 +2332,77 @@ def test_disabling_env_var_not_true(self):
...
+            status_code=400, body=b'{}')
+
+        fetcher = InstanceMetadataFetcher(config=config)
+        result = fetcher.retrieve_iam_role_credentials()
+        self.assertEqual(result, {})
+
+    def test_IPv6_IMDS_empty_config(self):
+        configs = [({'ec2_metadata_service_endpoint': ''},'http://169.254.169.254/'),
+                ({'imds_use_ipv6': ''}, 'http://169.254.169.254/'),
+                ({}, 'http://169.254.169.254/')]
```
nateprewitt, line 2398: I think we wanted to explicitly check `None` too since you aren't defaulting the config:
```
(None, 'http://169.254.169.254/')
```

tests/unit/test_utils.py
```diff
@@ -2316,20 +2332,77 @@ def test_disabling_env_var_not_true(self):
...
-        self.add_get_role_name_imds_response()
-        self.add_get_credentials_imds_response()
+        self.add_default_imds_responses()
 
         fetcher = InstanceMetadataFetcher(base_url=url, env=env)
         result = fetcher.retrieve_iam_role_credentials()
 
         self.assertEqual(result, self._expected_creds)
 
+    def test_IMDS_USE_IPV6(self):
```
nateprewitt, line 2342: Since we've got a couple more changes, this should really be lowercase. It's generally against convention to use capitals in [functions](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names). The same would probably apply for `is_valid_IPv6_endpoint_url`

nateprewitt (changes requested):

tests/unit/test_utils.py
```diff
@@ -2316,20 +2332,78 @@ def test_disabling_env_var_not_true(self):
...
+                'http://[fe80:ec2::010%eth0]/', 'http://192.168.1.1/']
+        for url in urls:
+            self.assertTrue(is_valid_uri(url))
+        
+    def test_ipv6_endpoint_no_brackets_env_var_set(self):
+        url = 'http://fe80:ec2::010/'
+        config = {'ec2_metadata_service_endpoint': url}
+        self.assertFalse(is_valid_ipv6_endpoint_url(url))
+
+    def test_IPv6_invalid_endpoint(self):
```
nateprewitt, line 2365: Looks like this one slipped through the cracks.

nateprewitt (commented):
Cool, I think we're looking much better. One minor comment and then if you can squash the commits down, I think we're set. @joguSD, do you mind doing another once over before we merge?

joguSD:
Overall looks good, let's add a change log entry as a feature so we do a minor version bump. We're pulling the `urllib3` version up quite a bit and changing default credential providers so I think it makes sense to do a minor version bump here.

joguSD (approved):

nateprewitt (approved):

PR #2185, opened 2020-10-19
Add support for SSLKEYLOGFILE environment variable

nielslaukens:
I am debugging an issue where S3 closes the connection during a download. To gather exact timing information, it would be useful to capture a network dump (e.g. tcpdump) and be able to inspect *inside* the TLS layer.

 A lot of applications (Firefox, Chrome) support the SSLKEYLOGFILE environment variable that will dump the needed key material in a standardized format to the given file. Since Python 3.8, Python supports this as well when using `ssl.create_default_context()`. Botocore, however, creates its own `SSLContext()` object, so this feature is not activated.

Would this be an acceptable feature to include in botocore?

References:
* https://pypi.org/project/sslkeylog/
* https://python.libhunt.com/urllib3-changelog (added in 1.25.10)

### DIFF
--- a/botocore/httpsession.py
+++ b/botocore/httpsession.py
@@ -2,6 +2,7 @@ import os.path
 import logging
 import socket
 from base64 import b64encode
+import sys
 
 from urllib3 import PoolManager, ProxyManager, proxy_from_url, Timeout
 from urllib3.util.retry import Retry
@@ -81,6 +82,14 @@ def create_urllib3_context(ssl_version=None, cert_reqs=None,
         # We do our own verification, including fingerprints and alternative
         # hostnames. So disable it here
         context.check_hostname = False
+
+    # Enable logging of TLS session keys via defacto standard environment variable
+    # 'SSLKEYLOGFILE', if the feature is available (Python 3.8+). Skip empty values.
+    if hasattr(context, 'keylog_filename'):
+        keylogfile = os.environ.get('SSLKEYLOGFILE')
+        if keylogfile and not sys.flags.ignore_environment:
+            context.keylog_filename = keylogfile
+
     return context
 
 


nielslaukens:
Based on the automatic CI stuff, I may need a little help to get the test coverage back up to what it was before.

zdutta:
Hi @nielslaukens, thanks for the PR! Unfortunately, this breaks backwards compatibility with Python2.7+ since `SSLKEYLOGFILE` is only supported in Python3.8+ as outlined [here](https://urllib3.readthedocs.io/en/latest/advanced-usage.html#decrypting-captured-tls-sessions-with-wireshark). Our goal is to maintain as much consistency between the two versions as possible.

nielslaukens:
Hi zdutta,

Is the problem that this PR actually breaks Python 2.7? Or is the problem that this would introduce a feature that will not work on <3.8?

zdutta:
@nielslaukens , the latter.

nielslaukens:
I don't understand why this would be a problem for a non-advertised debugging aid...

This PR simply passes through a feature from an underlying library. It's similar to TLS1.3-support that is also new since Python 2.7.

This PR will make botocore better for *some* people. You are correct that it will not make it better for everyone. So if it really is the policy that, if not everyone can enjoy a feature, no-one should get it, I can only understand your decision but respectfully disagree.

zdutta:
@nielslaukens we appreciate the feedback! I have taken a further look at the specs of the upstream source of urrlib3 and the PR that introduced this feature. This PR does have potential. In order to make this merge ready we would like to see this as close to upstream as possible. This also entails the comments.

botocore/httpsession.py
```diff
@@ -81,6 +82,13 @@ def create_urllib3_context(ssl_version=None, cert_reqs=None,
         # We do our own verification, including fingerprints and alternative
         # hostnames. So disable it here
         context.check_hostname = False
+
+    # OpenSSL 1.1.1 keylog file
+    if hasattr(context, 'keylog_filename'):
+        keylogfile = os.environ.get('SSLKEYLOGFILE')
+        if keylogfile and not sys.flags.ignore_environment:
```
zdutta, line 89: Can we get some validation that `sys.flags.ignore_environment` is supported on all Python versions?

zdutta (changes requested):

botocore/httpsession.py
nielslaukens: According to the docs, sys.flags.ignore_environment is present since Python 2.7: https://docs.python.org/2.7/library/sys.html

nielslaukens: I've updated the comments, and added a backlink to upstream in the commit message

zdutta: Upon further inspection, can we get testing to show that `sys.flags.ignore_enviorment` properly opts out of setting the keylog file path in the `SSLContext`. Requesting this since it is not part of the upstream implementation. Thanks!

nielslaukens: I'm not sure I understand your request...
Do you want me to add a test to verify that the if-statement on line 90 actually works for the `ignore_environment`-part?

zdutta: After running through the possible outcomes if something were to change in the future, I think everything looks good as is. I wanted to see if we could do anything more to make this robust but your question is fair. I will approve and get this merged.

zdutta (approved):

Issue #2189, opened 2020-10-21
RefreshableCredentials Class fetches credentials too eagerly


rp4fx12:
**Is your feature request related to a problem? Please describe.**
The current advisory_refresh_timeout for the RefreshableCredentials Class is 15 minutes which can prove to be problematic for various 3rd party tools (we ran into an issue where we were using KIAM that was sending tokens with 15 minute expiration times). This causes the botocore calls to consistently refresh the token which ends up in a failing request. There is actually a [similar issue](https://github.com/aws/aws-sdk-java/issues/1893) in the java aws sdk.

**Describe the solution you'd like**
It would be helpful to be able to override the advisory and mandatory timeout settings so that we can avoid the constant token refresh in the case where we are using a 3rd party tool that has a 15 minute timeout window (or lower). My personal preference (and not sure if this is the right approach) would be to use an environment variable to setup these values (with the default being 15 minutes and 10 minutes respectively which are the current values), so that we can override functionality. I was thinking using environment variables as we use libraries that inherit from botocore and it may be tricky to find all the places where refreshable credentials are used and update the advisory and mandatory timeout values.

kdaily:
Hi @rp4fx12, thanks for the feature request.

rittneje:
We actually have the reverse of this issue. When generating a presigned URL with temporary credentials, the signature is invalidated when the temporary credentials expire, even if the signature duration is longer than that. Consequently, when dealing with presigned URLs, it is important to be able to enforce that the temporary credentials still be valid for some time. So we would like to be able to set the mandatory refresh time at 15 minutes.

As a side note, currently we can ask if the credentials will expire within the next 15 minutes via the `refresh_needed` method. However, the API lacks a public `refresh` method to do anything in response.

PR #2190, opened 2020-10-21
Feature/overriding refresh token timeouts

rp4fx12:
There might be a better approach here (or this may not be desired or needed at all) but this should allow for users to override the advisory and mandatory refresh timeouts that are a part of the RefreshableToken Class - addresses this issue: #2189


botocore/credentials.py
```diff
@@ -406,6 +400,22 @@ def create_from_metadata(cls, metadata, refresh_using, method):
         )
         return instance
 
+    @property
+    def _advisory_refresh_timeout(self):
```
rp4fx12, line 404: Not sure if this is an acceptable way to handle this but just offering it as a potential suggestion.

rp4fx12: Might not also want to pull from the environment variables every time this property is called.

rittneje: Why not just save it off in the constructor rather than making a property method?

aBurmeseDev:
Thank you for submitting this PR. After team discussion, we're going to have to discuss with other SDKs for potential cross-SDK approach to keep it consistent and decided not to consider this one at this time.

PR #2192, opened 2020-10-21
update awscrt to fix bugs

graebm:
- awscrt now accepts duck typed streams (rather than requiring inheritance from IOBase).
- fixes signing bug involving "X-Amz-Security-Token"


kyleknap (approved):
Thanks! :ship:

PR #2197, opened 2020-10-27
Vendoring ipv6 regex patterns from urllib3

zdutta:
Vendoring ipv6 validation regex patterns from urllib3. https://github.com/urllib3/urllib3/blob/7e856c0/src/urllib3/util/url.py

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -27,7 +27,6 @@ import warnings
 
 import dateutil.parser
 from dateutil.tz import tzutc
-from urllib3.util.url import IPV6_ADDRZ_RE
 
 import botocore
 import botocore.awsrequest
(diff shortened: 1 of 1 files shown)


botocore/utils.py
```diff
@@ -141,6 +140,38 @@
     "tagging": "resource-groups-tagging-api"
 }
 
+# Vendoring ipv6 validation regex patterns from urllib3
```
nateprewitt, line 143: Minor nit: Let's use change ipv6 to IPv6. Otherwise, I think we're looking good.

nateprewitt (changes requested):

nateprewitt (approved):

PR #2200, opened 2020-10-29
Replace travis CI with GitHub Actions

stealthycoin:


nateprewitt (approved):
👍 from me if tests finish cleanly.

kyleknap (approved):
We should do a follow up PR to make sure Windows gets included, but this is fine for a substitute for Travis. :ship:

PR #2206, opened 2020-11-06
Add data/__init__.py

tschoonj:
When trying to load `endpoints.json` to extract the human-readable region names using `importlib.resource.read_text`, it failed since `data` is not recognised as a module. Simply adding an empty `__init__.py` fixes this.

With this patch the following works:

```python
import importlib.resources
importlib.resources.read_text('botocore.data', 'endpoints.json')
```


nateprewitt:
Hi @tschoonj,

Thanks for reaching out about this. Unfortunately, I don't think we want to add an \_\_init\_\_.py file in this directory. That denotes this is a Python module, which isn't really accurate. The data directory is a collection of ancillary JSON models for services, not intended for the Python path.

You should be able to accomplish the same thing with a function like:
```
import botocore
import os

def load_endpoints():
    botocore_base = os.path.dirname(os.path.abspath(botocore.__file__))
    endpoints_path = os.sep.join([botocore_base, 'data', 'endpoints.json'])
    with open(endpoints_path, 'r') as f:
        return f.read()

PR #2211, opened 2020-11-09
Rearranged exception handling for IMDS

zdutta:
Reordered exception handling for IMDS Fetcher

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -374,15 +374,15 @@ class IMDSFetcher(object):
                     raise BadIMDSRequestError(request)
             except ReadTimeoutError:
                 return None
+            except RETRYABLE_HTTP_ERRORS as e:
+                logger.debug(
+                    "Caught retryable HTTP exception while making metadata "
+                    "service request to %s: %s", url, e, exc_info=True)
             except HTTPClientError as e:
                 if isinstance(e.kwargs.get('error'), LocationParseError):
                     raise InvalidIMDSEndpointError(endpoint=url, error=e)
                 else:
                     raise
-            except RETRYABLE_HTTP_ERRORS as e:
-                logger.debug(
-                    "Caught retryable HTTP exception while making metadata "
-                    "service request to %s: %s", url, e, exc_info=True)
         return None
 
     def _get_request(self, url_path, retry_func, token=None):
--- a/tests/functional/test_utils.py
+++ b/tests/functional/test_utils.py
@@ -13,9 +13,12 @@
 import os
 import tempfile
 import shutil
-from tests import unittest
+from tests import unittest, mock
 
-from botocore.utils import FileWebIdentityTokenLoader
+from botocore.exceptions import (ConnectionClosedError, HTTPClientError,
+                                InvalidIMDSEndpointError)
+from botocore.utils import FileWebIdentityTokenLoader, InstanceMetadataFetcher
+from urllib3.exceptions import LocationParseError
 
 
 class TestFileWebIdentityTokenLoader(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


botocore/utils.py
```diff
@@ -374,15 +374,15 @@ def _fetch_metadata_token(self):
                     raise BadIMDSRequestError(request)
             except ReadTimeoutError:
                 return None
+            except RETRYABLE_HTTP_ERRORS as e:
```
nateprewitt, line 377: It doesn't look like we have a test that actually exercises this code path. Can we get something mocked up in the functional tests that validates this behavior is working correctly?

nateprewitt (changes requested):

tests/functional/test_utils.py
```diff
@@ -13,9 +13,14 @@
 import os
 import tempfile
 import shutil
-from tests import unittest
+from tests import unittest, mock, ClientHTTPStubber, BaseEnvVar, create_session
```
nateprewitt, line 16: Are you still using `create_session`?

tests/functional/test_utils.py
```diff
@@ -13,9 +13,14 @@
...
 
-from botocore.utils import FileWebIdentityTokenLoader
+from botocore.exceptions import (EndpointConnectionError, HTTPClientError,
+                                InvalidIMDSEndpointError)
+from botocore.utils import FileWebIdentityTokenLoader, InstanceMetadataFetcher
+import botocore.httpsession
+import botocore.session
```
nateprewitt, line 22: Is this import being used?

tests/functional/test_utils.py
```diff
@@ -40,3 +45,20 @@ def test_can_load_token(self):
...
         self.assertEqual(self.token, token)
+
+
+class TestInstanceMetadataFetcher(BaseEnvVar):
+    def test_catch_retryable_http_errors(self):
+        with mock.patch('botocore.httpsession.URLLib3Session.send') as send_mock:
+            fetcher = InstanceMetadataFetcher()
+            send_mock.side_effect = EndpointConnectionError(endpoint_url="foo")
+            fetcher.retrieve_iam_role_credentials()
+        self.assertEquals(send_mock.call_count, 2)
```
nateprewitt, line 56: Do we have any way to assert the second call is a retry? Maybe pulling a url off `call_args`, `call_args_list`, or `mock_calls`?

nateprewitt (changes requested):
Just a quick question about imports and how we're validating retries.

tests/functional/test_utils.py
```diff
@@ -40,3 +44,24 @@ def test_can_load_token(self):
...
+
+class TestInstanceMetadataFetcher(unittest.TestCase):
+    def test_catch_retryable_http_errors(self):
+        with mock.patch('botocore.httpsession.URLLib3Session.send') as send_mock:
+            fetcher = InstanceMetadataFetcher()
+            send_mock.side_effect = EndpointConnectionError(endpoint_url="foo")
+            fetcher.retrieve_iam_role_credentials()
+        self.assertEquals(send_mock.call_count, 2)
+        self.assertEquals(
+            urlparse(send_mock.call_args_list[0][0][0].url).netloc,
```
nateprewitt, line 57: Can we use `get_base_url` from the fetcher to assert this is actually the IMDS url we're calling?

tests/functional/test_utils.py
```diff
@@ -13,9 +13,13 @@
 import os
 import tempfile
 import shutil
-from tests import unittest
+from tests import unittest, mock
 
-from botocore.utils import FileWebIdentityTokenLoader
+from botocore.compat import urlparse
```
nateprewitt, line 18: I think we can remove this now.

nateprewitt (changes requested):
One import needs to be cleaned up but otherwise, it looks good.

nateprewitt (approved):
**Edit:** I pulled this down and it looks like one of the tests doesn't actually work as expected. Pulling back approval while we get that sorted.

tests/functional/test_utils.py
```diff
@@ -40,3 +43,24 @@ def test_can_load_token(self):
         loader = FileWebIdentityTokenLoader(self.token_file)
         token = loader()
         self.assertEqual(self.token, token)
+
+
+class TestInstanceMetadataFetcher(unittest.TestCase):
+    def test_catch_retryable_http_errors(self):
+        with mock.patch('botocore.httpsession.URLLib3Session.send') as send_mock:
+            fetcher = InstanceMetadataFetcher()
+            send_mock.side_effect = EndpointConnectionError(endpoint_url="foo")
```
nateprewitt, line 52: Ok a couple tweaks here now that I've played around with it. `EndpointConnectionError` is the only retyable exception that doesn't meet this criteria actually. This test will pass with or without the change in utils. Let's change the exception to `ConnectionClosedError` instead.

tests/functional/test_utils.py
```diff
@@ -40,3 +43,24 @@ def test_can_load_token(self):
...
+            fetcher.retrieve_iam_role_credentials()
```
nateprewitt, line 53: This will return an empty dictionary if it doesn't fail. Let's grab that for our assertion in some variable like `creds`.

tests/functional/test_utils.py
```diff
@@ -40,3 +43,24 @@ def test_can_load_token(self):
...
+        self.assertEquals(send_mock.call_count, 2)
+        call_one = send_mock.call_args_list[0][0][0]
```
nateprewitt, line 55: This may be simpler to write as:
```
# Validate all calls are to IMDS
for call_instance in send_mock.call_args_list:
    self.assertTrue(call_instance[0][0].url.startswith(fetcher.get_base_url()))

self.assertTrue(creds == {})

nateprewitt (changes requested):
Ok, a few more tweaks to make sure the tests prevent regression.

nateprewitt (approved):
Ok, I think we're looking better now.

PR #2212, opened 2020-11-10
Update pin for urllib3 1.26.0 release

nateprewitt:
urllib3 released 1.26.0 today. This will raise our pin to avoid pip conflicts for users installing the latest version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ if sys.version_info[:2] == (3, 4):
     # urllib3 dropped support for python 3.4 in point release 1.25.8
     requires.append('urllib3>=1.25.4,<1.25.8')
 else:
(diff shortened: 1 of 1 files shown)


zdutta (approved):

qbitacademy-github:
@nateprewitt shouldn't this be bumped too https://github.com/boto/botocore/blob/develop/setup.cfg#L9? Sorry I might miss something but it seems like my local tests are still conflicting on `urllib3`.

cletomartin:
@nateprewitt shouldn't this be bumped too https://github.com/boto/botocore/blob/develop/setup.cfg#L9? Sorry I might miss something but it seems like my local tests are still conflicting on urllib3.

nateprewitt:
Yep, great catch @cletomartin! We'll get that merged shortly.

PR #2219, opened 2020-11-16
Add support for HTTPS proxies

jalopezsilva:
Starting with 1.26 urllib3 now supports HTTPS proxies. To enable the support two changes are needed:

* An additional ``proxy_kwargs`` argument that can be passed from client config. This dictionary will be used to pass any arguments needed to the underlying httpsession.
* The ``use_forwarding_for_https`` mode requires us to send the absolute URI when enabled.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -104,7 +104,8 @@ class ClientArgsCreator(object):
             proxies=new_config.proxies,
             timeout=(new_config.connect_timeout, new_config.read_timeout),
             socket_options=socket_options,
-            client_cert=new_config.client_cert)
+            client_cert=new_config.client_cert,
+            proxies_kwargs=new_config.proxies_kwargs)
 
         serializer = botocore.serialize.create_serializer(
             protocol, parameter_validation)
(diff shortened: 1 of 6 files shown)


joguSD:
Thanks for starting this PR. I think we'll need to be a little more selectively and explicitly select which proxy configurations to make available rather than just forwarding the entire `proxy_kwargs` dictionary. This implementation won't be sufficient to get support for the AWS CLI so at the very least we'll need to make the following configurations available:

* Proxy CA bundle
* Proxy Client Cert
* Forwarding for HTTPS proxies

Are there any other configurations you would need to adjust on the proxy's SSL context?

jalopezsilva:
@joguSD thanks for the feedback. We would also need a proxy client key and that's it.  We don't need anything else for the time being.

How would like me to proceed? Should I remove ``proxy_kwargs`` from the config and add the following instead?

* ``proxy_ca_bundle``
* ``proxy_client_cert``
* ``proxy_client_key``
* ``proxy_use_forwarding_for_https``

Happy to make the changes, just let me know!

joguSD:
@jalopezsilva I think we'll be picking up this work so we'll handle the updates. Will make sure we include the `proxy_client_key` option as well. We'll likely expose this in the same manner as our current `client_cert` option which takes a tuple requests style.

jalopezsilva:
Awesome, thanks @joguSD. I'll keep an eye out for the changes then!

nateprewitt:
Hey @jalopezsilva,

I'm going to merge this into a feature branch to get it setup with the new config changes. If you want to run any testing on the new config that would be great. We'll track final testing and merging into a release from there.

Thanks again for the PR!

PR #2221, opened 2020-11-16
Fix Content-MD5 header for s3transfer

### LINKED ISSUE #1979: Can not PUT a 0 bytes object when Object Lock is enabled on s3 bucket

nateprewitt:
### Overview

This is a continuation of #1985 to resolve #1979. We're adding two additional tests to prevent future regression with empty bodies, and adding some more info on the underlying issue.

### Bug
While the initial PR was addressing a lack of `Content-MD5` headers being added for empty files going to s3, the bug is little more subtle. Currently, using the low level API for `PutBucket` will ensure all input, regardless of length, is transformed into a file-like object. When this is passed to our MD5 checksum function, it will always be `True` and get the header added correctly. These always evaluate to `True` because `bool` resolves "truthiness" to be if the file-object exists or not.

The bug enters when the higher level s3 APIs are being invoked like `s3 sync` in the AWS CLI. This replaces our standard file-like wrappers (StringIO, BytesIO) with s3transfer's [`ReadFileChunk`](https://github.com/boto/s3transfer/blob/b2b6c44cb283e134f6bf9fecad7ff5ee28df793d/s3transfer/utils.py#L387). Due to s3transfer implementing [`__len__`](https://github.com/boto/s3transfer/blob/b2b6c44cb283e134f6bf9fecad7ff5ee28df793d/s3transfer/utils.py#L519-L525) on what would otherwise appear to be a file-like object, we now get different criteria for what makes it truthy. `bool` will choose `__len__` > 0 as being more specific than if the file exists. That broke our previous assumptions on what would be coming through to `conditionally_calculate_md5`.

### Outcome
We probably *should* make `ReadFileChunk` adhere closer to the file-like object spec, but at this point it would be a backwards incompatible change. I think the original proposal from @wimglenn is the safest bet to make things operate as expected.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -25,7 +25,7 @@ import uuid
 
 from botocore.compat import (
     unquote, json, six, unquote_str, ensure_bytes, get_md5,
-    MD5_AVAILABLE, OrderedDict, urlsplit, urlunsplit, XMLParseError,
+    OrderedDict, urlsplit, urlunsplit, XMLParseError,
     ETree,
 )
 from botocore.docs.utils import AutoPopulatedParam
@@ -37,15 +37,11 @@ from botocore.signers import add_generate_db_auth_token
 from botocore.exceptions import ParamValidationError
 from botocore.exceptions import AliasConflictParameterError
 from botocore.exceptions import UnsupportedTLSVersionWarning
-from botocore.exceptions import MissingServiceIdError
 from botocore.utils import percent_encode, SAFE_CHARS
 from botocore.utils import switch_host_with_param
-from botocore.utils import hyphenize_service_id
 from botocore.utils import conditionally_calculate_md5
 
-from botocore import retryhandler
 from botocore import utils
-from botocore import translate
 import botocore
 import botocore.auth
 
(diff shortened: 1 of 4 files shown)


joguSD (approved):

zdutta (commented):
Looks good!

PR #2222, opened 2020-11-17
Use dict instead of OrderedDict for generating skeletons

vz10:


kyleknap (approved):
Looks good to me. :ship:

PR #2224, opened 2020-11-17
Adding state to to WaiterError exception

zdutta:
Improving WaiterError exception message with failure state. Addressing https://github.com/aws/aws-cli/issues/2918

### DIFF
--- a/botocore/waiter.py
+++ b/botocore/waiter.py
@@ -163,6 +163,21 @@ class AcceptorConfig(object):
         self.argument = config.get('argument')
         self.matcher_func = self._create_matcher_func()
 
+    @property
+    def explanation(self):
+        if self.matcher == 'path':
+            return 'For expression "%s" we matched expected path: "%s"' % (self.argument, self.expected)
+        elif self.matcher == 'pathAll':
+            return 'For expression "%s" all members matched excepted path: "%s"' % (self.argument, self.expected)
+        elif self.matcher == 'pathAny':
+            return 'For expression "%s" we matched expected path: "%s" at least once' % (self.argument, self.expected)
+        elif self.matcher == 'status':
+            return 'Matched expected HTTP status code: %s' % self.expected
+        elif self.matcher == 'error':
+            return 'Matched expected service error code: %s' % self.expected
+        else:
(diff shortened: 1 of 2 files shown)


botocore/waiter.py
```diff
@@ -309,23 +324,35 @@ def wait(self, **kwargs):
...
+                        ),
+                        last_response=response,
                     )
             if current_state == 'success':
                 logger.debug("Waiting complete, waiter matched the "
                              "success state.")
                 return
             if current_state == 'failure':
+                reason = 'Waiter encountered a terminal failure state: %s' % (
+                        acceptor.explanation
```
nateprewitt, line 339: This will potentially be None if we implement a new acceptor later and blow up with a `ValueError` on formatting. Should be set up a safer default in the event of no match?

tests/unit/test_waiters.py
```diff
@@ -392,6 +392,53 @@ def test_unspecified_errors_propagate_error_code(self):
...
+            ([{'state': 'failure', 'matcher': 'status', 'expected': 404}],
+             [{'ResponseMetadata': {'HTTPStatusCode': 404}}],
+             '404'),
+            ([{'state': 'failure', 'matcher': 'error', 'expected': 'FailError'}],
+             [{'Error': {'Code': 'FailError', 'Message': 'foo'}}],
+             'FailError'),
+            ([{'state': 'retry', 'matcher': 'error', 'expected': 'RetryMe'}],
+             [{'Success': False}]*4,
+             'Max attempts exceeded'),
+            ([{'state': 'retry', 'matcher': 'error', 'expected': 'RetryMe'}],
```
nateprewitt, line 430: Can we keep this test consistent with the one we deleted. That way we're sure we're not losing any edge cases.

nateprewitt (changes requested):

botocore/waiter.py
joguSD: @nateprewitt In practice it's probably unlikely. If an unknown waiter type is encountered you can't actually instantiate the `AcceptorConfig` object anyways. Though I suppose a default message long the lines of `'No explanation for unknown waiter type "%s"'`.

nateprewitt (approved):

PR #2226, opened 2020-11-18
Handle top level Error key for non-error responses

joguSD:
The `standard` retry mode will attempt to introspect any parsed response
for a top level `"Error"` key. If present the standard retry strategy
assumes that the response is an error response and that there will be a
nested dictionary under the `"Code"` key. This assumption is not
generally safe as standard responses can also contain a top level
`"Error"` key.

Fixes https://github.com/aws/aws-cli/issues/5644

### DIFF
--- a/botocore/retries/standard.py
+++ b/botocore/retries/standard.py
@@ -198,7 +198,10 @@ class RetryContext(object):
         """
         if self.parsed_response is None:
             return
-        return self.parsed_response.get('Error', {}).get('Code')
+        error = self.parsed_response.get('Error', {})
+        if not isinstance(error, dict):
+            return
+        return error.get('Code')
 
     def add_retry_metadata(self, **kwargs):
         """Add key/value pairs to the retry metadata.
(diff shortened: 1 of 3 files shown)


botocore/retries/standard.py
```diff
@@ -198,7 +198,10 @@ def get_error_code(self):
         """
         if self.parsed_response is None:
             return
-        return self.parsed_response.get('Error', {}).get('Code')
+        error = self.parsed_response.get('Error', {})
```
nateprewitt, line 201: This seems sensible for the least amount of change. Do you know if there are currently services returning 200 responses for known error cases? I wonder if it would be better to just avoid calling this entirely for non-error status codes. I wouldn't expect us to retry on 100, 200, or 300 status codes.

nateprewitt: We discussed this offline and there may be some edge cases that make this a non-starter. We'll revisit avoiding this logic for status_code < 400 later.

PR #2229, opened 2020-11-19
Add description for each service

swetashre:
Added the top level documentation string so that each service can have its description. 

Screen Shot 2020-11-20 at 2 30 51 PM


joguSD:
I wonder if we want to keep the original wording and then follow with the wording from the model. Taking a look at some popular services it looks like this top level description varies quite a bit between services so it might be useful to just keep the previous wording in there. Maybe try playing with it a little to see what looks best across multiple services (s3 and lambda are particularly interesting).

joguSD (approved):

joguSD:
Looks good!

PR #2235, opened 2020-12-01
feat: Adds keepends option to iter_lines in StreamingResponse

Darkheir:
This PR aims to add an optional `keepends` parameter to the `iter_lines` in `StreamingResponse`

From the python documentation `readline` and `readlines` should keep new line characters at the end of the string:

> f.readline() reads a single line from the file; a newline character (\n) is left at the end of the string, and is only omitted on the last line of the file if the file doesn’t end in a newline.

See https://docs.python.org/3/tutorial/inputoutput.html#methods-of-file-objects

Adding this option would make it easier to get this behavior for classes relying on `StreamingResponse`.

To not break the current implementation, this parameter is set to `False` by default.

### DIFF
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -101,7 +101,7 @@ class StreamingBody(object):
 
     next = __next__
 
-    def iter_lines(self, chunk_size=1024):
+    def iter_lines(self, chunk_size=1024, keepends=False):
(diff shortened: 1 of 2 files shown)


nateprewitt (approved):
Hi @Darkheir, I think this looks good. Thanks for putting this together!

PR #2237, opened 2020-12-02
Allow exclude extras from user agent

vz10:
Add ability to exclude extras from UserAgent

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -427,7 +427,7 @@ class Session(object):
                 'credential_provider').load_credentials()
         return self._credentials
 
-    def user_agent(self):
+    def user_agent(self, truncate=False):
         """
         Return a string suitable for use as a User-Agent header.
         The string will be of the form:
@@ -449,7 +449,12 @@ class Session(object):
         If ``user_agent_extra`` is not empty, then this value will be
         appended to the end of the user agent string.
 
+        If ``truncate`` is True we return only agent_name and agent_version
+        it can be done for security reasons
+
         """
+        if truncate:
+            return '%s/%s' % (self.user_agent_name, self.user_agent_version)
         base = '%s/%s Python/%s %s/%s' % (self.user_agent_name,
                                           self.user_agent_version,
                                           platform.python_version(),
--- a/tests/functional/test_credentials.py
+++ b/tests/functional/test_credentials.py
@@ -24,7 +24,9 @@ import sys
 from dateutil.tz import tzlocal
 from botocore.exceptions import CredentialRetrievalError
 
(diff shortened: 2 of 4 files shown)


kyleknap (commented):
Looks fine. Just had one comment about the interface/user agent value we produce. Other than that let's get the IMDS user agent updated for the credential provider and make sure to have a test for that.

botocore/session.py
```diff
@@ -457,7 +457,7 @@ def user_agent(self):
                                           platform.release())
         if os.environ.get('AWS_EXECUTION_ENV') is not None:
             base += ' exec-env/%s' % os.environ.get('AWS_EXECUTION_ENV')
-        if self.user_agent_extra:
```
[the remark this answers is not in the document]
stealthycoin: Yea I agree lets go as minimal as possible.

stealthycoin:
Will need to update here: https://github.com/boto/botocore/blob/v2/botocore/credentials.py#L83

tests/unit/test_credentials.py
```diff
@@ -1716,6 +1718,17 @@ def test_custom_cache(self):
         cache = resolver.get_provider('assume-role').cache
         self.assertIs(cache, custom_cache)
 
+    @mock.patch('botocore.awsrequest.AWSRequest')
+    def test_imds_use_truncated_user_agent(self, request):
+        session = Session()
```
stealthycoin, line 1723: Does it not work to use the pattern the prior tests use where we pull out a provider using 

```python
        resolver = credentials.create_credential_resolver(self.session)
        provider = resolver.get_provider('iam-role')
        provider.load()
```

I would imagine we shouldn't have to mock out the awsrequest and instead be able to inject mocks.

stealthycoin (commented):
The implementation looks good, I just had a comment about the tests.

stealthycoin (approved):
Looks good.

tests/functional/test_credentials.py
```diff
@@ -828,3 +828,16 @@ def test_credential_process_returns_error(self):
             reg = r"(?s)^((?!b').)*$"
             with self.assertRaisesRegexp(CredentialRetrievalError, reg):
                 session.get_credentials()
+
+
+class TestInstanceMetadataFetcher(unittest.TestCase):
```
kyleknap, line 833: Let's subclass from `BaseSessionTest` and use its session if we can. Mainly there are some environment variables that can be set to affect behavior of this provider and we should make sure they do not get inadvertently set.

kyleknap (commented):
Looks fine. Just had one more suggestion

kyleknap (approved):
Thanks! :ship:

PR #2240, opened 2020-12-04
Add proxies_config to botocore config

nateprewitt:
This is a continuation of #2219. Adding a subset of possible urllib3 configs to botocore's configuration mechanisms. This will allow AWS CLI to take advantage of these changes as well.


botocore/config.py
```diff
@@ -63,6 +63,26 @@ class Config(object):
...
+
+        * 'proxy_client_cert' -- The path to a certificate for proxy
+          TLS client authentication.
+
+          When a str is provided it is treated as a path to a proxy client
+          certificate. When a two element tuple is provided, it will be
+          interpreted as the path to the client certificate, and the path
+          to the certificate key.
+
+        * 'proxy_use_forwarding_for_https' -- For HTTPS proxies,
```
joguSD, line 81: Might want to note what type is expected here -- it's not immediately clear.

botocore/configprovider.py
```diff
@@ -127,7 +127,18 @@
...
+# used to configure how botocore interacts with proxy setups while
+# sending requests.
+DEFAULT_PROXIES_CONFIG_VARS = {
+    'proxy_ca_bundle': ('proxy_ca_bundle', 'AWS_PROXY_CA_BUNDLE', None, None),
+    'proxy_client_cert': (
+        'proxy_client_cert',
+        'AWS_PROXY_CLIENT_CERT', None, None),
+    'proxy_use_forwarding_for_https': (
+        'proxy_use_forwarding_for_https',
+        'AWS_PROXY_FORWARDING_FOR_HTTPS', None, None),
```
joguSD, line 140: Curious if we want to use `ensure_bool` for this one?

botocore/httpsession.py
```diff
@@ -224,6 +247,26 @@ def _setup_ssl_cert(self, conn, url, verify):
             conn.cert_reqs = 'CERT_NONE'
             conn.ca_certs = None
 
+    def _setup_proxy_ssl_context(self, proxies_settings):
+        proxy_ca_bundle = proxies_settings.get('proxy_ca_bundle')
+        proxy_cert = proxies_settings.get('proxy_client_cert')
+        if proxy_ca_bundle is None and proxy_cert is None:
+            return None
```
joguSD, line 254: What's the behavior around creating a proxy manager with an SSL context of `None`?

botocore/httpsession.py
```diff
@@ -224,6 +247,26 @@ def _setup_ssl_cert(self, conn, url, verify):
...
+
+        context = self._get_ssl_context()
+        try:
```
joguSD, line 257: Does urllib3 not have a similar `cert_file` and `key_file` kwarg for the proxies and that's why we're doing this ourselves?

botocore/httpsession.py
```diff
@@ -201,6 +223,7 @@ def _get_proxy_manager(self, proxy_url):
             proxy_headers = self._proxy_config.proxy_headers_for(proxy_url)
             proxy_manager_kwargs = self._get_pool_manager_kwargs(
                 proxy_headers=proxy_headers)
+            proxy_manager_kwargs.update(**self._proxies_kwargs)
```
joguSD, line 226: Does this get ignored if they're using an older version of `urllib3`? Do we want to be more opinionated on what happens in that case?

joguSD (commented):
Overall seems reasonable, will try and pull this down to test it more thoroughly.

botocore/configprovider.py
nateprewitt: I had that initially but I'm not sure we do. We start indiscriminately sending the kwarg to urllib3 as False. I defaulted us to sending nothing which seemed safer.

botocore/httpsession.py
nateprewitt: They're absorbed into ` **connection_pool_kw` on older versions of urllib3 and ignored. I tested against the versions we support and things worked as expected.

botocore/httpsession.py
nateprewitt: That's the default arg for urllib3, but we actually strip the entire kwarg in `_proxies_kwargs` before calling urllib3, so this is a no-op.

botocore/httpsession.py
nateprewitt: Right, they only have that on the PoolManager class. They expect a pre-constructed `SSLContext` for the ProxyManager.

botocore/httpsession.py
```diff
@@ -224,6 +247,26 @@ def _setup_ssl_cert(self, conn, url, verify):
...
+    def _setup_proxy_ssl_context(self, proxies_settings):
+        proxy_ca_bundle = proxies_settings.get('proxy_ca_bundle')
+        proxy_cert = proxies_settings.get('proxy_client_cert')
+        if proxy_ca_bundle is None and proxy_cert is None:
+            return None
+
+        context = self._get_ssl_context()
+        try:
+            if proxy_ca_bundle is not None:
+                context.load_verify_locations(capath=proxy_ca_bundle)
```
jalopezsilva, line 259: Did you mean ``cafile`` instead of ``capath`` given that we're talking about a bundle? (generally this a single file containing root/intermediary certificates)

jalopezsilva (commented):
Overall looks good to me! I found one bug on my end to end testing and left a comment. (related to ``capath`` usage vs ``cafile``)

joguSD (approved):

botocore/httpsession.py
nateprewitt: Great catch, @jalopezsilva. Latest revision should have that updated and I think we're pretty close to getting this merged.

shepazon:
Documentation questions, since I've been asked to update the "Shared Configuration and Credentials" guide:

* The `proxies` setting is not covered in this documentation, despite having been added something like four years ago. Since `proxies_config` works with that, does that mean that both need to be documented in this guide or neither do?
* If this does indeed need to be added to the shared configuration guide, I could use some guidance in terms of which services the proxy configuration apply to.

I can't really do much until I'm clear on these points.

PR #2250, opened 2020-12-14
Fix timestamp format for SSOCredentialFetcher

nateprewitt:
The timestamp format for SSO credentials was originally intended to be an [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) style timestamp in UTC. Due to a typo in the original AssumeRole implementation, the timezone is being converted to `UTC` instead of a literal `Z`. We're going to correct the SSO implementation since the current timestamp is incompatible with the other AWS SDKs. The AssumeRole timestamp will be left unchanged for now to preserve as much backwards compatibility as possible.

Customers that are upgrading should have their timestamp format updated on their next login or credential refresh with SSO.

### DIFF
--- a/.changes/next-release/bugfix-SSO-69789.json
+++ b/.changes/next-release/bugfix-SSO-69789.json
@@ -1,0 +1,5 @@
+{
+  "description": "Fixed timestamp format for SSO credential expirations",
+  "type": "bugfix",
+  "category": "SSO"
+}
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1980,6 +1980,8 @@ class CredentialResolver(object):
 
 
 class SSOCredentialFetcher(CachedCredentialFetcher):
+    _UTC_DATE_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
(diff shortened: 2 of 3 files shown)


benkehoe:
Technically, shouldn't `_serialize_rfc3339_timestamp` verify something like `value.tzinfo == utc()` before writing the zulu time format, and otherwise write it out with the offset from the tzinfo?

miketheman:
> Technically, shouldn't `_serialize_rfc3339_timestamp` verify something like `value.tzinfo == utc()` before writing the zulu time format, and otherwise write it out with the offset from the tzinfo?

It appears that the value being passed in I'd generated from a timestamp, and the tz is being set as utc. 
So long as this remains a private function, and is called with a value set correctly, then it should be fine, but it is good defensive practice to validate, or at least document in the function's docstring.

nateprewitt:
Agreed, we should add a docstring here for the private function so it's clear we're expecting the datetime to already be in UTC.

kyleknap (approved):
Thanks! :ship:

PR #2254, opened 2020-12-18
Add JSON dumps callback to JSONFileCache

nateprewitt:
Adding in optional serialization function for `json.dumps` to allow modifications to ambiguous data types like `datetime`.


joguSD (approved):
Looks good to me.

PR #2266, opened 2021-01-08
Add retry state for non-existence

swetashre:
Fixes https://github.com/boto/boto3/issues/2701

### DIFF
--- a/botocore/data/ssm/2014-11-06/waiters-2.json
+++ b/botocore/data/ssm/2014-11-06/waiters-2.json
@@ -53,6 +53,11 @@
           "matcher": "path",
           "state": "failure",
           "argument": "Status"
+        },
+        {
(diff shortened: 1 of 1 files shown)


2or3:
In this solution, If `Run Command Id` dose't exist , always retry 20 times.
Thus I recommend to add `time.sleep(sleep_amount) ` before waiter loop start as below point.
Please consider my proposal. Thanks.
https://github.com/boto/botocore/blob/23ee17f5446c78167ff442302471f9928c3b4b7c/botocore/waiter.py#L319

stobrien89:
Hey @swetashre,

Hope you're doing well! We're requesting these upstream now— I've sent the service team the request and created a new issue to track this [here](https://github.com/aws/aws-sdk/issues/86).

PR #2267, opened 2021-01-08
Use correct form PermissionModels value "SELF_MANAGED"

brandonramsey:
In the documentation for _CloudFormation: create_stack_set_ and _CloudFormation: update_stack_set_, the listed default value is incorrect -- says default is `SELF-MANAGED` but it should be `SELF_MANAGED`


swetashre:
@brandonramsey - Thank you for the contribution. This file comes from upstream. I would let the service team know about the issue.

Issue #2249, opened 2020-12-13
Enable setting tcp_keepalive and duration_seconds via botocore.config.Config object


eyager1:
**Is your feature request related to a problem? Please describe.**
Network connection parameters `tcp_keepalive` and `duration_seconds` are only settable via `~/.aws/config` file and not the `botocore.config.Config` object.  Projects like home-assistant can only configure boto3 via the python API and not any configuration files.

**Describe the solution you'd like**
Expand `tcp_keepalive` and `duration_seconds` parameters to be settable via `botocore.config.Config`.

swetashre:
@eyager1 - Thank you for your post. Marking it as a feature request.

sebtrack:
I wondered how is it, that it can't be set programmatically but in a global config file!

yuji-hatakeyama:
I hope `tcp_keepalive = True` programmatically too !

frederick-wang:
I also hope tcp_keepalive = True programmatically too

macsir:
Hi, is there any update on this? Thanks.

dlm6693:
@macsir I've opened a PR that adds `tcp_keepalive`. We've decided not to do the same for `duration_seconds`. If you wish to see that added as well please comment here and/or open up a new ticket.

macsir:
Hi, @dlm6693  Thanks for that. Yes, tcp_keepalive is the one I need.

dlm6693:
Hi all. tcp_keepalive is now accessible on the config object thanks to https://github.com/boto/botocore/pull/2766. If you would like to see duration_seconds added as well, please open up a new ticket.

PR #2268, opened 2021-01-11
Add `tcp_keepalive` as AWS_TCP_KEEPALIVE env var

ryansb:
Related to #2249 - adds tcp_keepalive as an environment variable via the new `config_store` way of handling variables.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -398,8 +398,12 @@ class ClientArgsCreator:
         scoped_keepalive = scoped_config and self._ensure_boolean(
             scoped_config.get("tcp_keepalive", False)
         )
-        # Enables TCP Keepalive if specified in client config object or shared config file.
-        if client_keepalive or scoped_keepalive:
+        envvar_keepalive = self._config_store.get_config_variable(
+            'tcp_keepalive'
+        )
+        # Enables TCP Keepalive if specified in client environment, config
+        # object, or shared config file.
+        if client_keepalive or scoped_keepalive or envvar_keepalive:
             socket_options.append((socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1))
         return socket_options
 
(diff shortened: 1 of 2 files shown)


miketheman:
cc @dlm6693 This may be of interest

devesch:
What's holding this to reach production?

HayesData:
I spent some time on a conference call /w AWS support last week and they openly acknowledged this needs to go out, would be super helpful if this made it to a stable release as this is the source of a lot of pain for some Lambda applications

dlm6693:
To everyone showing interest in this pull request, we cannot merge this until it can be supported in all AWS SDKs. It's our policy that  all AWS_ environment variables are supported in all SDKs.

I will touch base internally with the team and see if this something we can move forward with. Thank you!

miketheman:
@dlm6693 thanks for the update!
Would you consider a near term specific environment variable? Precedent has been set in the node ask with a similar value years ago `AWS_NODEJS_CONNECTION_REUSE_ENABLED`

This could help solve the pain folks have now while we wait for all SDKs to have this ability.

dlm6693:
@miketheman we consider adding it to the client config object as the short term fix. I don't think we'll consider a more specific environment variable as a short term option even with the similar one for JS.

ronyyosef:
Any news about it?

RyanFitzSimmonsAK:
Hey everyone, thanks for opening this pull request and all the discussion.  We're going to close this pull request as not planned.  As mentioned earlier, this would need to be a cross-SDK environment variable, and we'd like to avoid further expanding the long list of existing environment variables, particularly for an option that is already supported in the client configuration.

miketheman:
> particularly for an option that is already supported in the client configuration.

This kind of response makes me sad.

It assumes that client configuration is as easy to supply as an environment variable, which is most definitely is not.

Thanks for sitting on this for years before deciding that it's not going to happen.

PR #2276, opened 2021-01-21
Fall back to Transfer-Encoding 'chunked' 

vz10:
Fall back to Transfer-Encoding 'chunked' if AWSRequest body is not seekable stream

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -11,6 +11,7 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import io
 import sys
 import logging
 import functools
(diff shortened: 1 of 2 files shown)


botocore/awsrequest.py
```diff
@@ -404,12 +405,18 @@ def _determine_content_length(self, body):
...
-            body.seek(orig_pos)
-            return end_file_pos - orig_pos
-
+            try:
+                orig_pos = body.tell()
+                body.seek(0, 2)
+                end_file_pos = body.tell()
+                body.seek(orig_pos)
+                return end_file_pos - orig_pos
+            except io.UnsupportedOperation:
```
nateprewitt, line 414: Are we sure this is the right exception? We should never be receiving a Base IO class like io.BufferedIOBase since they're non-public constructors. Should this be one of the super classes (OSError or ValueError)?

vz10: If you do something like this:
```
import requests
req = requests.Request('GET', "http://amazon.com").prepare()
resp = session.send(req, stream=True)
```

and pass `resp.raw` as a request body you'll get to the state where

```
hasattr(resp.raw, 'seek')
True
```
but
```
resp.raw.seek(0, 2)
  File "<input>", line 1, in <module>
io.UnsupportedOperation: seek
```

nateprewitt: I think in this specific case, we're coding around an unintentional mistake in urllib3 that's lingering for backwards compatibility reasons.

This _might_ introduce some issues with the `Response` object in Requests going out of scope while we're still reading, resulting in truncated data. That's hopefully a slim edge case though, so it may not be an issue.

I think overall this looks straight forward enough.

nateprewitt (approved):

PR #2280, opened 2021-01-25
Add exponential backoff to retries in IMDSFetcher

eltoder:
As noted by @JordonPhillips[^1], retries should use a backoff. This
should help to avoid overwhelming IMDS and exacerbating problems that
lead to the initial failure.

[^1] https://github.com/boto/boto3/issues/1751#issuecomment-434496999

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -315,23 +315,25 @@ class IMDSFetcher(object):
     _RETRIES_EXCEEDED_ERROR_CLS = _RetriesExceededError
     _TOKEN_PATH = 'latest/api/token'
     _TOKEN_TTL = '21600'
+    _MAX_SLEEP_TIME = 1
 
     def __init__(self, timeout=DEFAULT_METADATA_SERVICE_TIMEOUT,
                  num_attempts=1, base_url=METADATA_BASE_URL,
-                 env=None, user_agent=None, config=None):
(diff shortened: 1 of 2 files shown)


eltoder:
Hi @kdaily,

Could you point me to any documentation where this recommendation is stated? The only documentation I found [1] does not explain why and when one would chose `AWS_METADATA_SERVICE_TIMEOUT` vs `AWS_METADATA_SERVICE_NUM_ATTEMPTS`. Does this mean that `AWS_METADATA_SERVICE_NUM_ATTEMPTS` should never be used? Is it deprecated?

As long as `AWS_METADATA_SERVICE_NUM_ATTEMPTS` is supported, is there any reason not to have back off? There's a similar logic already in `ContainerMetadataFetcher` below in the same file.

[1] https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html

eltoder:
@kdaily,
>In cases like these, retries will not occur as read timeouts are not retried for IMDS credential fetching.

As far as I can tell, `ReadTimeoutError` is not retried in `_fetch_metadata_token`, but it is retried in `_get_request`. The latter is used in `_get_iam_role` and `_get_credentials`. Additionally, these 2 functions have application-level validation of response that they got from IMDS (_needs_retry_for_role_name and _needs_retry_for_credentials) and will retry if they did not like it.

PR #2178, opened 2020-10-12
Remove inheritance from requests exceptions

vz10:

PR #2282, opened 2021-01-28
Remove integration test for old timeout exception

vz10:


kyleknap (approved):
Looks good. Just noting for reference that this test is no longer needed because of this change: https://github.com/boto/botocore/pull/2178

PR #2285, opened 2021-01-29
add debug logging statement to print certificate path

kdaily:
Print out the certificate path when in debug mode.

### DIFF
--- a/botocore/httpsession.py
+++ b/botocore/httpsession.py
@@ -45,7 +45,10 @@ def get_cert_path(verify):
     if verify is not True:
         return verify
 
-    return where()
+    cert_path = where()
+    logger.debug("Certificate path: {0}".format(cert_path))
+
+    return cert_path
 
 
 def create_urllib3_context(ssl_version=None, cert_reqs=None,


botocore/httpsession.py
```diff
@@ -45,6 +45,9 @@ def get_cert_path(verify):
     if verify is not True:
         return verify
 
+    cert_path = where()
```
vz10, line 48: Probably makes sense to return 'cert_path' instead of calling where() twice.

vz10 (commented):
Looks good. One small NIT comment.

botocore/httpsession.py
kdaily: 🤦🏻 forgot to push that, corrected. Thanks!

vz10 (approved):
🚢

botocore/httpsession.py
```diff
@@ -45,7 +45,10 @@ def get_cert_path(verify):
     if verify is not True:
         return verify
 
-    return where()
+    cert_path = where()
+    logger.debug("Certificate path: {0}".format(cert_path))
```
nateprewitt, line 49: Just as a stylistic note, we don't need to enumerate formatting args anymore and can just use `{}` or `{cert_path}`. The numbering requirement was dropped in Python 2.6.

nateprewitt (approved):

PR #2286, opened 2021-02-02
S3 arn custom endpoint support

joguSD:
This allows the S3 and S3 control clients to provide custom endpoints when providing arn based resources.


nateprewitt (approved):
:shipit:

PR #2292, opened 2021-02-09
Update service-2.json

rkulagowski:
Change the wording slightly: "can be called only" -> "can only be called"


kdaily:
Hi @rkulagowski, thanks for this contribution. This file is autogenerated, so any changes through PRs will get overwritten. I'll pass this on to the documentation team.

PR #2301, opened 2021-02-16
Unify json imports.

hellysmile:
Hey, I've found that `json` imports don't follow same approach

Sometimes it is directly `import json`, but sometimes `from botocore.compat import json`

This PR unifies json imports across all modules

Using local `compat` module looks better for me, as allows to change json behavior in one place if it is eventually needed

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -22,7 +22,6 @@ from operator import itemgetter
 import functools
 import time
 import calendar
-import json
 
 from botocore.exceptions import NoCredentialsError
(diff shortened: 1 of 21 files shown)


akx:
This would be pretty useful to be able to e.g. switch in the `orjson` JSON parser!

RyanFitzSimmonsAK:
Hi @hellysmile, thanks for this pull request and for your patience.  After review and consideration with the team, we've made the decision to close this pull request as not planned.

PR #2305, opened 2021-02-20
Create SECURITY.md

kdaily:
Add a security policy that will show up in the new issue template list.

Encourage users NOT to open an issue for something that is security related to be sure that it is responded to as quickly as possible.


nateprewitt (approved):
:shipit:

PR #2312, opened 2021-03-05
Add support for doubles in argument generation

joguSD:
This adds support for the shape type `double` in argument generation. This ensures things like `--generate-cli-skeleton output` work as otherwise the stubbed response fails validation due to these values being generated as `None` instead of `0.0`.


nateprewitt (approved):
Looks fine to me.

PR #2314, opened 2021-03-08
Fix requirements for WaiterError

zdutta:
Revise requirements for raising a WaiterError.

### DIFF
--- a/botocore/waiter.py
+++ b/botocore/waiter.py
@@ -75,6 +75,13 @@ def create_waiter_with_client(waiter_name, waiter_model, client):
     )
 
 
+def is_valid_waiter_error(response):
+    error = response.get('Error')
+    if isinstance(error, dict) and 'Code' in error:
+        return True
+    return False
+
+
 class NormalizedOperationMethod(object):
     def __init__(self, client_method):
         self._client_method = client_method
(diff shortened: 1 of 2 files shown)


botocore/waiter.py
```diff
@@ -324,12 +327,17 @@ def wait(self, **kwargs):
                 if 'Error' in response:
```
joguSD, line 327: I don't think it makes sense to treat any response with an `Error` key as an error In this case we should be treating it as a success response, since from an SDK/HTTP perspective the API call was successful it just happens to have a member named `Error`.

botocore/waiter.py
```diff
@@ -75,6 +75,12 @@ def wait(self, **kwargs):
     )
 
 
+def is_valid_waiter_error(response):
+    error = response.get('Error')
+    if isinstance(error, dict) and 'Code' in error:
+        return True
+
```
nateprewitt, line 82: Generally, it's best to have boolean functions like this to return explicitly `True` and `False`, instead of `True` and `None` in this case. Could we get an explicit default return value?

nateprewitt (changes requested):
Just one minor tweak.

nateprewitt (changes requested):

joguSD (approved):

nateprewitt (approved):

PR #2325, opened 2021-03-23
Escape line endings in S3 PutBucketLifecycleConfiguration requests

joguSD:
See previous PR for background: https://github.com/boto/botocore/pull/2309

This applies the same customization to PutBucketLifecycleConfiguration


nateprewitt (approved):
Looks good. Do we have any way to detect where this is needed as operation surface area grows for S3?

PR #2328, opened 2021-03-25
Update _retries.json for servicequotas

noahrossi:
We use the service-quotas API to enumerate all of the quotas in our account and noticed that boto3 was failing to retry after a `TooManyRequestsException` with HTTP status code `400`. After some digging, I realized that the retry checker was only checking for a `TooManyRequestsException` with code `429`, but not with code `400` which is what the API was giving us.

It looks like you've already added the configuration here for the `organizations` service, so I'm just adding `too_many_requests_400` as a default retry now.

Let me know what you think.

Thanks,
Noah

### DIFF
--- a/botocore/data/_retry.json
+++ b/botocore/data/_retry.json
@@ -287,6 +287,20 @@
           }
         }
       }
+    },
+    "servicequotas": {
+      "__default__": {
+        "policies": {
+          "too_many_requests_400": {
+            "applies_when": {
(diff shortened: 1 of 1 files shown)


zdutta:
Hi @noahrossi, can you share a snippet of the expanded exception you are receiving. Typically `TooManyRequestsException` exceptions should have a code of 429.

PR #2330, opened 2021-03-26
CRT Signer Integration

nateprewitt:
**Note: The CRT integration in botocore is currently a beta feature that will be seeing iteration in the near future. You can enable these features for testing with by pip installing `botocore[crt]`. This will add an additional depdendency ([`awscrt`](https://github.com/awslabs/aws-crt-python)) to your environment, enabling these new features. This should not be considered "stable" for production at this time.**

This PR will be the first integration of CRT functionality into Botocore. A new optional module for CRT-based signers is being added to `auth.py`, allowing us to utilize a more uniform set of functionality in the AWS SDKs. The end goal will be to provide more performant, streamlined, and consistent SDKs across AWS.


tests/unit/auth/test_sigv4.py
```diff
@@ -144,17 +144,35 @@ def _test_signature_version_4(test_case):
...
+                     test_case.raw_request, 'authheader')
+
+@requires_crt
+def _test_crt_signature_version_4(test_case):
+    test_case = _SignatureTestCase(test_case)
+    request = create_request_from_raw_request(test_case.raw_request)
 
+    # Use CRT logging to diagnose interim steps (canonical request, etc)
+    # import awscrt.io
+    # awscrt.io.init_logging(awscrt.io.LogLevel.Trace, 'stdout')
```
joguSD, line 173: Side question: Is CRT logging still limited to `stdout`/`stderr`?

tests/unit/crt/auth/test_crt_signers.py
```diff
@@ -0,0 +1,34 @@
...
+    TestS3SigV4Auth, TestSigV4Presign, TestSigV4Resign
+)
+
+
+@requires_crt
+class TestCrtS3SigV4Auth(TestS3SigV4Auth):
+    # Repeat TestS3SigV4Auth tests, but using CRT signer
+
+    def setUp(self):
+        AuthClass = botocore.crt.auth.CrtS3SigV4Auth
```
joguSD, line 14: Is there any reason to to this as part of `setUp` and not just as part of the class definition like we did for the super class?

botocore/compat.py
```diff
@@ -349,3 +349,10 @@ def get_tzinfo_options():
     from collections.abc import MutableMapping
 except ImportError:
     from collections import MutableMapping
+
+# Detect if CRT is available for use
+try:
+    import awscrt.auth
+    HAS_CRT = True
```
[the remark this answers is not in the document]
nateprewitt: Yep, completely agree on the user-agent addition. I think we can get that added. For the environment variable opt out, I think that's reasonable to give the option for a quick fix. You could similarly uninstall the crt from the workspace, but I guess there may be a future where the Python CRT is used across multiple tools, not just boto3.

We don't really have access to the providers here, so we could do a one off grab from os.environ inside of this try/except closure, otherwise I'll likely need to add a function somewhere else that we use instead of `HAS_CRT` to evaluate all required criteria for CRT use.

botocore/session.py
```diff
@@ -214,6 +214,17 @@ def _create_csm_monitor(self):
             return handler
         return None
 
+    def _get_crt_version(self):
+        try:
+            import pkg_resources
+            return pkg_resources.get_distribution("awscrt").version
```
joguSD, line 220: Hmm, a little unfortunate we have to do this. We should try and get the version directly on the module and rip this out when we get the chance.

joguSD (approved):

botocore/session.py
nateprewitt: Yep, I filed awslabs/aws-crt-python#257 to try to get us off of this syntax. It's our only option right now though unfortunately.

PR #2333, opened 2021-03-30
use modern print in example

sir-sigurd:


nateprewitt (approved):
Sure, this seems reasonable. Thanks @sir-sigurd!

PR #2336, opened 2021-04-02
Waiter for Stepfunction Execution

yasver3474:
#### Waiter for Step Function

Noticed stepfunctions do not have any waiters. Needed this waiter as I have a lambda function that calls the start_execution and needs to wait for the execution of the stepfunction to be completed.


stobrien89:
Hi @yasver3474,

Thanks for the pull request!  Responsibility for waiters and paginators has been shifted to service teams, as they know best what the appropriate number of timeouts, retries, etc. should be.  Additionally, service teams can distribute their waiters and paginators across all of the tools and SDKs that use them, so customers get a consistent experience regardless of language or environment.  

We've been moving waiter and paginator requests to our SDK shared repo for tracking and submitting internal tickets to service teams for implementation.  We typically like to confirm that requests are for widely used use-cases— would you be able to share in detail your specific use case for this?

PR #2340, opened 2021-04-08
check env config for ca bundle when verify is True

### LINKED ISSUE #2339: check for CA Bundle when True is specified for verify
**Describe the bug**

The `verify` argument when creating a client is documented to be of type `boolean/string` and defaults to `None`. While the longer description explains that it can take on a value of `False` or a path to a CA bundle, it currently does not allow you to pass a value of `True` -- in the case that `True` is passed, the current implementation will not check the `ca_bundle` configuration variable [here](https://github.com/boto/botocore/blob/develop/botocore/session.py#L801). 

This is exacerbated when using other tools such as MLFlow and DVC that allow you to swap out storage interfaces. In such tools you can also set a `verify` parameter to `True`, `False` or `path/to/ca/bundle` however you cannot specify `None`. In this case it is impossible to use Boto3 when a CA Bundle is required to be used.

The solution to this issue is extremely simple, just checking for `verify` being either `None` or `True` [here](https://github.com/boto/botocore/blob/develop/botocore/session.py#L801)

**Steps to reproduce**
If you're behind a corporate firewall that requires custom cert bundle, try to create a session with `verify` set to `True` and it will not load the CA bundle from the env vars.

**Expected behavior**
When a user passes `True` it should work the same as passing `None`

**Debug logs**

koelemay:
Closes #2339

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -747,6 +747,8 @@ class Session(object):
             * path/to/cert/bundle.pem - A filename of the CA cert bundle to
               uses.  You can specify this argument if you want to use a
               different CA cert bundle than the one used by botocore.
+            * True - check environment configuration for CA cert bundle
+            * None - check environment configuration for CA cert bundle
 
         :type endpoint_url: string
         :param endpoint_url: The complete URL to use for the constructed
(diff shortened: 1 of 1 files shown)


RyanFitzSimmonsAK:
Hi @koelemay, thanks for this pull request and for your patience.  We're going to be closing this pull request, as it is actually a potentially breaking change.  Changing the behavior for `True` introduces the possibility of getting back new, unexpected values if something else in the environment is using requests, which is breaking.

PR #2344, opened 2021-04-12
Add Python 2.7 deprecation notice to readme

joguSD:


kyleknap (approved):
Looks good :ship:

nateprewitt (commented):
We're two months past the 3.4/3.5 deprecation and our supported versions are pretty widely documented. Do we want to replace the existing notification with this one to avoid information overload in the README?

joguSD:
@nateprewitt I'm going to leave the note in for now and remove it once we actually drop Python 2.7.

PR #2366, opened 2021-04-27
Fix document CodeCommit get_file response

kzcat:
Fix https://github.com/boto/boto3/issues/2329

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/codecommit.html#CodeCommit.Client.get_file
```
fileContent (bytes) --
The base-64 encoded binary data object that represents the content of the file.
```
But `fileContent` returns as base64 decoded binary data.
```python
import boto3
client = boto3.client('codecommit')
client.put_file(repositoryName='repo', filePath='path', branchName='main', parentCommitId='', fileContent='test')
client.get_file(repositoryName='repo', filePath='path')['fileContent'] 
# => b'test'
```

### DIFF
--- a/botocore/data/codecommit/2015-04-13/service-2.json
+++ b/botocore/data/codecommit/2015-04-13/service-2.json
@@ -4414,7 +4414,7 @@
         },
         "fileContent":{
           "shape":"FileContent",
-          "documentation":"<p>The base-64 encoded binary data object that represents the content of the file.</p>"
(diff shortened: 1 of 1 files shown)


kdaily:
Hi @kzcat,

Thanks for identifying this. Unfortunately this documentation comes from an upstream source and is autogenerated. I will check with the CodeCommit team on this, though.

PR #2368, opened 2021-04-28
make StreamingBody file-like

mdavis-xyz:
This is a short-term solution to #879 .

It seems that converting StreamingBody to a subclass of an `io.IOBase` would take a while. In the mean time, let's just make `t = TextIOWrapper(StreamingBody(), encoding="utf-8")` work, by adding `readable()` and other attributes/methods.


mdavis-xyz:
This PR has been open for a while now.
Can someone please merge this or give me feedback on why it can't be merged?

e.g. @nateprewitt @kyleknap , thanks

PR #2370, opened 2021-04-28
Fix FIPS and global endpoint behavior for S3 ARNs

vz10:


botocore/utils.py
```diff
@@ -1569,6 +1571,45 @@ def set_endpoint(self, request, **kwargs):
...
+                )
+            else:
+                raise UnsupportedS3AccesspointConfigurationError(
+                    msg=(
+                        'Invalid configuration, cross region Access Point ARN'
+                    )
+                )
+
+    def _validate_global_regions(self, request):
+        if self._s3_config.get('use_arn_region'):
```
kyleknap, line 1602: I'd assume we would want to use `self._s3_config.get('use_arn_region', True)` here since by default we use the arn region? Otherwise, commands like this no longer work as they used to resolve to the us-east-1 endpoint:
```
$ aws s3api get-object --bucket arn:aws:s3:us-east-1:123456789012:accesspoint:myendpoint --key foo --region s3-external-1 /tmp/foo

Unsupported configuration when using S3 access-points: Invalid configuration, client region is not a regional endpoint
```

botocore/utils.py
```diff
@@ -1569,6 +1571,45 @@ def set_endpoint(self, request, **kwargs):
...
+            return
+        client_region = request.context.get('client_region', '')
+        if client_region in ['aws-global', 's3-external-1']:
+            raise UnsupportedS3AccesspointConfigurationError(
+                msg=(
+                    'Invalid configuration, client region '
+                    'is not a regional endpoint'
+                )
```
kyleknap, line 1610: For all of the error messages that involve using incompatible region, it may be good to say in the error message what the client region that was received so it is easier to troubleshoot.

botocore/utils.py
```diff
@@ -1569,6 +1571,45 @@ def set_endpoint(self, request, **kwargs):
     def _use_accesspoint_endpoint(self, request):
         return 's3_accesspoint' in request.context
 
+    def _validate_fips_supported(self, request):
+        client_region = request.context.get('client_region', '')
```
kyleknap, line 1575: Is there a particular reason why region is pulled from the request instead of use the `self._region` property? It seems like a fair amount of the logic here uses that for the client region.

botocore/utils.py
```diff
@@ -1569,6 +1571,45 @@ def set_endpoint(self, request, **kwargs):
...
+        if 'fips' not in client_region:
+            return
+        if 'outpost_name' in request.context['s3_accesspoint']:
+            raise UnsupportedS3AccesspointConfigurationError(
+                msg=(
+                    'Invalid configuration Outpost Access Points '
+                    'do not support Fips- regions'
+                )
+            )
+        client_region = client_region.replace('fips-', '')
```
kyleknap, line 1585: It may be good to look for the `<region>-fips` pattern as well. It seems like some fips regions can be defined in that fashion and looks like the Ruby implementation is doing that as well: https://github.com/aws/aws-sdk-ruby/pull/2495/files#diff-10a2f251424fdb0be98d76e54c194c4d2eeebedb018faab5db57c22ffbc92a61R70

botocore/utils.py
```diff
@@ -1569,6 +1571,45 @@ def set_endpoint(self, request, **kwargs):
...
 
+    def _validate_fips_supported(self, request):
+        client_region = request.context.get('client_region', '')
+        if 'fips' not in client_region:
+            return
+        if 'outpost_name' in request.context['s3_accesspoint']:
+            raise UnsupportedS3AccesspointConfigurationError(
+                msg=(
+                    'Invalid configuration Outpost Access Points '
+                    'do not support Fips- regions'
```
kyleknap, line 1582: Let's make the casing consistent here and make it `FIPS` instead of `Fips-`.

botocore/utils.py
```diff
@@ -1569,6 +1571,45 @@ def set_endpoint(self, request, **kwargs):
...
+            )
+        client_region = client_region.replace('fips-', '')
+        if request.context['s3_accesspoint']['region'] != client_region:
+            if self._s3_config.get('use_arn_region'):
+                raise UnsupportedS3AccesspointConfigurationError(
+                    msg=(
+                        'Invalid configuration, '
+                        'FIPS region does not match ARN region'
+                    )
+                )
```
kyleknap, line 1593: Should we be erroring out when the region does not match but you opted into using the arn region? The general behavior looks like it has always been just use the region from the arn unless it is outside the partition of the configured client: https://github.com/boto/botocore/blob/92c63ea94542c5d54d582c50509be02b96d7a443/botocore/utils.py#L1580-L1589

kyleknap (commented):
Looks like a good start. Just had some questions and suggestions.

botocore/utils.py
vz10: I'm not sure. Looks like here we use opposite logic https://github.com/boto/botocore/blob/92c63ea94542c5d54d582c50509be02b96d7a443/botocore/utils.py#L1811

botocore/utils.py
vz10: Yes, but since `endpoints.json` doesn't have such entry for S3 it won't work  in any case

botocore/utils.py
kyleknap: I double checked this, the default is `True` for the S3 client, but for the S3 control client it looks like it defaults to `False`. I think we need to stick with the default for the S3 client. Mainly if we made the default `False` here, we would introduce a regression where we would now fast fail if:
1. The user has `AWS_S3_US_EAST_1_REGIONAL_ENDPOINT=regional` set (In v2, this environment variable no longer exists and essentially defaults to the `regional` value).
2. The user does not set a region.

```
$ AWS_S3_US_EAST_1_REGIONAL_ENDPOINT=regional aws s3api get-object --bucket arn:aws:s3:us-east-1:123456789012:accesspoint:myendpoint --key foo  /tmp/foo
Unsupported configuration when using S3 access-points: Client is configured to use the global psuedo-region "aws-global". When providing access-point ARNs a regional endpoint must be specified.
```
I pushed up the update in the latest commit to fix this

kyleknap:
I pushed up a commit to address the feedback that I had for the previous commits. Everything else before that looked fine.

joguSD (approved):

PR #2375, opened 2021-05-03
Fix not using min value in skeleton generation

vz10:


tim-finnigan:
Closing per team decision, can revisit the proposed changes in the future. Linking related issue for reference: https://github.com/aws/aws-cli/issues/2337

PR #2381, opened 2021-05-07
Fix rejecting URLs with unsafe characters in is_valid_endpoint_url()

### LINKED ISSUE #2377: FAIL: test_no_new_lines (tests.unit.test_utils.TestIsValidEndpointURL) ; with Python versions patched for bpo-43882
**Describe the bug**
The recent fix to [bpo43882](https://bugs.python.org/issue43882) has caused the following test to start to fail:
```
FAIL: test_no_new_lines (tests.unit.test_utils.TestIsValidEndpointURL)
  File "/tmp/portage/dev-python/botocore-1.20.64/work/botocore-1.20.64/tests/unit/test_utils.py", line 713, in test_no_new_lines
    self.assertFalse(is_valid_endpoint_url('https://foo.bar.com\nbar/'))
AssertionError: <re.Match object; span=(0, 14), match='foo.bar.combar'> is not false
```

The relevant behavior change in CPython is that all newlines (as well as CRs and tabs) are stripped from the URL, so they're not present in the split components anymore.

**Steps to reproduce**
Install Python 3.10.0b1 or 3.9 from git branch (the change is going to be the part of the next release), and run tests with it.

**Expected behavior**
Tests passing ;-). I suppose you could either check URLs for newlines early on, and reject all newlines, or agree with the new behavior of stripping them. In the latter case, it might make sense to add stripping inside botocore too, for consistent behavior across Python versions.

**Debug logs**
n/a

mgorny:
Detect unsafe characters in is_valid_endpoint_url()
and is_valid_ipv6_endpoint_url() early, in order to fix rejecting
invalid URLs with Python 3.9.5+ and other versions carrying bpo-43882
fix.  In these versions, urlsplit() silently strips LF, CR and HT
characters while splitting the URL, effectively disarming the validator
in botocore.

The solution is based on a similar fix in Django.

Fixes #2377

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -173,6 +173,10 @@ ZONE_ID_PAT = "(?:%25|%)(?:[" + UNRESERVED_PAT + "]|%[a-fA-F0-9]{2})+"
 IPV6_ADDRZ_PAT = r"\[" + IPV6_PAT + r"(?:" + ZONE_ID_PAT + r")?\]"
 IPV6_ADDRZ_RE = re.compile("^" + IPV6_ADDRZ_PAT + "$")
 
+# These are the characters that are stripped by post-bpo-43882 urlparse().
+UNSAFE_URL_CHARS = frozenset('\t\r\n')
+
+
 def ensure_boolean(val):
     """Ensures a boolean value if a string or boolean is provided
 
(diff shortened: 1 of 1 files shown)


nateprewitt (approved):
Looking over https://github.com/django/django/pull/14361, this looks reasonable. It has a slight behavior change from before where we were returning `None` in some of these cases but now return `False`. Given potential security impact, I think we're alright with that change though. It actually brings us more inline with what we've documented the behavior to be (`True` or `False`).

Thanks for the contribution, @mgorny!

mgorny:
Thanks for the merge!

PR #2385, opened 2021-05-12
Pin jinja2 to <3.0

stealthycoin:
Jinja2 3.0 drops support for python 3.5 and Sphinx doens't have a pin.
Markupsafe was also unbounded by jinja2, which means an incompatible
version is now pulled in and we need to put a ceiling pin on that as
well.

markupsafe>=1.1,<2.0
jinja2>=2.3,<3.0


nateprewitt (approved):
It looks like they're using `python_requires` in the latest release of Jinja2, so it _should_ avoid this requirement with a modern version of `pip`. That may not be the case for all environments building docs right now though, so this seems reasonable for the time being.

PR #2386, opened 2021-05-13
Fix STS regional endpoint handling if no explicit endpoint is defined

reegnz:
When defining AWS_STS_REGIONAL_ENDPOINTS=regional as an environment
variable, the appropriate regional endpoint should be selected.

This was not the case before this change, if no endoint_url was
pre-defined, it defaulted to the global endpoint.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -267,8 +267,8 @@ class ClientArgsCreator(object):
     def _compute_sts_endpoint_config(self, **resolve_endpoint_kwargs):
         endpoint_config = self._resolve_endpoint(**resolve_endpoint_kwargs)
         if self._should_set_global_sts_endpoint(
-                resolve_endpoint_kwargs['region_name'],
-                resolve_endpoint_kwargs['endpoint_url']):
+                endpoint_config['region_name'],
+                endpoint_config['endpoint_url']):
             self._set_global_sts_endpoint(
                 endpoint_config, resolve_endpoint_kwargs['is_secure'])
         return endpoint_config


kdaily:
Hi @reegnz,

I can't reproduce this. If I run the following test:

```
import boto3
boto3.set_stream_logger('')
sts = boto3.client('sts')
sts.get_caller_identity()
```

If I have `AWS_STS_REGIONAL_ENDPOINT` unset (default), I get the global endpoint:

```
2021-06-10 14:47:35,158 urllib3.connectionpool [DEBUG] Starting new HTTPS connection (1): sts.amazonaws.com:443
2021-06-10 14:47:35,708 urllib3.connectionpool [DEBUG] https://sts.amazonaws.com:443 "POST / HTTP/1.1" 200 442
2021-06-10 14:47:35,709 botocore.parsers [DEBUG] Response headers: {'x-amzn-RequestId': '5ac8dad0-9fee-4c06-bcb3-4242b9d0f125', 'Content-Type': 'text/xml', 'Content-Length': '442', 'Date': 'Thu, 10 Jun 2021 21:47:35 GMT'}
```

If I set `AWS_STS_REGIONAL_ENDPOINT=regional` and I am using region `us-west-2`, I get:

```
2021-06-10 14:47:06,969 urllib3.connectionpool [DEBUG] Starting new HTTPS connection (1): sts.us-west-2.amazonaws.com:443
2021-06-10 14:47:07,746 urllib3.connectionpool [DEBUG] https://sts.us-west-2.amazonaws.com:443 "POST / HTTP/1.1" 200 442
2021-06-10 14:47:07,746 botocore.parsers [DEBUG] Response headers: {'x-amzn-RequestId': '415432bd-8f0b-400b-b34e-0ae1babebcd7', 'Content-Type': 'text/xml', 'Content-Length': '442', 'Date': 'Thu, 10 Jun 2021 21:47:07 GMT'}
```

reegnz:
In the meantime I have figured out that you can set the region with `AWS_DEFAULT_REGION=us-east-1` and then it'll use the regional endpoint.

So the PR is unnecessary, but the default region env variable needs to be set (if not set otherwise, eg. the profile).

PR #2398, opened 2021-05-28
Handling document types and adding C2J protocol tests

zdutta:
Introducing changes to handle document types. Enabling potential use of arbitrary JSON documents in request/response objects. Additional protocol testing included.

### DIFF
--- a/botocore/validate.py
+++ b/botocore/validate.py
@@ -127,6 +127,11 @@ class ValidationErrors(object):
         elif error_type == 'unable to encode to json':
             return 'Invalid parameter %s must be json serializable: %s' \
                 % (name, additional['type_error'])
+        elif error_type == 'invalid type for document':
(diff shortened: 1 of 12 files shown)


joguSD (approved):

botocore/validate.py
```diff
@@ -181,6 +188,25 @@ def _validate_jsonvalue_string(self, params, shape, errors, name):
         except (ValueError, TypeError) as e:
             errors.report(name, 'unable to encode to json', type_error=e)
 
+    def _validate_document(self, params, shape, errors, name):
```
mtdowling, line 191: +1, this method is critical for making sure we can evolve document types as needed. For example, we will likely eventually add support for serializing and (lazily) deserializing modeled values in and out of document types. I don't want us to be blocked on that feature due to any implicit behavior in how Python serializes values (like if it checks to_json methods or something implicit).

Issue #2397, opened 2021-05-28
S3 uses path-style URLs


AriFordsham:
**Describe the bug**
The models in botocore/data/s3/2006-03-01/service-2.json describe path-stype URLs, which are deprecated (<https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/>). This breaks [amazonka](https://github.com/brendanhay/amazonka) which relies on them, and cannot process the 301 redirects used in eu-west-2.

**Expected behavior**
The models should describe virtual-hosting URLs, which are current best practice for AWS.

AriFordsham:
Ah. I see. boto deals with this in code. https://github.com/boto/botocore/pull/52

kdaily:
Hi @AriFordsham,

As you've found this is implemented in the code. The models here come from upstream and are shared across all AWS SDKs. Can you clarify a line of the model that you're referring to? I'm not sure what I can do to make any changes here, as they come from the S3 team. Thanks!

AriFordsham:
https://github.com/boto/botocore/blob/2a04c69389446b6cffd6972f00aec5b74bdd5639/botocore/data/s3/2006-03-01/service-2.json#L20
https://github.com/boto/botocore/blob/2a04c69389446b6cffd6972f00aec5b74bdd5639/botocore/data/s3/2006-03-01/service-2.json#L35
https://github.com/boto/botocore/blob/2a04c69389446b6cffd6972f00aec5b74bdd5639/botocore/data/s3/2006-03-01/service-2.json#L46
...and ~91 others.

These  models describe path-style URLs, which are on the deprecation path. Boto corrects for this, and defaults to virtual-hosting style using hard-coded logic that corrects URLs before request.

The correction logic essentially does two tasks:
1. Prepends the bucket name to the hostname
2. Strips the bucket name prefix from the URL.

It would arguably cleaner design if this could be reflected in the model itself (This would also make it easier for third-party consumers of the models). Of course, this would be subject to the limitations of the model format - I don't know if this is documented anywhere.

(1) could be achieved by prepending the bucket name to https://github.com/boto/botocore/blob/2a04c69389446b6cffd6972f00aec5b74bdd5639/botocore/data/s3/2006-03-01/service-2.json#L6-L7, conditional on a URL style variable (or  prepending a variable which is conditionally set to either the bucket name or empty conditional on the path style).

[...]

PR #2400, opened 2021-05-30
Proposed change to reflect virtual-hosting style

AriFordsham:
Intended to demonstrate a proposal in #2397 - two-variable version


kdaily:
Hi @AriFordsham,

See comment here: https://github.com/boto/botocore/issues/2397#issuecomment-859923182

Issue #2401, opened 2021-05-31
NameError: name 'NoCredentialsError' is not defined


notoriou1pig:
Hi,
I think there is a oversight in botocore/crt/auth.py. It tries to raise NoCredenialsError, but because the exception is not included in the file. 

Here is the tail of my error log
```
[...]
  File "/path/to/python3.7/site-packages/botocore/signers.py", line 90, in handler
    return self.sign(operation_name, request)
  File "/path/to/python3.7/site-packages/botocore/signers.py", line 162, in sign
    auth.add_auth(request)
  File "/path/to/python3.7/site-packages/botocore/crt/auth.py", line 32, in add_auth
    raise NoCredentialsError()
NameError: name 'NoCredentialsError' is not defined
```

To fix it, I added the following line at top of my `botocore/crt/auth.py` file.
```
 from botocore.exceptions import NoCredentialsError
```
Here is the output after fixing the file :
```
[...]
  File "/path/to/python3.7/site-packages/botocore/signers.py", line 90, in handler
    return self.sign(operation_name, request)
  File "/path/to/python3.7/site-packages/botocore/signers.py", line 162, in sign
    auth.add_auth(request)
  File "/path/to/python3.7/site-packages/botocore/crt/auth.py", line 32, in add_auth
    raise NoCredentialsError()
botocore.exceptions.NoCredentialsError: Unable to locate credentials
```

You can easily reproduce with the following script :
```
import boto3
from botocore.config import Config

s3_config = Config(client_cert=("path/to/nothing", "path/to/something_else"))
s3_client = boto3.client('s3', config=s3_config)
s3_client.upload_file("/tmp/file", "bucket", "key") # Here, the file must exist else, another exception will raise. 
```

I hope you can fix it quick,
Thanks a lot,
notoriou1Pig =D

nateprewitt:
Resolved with #2402

PR #2402, opened 2021-05-31
Import NoCredentialsError in crt/auth.py

notoriou1pig:
Fix issue #2401 by importing the exception in the file.


nateprewitt (approved):
Great catch, thanks @notoriou1pig!

PR #2417, opened 2021-06-16
Add CRT auth package

kyleknap:
This pulls in CRT auth abstractions from the develop branch in order to have a starting to point to pull in future updates related to CRT auth. Note the code does not actually use any of the CRT auth abstractions when making API requests. The main difference in porting the package is that because CRT is a required dependency there is no branching logic for conditionally using CRT auth. Eventually, we should remove the pure-Python sigv4 auth abstractions and completely rely on the CRT auth abstractions for sigv4 signing.

For reference the changes should follow closely to the original PR that added CRT auth to the develop branch: https://github.com/boto/botocore/pull/2350


stealthycoin (approved):

nateprewitt (approved):
Looks like a clean port. Thanks!

botocore/crt/auth.py
```diff
@@ -0,0 +1,296 @@
...
+        return False
+
+    def _should_add_content_sha256_header(self, explicit_payload):
+        # Never add X-Amz-Content-SHA256 header
+        return False
+
+
+# Defined at the bottom of module to ensure all Auth
+# classes are defined.
+CRT_AUTH_TYPE_MAPS = {
```
nateprewitt, line 291: It doesn't look like we're actually using these anywhere? We'd need them to get wired up in auth.py, right?

nateprewitt: Nevermind, reread the initial PR. Forgot where we were at, so you can disregard this.

nateprewitt (approved):
Still looks good, thanks for rebasing!

PR #2429, opened 2021-06-29
Fix typo: sesssion -> session

gaisin:
Fix the typo `sesssion -> session`.

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -732,7 +732,7 @@ class Session(object):
 
         :type service_name: string
         :param service_name: The name of the service for which a client will
-            be created.  You can use the ``Sesssion.get_available_services()``
+            be created.  You can use the ``Session.get_available_services()``
             method to get a list of all available service names.
 
         :type region_name: string


botocore/data/devicefarm/2015-06-23/service-2.json
```diff
@@ -4260,7 +4260,7 @@
         },
         "endpoint":{
           "shape":"String",
-          "documentation":"<p>The endpoint for the remote access sesssion.</p>"
+          "documentation":"<p>The endpoint for the remote access session.</p>"
```
nateprewitt, line 4263: Would you mind removing this change? Anything under the `/data` directory is generally managed by the service team and will need to be updated upstream to avoid this change being overwritten in the next release.

nateprewitt (changes requested):
Great catch, @gaisin! We just need one minor change and can get this merged. We'll follow up with the service team to get the service documentation fixed.

botocore/data/devicefarm/2015-06-23/service-2.json
gaisin: @nateprewitt Sure, I have removed changes from `/data` directory. Thank you for checking out the PR :)

nateprewitt (approved):
Thanks @gaisin!

Issue #2442, opened 2021-07-18
Add ability to specify `signature_version=UNSIGNED` using a string


Ark-kun:
**Describe the bug**
It looks like boto3 does not support public buckets by default (why?), so everyone has to create a specially configured client:
`boto3.client('s3', config=Config(signature_version=UNSIGNED))`

The popular MLFlow library has the following client initialization code (simplified):
```python
def create_client():
   return boto3.client('s3', config=Config(signature_version=os.environ.get("S3_SIGNATURE_VERSION", "s3v4")))
```

Boto `Config` and `RequestSigner` attribute `signature_version` has `str` type, but there does not seem to be any string that would map to the `UNSIGNED` python singleton object.

This makes accessing public MLFlow artifact buckets impossible.

**Steps to reproduce**

```python
os.environ["S3_SIGNATURE_VERSION"] = "unsigned"
...
client = boto3.client('s3', config=Config(signature_version=os.environ.get("S3_SIGNATURE_VERSION", "s3v4")))
```
Get error:
```
UnknownSignatureVersionError: Unknown Signature Version: unsigned.
```

**Expected behavior**

I expect `client = boto3.client('s3', config=Config(signature_version="unsigned"))` to succeed and work as expected.

**Logs**
Full stack trace
```

/opt/conda/lib/python3.7/site-packages/mlflow/store/artifact/s3_artifact_repo.py in list_artifacts(self, path)
    113         paginator = s3_client.get_paginator("list_objects_v2")
    114         results = paginator.paginate(Bucket=bucket, Prefix=prefix, Delimiter="/")
--> 115         for result in results:
    116             # Subdirectories will be listed as "common prefixes" due to the way we made the request
    117             for obj in result.get("CommonPrefixes", []):

/opt/conda/lib/python3.7/site-packages/botocore/paginate.py in __iter__(self)
    253         self._inject_starting_params(current_kwargs)
    254         while True:
--> 255             response = self._make_request(current_kwargs)
    256             parsed = self._extract_parsed_response(response)
    257             if first_request:

/opt/conda/lib/python3.7/site-packages/botocore/paginate.py in _make_request(self, current_kwargs)
    330 
    331     def _make_request(self, current_kwargs):
--> 332         return self._method(**current_kwargs)
    333 
    334     def _extract_parsed_response(self, response):

/opt/conda/lib/python3.7/site-packages/botocore/client.py in _api_call(self, *args, **kwargs)
[...]

PR #2448, opened 2021-07-28
Made it possible to specify UNSIGNED using a string

Ark-kun:
This fixes some incorrect type signatures and makes botocore easier to use in CLIs and other libraries that deal with non-Python input.

Fixes https://github.com/boto/botocore/issues/2442


tim-finnigan:
Thanks for creating this PR, but the team decided this should be closed. Followed up on corresponding issue here: https://github.com/boto/botocore/issues/2442#issuecomment-2203979888

PR #2455, opened 2021-08-12
fix-credential-process-without-stdin

kernd:
in the ProcessProvider class, the _retrieve_credentials_using method uses the self._popen to open a subprocess. If there is no stdin attached to currently running program, then this call can result in an "handle is invalid" error on Windows. Adding the parameter stdin=subprocess.PIPE fixes the issue for this case.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -980,6 +980,7 @@ class ProcessProvider(CredentialProvider):
         # command and all arguments as a list.
         process_list = compat_shell_split(credential_process)
         p = self._popen(process_list,
+                        stdin=subprocess.PIPE,
                         stdout=subprocess.PIPE,
                         stderr=subprocess.PIPE)
         stdout, stderr = p.communicate()


kernd:
here is an example traceback to show how this call can fail:

```  File "boto3\__init__.py", line 93, in client
  File "boto3\session.py", line 258, in client
  File "botocore\session.py", line 839, in create_client
  File "botocore\session.py", line 441, in get_credentials
  File "botocore\credentials.py", line 1975, in load_credentials
  File "botocore\credentials.py", line 961, in load
  File "botocore\credentials.py", line 980, in _retrieve_credentials_using
  File "subprocess.py", line 804, in __init__
  File "subprocess.py", line 1142, in _get_handles
OSError: [WinError 6] The handle is invalid```

kernd:
@nateprewitt do you know what are the next steps to get this looked at?

tim-finnigan:
Checking in as there has not been any activity here in a while. I could not find any other reports of this issue, is it still something you're encountering on new versions of Boto3? If so please create a issue for further review and investigation.

PR #2463, opened 2021-08-20
Return generic error response for errors with non-xml bodies

chenry3:
Some proxies may return a custom response in the HTTP body when
encountering an error response such as an HTTP/503.

Currently we assume that any error response with a body that does not
begin with "<html>" is XML which can lead to unhandled ResponseParserError
errors on proxies with custom body responses for things like a transient timeout.

Lets expand what we consider a "generic" error to include responses whose
body is not XML by ensuring it must start with a leading "<".
`
The other implementation option is to handle ResponseParserErrors in
botocore.parsers.QueryParser._do_error_parse and return _do_generic_error_parse
instead which I'd be happy to implement instead of preferred

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -281,7 +281,11 @@ class ResponseParser(object):
                 return True
 
             body = response['body'].strip()
-            return body.startswith(b'<html>') or not body
+            return (
+                body.startswith(b"<html>")
+                or not body.startswith(b"<")
+                or not body
+            )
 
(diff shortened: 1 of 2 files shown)


RyanFitzSimmonsAK:
Hey @chenry3, thanks for this pull request and for your patience.  I'm going to close this pull request, not because it doesn't have value, but because the implementation details are something we want to handle internally.  I'll be opening a backlog item for this.  Thanks again for contributing!

Issue #2468, opened 2021-08-26
get_object_lock_configuration computes wrong signature when using v2 authentication


lucasso:
**Describe the bug**
wrong StringToSign is produced by HmacV1Auth class when signing GetObjectLockConfiguration request

**Steps to reproduce**
```
import logging, botocore, botocore.session
session = botocore.session.get_session()
session.set_stream_logger('botocore', logging.DEBUG)
cl = session.create_client('s3', config=botocore.config.Config(signature_version="s3"))
cl.get_object_lock_configuration(Bucket="aaa")
```
gives **SignatureDoesNotMatch**
`botocore.exceptions.ClientError: An error occurred (SignatureDoesNotMatch) when calling the GetObjectLockConfiguration operation: The request signature we calculated does not match the signature you provided. Check your key and signing method.`

When using `signature_version="s3v4"` code works correctly and gives **AccessDenied**
 `botocore.exceptions.ClientError: An error occurred (AccessDenied) ` as expected (because "aaa" bucket is old but not mine)

The problem is that word "?object-lock" is not included in StringToSign by botocore, but it is included by AWS when computing signature. I've created a pull request which fixes this issue (see link below). Please review and merge.

**Expected behavior**
AccessDenied error instead of SignatureDoesNotMatch error

**Debug logs**
```
2021-08-26 07:33:19,183 - botocore.hooks - DEBUG - Event before-parameter-build.s3.GetObjectLockConfiguration: calling handler <function validate_bucket_name at 0x7f80a2691d50>
2021-08-26 07:33:19,183 - botocore.hooks - DEBUG - Event before-parameter-build.s3.GetObjectLockConfiguration: calling handler <bound method S3RegionRedirector.redirect_from_cache of <botocore.utils.S3RegionRedirector object at 0x7f80a0b84590>>
2021-08-26 07:33:19,183 - botocore.hooks - DEBUG - Event before-parameter-build.s3.GetObjectLockConfiguration: calling handler <bound method S3ArnParamHandler.handle_arn of <botocore.utils.S3ArnParamHandler object at 0x7f80a0b845d0>>
2021-08-26 07:33:19,184 - botocore.hooks - DEBUG - Event before-parameter-build.s3.GetObjectLockConfiguration: calling handler <function generate_idempotent_uuid at 0x7f80a2691bd0>
[...]

PR #2469, opened 2021-08-26
Fixes boto/botocore#2468 issue, corrected v2 signature of get_object_lock_configuration

lucasso:
"?object-lock" part of request URL is now correctly included in "string_to_sign" in HmacV1Auth signer

### DIFF
--- a/tests/unit/auth/test_signers.py
+++ b/tests/unit/auth/test_signers.py
@@ -98,7 +98,8 @@ class TestHMACV1(unittest.TestCase):
         # specified as query strings end up in the canonical resource.
         operations = ('acl', 'cors', 'lifecycle', 'policy',
                       'notification', 'logging', 'tagging',
-                      'requestPayment', 'versioning', 'website')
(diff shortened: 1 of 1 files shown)


nateprewitt (changes requested):
Hey @lucasso, thanks for the PR! Would you be able to add a test or two codifying this change? That will help ensure this behavior doesn't regress in the future.

lucasso:
Working on a unittest. I should finish in a week.

lucasso:
Added a unittest. Fails without a change with:
```
FAIL: test_bucket_operations (tests.unit.auth.test_signers.TestHMACV1)
  File "/home/vpnuser/repos/botocore/tests/unit/auth/test_signers.py", line 107, in test_bucket_operations
    self.assertEqual(cr, '/quotes?%s' % operation)
AssertionError: '/quotes' != '/quotes?object-lock'
- /quotes
+ /quotes?object-lock
```
and passes with the change

nateprewitt (approved):
Thanks @lucasso! I think this looks good.

PR #2472, opened 2021-08-27
Add automatic linting and formatting requirements to code base

nateprewitt:
This is a continuation of #2478 to add some more substantial framing for how we handle code structure in the boto projects. This PR is currently a draft for comment and may be broken up into smaller parts for easier review. The current proposal adds 3 new tools to ensure we're cleaning up old code and keeping what we have currently consistent. This also helps expedite PR review by removing the need for iterations on minor style fixes.

### Proposed additions:
* ~**isort** - This will order our imports and ensure consistent layout. This avoids duplicate imports by keeping everything in the same spot and makes it easier to find what's already available.~ (#2642)
* **pyupgrade** - This will automatically "upgrade" existing code to use the feature set configured for the lowest supported version (in this case Python 3.6). This is a more significant change that most version bumps because it's cleaning up all of our Python 2.6, 2.7 and 3.0-3.5 code.
* **black** -  This is the last one and maybe most contentious. We're adopting black with two caveats, we're not going to enforce quote normalization for now, and we're going to keep line length at 79 for the time being.


kojiromike:
To jump into the conversation head first: Have you thought about adding a [.pre-commit-config.yaml](https://pre-commit.com/) with pre-commit checks for e.g. isort and black, and enabling [pre-commit.ci](https://pre-commit.ci/) on the project? The former (yaml) would provide a standard way for users to keep their code formatted and linted consistently/locally, and the latter would use the same yaml to ensure pull requests painlessly maintain the same standards.

pre-commit.ci is free for open source projects, so I'd hope the administrative overhead of trying it out would be cheap. (Of course, I'd hope AWS would be willing to chip in and support that awesome project financially, too.)

nateprewitt:
This is great feedback, @kojiromike! We were talking about something similar last week but hadn't dove into options. We'll take a look at whether this will fit our needs. Thanks!

nateprewitt:
Cross linking with boto/boto3#2959.

nateprewitt:
Resolving this now that the last PR is merged in #2675.

PR #2475, opened 2021-09-02
Add support for S3 Multi-Region Access Points

nateprewitt:
This PR will add support for Botocore for S3 Mutli-Region Access Points and integrates support for the SigV4a signing algorithm.


kyleknap (approved):
:ship: assuming CI passes

PR #2490, opened 2021-09-19
:book: update organizations document typo

umeshkumhar:
**Issue**# https://github.com/boto/boto3/issues/2576

Additional literal "/n/n" within the text is removed.

### DIFF
--- a/botocore/data/organizations/2016-11-28/examples-1.json
+++ b/botocore/data/organizations/2016-11-28/examples-1.json
@@ -510,7 +510,7 @@
           "output": {
           }
         },
-        "description": "The following example shows how to request information about the current user's organization:/n/n",
+        "description": "The following example shows how to request information about the current user's organization:",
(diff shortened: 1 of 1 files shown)


umeshkumhar:
@nateprewitt Hey, can you please approve workflow approval? So we can validate the stuff

stobrien89:
Hi @umeshkumhar,

Thanks for pointing this out!  Many of the examples are handed to us by the service documentation writers, so we actually need to verify that these are fixed on their end first before we can accept any changes here.  I think we'll be able to pull this in once we confirm, but it could take some time to hear back from them, so we'll keep you posted with any updates we receive.

stobrien89:
Hi @umeshkumhar,

I confirmed with the team that we won't be able to take a change here, as much as we'd like to.  We'll definitely need to pull the changes from upstream, so any changes made manually here will unfortunately be overwritten.  Sorry about that!

Issue #2304, opened 2021-02-18
NoCredentialsError due to wrongly breaking the retry loop


attilasimon72:
**Description**
In some cases [IMDSFetcher._fetch_metadata_token](https://github.com/boto/botocore/blob/e1b6e6b58cc157a0197885c7a07c18fc3b73d33e/botocore/utils.py#L366) can ignore your `AWS_METADATA_SERVICE_NUM_ATTEMPTS` [configuration](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html), and you will end up getting `NoCredentialsError` on the first attempt.

**Details**
Our web app runs on EC2 with role-based authentication and relies on `InstanceMetadataFetcher`.

We have this config for the metadata service:

> AWS_METADATA_SERVICE_TIMEOUT=5
> AWS_METADATA_SERVICE_NUM_ATTEMPTS=3

When the load goes high and autoscaling launches a new EC2 instance, the web app of the new instance starts to bomb the metadata service with credential requests. (Boto sessions are created upon the first need, then they persist for the lifetime of the workers, but due to the high load on startup all the workers are busy getting the credentials in parallel).
As a result of this scenario, a low number of boto sessions fail to get credential, and due to #1791 they get stuck in repeatedly raising `NoCredentialsError`.

We attempted a trivial workaround by setting the number of retries higher:

> AWS_METADATA_SERVICE_TIMEOUT=5
> AWS_METADATA_SERVICE_NUM_ATTEMPTS=12

But no change, the same chain of `NoCredentialsError` again.

By adding logs to our code we identified that instead of exhausting all the configured 12 attempts, `InstanceMetadataFetcher` gave up after the first one.

Looking at the code it was easy to point out the reason in [IMDSFetcher._fetch_metadata_token](https://github.com/boto/botocore/blob/e1b6e6b58cc157a0197885c7a07c18fc3b73d33e/botocore/utils.py#L366):

If the first attempt ends with `ReadTimeoutError` in the for loop of `_fetch_metadata_token` then the loop will break prematurely, failing to exhaust configured `AWS_METADATA_SERVICE_NUM_ATTEMPTS`. If you look at the definition of `RETRYABLE_HTTP_ERRORS` at the top of the module, you can find it also has `ReadTimeoutError`, which indicates it is planned as retriable.

```python
RETRYABLE_HTTP_ERRORS = (
    ReadTimeoutError, EndpointConnectionError, ConnectionClosedError,
[...]

Issue #1897, opened 2019-11-20
Getting "botocore.exceptions.NoCredentialsError: Unable to locate credentials" on ec2 instance as of botocore-1.13.21


fchorney:
It seems like as of botocore-1.13.21, using boto3 on my ec2 instances is failing with `botocore.exceptions.NoCredentialsError: Unable to locate credentials` where it was working perfectly fine beforehand. 

```
Traceback (most recent call last):
  File "<string>", line 16, in <module>
  File "/foo/lib64/python3.7/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/foo/lib64/python3.7/site-packages/botocore/client.py", line 648, in _make_api_call
    operation_model, request_dict, request_context)
  File "/foo/lib64/python3.7/site-packages/botocore/client.py", line 667, in _make_request
    return self._endpoint.make_request(operation_model, request_dict)
  File "/foo/lib64/python3.7/site-packages/botocore/endpoint.py", line 102, in make_request
    return self._send_request(request_dict, operation_model)
  File "/foo/lib64/python3.7/site-packages/botocore/endpoint.py", line 132, in _send_request
    request = self.create_request(request_dict, operation_model)
  File "/foo/lib64/python3.7/site-packages/botocore/endpoint.py", line 116, in create_request
    operation_name=operation_model.name)
  File "/foo/lib64/python3.7/site-packages/botocore/hooks.py", line 356, in emit
    return self._emitter.emit(aliased_event_name, **kwargs)
  File "/foo/lib64/python3.7/site-packages/botocore/hooks.py", line 228, in emit
    return self._emit(event_name, kwargs)
  File "/foo/lib64/python3.7/site-packages/botocore/hooks.py", line 211, in _emit
    response = handler(**kwargs)
  File "/foo/lib64/python3.7/site-packages/botocore/signers.py", line 90, in handler
    return self.sign(operation_name, request)
  File "/foo/lib64/python3.7/site-packages/botocore/signers.py", line 157, in sign
    auth.add_auth(request)
  File "/foo/lib64/python3.7/site-packages/botocore/auth.py", line 357, in add_auth
    raise NoCredentialsError
botocore.exceptions.NoCredentialsError: Unable to locate credentials
```

The code itself is pretty simple:

```python
import boto3
region = "foo"
id = 123456789
secret_name = "bar"
asm = boto3.client(
    service_name='secretsmanager',
    region_name=region
)
[...]

Issue #1892, opened 2019-11-19
Regression in 1.13.21: cannot assume role using EC2 instance metadata


jbergknoff-rival:
Hi, we just started running into issues assuming roles in Jenkins, where the principal credentials come from EC2 instance metadata. This appears to be a regression in 1.13.21.

Here's a reproduction, running in an EC2 instance with permission to assume the role specified in the `aws-config` file.

```
$ docker run -it --rm python:3.7.5-alpine3.10 sh
/ # cat > aws-config
[profile role-to-assume]
role_arn = arn:aws:iam::...:role/...
credential_source = Ec2InstanceMetadata

/ # pip install awscli
...
Successfully installed PyYAML-5.1.2 awscli-1.16.285 botocore-1.13.21 colorama-0.4.1 docutils-0.15.2 jmespath-0.9.4 pyasn1-0.4.8 python-dateutil-2.8.0 rsa-3.4.2 s3transfer-0.2.1 six-1.13.0 urllib3-1.25.7

/ # AWS_PROFILE=role-to-assume AWS_CONFIG_FILE=aws-config aws sts get-caller-identity

Error when retrieving credentials from Ec2InstanceMetadata: No credentials found in credential_source referenced in profile role-to-assume

/ # pip install botocore==1.13.20
...
ERROR: awscli 1.16.285 has requirement botocore==1.13.21, but you'll have botocore 1.13.20 which is incompatible.
Installing collected packages: botocore
  Found existing installation: botocore 1.13.21
    Uninstalling botocore-1.13.21:
      Successfully uninstalled botocore-1.13.21
Successfully installed botocore-1.13.20

/ # AWS_PROFILE=role-to-assume AWS_CONFIG_FILE=aws-config aws sts get-caller-identity
{
    "UserId": "...",
    "Account": "...",
    "Arn": "arn:aws:sts::...:assumed-role/.../botocore-session-1574199910"
}
```

crgwbr:
This just broke a bunch of CI pipelines for me. Reverting from aws-cli `1.16.285` to `1.16.284` seems to be a temp fix.

vurt007:
same ... just broke our deployment pinned to 1.13.20 for now

scottschreckengaust:
https://github.com/aws/aws-cli/issues/4682

joguSD:
Thanks for bringing this to our attention. We'll be reverting this ASAP and cutting an additional release today.

itwasntandy:
[...]

PR #2499, opened 2021-09-24
Retry on ReadTimeoutError when fetching metadata tokens

jchorl:
`ReadTimeoutError` is already covered in the `RETRYABLE_HTTP_ERRORS` case. I looked at the [PR that added it](https://github.com/boto/botocore/pull/1895).

At the time, `_fetch_metadata_token` raised `self._RETRIES_EXCEEDED_ERROR_CLS()` if retries were exhausted. It doesn't do that anymore, it just returns `None`. So the behaviour should be the same, but with retries (which are disabled on IMDS by default).

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -409,8 +409,6 @@ class IMDSFetcher(object):
                     return None
                 elif response.status_code in (400,):
                     raise BadIMDSRequestError(request)
-            except ReadTimeoutError:
-                return None
             except RETRYABLE_HTTP_ERRORS as e:
                 logger.debug(
                     "Caught retryable HTTP exception while making metadata "


RyanFitzSimmonsAK:
Hi @jchorl, thanks for reaching out and for opening this pull request.  I'm going to close this as a potentially breaking change.  For a bit of background, I'd recommend you take a look at [this issue](https://github.com/boto/botocore/issues/2304), specifically [this](https://github.com/boto/botocore/issues/2304#issuecomment-926203508) comment.

> My apologies for the delayed response. I saw you probably had a chance to look at https://github.com/boto/botocore/pull/1895— this change was made shortly after the rollout of imdsV2, where changes that we made to the credential-fetching strategy broke users whose deployments were incompatible with imdsV2 (see https://github.com/boto/botocore/issues/1892, https://github.com/boto/botocore/issues/1897 and https://github.com/aws/aws-cli/issues/4682).
> 
> At the time, ReadTimeout errors were retry-able, but in each case of incompatibility, customers were hitting a ReadTimeoutError, so the thought was to have this fail fast so they could quickly fall back to imdsvV1. Reverting this change would likely break people again.

ReadTimeoutError is purposefully being fast-failed to avoid the described behavior.

PR #2507, opened 2021-09-30
Replace deprecated inspect.formatargspec with inspect.signature

hugovk:
`inspect.formatargspec` is deprecated:

> _Deprecated since version 3.5:_ Use `signature()` and Signature Object, which provide a better introspecting API for callables.

https://docs.python.org/3/library/inspect.html#inspect.formatargspec

https://github.com/python/cpython/pull/28618


nateprewitt:
Thanks for the PR @hugovk! We'd started some of this work but weren't sure what the impact of this change is on documentation generation. I think the only outstanding check is validating we're not adding additional signature footprint (specifically with the inclusion of `self`) in this change.

I believe the move to Signature, and no longer using `[1:]` for args, may be an issue.

mgorny:
Ping. This is now breaking botocore on Python 3.11, where `inspect.formatargspec()` was removed.

mgorny:
I've submitted #2673 as a little more elaborate approach that preserves the current behavior of skipping "self" (or more precisely, the first positional parameter). I hope you don't mind me modifying your code.

hugovk:
Absolutely not, go for it!

nateprewitt:
Thanks @mgorny, we’re waiting for the last reformatting PR to get merged in the next week or two. Then we plan to drop Python 3.6 on May 30th, at which point we’ll merge this and start testing on the 3.11dev build.

mgorny:
Thanks. It seems that I need to work on it some more but I should finish it today.

dlm6693 (approved):

dlm6693:
@mgorny these are a few manual tests that I ran:

```
import inspect
from botocore.loaders import Loader

class Foo:
    def test(self, a, b):
        pass

f = Foo()
inspect.signature(f.test)
<Signature (a, b)>

l = Loader()
inspect.signature(l._potential_locations)
> <Signature (name=None, must_exist=False, is_dir=False)>
```
Note that `self` is not include in either `Signature` object

PR #2511, opened 2021-10-01
Update rest-json protocol handling

nateprewitt:
This PR will bring us inline with new standards for rest-json serialization against AWS services. The end goal is to address inconsistencies between SDKs and service implementations by setting a clear set of defaults around when `Content-Length`, `Content-Type` and empty JSON bodies (`{}`) are included in requests.


botocore/serialize.py
```diff
@@ -593,9 +595,64 @@ def _convert_header_value(self, shape, value):
...
             return value
 
+    def _finalize_serialization(self, serialized, operation_model):
+        """Perform any required post-processing steps for serialization."""
+        self._prepare_additional_traits(serialized, operation_model)
+
 
 class RestJSONSerializer(BaseRestSerializer, JSONSerializer):
 
+    def _finalize_serialization(self, serialized, operation_model):
```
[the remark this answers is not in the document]
nateprewitt: `_prepare_additional_traits` is related to model trait parsing. I'm not sure it makes sense to try and squeeze protocol specific logic into a model function. We should try to keep functions specific to one task where possible.

tests/unit/auth/aws4_testsuite/post-content-length/post-content-length.creq
```diff
@@ -0,0 +1,10 @@
+POST
+/
+
+content-length:3
+host:example.amazonaws.com
+my-header1:VALUE1
+x-amz-date:20150830T123600Z
+
+content-length;host;my-header1;x-amz-date
+e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```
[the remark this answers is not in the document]
nateprewitt: This is intentional, we're parsing for exact equivalence in the tests and cannot have newlines. All of the existing tests also follow this paradigm.

botocore/crt/auth.py
```diff
@@ -73,10 +74,12 @@ def add_auth(self, request):
             signed_body_value=explicit_payload,
             signed_body_header_type=body_header,
             expiration_in_seconds=self._expiration_in_seconds,
-            )
+        )
         crt_request = self._crt_request_from_aws_request(request)
         future = awscrt.auth.aws_sign_request(crt_request, signing_config)
         future.result()
+        if stripped_headers:
```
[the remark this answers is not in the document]
nateprewitt: Good catch, that seems like a fine optimization.

botocore/serialize.py
```diff
@@ -593,9 +595,64 @@ def _convert_header_value(self, shape, value):
...
+    def _set_default_json_body(self, serialized, operation_model):
+        """Set default empty body ({}) for RestJSON requests if needed."""
+
+        # If we've already computed a body or will never serialize one,
+        # we can skip this. We will also skip modifying the body for
+        # blob or string payload shapes.
+        if (
+            serialized['body'] != b''
+            or not operation_model.can_serialize_body
+            or self._has_blob_or_string_payload(operation_model)
```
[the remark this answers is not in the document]
nateprewitt: Sure, I think this is more of a stylistic decision. The goal of `_finalize_serialization` is to provide a discrete set of ordered steps to perform in finalization. Ideally this keeps the logic minimal and avoids leakage out of the pluggable steps.

I'm not tied to that, but it was the goal with the current layout decision. Let me know if you have thoughts.

botocore/serialize.py
```diff
@@ -596,9 +597,64 @@ def _convert_header_value(self, shape, value):
...
 
+    def _finalize_serialization(self, serialized, operation_model):
+        """Perform any required post-processing steps for serialization."""
+        self._prepare_additional_traits(serialized, operation_model)
+
 
 class RestJSONSerializer(BaseRestSerializer, JSONSerializer):
 
+    def _finalize_serialization(self, serialized, operation_model):
+        super()._finalize_serialization(serialized, operation_model)
```
joguSD, line 608: This super call will do the `_prepare_additional_traits` call, which currently only calculates the `Content-MD5` checksum. There's a potential edge case where the `Content-MD5` is calculated only for the body to be modified later on this function. I don't think this is actually possible as there's no JSON services that use the `httpChecksum` trait but might be something worth noting or delaying the super call?

On that note, `_finalize_serialization` sounds like something that probably *shouldn't* be modifying the body, but rather things that would rely on the body being completely serialized (like checksums or content-length).

botocore/serialize.py
```diff
@@ -596,9 +597,64 @@ def _convert_header_value(self, shape, value):
...
+        has_body = serialized['body'] != b''
+        has_content_type = has_header('Content-Type', serialized['headers'])
+        if has_body and not has_content_type:
+            serialized['headers']['Content-Type'] = 'application/json'
+
+    def _set_content_length(self, serialized, operation_model):
+        headers = serialized['headers']
+        has_content_length = has_header('Content-Length', headers)
+
+        if not has_content_length and operation_model.can_serialize_body:
```
joguSD, line 646: Why not just check if we have a non-zero length body? At this point we've actually ran the serializer and the serializer has produced a body, whether or not the model conceptually says a body is possible doesn't seem relevant anymore.

botocore/serialize.py
```diff
@@ -596,9 +597,64 @@ def _convert_header_value(self, shape, value):
...
+        self._prepare_additional_traits(serialized, operation_model)
+
 
 class RestJSONSerializer(BaseRestSerializer, JSONSerializer):
 
+    def _finalize_serialization(self, serialized, operation_model):
+        super()._finalize_serialization(serialized, operation_model)
+
+        self._set_default_json_body(serialized, operation_model)
+        self._set_default_content_type_header(serialized, operation_model)
```
joguSD, line 611: I think the current implementation is correct, but I'm concerned that we're going about this from the wrong side. I'm not sure it makes sense to set a "default" retroactively like this. It feels like we're building up this list of predicates based off of the serializer's current behavior and finding the hole where we should put the default JSON body. This feels like a hard pattern to maintain because now you have to update the serializer's logic, but also make sure the assumptions being made in this finalization step still hold.

Can we not handle this more proactively before / during serialization? My hunch is that we should be able to set this payload/content-type up front and then have the serializer override the body/content-type if it knows better. This seems preferential to having the serializer implement some logic `A` around the model and then have this post serialization step where we check for `not A`.

botocore/serialize.py
nateprewitt: We don't want to emit a content-length when it's not relevant, e.g. if it's a GET request we shouldn't be emitting the Content-Type. This matches our current behavior in httpsession.py which would deviate without this check.

botocore/serialize.py
nateprewitt: >  It feels like we're building up this list of predicates based off of the serializer's current behavior and finding the hole where we should put the default JSON body. This feels like a hard pattern to maintain because now you have to update the serializer's logic, but also make sure the assumptions being made in this finalization step still hold.

Completely agree, I'm not a fan of this pattern, but it's unfortunately a necessity with our current inheritance model. We've made it difficult to modify rest-json specific behavior without moving almost all of `BaseRestSerializer` down into this class. I'll see if we can pull a bit more of this up by adding some inline helper functions to `serialize_to_request` and the other methods.

botocore/serialize.py
nateprewitt: Same comment as below. I'll look at getting the body modified earlier to avoid this conflict.

PR #2512, opened 2021-10-01
Remove redundant six library

hugovk:
Botocore supports Python 3.6+ so the six library is no longer needed.

Remove the vendored library and replace its uses with the Python 3 code. In part I used https://github.com/asottile/pyupgrade/ for this.

`compat.py` can be tidied/refactored further, but this PR is already quite big so that can come later.


nateprewitt:
Thanks for the PR, @hugovk! We also have this slated as clean up work we'd like to do. There's still some questions about when/if we want to remove six, so this just comes down to scheduling. I'll reply to the PR when we have a clearer time for merging a change like this.

hugovk:
This has got a bit merge conflicty, let's close and redo when the time is right 👍

PR #2529, opened 2021-10-12
Update awscrt version to 0.12.4

justindho:
This PR updates the botocore v2 branch to use awscrt 0.12.4, which supports the `HTTPS_PROXY` environment variable for S3 transfers when the `preferred_transfer_client` is set to `crt`


kyleknap (approved):
Looks good! :ship:

PR #2543, opened 2021-10-26
Mask proxy url credentials

### LINKED ISSUE #2422: Mask proxy credentials before raising exception
**Describe the bug**
When the proxy fails, an exception is raised which displays the proxy_url. For secured proxies, the proxy_url may contain the credentials in the format https://appId:password@myproxy.com.  In the pipeline, the credentials will be provided as variables so they are not visible in the logs but if an exception happens, httpSession.py raises an exception that includes the true proxy url in the error message.

Example:

In pipeline the proxy url would looke like this: https://$APP_ID:$PASSWORD@myproxy.com 

In the exception it looks like this:

`Failed to connect to proxy URL: https://sampleAppId:monkey123@myproxy.com`

Connection issues can be caused by network problems as well as invalid id and password. 

**Steps to reproduce**
Try to hit an endpoint through a secured proxy but put in a bogus id/password. 

I ran into this in a Gitlab pipeline executing the aws cli which uses botocore.

**Expected behavior**

The same exception is produced but the credentials are masked. This prevents exposing credentials in Git pipelines when there are issues with the proxy.

Example:

`Failed to connect to proxy URL: https://****:******@myproxy.com`

**Debug logs**
N/A

zhan9san:
Fix #2422

Could you kindly review this PR?

### DIFF
--- a/.changes/next-release/enhancement-Exceptions-94766.json
+++ b/.changes/next-release/enhancement-Exceptions-94766.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Exceptions",
+  "description": "ProxyConnectionError previously provided the full proxy URL. User info will now be appropriately masked if needed."
+}
--- a/botocore/httpsession.py
+++ b/botocore/httpsession.py
@@ -108,6 +108,23 @@ def ensure_boolean(val):
         return val.lower() == 'true'
 
 
+def mask_proxy_url(proxy_url):
+    """
+    Mask proxy url credentials
+
+    :type proxy_url: str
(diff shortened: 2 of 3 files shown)


jonchapmansai:
How do we get this reviewed?

nateprewitt:
Hi @zhan9san, thanks for the PR! We've added some minor changes and tests. We're working on doing final validation and will look to get this merged in an upcoming release.

joguSD (approved):

jonchapmansai:
Thank you @zhan9san @nateprewitt @joguSD
I really appreciate your work improving that

PR #2545, opened 2021-10-30
Code highlighting

ilhanyumer:
Looks better

### DIFF
--- a/docs/source/tutorial/index.rst
+++ b/docs/source/tutorial/index.rst
@@ -28,7 +28,9 @@ Using Botocore
 ==============
 
 The first step in using botocore is to create a ``Session`` object.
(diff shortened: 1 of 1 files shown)


kdaily:
Hi @ilhanyumer,

Thanks for the PR! I think that if we wanted to make this change, we would need to make it in all reStructuredText files across our documentation and examples; for example, in the [topics](https://github.com/boto/botocore/tree/develop/docs/source/topics) documentation.

Feel free to open an issue if you feel that this would be useful when rendering the documentation, and we'll see if more interest can be generated from other members of the community.. Having a description or example of the difference with and without the change would be useful.

Closing this for now as we wouldn't change this one single file.

PR #2554, opened 2021-11-02
Hide generate_presigned_url for all clients but S3

joguSD:
The generate_presigned_url is generically added to all clients but not all
services properly support presigned URLs. This adds a filtering mechanism to
generating client documentation and filters out documenting the method for all
clients but S3. The method is still present on all clients but will not be
documented unless we have confirmation that the service supports it.


nateprewitt (approved):
:rocket:

PR #2558, opened 2021-11-05
Adding modeled endpoint variants support

zdutta:
Introducing changes for supporting modeled endpoint variant configurations. The two new configuration options are: `use_dualstack_endpoint` and `use_fips_endpoint`.


botocore/client.py
```diff
@@ -82,9 +82,30 @@ def create_client(self, service_name, region_name, is_secure=True,
         service_name = first_non_none_response(responses, default=service_name)
         service_model = self._load_service_model(service_name, api_version)
         cls = self._create_client_class(service_name, service_model)
+        if region_name is not None:
```
joguSD, line 85: It might be worth breaking out the region normalization into it's own function but we can always refactor that later.

botocore/session.py
```diff
@@ -801,7 +802,7 @@ def create_client(self, service_name, region_name=None, api_version=None,
         elif default_client_config is not None:
             config = default_client_config
 
-        region_name = self._resolve_region_name(region_name, config)
+        region_name = self._resolve_region_name(region_name, config, service_name)
```
joguSD, line 805: Are we using `service_name` here anymore?

joguSD (approved):

botocore/config.py
```diff
@@ -169,6 +169,18 @@ class Config(object):
...
         into the prefix of the hostname. This is useful for clients providing
         custom endpoints that should not have their host prefix modified.
+
+    :type use_dualstack_endpoint: bool
+    :param use_dualstack_endpoint: Setting to True enables fips
+        endpoint resolution.
+
+        Defaults to None.
+
+    :type use_fips_endpoint: bool
```
lorengordon, line 179: Are these docstrings correct? Looks mixed up...

lorengordon: Oops, fixed in follow on PR...

Issue #2562, opened 2021-11-08
ImportError: cannot import name 'OP_NO_TICKET' from 'urllib3.util.ssl_'


barrykan:
**Describe the bug**
The just released version 1.23.0 will spit out the error in the title if you try to import `botocore.utils`, which precludes almost any use of the library.

**Steps to reproduce**
My environment:
OS: macOS Big Sur (11.4)
Python: 3.7.12 (virtualenv ultimately from brew installed version)
pyOpenSSL: 20.0.0

Install botocore 1.23.0 with `pip3 install botocore==1.23.0`. Run `import botocore.utils`. 

**Expected behavior**
There should not be an import error (and I believe this is the first version where this happens as it doesn't in 1.22.12).

**Debug logs**
(I cannot set the debug logger without getting the import error, so I can only show the import error.)

```
In [1]: import botocore.utils
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-918253be0bba> in <module>
----> 1 import botocore.utils

~/docs/venv/barrypy3/lib/python3.7/site-packages/botocore/utils.py in <module>
     30 import botocore
     31 import botocore.awsrequest
---> 32 import botocore.httpsession
     33 from botocore.compat import (
     34     json, quote, zip_longest, urlsplit, urlunsplit, OrderedDict,

~/docs/venv/barrypy3/lib/python3.7/site-packages/botocore/httpsession.py in <module>
      8 from urllib3 import PoolManager, proxy_from_url, Timeout
      9 from urllib3.util.retry import Retry
---> 10 from urllib3.util.ssl_ import (
     11     ssl, OP_NO_SSLv2, OP_NO_SSLv3, OP_NO_COMPRESSION, OP_NO_TICKET,
     12     PROTOCOL_TLS, PROTOCOL_TLS_CLIENT, DEFAULT_CIPHERS,

ImportError: cannot import name 'OP_NO_TICKET' from 'urllib3.util.ssl_' (/Users/barry/docs/venv/barrypy3/lib/python3.7/site-packages/urllib3/util/ssl_.py)
```

nateprewitt:
Hi @barrykan, thanks for opening this issue. It looks like we may have a gap in the imports for users on old versions of urllib3. Can you please provide the urllib3 version you're using locally?

For immediate mitigation while we investigate, we'd recommend making sure you're on `urllib3 >= 1.26` or pinning to `botocore < 1.23`.

[...]

PR #2563, opened 2021-11-08
Add fallback for OP_NO_TICKET import on older versions of urllib3

nateprewitt:
Fix for #2562, adding fallback for OP_NO_TICKET import for older versions of urllib3.


jturner:
FYI, PROTOCOL_TLS, PROTOCOL_TLS_CLIENT are also not present in older urllib3.

joguSD (approved):

PR #2566, opened 2021-11-10
Add protocol support for lists in headers

joguSD:
This updates the `rest-xml` and `rest-json` protocol serializers/parsers to support lists mapped to a header.

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -941,6 +941,13 @@ class BaseRestParser(ResponseParser):
             parsed = json.loads(decoded)
         return parsed
 
+    def _handle_list(self, shape, node):
+        location = shape.serialization.get('location')
+        if location == 'header' and not isinstance(node, list):
+            # List in headers may be a comma separated string as per RFC7230
+            node = [e.strip() for e in node.split(',')]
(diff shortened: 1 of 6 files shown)


botocore/parsers.py
```diff
@@ -941,6 +941,13 @@ def _handle_string(self, shape, value):
             parsed = json.loads(decoded)
         return parsed
 
+    def _handle_list(self, shape, node):
+        location = shape.serialization.get('location')
+        if location == 'header' and not isinstance(node, list):
+            # List in headers may be a comma separated string as per RFC7230
+            node = [e.strip() for e in node.split(',')]
```
nateprewitt, line 948: Only thought here is do we want to handle any erroneous commas/white space here? We could theoretically end up with a value of `""`. I don't know if there's any semantic meaning to those values. Adding a `if e.strip() != ""` at the end would address that.

joguSD: This is a good callout. It's a little ambiguous what the "correct" behavior here is. If it's ever a problem I think we fix it when we know there's a use case and what the actual expected behavior is.

nateprewitt (approved):

PR #2578, opened 2021-12-18
Fix grammatical error in ACM documentation

Spencer-Doak:
Hey, folks! I happened to notice this grammatical error while perusing the ACM service JSON file earlier today. Figured I'd put in a pull request. I imagine this issue might exist across multiple repos (e.g., all the other language's SDKs/libraries), so maybe there is an upstream where this change should be applied instead.

### DIFF
--- a/botocore/data/acm/2015-12-08/service-2.json
+++ b/botocore/data/acm/2015-12-08/service-2.json
@@ -794,7 +794,7 @@
       "members":{
         "message":{"shape":"String"}
       },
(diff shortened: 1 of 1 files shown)


kdaily:
Hi @Spencer-Doak,

Thanks for noticing! This would indeed need to be addressed upstream, these API models are generated by service teams, so any changes here would be reverted the next time we get them. I will let the ACM team know!

Issue #2580, opened 2021-12-22
Cannot import name 'PROTOCOL_TLS' from 'urllib3.util.ssl_'


thehunt33r:
**Describe the bug**

This issue is a direct follow-up from https://github.com/boto/botocore/issues/2562

When using older versions of urllib3 (often vendored), importing botocore.utils (or using botocore in any consideration) will fail.  

**Steps to reproduce**

```shell
$ pip install --upgrade botocore==1.23.26
$ pip install --upgrade urllib3==1.22.0
$ python
>>> import botocore.utils
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/matthieu/.virtualenvs/reproavm/lib/python3.8/site-packages/botocore/utils.py", line 32, in <module>
    import botocore.httpsession
  File "/Users/matthieu/.virtualenvs/reproavm/lib/python3.8/site-packages/botocore/httpsession.py", line 10, in <module>
    from urllib3.util.ssl_ import (
ImportError: cannot import name 'PROTOCOL_TLS' from 'urllib3.util.ssl_' (/Users/matthieu/.virtualenvs/reproavm/lib/python3.8/site-packages/urllib3/util/ssl_.py)
```

**Expected behavior**

botocore should fallback to importing PROTOCOL_TLS from `ssl`.

stobrien89:
Hi @thehunt33r,

Thanks for raising this!  I don't see anything wrong with your change at surface-level, but I'll review with the team to get their thoughts.

stobrien89:
Hi @thehunt33r,

I saw you mentioned using vendored older versions of `urllib3`, but is there a particular reason you need to use such an old version?  You may already be aware, but our minimum version required at this time is `1.25.4`.

thehunt33r:
Hey!

Sorry yeah I was using an old version of an AWS solution (LandingZone) I'm stuck with for the foreseeable future.

I did see the minimum requirement for the version of urllib3 but since an earlier similar PR was merged I thought it could be worth it to do this change? If you don't think that's worth it feel free to close the PR :-)

Thank you for taking the time to look through it. 

Matthieu

stobrien89:
Hi @thehunt33r,

That makes sense.  I discussed this with a few team members yesterday and they were not inclined to merge the PR due to the problem occurring well below our `urllib3` floor, but we were missing a few other team members due to the holidays— I'll bring this up again once we have the full group and see if we can get a better consensus.

[...]

PR #2581, opened 2021-12-22
Add fallback for PROTOCOL_TLS when using older urllib3 version

thehunt33r:
This PR is a follow-up to https://github.com/boto/botocore/pull/2563 and it resolves https://github.com/boto/botocore/issues/2580


thehunt33r:
Hi @stobrien89 

Thanks for your answer and your feedback. Makes a lot of sense, and I'm happy with that! Thanks!

PR #2591, opened 2022-01-14
Adding smart default support

zdutta:
Adding support for `defaults_mode` client configuration option. The `defaults_mode` will be used to determine how certain default configuration options are resolved in the SDK. Configurations values impacted include: `retry_mode`, `sts_regional_endpoints` , s3 `us_east_1_regional_endpoint` and `connect_timeout`.

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -32,8 +32,10 @@ from botocore.configprovider import ConfigValueStore
 from botocore.configprovider import ConfigChainFactory
 from botocore.configprovider import create_botocore_default_config_mapping
 from botocore.configprovider import BOTOCORE_DEFAUT_SESSION_VARIABLES
+from botocore.configprovider import DefaultConfigResolver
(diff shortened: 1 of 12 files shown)


joguSD (approved):

botocore/session.py
```diff
@@ -870,6 +899,25 @@ def _resolve_region_name(self, region_name, config):
...
+    def _resolve_defaults_mode(self, client_config, config_store):
+        mode = config_store.get_config_variable('defaults_mode')
+
+        if client_config and client_config.defaults_mode:
+            mode = client_config.defaults_mode
+
+        default_config_resolver = self._get_internal_component(
+            'default_config_resolver')
+        default_modes = default_config_resolver.get_default_modes()
+        default_modes.extend(['legacy', 'auto'])
```
joguSD, line 911: It might make more sense to just have these be the base in the `get_default_modes` method and then extend it by the keys we find in the model file.

nateprewitt (approved):

PR #2595, opened 2022-01-19
Accept proxy config via environment variables

mogul:
Makes these proxy settings easily configurable via environment variables.

See https://github.com/aws/aws-sdk/issues/357 for context.

### DIFF
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -149,10 +149,10 @@ DEFAULT_S3_CONFIG_VARS = {
 # used to configure how botocore interacts with proxy setups while
 # sending requests.
 DEFAULT_PROXIES_CONFIG_VARS = {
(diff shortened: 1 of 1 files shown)


tim-finnigan:
Thanks @mogul for the PR. One issue with this PR is that those variable names are specific to S3. They would have to be generalized as they could apply to other services. Another thing is that we would need to collaborate with other SDKs if we're considering 'AWS_*' prefixed variable names. But first we want to get more community feedback on the issue you created. I’m going to set this PR to Draft but we can keep your issue open for tracking the request.

tim-finnigan:
Thanks for your patience, as mentioned earlier the work involving environment variables would require coordination across SDK teams. We're continuing to track https://github.com/aws/aws-sdk/issues/357 for consideration but as of now there are still no plans for this. Going to close this PR as not planned at this time.

PR #2596, opened 2022-01-22
Dox fix: default role used.

JulienPalard:
Don't mind me, I spotted while working on an Sphinx linter.

For context, simple backticks in Sphinx are [default roles](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html), the Sphinx doc explicitly allows using default role for whatever you want, so I'm fixing nothing here.

But as it's the only default role in botocore whole doc it feels a small inconsistency that can be fixed.


nateprewitt (approved):
Thanks @JulienPalard!

PR #2597, opened 2022-01-23
Doc fix: no newline at end of file.

JulienPalard:
Don't mind me, I spotted it while working on an Sphinx linter, as #2596.


nateprewitt (approved):
Thanks again, @JulienPalard!

PR #2600, opened 2022-01-26
Tolerate custom IMDS endpoints without trailing slash

nateprewitt:
When using `AWS_EC2_METADATA_SERVICE_ENDPOINT` or `ec2_metadata_service_endpoint`, we use the url as-is without checking how it's formatted. This leads to issues with appending an IMDS path for URLs that don't end in a trailing slash. Most users aren't going to think to add that when using the override and this causes unnecessary failures.

This fix will maintain the backwards compatibility of `get_base_url` while ensuring paths are constructed correctly.

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -416,9 +416,15 @@ class IMDSFetcher(object):
 
         return chosen_base_url
 
+    def _construct_url(self, path):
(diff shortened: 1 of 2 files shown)


zdutta (approved):
Looks good

PR #2601, opened 2022-01-26
Fix IPv6 endpoints with non-default port

drake127:
This commit adds brackets to IPv6 addresses in auth.py. Otherwise the signature is not calculated properly.

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -27,7 +27,7 @@ from botocore.compat import (
     six, unquote, urlsplit, urlunsplit, HAS_CRT
 )
 from botocore.exceptions import NoCredentialsError
-from botocore.utils import normalize_url_path, percent_encode_sequence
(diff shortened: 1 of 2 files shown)


drake127:
E.g.: `http://[::1]:8080` doesn't currently work.

nateprewitt:
Hi @drake127, thanks for the PR! This looks reasonable to me, would you mind adding a test to ensure the behavior continues working as expected?

drake127:
I am willing to do it if you give me some hints how to add tests in your project. I tested it against Scality docker endpoint but I suspect I cannot do it here. Do you have some endpoint in your test infra (that verifies V4 signatures)?

nateprewitt (approved):
Yep, those are great! I think this should be set, let's wait for CI and we'll be ready to merge.

PR #2605, opened 2022-02-02
Add RequestId to error message

tdriscoll:
This adds the RequestID to the error string. The RequestID is useful/required when raising tickets with AWS support. 

It's possible to catch these errors and explicitly log the RequestID but that means handling exceptions for every API call. I believe it is also possible with DEBUG logging but that logs much more than errors. 

My experience was that I went to report an error but did not have the foresight to log the request id. In addition, it was a difficult to reproduce error. This is why I like the idea of including the RequestID by default for errors.

I tried to keep to the style of the repo but let me know if you want me to clean anything up.

### DIFF
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -454,16 +454,18 @@ class UnsupportedSignatureVersionError(BotoCoreError):
 class ClientError(Exception):
     MSG_TEMPLATE = (
         'An error occurred ({error_code}) when calling the {operation_name} '
-        'operation{retry_info}: {error_message}')
+        'operation{retry_info}: {error_message}{request_id_info}')
 
     def __init__(self, error_response, operation_name):
         retry_info = self._get_retry_info(error_response)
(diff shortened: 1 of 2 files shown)


kdaily:
Thanks for the PR, @tdriscoll. I note that this would close this issue (you also requested, thanks!): 

https://github.com/boto/boto3/issues/1143

I don't have a timeline when this will get a review.

tim-finnigan:
Thanks for creating this PR. I brought it up for discussion with the team, and the consensus was that this should be closed as not planned. You can [already get](https://github.com/boto/boto3/issues/1143#issuecomment-310130166) the Request ID from the `ResponseMetadata` without too much trouble. There is also documentation referenced [here](https://github.com/boto/boto3/issues/1143#issuecomment-1186075811) on logging the Request ID.

tdriscoll:
Understood. Thanks for taking the time to consider it!

PR #2614, opened 2022-02-10
Parsing int/long in rest-json response header

zdutta:
Adding support for parsing int/long types in rest-json response headers.


Bartascnk (approved):

tests/unit/protocols/output/rest-json.json
```diff
@@ -1257,5 +1257,56 @@
...
+          "name": "OperationName"
+        },
+        "result": {
+          "IntegerMember": 123,
+          "LongMember": 200
+        },
+        "response": {
+          "status_code": 200,
+          "headers": {
+              "x-amz-integer-member": 123,
```
nateprewitt, line 1304: Headers should always be bytes or strings. We should be able to specify these as strings and get the correct output still, right?

nateprewitt (approved):

PR #2616, opened 2022-02-11
One-liner: Eliminate Magic Number as Default Argument to iter_lines method of StreamingBody

svidovich:
Hello,

I was rummaging through botocore to investigate how `StreamingBody`'s `iter_lines` worked, and noticed that this took `1024` as a default instead of the provided constant, so I thought I would make a PR. This is my first time making a PR here; I'm glad it's small.

Anyway, I ran the pre-commit and it worked out nicely:

```
pre-commit run --all-files
Check Yaml...............................................................Passed
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
flake8...................................................................Passed
```

What do you think?

Regards,
SamV

### DIFF
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -124,7 +124,7 @@ class StreamingBody(IOBase):
 
     next = __next__
 
-    def iter_lines(self, chunk_size=1024, keepends=False):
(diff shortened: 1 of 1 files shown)


nateprewitt (approved):
This seems reasonable to me, thanks @svidovich!

svidovich:
Yay!

PR #2621, opened 2022-02-18
Fix: Add missing botocore.parsers import to botocore.args

whardier:
Found while vendoring botocore for an internal dev tool.


nateprewitt:
This should go out in our next release, thanks again @whardier!

Issue #2365, opened 2021-04-27
Store schemas compressed on disk.


adamchainz:
**Is your feature request related to a problem? Please describe.**

The `data` directory of a `botocore` install is over 50MB. The JSON inside compresses really well - we can see as the PyPI packages are just 7MB.

**Describe the solution you'd like**

It would be good to keep the schemas compressed on disk and only decompress them when reading into memory. This would save disk space, and probably a little time too since the decompression step is likely to be faster that reading all the bytes from disk.

Python's zlib or zip modules in the standard library can be used.

For an example of a library shipping data in a zip file, see my heroicons package: https://github.com/adamchainz/heroicons/blob/main/src/heroicons/__init__.py

stobrien89:
Hi @adamchainz,

Thanks for the feature request!  I'll review this with the team, although I can't make any guarantees as to when/if this will be implemented.

kdaily:
@adamchainz,

This is an interesting idea. This has been noted previously in a similar scenario with the AWS CLI as well:

https://github.com/aws/aws-cli/issues/5725

The AWS SDKs consume the API models from upstream. Changing the way that they are stored and accessed would be a significant feature. One drawback would be the lack of direct human readability of the API models that are currently available in the Python SDK. It would be difficult to see where API changes were introduced between versions of the SDK. For example, removing the documentation strings from the models would cut 20MB off of the size, which might be useful in a CI/CD environment.

Do you have specific scenarios of your own that a slimmed down version?

adamchainz:
> It would be difficult to see where API changes were introduced between versions of the SDK.

One can use the [`textconv` git attribute](https://www.git-scm.com/docs/gitattributes#_performing_text_diffs_of_binary_files) in the repo to have git decompress the files before comparing them.

> Do you have specific scenarios of your own that a slimmed down version?

This affects me in a couple ways:

1. I bundle `boto3` into my lambda functions so I can pin an exact version. The occasional API change can break code. Bundling botocore takes a function over the 50MB limit, which requires ann upload to S3 rather than directly to Lambda, and prevents the console code editor from working.

2. I have maybe 30 projects using boto3/botocore, each with their own virtual environment. This means I have 1.5GB of botocore, which isn't a great use of disk space.

benkehoe:
[...]

PR #2628, opened 2022-03-03
Support loading gzip compressed models

joguSD:
This adds support for loading gzip compressed models (i.e. `.json.gz` files) which unblocks generating more optimized wheel builds of botocore. See #2365 for more discussion. As `gzip` may not be available on all platforms support for this _should_ gracefully fallback to the old behavior of only checking for `.json` files.

Ideally, we'll be able to generate and distribute more optimized wheels where all models are compressed and will use these loaders instead.


botocore/loaders.py
```diff
@@ -107,10 +107,21 @@
 from botocore import BOTOCORE_ROOT
 from botocore.compat import json
 from botocore.compat import OrderedDict
+from botocore.compat import HAS_GZIP
 from botocore.exceptions import DataNotFoundError, UnknownServiceError
 from botocore.utils import deep_merge
 
 
+_JSON_OPEN_METHODS = {
```
nateprewitt, line 115: Very minor nit, but should we have this be more general than JSON? It's just a suffix to open function map.

nateprewitt (approved):

botocore/loaders.py
joguSD: I had it as `_OPEN_METHODS` but it's only used / specific to the `JSONFileLoader` so I decided to rename it. I think if we want to expand this to support more formats + Gzip on top of them we should look at a more serious refactor where support for compression is closer to a mixin that sits on top of the base loader class.

stealthycoin (approved):

nateprewitt (approved):
:shipit:

PR #2642, opened 2022-03-22
Run isort over botocore and tests

nateprewitt:
Breaking #2472 into slightly smaller pieces so we can have a more reasonable chance at properly reviewing the changes. This will be a first pass to add isort to keep our imports organized and fix our redundant/unnecessary imports.


scripts/get-model-filename
```diff
@@ -21,15 +21,17 @@ will be created if it does not exist.
   Copied: /tmp/myfile.json -> /Users/foo/botocore/data/aws/cloudwatch/2010-08-01.normal.json
 
 """
+import json
+
 # Note we're using optparse for 2.6 compat.
```
joguSD, line 26: Hm, the comment around the import causing a space is a little strange. I don't feel too strongly on this though... thoughts? Also, did we mean to hit the `scripts` directory as well?

joguSD (approved):

scripts/get-model-filename
nateprewitt: Yeah, black may clean this up later, but I don't think we have a way to disable it. I'm also relatively indifferent on it for the moment. I don't think it particularly hurts readability.

For scripts, we were already including it as part of the .pre-commit hooks. It seems reasonable to keep those files formatted too, but it's not strictly required.

joguSD (approved):

Issue #2648, opened 2022-03-28
[s3] Two region names are missing in S3 `BucketLocationConstraint`


vemel:
You can see that `ap-southeast-3` and `us-east-1` are missing in S3 schema: https://github.com/boto/botocore/blob/develop/botocore/data/s3/2006-03-01/service-2.json#L1467

However, S3 service supports both `ap-southeast-3` and `us-east-1` regions. Also `EU` region probably has to be removed from the schema.

tim-finnigan:
Thanks for reporting this @vemel. As mentioned [here](https://github.com/boto/botocore/pull/2649#issuecomment-1081059823) these changes would need to be made by the S3 team directly. I created an issue here in our cross-SDK repository since this affects other SDKs: https://github.com/aws/aws-sdk/issues/217

I reached out to the S3 team and will post an update there when I hear back.

PR #2649, opened 2022-03-28
[s3] Update region list in `BucketLocationConstraint`

vemel:
Fix for #2648 

- Added `ap-southeast-3`
- Added `us-east-1`
- Removed `EU`


nateprewitt:
Hi @vemel, thanks for the PR! Unfortunately this file is authored elsewhere and simply copied into Botocore. We won't be able to make these changes directly in Botocore and will need to be made by the S3 team directly. We'll track the outcome of this in #2648.

vemel:
Thank you!

Yes, I expected that `botocore` schemas are managed somewhere externally. I created the PR just as an example of the changes I propose.

Issue #2645, opened 2022-03-24
Http request fails with ValueError behind https proxy, when proxy specified is an Ip Address.


vabhasin:
**Describe the bug**
Boto Client configured with proxy settings fails when sending urllib3 http request behind a https proxy.
Request fails with error `"ValueError: check_hostname requires server_hostname"` and happens only when you specify proxy address using ip address. I was able to confirm that the same code works when you provide proxy server with a hostname instead of Ip Address.

Client creation code:
```
client = boto3.client('cloudwatch', 'us-east-1', config=config.Config(proxies={'https': 'https://1.2.3.4:5678'}, 
proxies_config={'proxy_ca_bundle': '/home/user/ca.pem'}))
response = client.put_metric_data(Namespace='namespace',
                                  MetricData=sample_metric)
```

Looking into the stack trace the error comes from [ssl.py](https://github.com/python/cpython/blob/a43fd45918cef48187053d7c56d3b9bb902151d0/Lib/ssl.py#L996-L997).
```
if context.check_hostname and not server_hostname:
            raise ValueError("check_hostname requires server_hostname")
```

`check_hostname` in the ssl_context has been set to `True` by botocore here.
https://github.com/boto/botocore/blob/0cc671363e27bd620917414215f6d71026384682/botocore/httpsession.py#L341-L343

[urllib3](https://github.com/urllib3/urllib3/blob/6de3330eb54f73a57c7860f75123bde8b043dbd2/src/urllib3/util/ssl_.py#L428-L454) doesn't pass `server_hostname` to `_ssl_wrap_socket_impl` as the `server_hostname` is an **Ip address** in our case.
This was changed in a recent [fix](https://github.com/urllib3/urllib3/issues/2400) in urllib3, according to which SNI should only be sent if the proxy is a hostname, not an Ip address.
```
    # If we detect server_hostname is an IP address then the SNI
    # extension should not be used according to RFC3546 Section 3.1
    use_sni_hostname = server_hostname and not is_ipaddress(server_hostname)
    # SecureTransport uses server_hostname in certificate verification.
    send_sni = (use_sni_hostname and HAS_SNI) or (
        IS_SECURETRANSPORT and server_hostname
    )
    # Do not warn the user if server_hostname is an invalid SNI hostname.
    if not HAS_SNI and use_sni_hostname:
        warnings.warn(
            "An HTTPS request has been made, but the SNI (Server Name "
            "Indication) extension to TLS is not available on this platform. "
[...]

PR #2652, opened 2022-03-29
Enable "check_hostname" only if proxy url is not Ip adress

vabhasin:
**Describe the changes**
This change uses `prox_url` to create `proxy_ssl_context ` as we don't need to enable `check_hostname` if the provided proxy url host is an ip address. This uses `is_ipaddress` similar to how urllib3 uses to determine if the provided hostname is an ip address, if not `check_hostname` is enabled.
The change is to fix issue https://github.com/boto/botocore/issues/2645, where using ip address in proxy url inside proxy config fails to send http request due to error `"ValueError: check_hostname requires server_hostname"`.

**Testing**
Unit tests


tests/unit/test_http_session.py
```diff
@@ -353,7 +353,7 @@ def test_urllib3_proxies_kwargs_included(self):
             )
             self.assert_request_sent()
 
-    def test_proxy_ssl_context_uses_check_hostname(self):
+    def test_proxy_ssl_context_use_check_hostname_if_not_ip_address(self):
```
nateprewitt, line 356: We can probably leave this name as is.

tests/unit/test_http_session.py
```diff
@@ -367,6 +367,20 @@ def test_proxy_ssl_context_uses_check_hostname(self):
             last_call = self.proxy_manager_fun.call_args[-1]
             self.assertIs(last_call['ssl_context'].check_hostname, True)
 
+    def test_proxy_ssl_context_does_not_use_check_hostname_if_ip_address(self):
```
nateprewitt, line 370: Let's make sure we're testing this for both IPv4 and IPv6 proxies.

botocore/httpsession.py
```diff
@@ -317,7 +318,7 @@ def _get_proxy_manager(self, proxy_url):
             proxy_headers = self._proxy_config.proxy_headers_for(proxy_url)
             proxy_manager_kwargs = self._get_pool_manager_kwargs(
                 proxy_headers=proxy_headers)
-            proxy_manager_kwargs.update(**self._proxies_kwargs)
+            proxy_manager_kwargs.update(**self._proxies_kwargs(proxy_url))
```
nateprewitt, line 321: If we're going to start handing down arguments, let's decouple this a bit. It is likely better to keep `_proxies_kwargs` as our static settings and then have an additional step that creates and adds the `proxy_ssl_context` in a separate step.

We don't want to have to require access to every possible SSLContext configuration option when we're pulling static args such as line 389.

botocore/httpsession.py
```diff
@@ -341,17 +342,19 @@ def _setup_ssl_cert(self, conn, url, verify):
...
         if proxy_ca_bundle is None and proxy_cert is None:
             return None
 
         context = self._get_ssl_context()
         try:
+            u = parse_url(proxy_url)
             # urllib3 disables this by default but we need
-            # it for proper proxy tls negotiation.
-            context.check_hostname = True
+            # it for proper proxy tls negotiation when proxy_url is not an Ip Address
```
nateprewitt, line 355: nit
```suggestion
            # it for proper proxy tls negotiation when proxy_url is not an IP Address
```

botocore/httpsession.py
```diff
@@ -20,7 +20,9 @@
     OP_NO_SSLv2,
     OP_NO_SSLv3,
     ssl,
+    is_ipaddress,
```
nateprewitt, line 23: What's the minimum version of urllib3 this is available in?

botocore/httpsession.py
```diff
@@ -341,17 +342,19 @@ def _setup_ssl_cert(self, conn, url, verify):
...
-    def _setup_proxy_ssl_context(self, proxies_settings):
+    def _setup_proxy_ssl_context(self, proxies_settings, proxy_url):
         proxy_ca_bundle = proxies_settings.get('proxy_ca_bundle')
         proxy_cert = proxies_settings.get('proxy_client_cert')
         if proxy_ca_bundle is None and proxy_cert is None:
             return None
 
         context = self._get_ssl_context()
         try:
+            u = parse_url(proxy_url)
```
nateprewitt, line 353: We likely need to be accounting for `LocationParseError` if we're going to add this.

nateprewitt (changes requested):
Thanks for the PR, @vabhasin! I've left some initial feedback.

tests/unit/test_http_session.py
vabhasin: Sure.

botocore/httpsession.py
vabhasin: Yes, makes sense. I can separate them out

botocore/httpsession.py
vabhasin: Thanks. I will add

tests/unit/test_http_session.py
vabhasin: Thanks @nateprewitt, that makes sense.
But now when I tried testing IPv6 here, I am seeing a problem. So, I used the IPv6 address `https://[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:5678` which was one of the examples in [urllib3](https://github.com/urllib3/urllib3/blob/fab33936db9281e96b6325b4fafe2fbb852fa458/test/test_ssl.py#L18) to test is_ipaddress. I had to use the braces format here so that I can pass in the port (not passing the port get `parse_url` to fail.). Now, the `is_ipaddress` that we use here from [urllib3](https://github.com/urllib3/urllib3/blob/fab33936db9281e96b6325b4fafe2fbb852fa458/src/urllib3/util/ssl_.py#L571) only matches for `_BRACELESS_IPV6_ADDRZ_RE`. is it that hostname at that point is expected to be without braces?

vabhasin: Using `IPV6_ADDRZ_RE` instead, works.
Also, not having a port name doesn't create a problem when using braces.

botocore/httpsession.py
vabhasin: Looks like it was added into 1.23 release ([commit](https://github.com/urllib3/urllib3/commit/7c216f433e39e184b84cbfa49e41135a89e4baa0))

tests/unit/test_http_session.py
nateprewitt: Thanks for bringing this to my attention, @vabhasin! I'm currently discussing options with the other urllib3 maintainers. I think ideally we want to upstream this so we're not maintaining it in two places. That will unfortunately not be in all of our supported versions, so we'll need a temporary wrapper.

I'll follow up shortly with more information on how we want to proceed. I think the current vending isn't going to be the path forward though.

botocore/httpsession.py
```diff
@@ -177,6 +185,21 @@ def mask_proxy_url(proxy_url):
...
+
+    :param str host: Hostname to examine.
+    :return: True if the hostname is an IP address, False otherwise.
+    """
+    if not six.PY2 and isinstance(host, bytes):
+        # IDN A-label bytes are ASCII compatible.
+        host = host.decode("ascii")
+    return bool(IPV4_RE.match(host) or IPV6_ADDRZ_RE.match(host))
+
+
```
nateprewitt, line 202:
```suggestion
    """Wrapper for urllib3's is_ipaddress to provide support for bracketed IPv6 addresses.

    :param str host: Hostname to examine.
    :return: True if the hostname is an IP address, False otherwise.
    """
    return is_ipaddress(host) or bool(IPV6_ADDRZ_RE.match(host))
```

You'll also want to update this import to be `from botocore.utils import IPV6_ADDRZ_RE` for backwards compatibility.

botocore/httpsession.py
```diff
@@ -21,6 +26,9 @@
     OP_NO_SSLv3,
     ssl,
 )
+from urllib3.util.url import IPV4_RE, IPV6_ADDRZ_RE, parse_url
+
+from botocore.vendored import six
```
nateprewitt, line 31: Let's remove this, we don't support Python2 or six anymore.

tests/unit/test_http_session.py
```diff
@@ -367,6 +367,26 @@ def test_proxy_ssl_context_uses_check_hostname(self):
...
+        cert = ('/some/cert', '/some/key')
+        proxies_config = {'proxy_client_cert': "path/to/cert"}
+        urls = ['https://1.2.3.4:5678',
+                'https://4.6.0.0',
+                'https://[FE80::8939:7684:D84b:a5A4%251]:1234',
```
nateprewitt, line 375: Let's make sure we're doing IP's without a zone identifier or port too.

i.e.
* https://[FE80::8939:7684:D84b:a5A4]
* https://[FE80::8939:7684:D84b:a5A4]:1234

tests/unit/test_http_session.py
```diff
@@ -367,6 +367,26 @@ def test_proxy_ssl_context_uses_check_hostname(self):
...
+                'https://[FE80::8939:7684:D84b:a5A4%251]',
+                'https://[::1]:789']
+        for proxy_url in urls:
+            with mock.patch('botocore.httpsession.SSLContext'):
```
nateprewitt, line 379: It looks like there are some issues with your current Mock usage causing this test to fail. Can you take a look at that?

nateprewitt (changes requested):
Thanks for taking a look at the previous feedback. I added a proposal for the current IPv6 issue, please let me know what you think.

Otherwise, there are some minor notes on the test failures that will need to be addressed.

botocore/httpsession.py
vabhasin: Thanks , I've added this now.

tests/unit/test_http_session.py
vabhasin: I can see the tests are good now, until I change the import to `from botocore.utils import IPV6_ADDRZ_RE` instead of urllib3.
pytest is not able to import from utils somehow, whereas python can. I committed anyways as I wanted to see if this was something with my workspace.

botocore/httpsession.py
```diff
@@ -19,8 +24,12 @@
     PROTOCOL_TLS,
     OP_NO_SSLv2,
     OP_NO_SSLv3,
+    is_ipaddress,
     ssl,
 )
+from urllib3.util.url import parse_url
+
+from botocore import utils
```
nateprewitt, line 32: We merged #2660 to fix this. Let's change the path to `from botocore.compat import IPV6_ADDRZ_RE` which will handle this. I don't think we'll want to add a generic `utils` namespace.

nateprewitt (changes requested):

botocore/httpsession.py
vabhasin: Ok, I didn't realize that. let me revert that back.

botocore/httpsession.py
```diff
@@ -318,6 +333,11 @@ def _get_proxy_manager(self, proxy_url):
             proxy_manager_kwargs = self._get_pool_manager_kwargs(
                 proxy_headers=proxy_headers)
             proxy_manager_kwargs.update(**self._proxies_kwargs)
+            proxy_ssl_context = self._setup_proxy_ssl_context(self._proxy_config.settings, proxy_url)
+            if proxy_ssl_context:
```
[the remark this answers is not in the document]
vabhasin: Before my change, the `proxies_kwargs` added `proxy_ssl_context` only when it was not `None`. This was also the problem behind tests failing in my last commit. `proxy_ssl_context` should not be in pool manager kwargs if it's None.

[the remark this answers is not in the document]
vabhasin: @nateprewitt let me know if it's good now.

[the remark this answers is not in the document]
nateprewitt: Thanks @vabhasin, I've added a couple minor updates but I think we're set now. I've added @joguSD to do a final sign off since I've added changes. Then we'll look at getting the change merged.

nateprewitt (approved):

joguSD (approved):

Issue #1197, opened 2017-05-15
Make StreamingBody a context manager


cbare:
Would it be helpful to make StreamingBody a context manager by adding  #an `__exit__(...)` method?

stealthycoin:
Marking this as a feature request.

CMCDragonkai:
This would be useful.

dlm6693:
This functionality has now been added. See https://github.com/boto/botocore/pull/2677 for implementation details.

PR #2677, opened 2022-05-18
allow StreamingBody class to be used as a context manager

dlm6693:
This is addressing a feature request originally made in https://github.com/boto/botocore/issues/1197. It allows the `StreamingBody` class to be instantiated as a context manager (i.e. able to use in a `with` statement). It's implemented in the standard way by adding `__enter__` and `__exit__` magic methods.


nateprewitt (approved):
:shipit:

Issue #2667, opened 2022-05-04
Unable to use AWSRequest params with duplicate query keys


zrsmithson:
### Describe the bug

When trying to make a generic `AWSRequest` so I can add a `SigV4Auth` from IAM policy (which is required as per #1784), I'm unable to use the `params` attribute with duplicate keys because it only works with `dict` objects. Instead I have manually build the query string and append it to the url before instantiating the `AWSRequest` object

This might actually be a feature request, but the documentation seems to indicate that it should behave similarly to `requests`:
> This class was originally inspired by requests.models.Request, but has been boiled down to meet the specific use cases in botocore. That being said this class (even in requests) is effectively a named-tuple.

`requests` allows `params` to be a dictionary or list of tuples, which allows for keys to be repeated.

### Expected Behavior

``` python
import boto3
from botocore.awsrequest import AWSRequest
from botocore.auth import SigV4Auth
from botocore.endpoint import URLLib3Session

url = "http://some-api-endpoint.com"
creds = boto3.Session().get_credentials()
region = "us-west-1"
query_params = [
    ("eq", "attr1:value1"),
    ("eq", "attr2:value2"),
]

request = AWSRequest(method="GET", url=url, params=query_params)
SigV4Auth(creds, "execute-api", region).add_auth(request)
response = URLLib3Session().send(request.prepare())
```

### Current Behavior

As it is, I have to use `urlencode` to manually add a "?" and the query string to the end of the url

``` python
import boto3
from botocore.awsrequest import AWSRequest
from botocore.auth import SigV4Auth
from botocore.endpoint import URLLib3Session
from botocore.compat import urlencode

url = "http://some-api-endpoint.com"
creds = boto3.Session().get_credentials()
region = "us-west-1"
query_params = [
    ("eq", "attr1:value1"),
    ("eq", "attr2:value2"),
]

encoded_url = f"{url}?{urlencode(dt)}"

request = AWSRequest(method="GET", url=encoded_url)
SigV4Auth(creds, "execute-api", region).add_auth(request)
response = URLLib3Session().send(request.prepare())
```

### Reproduction Steps

See the examples above, but note that an actual api endpoint url is required to test it.

### Possible Solution

There might be other places that would need to change, but currently it fails because it tries to make a list of key/value tuples in `botocore.auth.SigV4Auth._canonical_query_string_params` without handling the case that they already are in the correct form. There is a comment that mentions repeated keys, but it would not be possible the way it is parsed.

Current:
``` python
[...]

PR #2679, opened 2022-05-19
add list of tuples functionality to AWSRequest params attribute

dlm6693:
Addresses https://github.com/boto/botocore/issues/2667. Allows request parameters to be passed to the low level interface `AWSRequest` as a list of tuples (`params` attribute). Currently it can only accept dictionaries. This will also allow the user to send duplicate queries should they wish to do so.

### DIFF
--- a/.changes/next-release/enhancement-AWSRequest-24612.json
+++ b/.changes/next-release/enhancement-AWSRequest-24612.json
@@ -1,5 +1,0 @@
-{
-  "type": "enhancement",
-  "category": "AWSRequest",
-  "description": "Allows AWSRequest to accept list of tuples in params attribute"
-}
--- a/.changes/next-release/enhancement-AWSRequest-24612.json
+++ b/.changes/next-release/enhancement-AWSRequest-24612.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "AWSRequest",
+  "description": "Allows AWSRequest to accept list of tuples in params attribute"
+}
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -19,6 +19,7 @@ import hmac
 import json
 import logging
 import time
+from collections.abc import Mapping
 from email.utils import formatdate
 from hashlib import sha1, sha256
 from operator import itemgetter
@@ -245,10 +246,11 @@ class SigV4Auth(BaseSigner):
     def _canonical_query_string_params(self, params):
         # [(key, value), (key2, value2)]
         key_val_pairs = []
-        for key in params:
-            value = str(params[key])
+        if isinstance(params, Mapping):
+            params = params.items()
+        for key, value in params:
             key_val_pairs.append(
-                (quote(key, safe='-_.~'), quote(value, safe='-_.~'))
+                (quote(key, safe='-_.~'), quote(str(value), safe='-_.~'))
             )
         sorted_key_vals = []
         # Sort by the URI-encoded key names, and in the case of
--- a/tests/unit/test_awsrequest.py
+++ b/tests/unit/test_awsrequest.py
@@ -134,6 +134,26 @@ class TestAWSRequest(unittest.TestCase):
             prepared_request.url, 'http://example.com/?bar=foo&foo=bar'
         )
 
+    def test_can_use_list_tuples_for_params(self):
+        request = AWSRequest(
+            url='http://example.com/', params=[('foo', 'bar')]
+        )
+        prepared_request = request.prepare()
+        self.assertEqual(
+            prepared_request.url, 'http://example.com/?foo=bar'
+        )
+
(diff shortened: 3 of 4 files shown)


tests/unit/test_awsrequest.py
```diff
@@ -134,6 +134,26 @@ def test_can_prepare_url_params_with_existing_query(self):
...
+    def test_can_use_list_tuples_for_params(self):
+        request = AWSRequest(
+            url='http://example.com/', params=[('foo', 'bar')]
+        )
+        prepared_request = request.prepare()
+        self.assertEqual(
+            prepared_request.url, 'http://example.com/?foo=bar'
+        )
+
+    def test_can_use_list_duplicate_tuples_for_params(self):
```
nateprewitt, line 146: We can probably use `pytest.parametrize` to collapse this into a single test with some more cases. Off the top of my head, we'll probably want to evaluate cases like:

* `[("foo", None)]`
* `[(,)]`
* `[(None, None)]`
* `[("foo", "bar"), ("foo", "baz"), ("fizz", "buzz")]`
* `[("food", "bar"), ("foo", None)]`

.changes/next-release/enhancement-AWSRequest-24612.json
```diff
@@ -0,0 +1,5 @@
+{
```
nateprewitt, line 1: While we generally want Changelog entries, `AWSRequest` isn't a public interface and is omitted from our public docs. We should probably keep this out of the Changelog to avoid any confusion.

botocore/auth.py
```diff
@@ -245,10 +246,11 @@ def canonical_query_string(self, request):
...
         # [(key, value), (key2, value2)]
         key_val_pairs = []
-        for key in params:
-            value = str(params[key])
+        if isinstance(params, Mapping):
+            params = params.items()
+        for key, value in params:
             key_val_pairs.append(
-                (quote(key, safe='-_.~'), quote(value, safe='-_.~'))
+                (quote(key, safe='-_.~'), quote(str(value), safe='-_.~'))
```
nateprewitt, line 253: What was the reason for casting the value to str here?

nateprewitt (changes requested):
This looks like a good start! I've left some minor notes, let me know what you think.

botocore/auth.py
dlm6693: I'm not sure, but its currently being cast to a string: https://github.com/boto/botocore/blob/develop/botocore/auth.py#L249
So I didn't want to change that underlying behavior since it wasn't in the scope of this ticket. Do you think I should remove it?

.changes/next-release/enhancement-AWSRequest-24612.json
dlm6693: Ah got it I thought it was needed for everything! I'll remove it.

botocore/auth.py
nateprewitt: Apologies, I read the diff too quickly. That's fine.

tests/unit/test_awsrequest.py
dlm6693: Oh cool! Didn't know about that feature. Very handy. The second option you provided raises a syntax error. Should I do something like two empty strings instead or some other falsy values?

nateprewitt: We were looking for an empty tuple. That's likely a failure case (which may be a different test) but I wanted to make sure we understand the failure states the list is introducing. We can do empty strings too (an empty key is what we specifically care about).

dlm6693: ok cool thank you for clarifying

nateprewitt: Also as a note, you'll need to move this out of the test class to use the `pytest.mark.parametrize` decorator. You can find an example [here](https://github.com/boto/botocore/blob/develop/tests/functional/test_model_completeness.py#L28-L31). You also don't need that generator function, you can write the cases inline like:

```python
@pytest.mark.parametrize(
    "params, expected_url",
    ([("foo", None)], "...?foo"),
    ([()], "...<no query string?>"),
    ...
)
```

In general we should be moving in that direction, the current testing layouts are due to our former test runner, `nosetest`.

tests/unit/test_awsrequest.py
```diff
@@ -268,6 +269,41 @@ def seek(self, where):
...
+    prepared_request = request.prepare()
+    assert prepared_request.url == expected_output
+
+
+def test_empty_list_tuples_value_error_for_params():
+    request = AWSRequest(
+        url='http://example.com/', params=[()]
+    )
+    with pytest.raises(ValueError):
+        request.prepare()
```
dlm6693, line 304: @nateprewitt since you mentioned we want to move away from test case classes, I put this one outside as well as the parametrized one.

nateprewitt (approved):
⛵

Issue #2396, opened 2021-05-28
PageIterator is not an iterator


Mateuszj111:
**Is your feature request related to a problem? Please describe.**
Hello,
my point is that PageIterator's name is quite confusing - it is an iterable not an iterator.
```python
pagination = conn.get_paginator("list_objects_v2").paginate(Bucket=bucket_name, Prefix=some_prefix)
next(pagination)

TypeError: 'PageIterator' object is not an iterator
```

**Describe the solution you'd like**
Either add ```__next__``` to PageIterator or change name to PageIterable.

kdaily:
Will check on this, @Mateuszj111, thanks!

PR #2685, opened 2022-05-25
Add doc string to `PageIterator`

dlm6693:
Addressing https://github.com/boto/botocore/issues/2396

While ideally we would add `next` functionality to `PageIterator` or simply update the name of the class to prevent confusion, both would cause breaking changes. As a result all we really can do is add a doc string with additional info. Perhaps using a warning here would be warranted, but decided against it.

Let me know if there's anything that should be added to the doc string.

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -187,7 +187,8 @@ class PaginatorModel:
 class PageIterator:
     """An iterable object to pagiante API results.
     Please note it is NOT a python iterator.
-    Use ``iter`` to wrap this as a generator"""
+    Use ``iter`` to wrap this as a generator.
+    """
 
     def __init__(
         self,
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -185,6 +185,9 @@ class PaginatorModel:
 
 
 class PageIterator:
+    """An iterable object to pagiante API results.
+    Please note it is NOT a python iterator"""
+
     def __init__(
         self,
         method,
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -186,7 +186,8 @@ class PaginatorModel:
 
 class PageIterator:
     """An iterable object to pagiante API results.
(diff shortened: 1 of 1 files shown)


nateprewitt (changes requested):
Minor clarification and formatting tweaks. Otherwise, looks good!

botocore/paginate.py
```diff
@@ -185,6 +185,9 @@ def get_paginator(self, operation_name):
 
 
 class PageIterator:
+    """An iterable object to pagiante API results.
+    Please note it is NOT a python iterator"""
+
```
nateprewitt, line 190:
```suggestion
    """An iterable object to pagiante API results.
    
    Please note it is NOT a python iterator. Use ``iter`` to wrap this as a generator.
    """
```

dlm6693: done

botocore/paginate.py
```diff
@@ -186,7 +186,8 @@ def get_paginator(self, operation_name):
 
 class PageIterator:
     """An iterable object to pagiante API results.
-    Please note it is NOT a python iterator"""
+    Please note it is NOT a python iterator.
+    Use ``iter`` to wrap this as a generator"""
```
nateprewitt, line 190: Can we make sure to include a period at the end of the last sentence? We'll also want to make sure the closing `"""` is on its own line.

nateprewitt: No we can keep it on the same line. The example in the first comment should have what's considered standard formatting for doc strings.

While we don't adhere strictly to PEP8 anymore, the [conventions for docstrings](https://peps.python.org/pep-0008/#documentation-strings) are still widely used. Generally to the effect of:

```python
class MyClass:
    """This is my class.
    
    Extra info beyond the class description.
    """
```

dlm6693: Got it thanks! Pushing up now

nateprewitt (approved):

PR #2689, opened 2022-05-31
Update examples

stealthycoin:


botocore/data/lambda/2015-03-31/examples-1.json
```diff
@@ -1,1513 +1,5 @@
 {
   "version": "1.0",
   "examples": {
-    "AddLayerVersionPermission": [
```
nateprewitt, line 4: This is deleting all of our Lambda examples. We may want to double check on this before merging.

botocore/data/pricing/2017-10-15/examples-1.json
```diff
@@ -62,42 +63,6 @@
         "id": "to-retreive-attribute-values",
         "title": "To retrieve a list of attribute values"
       }
-    ],
-    "GetProducts": [
```
nateprewitt, line 66: This is another one where we're completely deleting an operation example. Does this need to be upstreamed?

PR #2673, opened 2022-05-14
Replace deprecated inspect.formatargspec() with inspect.signature()

mgorny:
Originally submitted by Hugo van Kemenade as #2507.  Modified by me
to remove the first positional parameter like the old code did.

PR #2693, opened 2022-06-10
Add formal suport for Python 3.11

nateprewitt:
This PR is the start of getting us prepared for Python 3.11's release in October. The codebase _should_ currently run without issue, but our tests are broken due to deprecations in inspect.signature(). There is an existing PR (#2673) that should unblock this and is now mergable now that the 3.6 deprecation is done.

We'll leave this as a draft until that's merged and then follow up on final testing at that time.

### DIFF
--- a/.github/workflows/run-crt-test.yml
+++ b/.github/workflows/run-crt-test.yml
@@ -11,8 +11,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11-dev']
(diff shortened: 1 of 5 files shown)


nateprewitt:
Progress! Now that we've resolved the signature issues, we're finding other problems. The current release of JMESPath doesn't appear to be compatible with Python 3.11. Fortunately, the fix for the issue was already merged in jmespath/jmespath.py#217, it just hasn't been released yet.

I've reached out to James (of JMES fame) and requested a new release. Pending that, we _appear_ to be set to merge.

nateprewitt:
Alright, jmespath 1.0.1 is released and I think we're set 🎉

dlm6693 (approved):
🚢 pending tests passing

PR #2698, opened 2022-06-16
Replace deprecated inspect.formatargspec() with inspect.signature()

nateprewitt:
Porting the fix from #2673 forward to account for customizations in resources where `self` may still be included in the function signature.


dlm6693 (approved):
🚢

PR #2700, opened 2022-06-17
sse for SelectObjectCount s3 op

dlm6693:
This PR addresses https://github.com/boto/boto3/discussions/3268. Discussion about supporting SSE for all s3 ops was discussed, but the service only supports it for 11 out of its 97 APIs/ops.

A manual integration test was performed before and after registering the handler for `SelectObjectContent`. Before the same exception `botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the SelectObjectContent operation: Access Denied` was raised that the user in the original issue highlighted. After, the call succeeded.


nateprewitt (approved):
🚢

Issue #2376, opened 2021-05-04
EMR clients are not using the correct endpoint for certain regions


bitbier:
**Describe the bug**

For certain regions (namely us-west-2), the EMR client is not using the correct service endpoint URL. [According to the documentation](https://docs.aws.amazon.com/general/latest/gr/emr.html), the endpoint should be `elasticmapreduce.us-west-2.amazonaws.com`; however, the client is configured to use `us-west-2.elasticmapreduce.amazonaws.com`. This is normally not a problem until you use VPC endpoints from a private subnet. The DNS names for these VPC endpoints do not match the URL configured and so you have to force via the `endpoint_url` option when you create the client.

https://github.com/boto/botocore/blob/develop/botocore/data/endpoints.json#L2605-L2607

**Steps to reproduce**

1. Create a Botocore session
2. Configure an EMR client for us-west-2
3. Inspect endpoint

```
import botocore.session

session = botocore.session.Session()
session.set_debug_logger("")

emr = session.create_client("emr")
emr._endpoint
```

**Expected behavior**
I expected to the EMR services to hit the `elasticmapreduce.us-west-2.amazonaws.com` instead it hits the `us-west-2.elasticmapreduce.amazonaws.com`.

**Debug logs**

```
>>> emr = session.create_client("emr")
2021-05-03 18:43:09,936 - botocore.utils - DEBUG - IMDS ENDPOINT: http://169.254.169.254/
2021-05-03 18:43:09,942 - botocore.credentials - DEBUG - Looking for credentials via: env
2021-05-03 18:43:09,943 - botocore.credentials - DEBUG - Looking for credentials via: assume-role
2021-05-03 18:43:09,943 - botocore.credentials - DEBUG - Looking for credentials via: assume-role-with-web-identity
2021-05-03 18:43:09,943 - botocore.credentials - DEBUG - Looking for credentials via: sso
2021-05-03 18:43:09,943 - botocore.credentials - DEBUG - Looking for credentials via: shared-credentials-file
2021-05-03 18:43:09,943 - botocore.credentials - DEBUG - Looking for credentials via: custom-process
2021-05-03 18:43:10,347 - botocore.loaders - DEBUG - Loading JSON file: .../python3.6/site-packages/botocore/data/endpoints.json
2021-05-03 18:43:10,357 - botocore.hooks - DEBUG - Event choose-service-name: calling handler <function handle_service_name_alias at 0x10873cbf8>
[...]

PR #2704, opened 2022-06-21
Warn users of `sslCommonName` deprecation.

dlm6693:
This somewhat addresses https://github.com/boto/botocore/issues/2376. Since we changes to endpoints are coming soon, we aren't addressing the root of the problem at the moment, but sending a warning that it will be soon.


tests/unit/test_client.py
```diff
@@ -1607,6 +1607,25 @@ def test_client_close_context_manager(self):
...
+            ssl = self.endpoint_data['sslCommonName']
+            host = self.endpoint_data['hostname']
+            expected_warning_message = (
+                "The myservice client is currently using a deprecated endpoint:"
+                f" {ssl}. In the next minor version this will be moved to "
+                f"{host}. See https://github.com/boto/botocore/issues/2376 for "
+                "more details."
+            )
+            self.assertEqual(str(warning_message), expected_warning_message)
+
```
nateprewitt, line 1628: Let's avoid hardcoding the warning message here so we're not having to update the test every time it changes. One additional test we can do though is a second client creation and asserting it doesn't create a second warning.

```python
with pytest.warns(None) as warning:
    creator.create_client(
        'myservice', 'us-west-2', credentials=self.credentials
    )
self.assertEqual(len(warning.warnings), 0)
```

botocore/client.py
```diff
@@ -606,8 +608,21 @@ def _create_endpoint(
             resolved, region_name, endpoint_url
         )
         if endpoint_url is None:
-            # Use the sslCommonName over the hostname for Python 2.6 compat.
-            hostname = resolved.get('sslCommonName', resolved.get('hostname'))
+            sslCommonName = resolved.get('sslCommonName')
+            hostname = resolved.get('hostname')
+            disable = ensure_boolean(
```
nateprewitt, line 613: nit; looking at this again, maybe `is_disabled` is more consistent with our other booleans.

nateprewitt (changes requested):
Just a quick note on testing and a minor nit. Otherwise, looks good.

tests/unit/test_client.py
dlm6693: per our offline conversation `pytest.warns` doesn't work as expected and seemingly warnings are still thrown under the hood, just suppressed/not printed to the console. so I didn't add a new test

botocore/client.py
```diff
@@ -606,8 +608,21 @@ def _create_endpoint(
...
+            hostname = resolved.get('hostname')
+            is_disabled = ensure_boolean(
+                os.environ.get('BOTO_EXPERIMENTAL__DISABLE_COMMONNAME', '')
+            )
+            if not is_disabled and sslCommonName is not None:
+                warnings.warn(
+                    f'The {service_name} client is currently using a '
+                    f'deprecated endpoint: {sslCommonName}. In the next '
+                    f'minor version this will be moved to {hostname}. '
+                    'See https://github.com/boto/botocore/issues/2376 '
```
joguSD, line 621: Perhaps it might be better if we create an issue that's a bit more explicit on the what/why surrounding this deprecation. It feels a little weird to link to a closed issue that doesn't really explain the deprecation beyond describing one of the affected services.

dlm6693: Yeah that makes sense I'll make one

dlm6693: https://github.com/boto/botocore/issues/2705

botocore/client.py
```diff
@@ -606,8 +608,21 @@ def _create_endpoint(
             resolved, region_name, endpoint_url
         )
         if endpoint_url is None:
-            # Use the sslCommonName over the hostname for Python 2.6 compat.
-            hostname = resolved.get('sslCommonName', resolved.get('hostname'))
+            sslCommonName = resolved.get('sslCommonName')
+            hostname = resolved.get('hostname')
+            is_disabled = ensure_boolean(
+                os.environ.get('BOTO_DISABLE_COMMONNAME', '')
```
nateprewitt, line 614: Let's make this explicitly False if the value doesn't exist, that will let us short circuit the logic in `ensure_boolean`.

nateprewitt (changes requested):
One last tweak and I think this is set!

PR #2708, opened 2022-06-24
Allow botocore to be used as a zipped module

dlm6693:
This pr addresses https://github.com/boto/boto3/issues/1770. The `Loader` interface was originally built to only be able to handle JSON files stored on disk. However, zipping dependencies is a hard requirement for python runtime environments using `pyspark`. As such, this interface has been refactored to handle BOTH paths on disk as well as zipped paths. It accomplishes this by dynamically creating `zipfile.Path` and `pathlib.Path` objects at runtime, whereas before it used hardcoded string paths. Numerous unit tests and test data have been added and updated to validate that zipped files can be loaded. Additionally, new CI test run has been added that zips botocore and runs all unit tests as well as additional functional tests that validate a zipped context.


botocore/loaders.py
```diff
@@ -233,29 +307,27 @@ class Loader:
...
     FILE_LOADER_CLASS = JSONFileLoader
     # The included models in botocore/data/ that we ship with botocore.
-    BUILTIN_DATA_PATH = os.path.join(BOTOCORE_ROOT, 'data')
+    BUILTIN_DATA_PATH = BOTOCORE_ROOT.joinpath('data')
     # For convenience we automatically add ~/.aws/models to the data path.
-    CUSTOMER_DATA_PATH = os.path.join(
-        os.path.expanduser('~'), '.aws', 'models'
-    )
+    # this filepath can never be zipped and will always be on disk
+    CUSTOMER_DATA_PATH = pathlib.Path('~', '.aws', 'models').expanduser()
```
nateprewitt, line 319: This is subtly different since we're deferring expansion. That's likely not a problem, but it may warrant a spot check on at least Linux/macOS/Windows.

botocore/loaders.py
```diff
@@ -233,29 +307,27 @@ class Loader:
 
     FILE_LOADER_CLASS = JSONFileLoader
     # The included models in botocore/data/ that we ship with botocore.
-    BUILTIN_DATA_PATH = os.path.join(BOTOCORE_ROOT, 'data')
+    BUILTIN_DATA_PATH = BOTOCORE_ROOT.joinpath('data')
     # For convenience we automatically add ~/.aws/models to the data path.
-    CUSTOMER_DATA_PATH = os.path.join(
-        os.path.expanduser('~'), '.aws', 'models'
-    )
+    # this filepath can never be zipped and will always be on disk
```
nateprewitt, line 318: This is useful for us to know while implementing this change but may not warrant a comment.

botocore/loaders.py
```diff
@@ -233,29 +307,27 @@ class Loader:
...
+    CUSTOMER_DATA_PATH = pathlib.Path('~', '.aws', 'models').expanduser()
+
     BUILTIN_EXTRAS_TYPES = ['sdk']
 
     def __init__(
         self,
         extra_search_paths=None,
         file_loader=None,
-        cache=None,
```
nateprewitt, line 247: This is a breaking change and will cause positional args to map to different values. Is there a reason this needed to be removed?

botocore/loaders.py
```diff
@@ -302,19 +374,18 @@ def list_available_services(self, type_name):
...
-            for service_name in possible_services:
-                full_dirname = os.path.join(possible_path, service_name)
-                api_versions = os.listdir(full_dirname)
-                for api_version in api_versions:
-                    full_load_path = os.path.join(
-                        full_dirname, api_version, type_name
-                    )
-                    if self.file_loader.exists(full_load_path):
-                        services.add(service_name)
+            for service_name_path in possible_services:
```
nateprewitt, line 387: nit; `service_path`

botocore/loaders.py
```diff
@@ -302,19 +374,18 @@ def list_available_services(self, type_name):
...
+                full_load_paths = [
+                    path_obj.joinpath(type_name)
```
nateprewitt, line 389: Is this actually grabbing all the versions of the APIs with iterdir?

requirements-dev-lock.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.7
+# This file is autogenerated by pip-compile with python 3.10
```
nateprewitt, line 2: We need to make sure this lockfile is generated without our lowest supported version.

requirements-dev.txt
```diff
@@ -2,6 +2,8 @@ wheel==0.37.0
 behave==1.2.5
 jsonschema==2.5.1
 coverage==5.5
+importlib_resources==5.8.0
```
nateprewitt, line 5: I don't think we want to add this to the dev dependencies. Was there something specific we needed?

requirements-dev.txt
```diff
@@ -2,6 +2,8 @@ wheel==0.37.0
...
+zipp==3.1.0
```
nateprewitt: Same note.

tests/functional/test_zipped_botocore.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
nateprewitt, line 1: We need to make sure we're using accurate copyright dates so we're not misattributing new code to the past. Overall though, we should probably keep this out of the normal test suite. Let's try to move these checks to `run-zip-tests` to establish state before we run checks.

tests/unit/data/foo/ec2/2010-01-01/service-2.json
```diff
@@ -0,0 +1 @@
+{"foo": "bar"}
```
nateprewitt, line 1: For these new service files, is there something particular we're trying to validate with each?

tests/unit/test_loaders.py
```diff
@@ -22,12 +22,19 @@
 import contextlib
 import copy
 import os
+import pathlib
+
+try:
+    from zipfile import Path as ZipPath
+except ImportError:
+    from zipp import Path as ZipPath
```
nateprewitt, line 30: What situation will zipfile not be present?

tests/unit/test_loaders.py
```diff
@@ -36,42 +43,80 @@
...
         data = self.file_loader.load_file(self.valid_file_path)
         self.assertEqual(len(data), 3)
         self.assertTrue('test_key_1' in data)
 
+        zip_data = self.file_loader.load_file(self.zip_valid_file_path)
+        self.assertEqual(len(zip_data), 3)
+        self.assertTrue('test_key_1' in zip_data)
+
+        self.assertEqual(data, zip_data)
+
```
nateprewitt, line 66: This should be a distinct test so it's clear what's failing, i.e. `test_load_file_zip`

tests/unit/test_loaders.py
```diff
@@ -36,42 +43,80 @@
...
         data = self.file_loader.load_file(self.compressed_file_path)
         self.assertEqual(len(data), 3)
         self.assertTrue('test_key_1' in data)
 
+        zip_data = self.file_loader.load_file(self.zip_compressed_file_path)
+        self.assertEqual(len(zip_data), 3)
+        self.assertTrue('test_key_1' in data)
+
+        self.assertEqual(data, zip_data)
+
```
nateprewitt, line 77: Same note here about individual tests.

tests/unit/test_loaders.py
```diff
@@ -36,42 +43,80 @@
...
     def test_load_compressed_file_exists_check(self):
         self.assertTrue(self.file_loader.exists(self.compressed_file_path))
+        self.assertTrue(self.file_loader.exists(self.zip_compressed_file_path))
 
     def test_load_json_file_does_not_exist_returns_none(self):
         # None is used to indicate that the loader could not find a
         # file to load.
-        self.assertIsNone(self.file_loader.load_file('fooasdfasdfasdf'))
+        self.assertIsNone(
+            self.file_loader.load_file(pathlib.Path('fooasdfasdfasdf'))
```
nateprewitt, line 86: Do we need to make this a pathlib object? Presumably a string should still work with this since it's a public interface.

tests/unit/test_loaders.py
```diff
@@ -87,39 +132,63 @@ def test_default_search_paths(self):
...
+            any(str(p).endswith(home_dir_path) for p in loader.search_paths)
         )
+        self.assertIsInstance(loader.search_paths, SearchPathList)
 
     def test_can_add_to_search_path(self):
         loader = Loader()
         loader.search_paths.append('mypath')
-        self.assertIn('mypath', loader.search_paths)
+        self.assertIn(pathlib.Path('mypath'), loader.search_paths)
+        self.assertIsInstance(loader.search_paths, SearchPathList)
```
nateprewitt, line 143: This seems like an implementation detail, we know if we add a string and it results in a Path object the conversion is working. Whether that's a SearchPathList isn't actually important to us.

tests/unit/test_loaders.py
```diff
@@ -87,39 +132,63 @@ def test_default_search_paths(self):
...
+        bad_types = (
+            None,
+            [1],
+            1,
+            1.5,
+            {'bad': 'type'},
+            b'bad_type',
+            False,
+            {'bad_type'},
+        )
```
nateprewitt, line 172: Having these check lists inside tests usually makes debugging failures difficult. Can we move this out to a parametrize decorator? It's not actually using any of the instance state for the test.

nateprewitt (changes requested):
Thanks for the PR! I haven't been able to go through everything but here's some high level initial feedback that will help move the pull request forward.

botocore/__init__.py
```diff
@@ -13,9 +13,14 @@
...
 import logging
-import os
 import re
 
+try:
+    from importlib.resources import files
+# python < 3.9 compatability
+except ImportError:
+    from importlib_resources import files
+
```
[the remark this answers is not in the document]
dlm6693: removed and updated to not support < 3.9/create a `pathlib.Path` object for `BOTOCORE_ROOT` instead

botocore/loaders.py
```diff
@@ -156,42 +155,48 @@ class JSONFileLoader:
...
             return
 
-        # By default the file will be opened with locale encoding on Python 3.
-        # We specify "utf8" here to ensure the correct behavior.
-        with open_method(full_path, 'rb') as fp:
-            payload = fp.read().decode('utf-8')
+        with full_path.open('rb') as fp:
+            payload = fp.read()
+            if ext == '.json.gz':
+                payload = decompress(payload)
```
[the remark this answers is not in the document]
dlm6693: sure i'll update

botocore/loaders.py
```diff
@@ -221,6 +226,75 @@ def create_loader(search_path_string=None):
     return Loader(extra_search_paths=paths)
 
 
+def _create_path(path):
+    """Construct a zipfile.Path object by splitting the root (path to the zip)
+    and the path to the subdirectory/file e.g.
+    '/Users/user123/Documents/Archive.zip/data' -->
+    zipfile.Path('/Users/user123/Documents/Archive.zip', '/data')
+    If the zipped path doesn't exist, return a pathlib.Path object instead
```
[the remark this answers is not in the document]
dlm6693: got it

botocore/loaders.py
```diff
@@ -221,6 +226,75 @@ def create_loader(search_path_string=None):
...
+        :return: A zipfile.Path or pathlib.Path object
+
+        """
+        if isinstance(__object, str):
+            return _create_path(__object)
+
+        if not any(
+            isinstance(__object, _type) for _type in self.ALLOWED_TYPES
+        ):
+            repr_types = ', '.join(repr(_type) for _type in self.ALLOWED_TYPES)
```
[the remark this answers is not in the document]
dlm6693: ok i'll change to `allowed_type_string`

botocore/loaders.py
```diff
@@ -221,6 +226,75 @@ def create_loader(search_path_string=None):
...
+    return pathlib.Path(path)
+
+
+class SearchPathList(list):
+    """A simple wrapper for a list. Used to mutate objects that are appended
+    to Loader.search_paths to be pathlib.Path or zipfile.Path objects
+
+    """
+
+    ALLOWED_TYPES = (str, zipfile.Path, pathlib.Path)
```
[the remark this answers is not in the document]
dlm6693: they'll make it to `Loader.search_paths` as is and will raise an `AttributeError` in `_potential_locations`. Either way an error is going to be raised at some point. This was based off a suggestion from @jamesls in the design doc

botocore/loaders.py
dlm6693: ok i'll remove thanks

botocore/loaders.py
dlm6693: was going to comment that i didn't see it being used anywhere in the code so thought it made to remove. there is a private `_cache` variable that is created at initialization, but does not reference this arg at all. Nor does the `instance_cache` decorator. i'll add it back

botocore/loaders.py
dlm6693: Yes. Here is an example for `ec2` from my machine with the first part of the directory path redacted
```python
type_name = 'service-2'
service_path = `PosixPath('..../botocore/botocore/data/ec2')
[p.joinpath('service-2') for p in service_path.iterdir()]
[PosixPath('..../botocore/botocore/data/ec2/2015-10-01/service-2'),
 PosixPath('..../botocore/botocore/data/ec2/2016-11-15/service-2'),
 PosixPath('....botocore/botocore/data/ec2/2015-04-15/service-2'),
 PosixPath('..../botocore/botocore/data/ec2/2016-04-01/service-2'),
 PosixPath('..../botocore/botocore/data/ec2/2014-10-01/service-2'),
 PosixPath('..../botocore/botocore/data/ec2/2016-09-15/service-2'),
 PosixPath('..../botocore/botocore/data/ec2/2014-09-01/service-2'),
 PosixPath('..../botocore/botocore/data/ec2/2015-03-01/service-2')]

File: requirements-dev-lock.txt (hunk 6 above)
Comment (dlm6693) [reply requirements-dev-lock.txt:2]: got it i'll revert to what's in `develop` since we aren't adding any new dependencies or updating versions

File: requirements-dev.txt (hunk 7 above)
Comment (dlm6693) [reply requirements-dev.txt:5]: yeah im removing now

File: tests/functional/test_zipped_botocore.py (hunk 9 above)
Comment (dlm6693) [reply tests/functional/test_zipped_botocore.py:1]: yep i've removed this file and move some state checks to `run-zip-tests`. I've removed any attempts to load data or service models in there as the unit tests will handle those checks after establishing a zipped state.

File: tests/unit/test_loaders.py (hunk 11 above)
Comment (dlm6693) [reply tests/unit/test_loaders.py:30]: `zipfile.Path` only exists as of python 3.8. Since we are only supporting python 3.9+, the exception handling will remain, but instead a dummy `ZipPath` class will be created that raises a `RuntimeError` if its ever initialized

File: tests/unit/test_loaders.py (hunk 14 above)
Comment (dlm6693) [reply tests/unit/test_loaders.py:86]: ah I was under the impression the file loader should be treated as a private interface even though it doesn't start with `_`. i'll update it

File: tests/unit/test_loaders.py (hunk 16 above)
Comment (dlm6693) [reply tests/unit/test_loaders.py:172]: yup will do

File: botocore/loaders.py [hunk 22]
```diff
@@ -156,42 +155,48 @@ class JSONFileLoader:
...
 
         :return: The loaded data if it exists, otherwise None.
 
         """
-        for (ext, open_method) in _JSON_OPEN_METHODS.items():
-            data = self._load_file(file_path + ext, open_method)
+        for ext in _JSON_EXTENSIONS:
+            # pathlib.Path.with_suffix would be useful here
+            # but zipfile.Path does not have anything comparable
+            with_ext = path_obj.parent.joinpath(path_obj.name + ext)
```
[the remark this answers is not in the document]
dlm6693: sure. the existing version concatenated strings this way so wanted to match style, but i'll change it

botocore/loaders.py
dlm6693: would you rather i change it to `pathlib.Path('~').expanduser().joinpath('.aws', 'models')` or add a simple unit test?

scripts/ci/run-tests
```diff
@@ -36,6 +36,8 @@ def process_args(args):
         test_args += "-n auto --dist=loadfile --maxprocesses=4 "
     if args.with_cov:
         test_args += f"--cov={PACKAGE} --cov-report xml "
+    if args.zipped:
+        test_args += "--import-mode=append "
```
dlm6693, line 40: this is required for the test runner to check paths in the correct order for loading modules (i.e. prioritizing `botocore.zip` over botocore on disk

scripts/ci/run-zip-tests
```diff
@@ -0,0 +1,61 @@
...
+
+
+def check_state():
+    zip_loader = isinstance(botocore.__loader__, zipimport.zipimporter)
+    zip_file = 'botocore.zip/botocore/__init__.py' in botocore.__file__
+    zip_path = isinstance(botocore.BOTOCORE_ROOT, zipfile.Path)
+    zipped_state = (zip_loader, zip_file, zip_path)
+    if not all(zipped_state):
+        print("Botocore is not loaded in a zipped context")
+        sys.exit(1)
```
dlm6693, line 54: we probably don't need to do all these checks. maybe just one or two, but perhaps there are others we could/should add?

botocore/loaders.py
```diff
@@ -103,21 +103,34 @@
...
+
+    class ZipPath:
+        def __init__(self):
+            raise RuntimeError(
+                "Using botocore in a zipped context "
+                "is not supported in python < 3.9"
+            )
+
+else:
+    from zipfile import Path as ZipPath
```
dlm6693, line 121: Since botocore will not support zips in python < 3.9, but `zipfile.Path` objects are native as of python 3.8, this check needs to be performed. I'm not a huge fan of branching logic on imports, but I'm not sure of how else to do this. If anyone knows of a more elegant way to check, let me know.

botocore/loaders.py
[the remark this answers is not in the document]
dlm6693: should've checked before commenting and resolving, but unfortunately `zipfile.Path` does not have a `suffixes` interface. `pathlib.Path` has far more functionality they don't provide identical APIs

tests/unit/test_loaders.py
```diff
@@ -36,46 +41,128 @@
...
+        self.valid_file_path = self.data_path.joinpath('foo')
+        self.compressed_file_path = self.data_path.joinpath('compressed')
+        if sys.version_info >= (3, 9):
+            self.zip_data_path = ZipPath(
+                self.data_path.joinpath('Archive.zip')
+            )
+            self.zip_valid_file_path = self.zip_data_path.joinpath('foo')
+            self.zip_compressed_file_path = self.zip_data_path.joinpath(
+                'compressed'
+            )
```
nateprewitt, line 55: Ideally we don't want to leak these details into the `setUp`. I think we'd talked about doing an explicit decorator for zip functionality. You can take a look at some examples [here](https://github.com/boto/botocore/blob/develop/tests/__init__.py#L56-L79).

We can either do this potentially as a pytest fixture or separate function. It doesn't look like the zip tests actually use anything from the class anyways, so the grouping may not be appropriate. We could consider moving these to their own test class.

tests/unit/test_loaders.py
```diff
@@ -206,9 +294,116 @@ def test_load_service_model_uses_provided_type_name(self):
...
+
+    @pytest.mark.skipif(
+        sys.version_info >= (3, 9), reason="Python version >= 3.9"
+    )
+    def test_zips_not_suppored_python_lt_39(self):
+        search_path = os.path.join(
+            os.path.dirname(__file__), 'data', 'Archive.zip', 'foo'
+        )
+        with self.assertRaises(RuntimeError):
+            create_loader(search_path)
```
nateprewitt, line 309: I don't know if we need to explicitly assert this for <3.9. We may want to make the check more explicit that if zipfile.Path doesn't exist, we can't load the zip directory.

tests/unit/test_loaders.py
```diff
@@ -206,9 +294,116 @@ def test_load_service_model_uses_provided_type_name(self):
...
+        zip_path = ZipPath(os.path.dirname(search_path)).joinpath('foo')
+        # two identical ZipPath objects do not 'equal' each other
+        self.assertIn(
+            str(zip_path), (str(path) for path in loader.search_paths)
+        )
+        # sanity check that one of the paths is a ZipPath
+        # since it can't be checked directly
+        self.assertTrue(
+            any(isinstance(path, ZipPath) for path in loader.search_paths)
+        )
```
nateprewitt, line 331: This may be a simpler approach and more exact in what we're checking.
```suggestion
        # two identical ZipPath objects do not 'equal' each other
       matching_zips = [path for path in loader.search_paths if str(path) == str(zip_path)]
       assert len(matching_zips) == 1
       assert isinstance(matching_zips[0], ZipPath)
```

tests/unit/test_loaders.py
```diff
@@ -206,9 +294,116 @@ def test_load_service_model_uses_provided_type_name(self):
...
+    @pytest.mark.skipif(
+        sys.version_info < (3, 9), reason="Python version < 3.9"
+    )
+    def test_zipped_list_api_versions(self):
+        path = os.path.join(
+            os.path.dirname(__file__), 'data', 'Archive.zip', 'foo'
+        )
+        loader = Loader(
+            extra_search_paths=[path], include_default_search_paths=False
+        )
```
nateprewitt, line 404: There's a few repeating patterns here. When we find that in test cases, it's usually a good sign to move it into a single function or fixture. That will keep the codebase slimmer, make future changes easier to organize, and reduce the risk of accidental bugs being missed.

scripts/ci/run-zip-tests
```diff
@@ -0,0 +1,66 @@
...
+    zip_file = 'botocore.zip/botocore/__init__.py' in botocore.__file__
+    print(f'file: {botocore.__file__}')
+    zip_path = isinstance(botocore.BOTOCORE_ROOT, zipfile.Path)
+    print(f'type: {botocore.BOTOCORE_ROOT.__module__}')
+    zipped_state = (zip_loader, zip_file, zip_path)
+    if not all(zipped_state):
+        print("Botocore is not loaded in a zipped context")
+        print(os.environ['GITHUB_WORKSPACE'])
+        print(os.listdir())
+        sys.exit(1)
```
nateprewitt, line 59: It may be easier to read if the prints and variables are grouped together. The final failed state output may be able to be included in a single print statement for legibility.

nateprewitt (changes requested):

scripts/ci/run-tests
dlm6693: Sure. I think I'd prefer the latter since that reflects the style of the other existing flags

.github/workflows/run-tests.yml
```diff
@@ -27,6 +27,13 @@ jobs:
       - name: Run tests
         run: |
           python scripts/ci/run-tests --with-cov --with-xdist
+      - name: Run tests in zipped context
+        env:
+          PYTHONPATH: ${{ env.GITHUB_WORKSPACE }}/botocore.zip
+        run: |
+          zip -r botocore.zip botocore/
+          python scripts/ci/run-tests --with-cov --with-xdist --zipped
+          rm botocore.zip
```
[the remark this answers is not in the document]
dlm6693: not sure why you were seeing this in the diff. its already been removed

botocore/loaders.py
```diff
@@ -103,21 +103,26 @@
...
 import logging
 import os
+import pathlib
+from zipfile import is_zipfile
+
+try:
+    from zipfile import Path as ZipPath
+except ImportError:
+    # python <3.8 compatibility
+    from zipp import Path as ZipPath
```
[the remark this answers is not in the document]
dlm6693: same thing here. this code has also been updated already

.python-version
```diff
@@ -0,0 +1 @@
+3.10.4
```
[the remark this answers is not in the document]
dlm6693: 😞

botocore/__init__.py
```diff
@@ -13,9 +13,19 @@
...
+    from importlib.resources import files
+# python < 3.9 compatability
+except ImportError:
+    import os
+    import pathlib
+
+    def files(path):
+        return pathlib.Path(os.path.dirname(os.path.abspath(__file__)))
+
+
```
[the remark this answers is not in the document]
dlm6693: `files` will construct a `zipfile.Path` object if it detects a zipped context for botocore.  As far as I can tell, `files` expects the name of a package, not the path to it, which is why `__file__` isn't called. `pathlib` is imported in `loaders.py` so I'll move both imports to the top.

botocore/loaders.py
```diff
@@ -103,21 +103,34 @@
...
+
+    class ZipPath:
+        def __init__(self, path):
+            raise RuntimeError(
+                "Using botocore in a zipped context "
+                "is not supported in python < 3.9"
+            )
+
+else:
+    from zipfile import Path as ZipPath
```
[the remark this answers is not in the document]
dlm6693: i had that in an earlier version. my main concern with that approach is the import will work for python 3.8. Are we ok with 3.7 and 3.8 breaking for slightly different reasons if a user attempts to use zips?

botocore/loaders.py
```diff
@@ -302,19 +396,18 @@ def list_available_services(self, type_name):
...
-                    if self.file_loader.exists(full_load_path):
-                        services.add(service_name)
+            for service_path in possible_services:
+                full_load_paths = [
+                    path_obj.joinpath(type_name)
+                    for path_obj in service_path.iterdir()
+                ]
+                for path in full_load_paths:
+                    if self.file_loader.exists(path):
+                        services.add(service_path.name)
```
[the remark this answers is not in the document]
dlm6693: iirc i was trying to mimic the existing logic as much as possible. i'll update to something like

```python
for path in service_path.iterdir():
    full_path = path.joinpath(type_name)
    if self.file_loader.exists(full_path):
    services.add(service_path.name)
    break

File: tests/unit/test_loaders.py (hunk 26 above)
Comment (dlm6693) [reply tests/unit/test_loaders.py:55]: Ah ok makes sense! Yeah I was thinking some sort of decorator like `pytest.mark.skipif` would be useful here, but wasn't sure how it'd work without splitting everything into a separate class. This was definitely more of a lazy approach. I'll see what I can cobble together

File: tests/unit/test_loaders.py (hunk 27 above)
Comment (dlm6693) [reply tests/unit/test_loaders.py:309]: ok should there be no assertion if it does have `Path`?

File: scripts/ci/run-zip-tests (hunk 30 above)
Comment (dlm6693) [reply scripts/ci/run-zip-tests:59]: i actually included all of these print statements solely for debugging purposes.  zip test runs are failing on gh actions here and i'm trying to understand why. i've tested the workflow locally using `act` and it was working as expected

File: tests/unit/test_loaders.py (hunk 29 above)
Comment (dlm6693) [reply tests/unit/test_loaders.py:404]: sure. i came from a job previously where a lot of folks felt the opposite. that hard coding tests with repeatable information made it easier to read, debug and catch bugs because of all the hand writing required. I don't necessarily feel strongly either way, but I see the merits of both sides. Regardless, if its preferred to make tests as dry as possible, I'm happy to do so.

File: botocore/loaders.py [hunk 37]
```diff
@@ -156,42 +168,51 @@ class JSONFileLoader:
...
+        for ext in _JSON_DECOMPRESSION_METHODS:
+            path = file_path.parent.joinpath(f'{file_path.name}{ext}')
+            if path.is_file() and path.exists():
                 return True
         return False
 
-    def _load_file(self, full_path, open_method):
-        if not os.path.isfile(full_path):
+    def _load_file(self, full_path, ext):
+        if not full_path.is_file() or not full_path.exists():
```
dlm6693, line 188: same issue for `is_file` here.

botocore/loaders.py
```diff
@@ -221,6 +242,90 @@ def create_loader(search_path_string=None):
...
+    additional details.
+
+    """
+
+    def joinpath(self, *other):
+        next = posixpath.join(self.at, *other)
+        return self._next(self.root.resolve_dir(next))
+
+    def _next(self, at):
+        return self.__class__(self.root, at)
```
dlm6693, line 281: added this because `joinpath` was returning a `zipfile.Path` type instead of `BotoZipPath`, which caused some tests to fail.

.github/workflows/run-zip-test.yml
```diff
@@ -0,0 +1,36 @@
...
+        run: |
+          if [ "$RUNNER_OS" == "Windows" ]; then
+            export PYTHONPATH='${{ github.workspace }}\botocore.zip'
+          else
+            export PYTHONPATH=${{ github.workspace }}/botocore.zip
+          fi
+          python -m zipfile -c botocore.zip botocore/
+          python scripts/ci/run-zip-tests --with-cov --with-xdist --append-import
+          rm botocore.zip
+        shell: bash
```
dlm6693, line 36: this took a number of iterations to figure out for windows. setting an environment variable in different steps via
```bash
echo "PYTHONPATH=${{ github.workspace }}\botocore.zip" >> $GITHUB_ENV
```
worked fine for mac/linux, but consistently had issues with Windows. I think it has to do with how gh actions treats forward slashes as escapes. Passing it between steps would result in it disappearing so I would see something like this
```bash
echo ${{ env.PYTHONPATH}}
D:\a\botocore\botocorebotocore.zip
```

botocore/loaders.py
```diff
@@ -461,16 +564,17 @@ def _potential_locations(self, name=None, must_exist=False, is_dir=False):
         # Will give an iterator over the full path of potential locations
         # according to the search path.
         for path in self.search_paths:
-            if os.path.isdir(path):
+            if path.is_dir():
                 full_path = path
                 if name is not None:
-                    full_path = os.path.join(path, name)
+                    name_parts = name.split(os.sep)
+                    full_path = path.joinpath(*name_parts)
```
dlm6693, line 571: this is required to handle windows paths i.e `\foo\bar\test.` Even on windows, `zipfile` uses posix path separators `/`

tests/__init__.py
```diff
@@ -79,6 +79,17 @@ def decorator(func):
...
 
+def skip_if_lt_39(func):
+    def decorator(*args, **kwargs):
+
+        if sys.version_info < (3, 9):
+            pass
+        else:
+            func(*args, **kwargs)
+
+    return decorator
```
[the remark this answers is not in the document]
nateprewitt: Ok after so more thought, here's what I'd propose. Let's concretely define the closure of requirements. We know we need:

1. from importlib.resources  import files
2. from zipfile import Path

With both of those, are there any other concrete requirements? I may be missing one but let's define something like this in the project `__init__`.

```python
try:
    # These imports are generally only support in Python 3.9+
    from importlib.resources  import files
    from zipfile import Path
    HAS_ZIP_SUPPORT = True
except ImportError:
    HAS_ZIP_SUPPORT = False
```

Now we'll use this elsewhere to create a consistent restriction so we're not stuck with multiple disparate conditionals. I'll write them inline as separate comments.

Starting here, let's do something like:
```python
def requires_zip_support():
    reason = "Test requires zip dependencies available in Python 3.9+"
    def decorator(func):
        return unittest.skipIf(not HAS_ZIP_SUPPORT, reason)(func)
    return 
```

scripts/ci/run-zip-tests
```diff
@@ -0,0 +1,51 @@
...
+
+
+def check_state():
+    zip_loader = isinstance(botocore.__loader__, zipimport.zipimporter)
+    zip_file = os.path.join('botocore.zip', 'botocore', '__init__.py') in botocore.__file__
+    zip_path = isinstance(botocore.BOTOCORE_ROOT, zipfile.Path)
+    zipped_state = (zip_loader, zip_file, zip_path)
+    if not all(zipped_state):
+        print("Botocore is not loaded in a zipped context")
+        sys.exit(1)
```
nateprewitt, line 44: These are good, let's also make sure we're verifying `HAS_ZIP_SUPPORT` is set to True.

botocore/loaders.py
```diff
@@ -221,6 +242,93 @@ def create_loader(search_path_string=None):
...
+
+    :return: A BotoZipPath/pathlib.Path object consisting of the root and
+    the path to a file or subdirectory
+
+    """
+    path_obj = pathlib.Path(path)
+    resolved_path = path_obj.resolve()
+    parts = resolved_path.parts
+    new_path = ''
+    while len(parts) > 0:
```
nateprewitt, line 264: This lets us short circuit the whole zip construction flow and solves the Python 3.7/3.8 issue I asked about earlier.

```suggestion
    while parts and HAS_ZIP_SUPPORT:
```

botocore/loaders.py
```diff
@@ -103,21 +103,33 @@
...
+try:
+    from zipfile import Path as ZipPath
+except ImportError:
+
+    class ZipPath:
+        def __init__(self, path):
+            raise RuntimeError(
+                "Using botocore in a zipped context "
+                "is not supported in python < 3.9"
+            )
```
nateprewitt, line 119: Lets tie this directly to `HAS_ZIP_SUPPORT`. If we need to validate `zipfile` has Path available (I'm assuming all versions support?), move that into the initialization requirements for `HAS_ZIP_SUPPORT`.

botocore/__init__.py
```diff
@@ -14,10 +14,20 @@
...
 
+try:
+    # only available in python 3.9+
+    from importlib.resources import files
+
+    BOTOCORE_ROOT = files('botocore')
+except ImportError:
+
+    BOTOCORE_ROOT = pathlib.Path(os.path.dirname(os.path.abspath(__file__)))
+
```
nateprewitt, line 30: Same note here, no try/except. Just if/else on `HAS_ZIP_SUPPORT`.

nateprewitt (changes requested):

botocore/loaders.py
```diff
@@ -156,42 +168,51 @@ class JSONFileLoader:
...
 
         """
-        for ext in _JSON_OPEN_METHODS:
-            if os.path.isfile(file_path + ext):
+        if isinstance(file_path, str):
+            file_path = _create_path(file_path)
+
+        for ext in _JSON_DECOMPRESSION_METHODS:
+            path = file_path.parent.joinpath(f'{file_path.name}{ext}')
+            if path.is_file() and path.exists():
```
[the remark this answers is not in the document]
dlm6693: I did some digging and this appears to be a bug with `zipfile.Path.is_file` in python 3.9:

```python
# python 3.9
def is_file(self):
    return not self.is_dir()

```python

File: .github/workflows/run-zip-test.yml (hunk 39 above)
Comment (dlm6693) [reply .github/workflows/run-zip-test.yml:36]: i was getting a parsing error with the if statement, but it doesn't appear to be an issue now

File: botocore/loaders.py [hunk 47]
```diff
@@ -156,42 +168,51 @@ class JSONFileLoader:
...
         return False
 
-    def _load_file(self, full_path, open_method):
-        if not os.path.isfile(full_path):
+    def _load_file(self, full_path, ext):
+        if not full_path.is_file() or not full_path.exists():
             return
 
-        # By default the file will be opened with locale encoding on Python 3.
-        # We specify "utf8" here to ensure the correct behavior.
```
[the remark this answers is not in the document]
dlm6693: no. the current loader only ever decodes using utf-8. i can write some tests, but since its not behavior we supported before, is it necessary?

botocore/loaders.py
```diff
@@ -221,6 +242,93 @@ def create_loader(search_path_string=None):
...
+    """
+    path_obj = pathlib.Path(path)
+    resolved_path = path_obj.resolve()
+    parts = resolved_path.parts
+    new_path = ''
+    while len(parts) > 0:
+        new_path = os.path.join(new_path, parts[0])
+        parts = parts[1:]
+        if is_zipfile(new_path):
+            return BotoZipPath(new_path).joinpath(*parts)
```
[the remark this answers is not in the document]
dlm6693: quick update on benchmarking: turning `parts` into a list and using `pop` on it in an f-string did improve performance by roughly 18-20% on zip paths. For non-zip paths, there was a very slight, but much smaller boost (3-4%)

tests/__init__.py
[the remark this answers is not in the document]
dlm6693: The reason why I created this custom decorator was because I applied it to a unit test helper function, not a test itself. IIRC it wasn't working correctly, but maybe I'm misremembering. I'll refactor based on your suggestions.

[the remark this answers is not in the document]
dlm6693: fwiw `importlib.resources` is only native as of 3.9 while `zipfile.Path` is as of 3.8 so that import line will never be reached when we expect the import error to be raised. should it still be included?

botocore/loaders.py
dlm6693: I'm going to keep `while len(parts) > 0` unless you had a specific reason to remove?

botocore/loaders.py
dlm6693: no `zipfile.Path` is only native as of python 3.8. regardless, because we try to import `importlib.resources.files` which is only a part of 3.9+, it will have the desired effect

botocore/__init__.py
dlm6693: not sure i understand this comment. this try/except establishes whether or not `HAS_ZIP_SUPPORT` is true or false. Did you want it defined somewhere else, perhaps `compat.py` and imported here?

botocore/loaders.py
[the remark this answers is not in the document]
dlm6693: sure that makes sense. maybe we can have a whole suite of files using encodings from a given list and test decoding all of those?

tests/unit/test_loaders.py
```diff
@@ -22,62 +22,214 @@
...
+    ],
+)
+def test_load_different_locales(file_name, expected_unicode):
+    base_path = pathlib.Path(__file__).parent.joinpath('data', 'locales')
+    file_loader = JSONFileLoader()
+    data = file_loader.load_file(base_path.joinpath(file_name))
+    encoded = data['name'].encode('utf-8')
+    # it appears that pytest decodes unicode implicitly
+    expected_encoded = expected_unicode.encode('utf-8')
+    assert encoded == expected_encoded
```
dlm6693, line 222: @nateprewitt is this what you had in mind for locales/encodings? I can try encoding in utf-16 and utf-32 since those are the two other formats json supports.

PR #2715, opened 2022-07-05
Use alternative credential provider if shared credential has been expired

sedyn:
I am using `saml2aws` which returns expirable access token for `credential_process`.

When token is expired, botocore does not re-run `credential_process`.

I revised `SharedCredentialProvider` class so it can check the expiration of access token and make alternative providers to load credential.

ref. https://github.com/aws/aws-toolkit-vscode/issues/2664

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1291,6 +1291,8 @@ class SharedCredentialProvider(CredentialProvider):
     # so we support both.
     TOKENS = ['aws_security_token', 'aws_session_token']
 
+    TOKEN_EXPIRES_KEY = 'x_security_token_expires'
+
     def __init__(self, creds_filename, profile_name=None, ini_parser=None):
         self._creds_filename = creds_filename
         if profile_name is None:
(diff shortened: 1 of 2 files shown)


simnalamburt:
*Call to maintainers*

It's been almost a year since this PR was submitted, and no one has reviewed it. I need this patch. If there is anything I need to do to get it reviewed, I will help. Is there anything I can do to help make this patch get merged?

The rationale behind this patch is simple: If an existing token in `~/.aws/credentials` expires, it will not be used and the next authentication method will be used. If you have concerns about this approach and are hesitant to move forward, we'd love to talk about it. I'm willing to work around this in a way that doesn't cause maintainers to worry.

RyanFitzSimmonsAK:
Hey, thanks for this pull request and for your patience.  After discussing with the team, we've decided to close this as not planned.  This change would definitely be considered breaking, and could potentially lead to users using credentials that they don't intend to.  It's also a little unintuitive; I think most users would expect expired credentials to fail with an error, rather than use different ones.

PR #2721, opened 2022-07-22
fix: resolve potential race condition in JSONFileCache

msw-kialo:
The `JSONFileCache.__setitems__` might raise a `FileExistsError` if it is
run in parallel by two threads / processes:
1. Both processes discover the working dir hasn't been created yet.
2. The first process creates the working dir.
3. The second process tries to create the working dir, too and fails
   as it already exists.

Accept this potential race — `exist_ok=True` will prevent `os.makedirs`
to raise an exception.

The `os.path.isdir` condition is technically no longer needed but kept
keeping the change small. Furthermore, it should be slightly faster
as the cache should exist most of the time.

The `exist_ok` parameter is supported since Python 3.2 / fixed in 3.4.1 
and therefore available in all needed python versions.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -342,7 +342,8 @@ class JSONFileCache:
                 f"JSON serializable: {value}"
             )
         if not os.path.isdir(self._working_dir):
-            os.makedirs(self._working_dir)
+            # a racing parallel thread / process might just have created the working dir:
+            os.makedirs(self._working_dir, exist_ok=True)
         with os.fdopen(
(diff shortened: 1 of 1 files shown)


dlm6693 (dismissed):
Thanks for your submission @msw-kialo! This looks good to me, but we need more time to review this internally.

msw-kialo:
The fix was in the mean time duplicated in https://github.com/boto/botocore/pull/3494 and there was actually merged (and not only tested). So I will just close this.

PR #2727, opened 2022-07-28
Fix Trace ID environment variable

kyleknap:
There was a typo in expected env var. It should be `_X_AMZN_TRACE_ID` and not `_X_AMZ_TRACE_ID`.

PR #2733, opened 2022-08-04
Fix encoding scheme for trace ids

nateprewitt:
This is a follow up to #2727 to fix the encoding scheme for our trace IDs. The change omits characters that cannot be encoded to preserve content semantics with services parsing the header.


kyleknap (approved):
Looks good :ship:

PR #2735, opened 2022-08-08
Replace default shape enum behavior from random to first element

hssyoo:
Related issue: https://github.com/aws/aws-cli/issues/6695

Currently, if a `shape` object has the `enum` attribute, then it'll return a randomly sampled element from the list. This is confusing users who expect consistent outputs from their inputs. This PR takes the first element of the list instead in order to provide consistent reproducibility.


hssyoo:
Closing this after discussing with team and deciding to implement this change in CLI v2 branch.

PR #2738, opened 2022-08-16
Deprecate ssl common name

dlm6693:
Now that https://github.com/aws/aws-cli/pull/7114 is close to full approval, this should be re-opened and merged to deprecate the SSL common name paradigm in botocore. This is a copy of https://github.com/boto/botocore/pull/2704/files which had to be rolled back because of unforeseen issues it caused in aws-cli v1.


botocore/client.py
```diff
@@ -606,8 +608,21 @@ def _create_endpoint(
...
+            if not is_disabled and sslCommonName is not None:
+                warnings.warn(
+                    f'The {service_name} client is currently using a '
+                    f'deprecated endpoint: {sslCommonName}. In the next '
+                    f'minor version this will be moved to {hostname}. '
+                    'See https://github.com/boto/botocore/issues/2705 '
+                    'for more details.',
+                    category=FutureWarning,
+                )
+                hostname = sslCommonName
```
nateprewitt, line 625: Did we confirm this warning does not trigger for the CLI?

dlm6693: so it should never be triggered even if the environment variable isn't set? that PR doesn't suppress any warnings so I guess I'm wondering how is this deprecation supposed to work at all in the CLI. It will still hit this code path, but will silently change it back to the common name with the mapping. let me double check again

dlm6693: looks like we're good since `endpoint_url` will never be `None` in the CLI v1 since we set it in on `parsed_globals`

dlm6693: feel free to test yourself as well, but i didn't get any warnings

nateprewitt (approved):
Looks good, thanks!

PR #2739, opened 2022-08-17
Add awsQueryCompatible trait support to service

eddy-aws:
When awsQueryCompatible trait is found, it's made available in the service model. When this property is found,
the error code is returned by looking up the mapping. If the error_code is not found in the mapping, original error is returned.This is a pre-requisite for migrating services from
AWSQuery wire protocol to AWSJson.

A test service has been updated and tested


nateprewitt:
Resolving this in favor of the merged #2768.

PR #2740, opened 2022-08-18
Support for context params in service models

jonemo:
Service model files (`/botocore/data/*/*/service-2.json`) will soon contain information about three types of "context params": 

1. static context parameters, optional for each operation
2. (dynamic) context parameters, optional for each operation
3. client context parameters, optional for each service

This PR adds properties for accessing context parameters information in the `ServiceModel` class.

### DIFF
--- a/botocore/model.py
+++ b/botocore/model.py
@@ -454,7 +454,7 @@ class ServiceModel:
             )
 
     # Signature version is one of the rare properties
-    # than can be modified so a CachedProperty is not used here.
+    # that can be modified so a CachedProperty is not used here.
 
     @property
     def signature_version(self):
--- a/botocore/model.py
+++ b/botocore/model.py
@@ -85,6 +85,7 @@ class Shape:
         'document',
         'union',
         'contextParam',
+        'clientContextParams',
     ]
     MAP_TYPE = OrderedDict
 
@@ -275,6 +276,10 @@ StaticContextParameter = namedtuple(
 
 ContextParameter = namedtuple('ContextParameter', ['name', 'member_name'])
 
+ClientContextParameter = namedtuple(
+    'ClientContextParameter', ['name', 'type', 'documentation']
+)
+
 
 class ServiceModel:
     """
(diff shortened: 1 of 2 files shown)


botocore/model.py
```diff
@@ -267,6 +269,18 @@ def enum(self):
...
+)
+
+
+ContextParameter = namedtuple('ContextParameter', ['name', 'member_name'])
+
+ClientContextParameter = namedtuple(
+    'ClientContextParameter', ['name', 'type', 'documentation']
+)
+
+
```
dlm6693, line 283: i personally prefer the syntax of `typing.NamedTuple` and I think @nateprewitt just merged a minor PR that used it to replace a data class so maybe lets switch for consistency.

tests/unit/test_model.py
```diff
@@ -428,6 +496,37 @@ def test_http_checksum_present(self):
...
+        self.assertEqual(static_ctx_param1.name, 'stringStaticContextParam')
+        self.assertEqual(static_ctx_param1.value, 'Static Context Param Value')
+        self.assertEqual(static_ctx_param2.name, 'booleanStaticContextParam')
+        self.assertEqual(static_ctx_param2.value, True)
+
+    def test_static_context_parameter_abent(self):
+        service_model = model.ServiceModel(self.model)
+        operation = service_model.operation_model('OperationTwo')
+        self.assertIsInstance(operation.static_context_parameters, list)
+        self.assertEqual(len(operation.static_context_parameters), 0)
```
dlm6693, line 528: i think we should add some functional tests to make sure we can create a client with changes to the structure of service models

jonemo: I would argue that the test coverage you ask for already exists:

1. Any place that loads a service model should do so using `model.ServiceModel`. The unit tests in this PR already confirm that this class can handle service models with the new fields. (Note that there is no change in structure.)
2. Many functional test and all integration tests already require successful loading of a service model. I have already been running these tests locally with service model files that contain the new fields.

nateprewitt (approved):
Looks good, thanks!

PR #2741, opened 2022-08-18
Generalize set_operation_specific_signer to recognize "s3v4a" case

jonemo:
Refactors the `set_operation_specific_signer` event handler to work independently of other S3-specific event handlers. This is in preparation for future changes to S3-specific logic. 

After this change, when the function is called with `signing_name == 's3'` and `auth_type='v4a'`, it will now return `s3v4a` instead of previously `v4a`. This does not change any current behavior because the function is never called with this combination of values (see below for why).

Context: The function `set_operation_specific_signer` usually gets called when a `choose-signer.*` event is emitted. However, it currently does _not_ get called for the `choose-signer.s3.GetObjectAttributes` event when a global access point is used because the handler `S3EndpointSetter.set_signer` [is subscribed to the same event](https://github.com/boto/botocore/blob/2a672d6753af1c59b83f2d783a28c4c59b07dfd1/botocore/utils.py#L1802) and [returns a value first](https://github.com/boto/botocore/blob/f4ed130b78076fb5683e4384c7df007e82dda71d/botocore/utils.py#L1978). This is also the reason why the `if is_global_accesspoint` branch can be removed from the function.

Separately, a new unit test is added to add test coverage for the scenario where the `context['signing']` is already set.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -59,7 +59,6 @@ from botocore.signers import (
 from botocore.utils import (
     SAFE_CHARS,
     conditionally_calculate_md5,
-    is_global_accesspoint,
     percent_encode,
     switch_host_with_param,
 )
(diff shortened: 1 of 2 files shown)


nateprewitt (approved):
🚢 Cool, I think this looks like a good optimization!

PR #2746, opened 2022-08-23
Add readinto to StreamingBody

aJanuary:
This adds a readinto implementation to StreamingBody, following the
contract outlined in the Python io package.

The primary motivation for this is that it allows consumers to wrap the
StreamingBody in an io.BufferedReader. This provides a nice performance
boost for parsing streaming binary data formats, where you want to be
able to read small numbers of bytes in your parsing algorithm, but
benefit from reading large chunks of data from upstream.

### DIFF
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -110,6 +110,22 @@ class StreamingBody(IOBase):
             self._verify_content_length()
         return chunk
 
+    def readinto(self, b):
+        """Read bytes into a pre-allocated, writable bytes-like object b, and return the number of bytes read."""
+        try:
+            amount_read = self._raw_stream.readinto(b)
+        except URLLib3ReadTimeoutError as e:
+            # TODO: the url will be None as urllib3 isn't setting it yet
+            raise ReadTimeoutError(endpoint_url=e.url, error=e)
+        except URLLib3ProtocolError as e:
+            raise ResponseStreamingError(error=e)
+        self._amount_read += amount_read
+        if amount_read == 0 and len(b) > 0:
+            # If the server sends empty contents then we know we need to verify
+            # the content length.
+            self._verify_content_length()
+        return amount_read
+
     def readlines(self):
         return self._raw_stream.readlines()
 
--- a/botocore/httpchecksum.py
+++ b/botocore/httpchecksum.py
@@ -243,6 +243,17 @@ class StreamingChecksumBody(StreamingBody):
             self._validate_checksum()
         return chunk
 
+    def readinto(self, b):
+        amount_read = super().readinto(b)
+        if amount_read == len(b):
+            view = b
+        else:
+            view = memoryview(b)[:amount_read]
+        self._checksum.update(view)
+        if amount_read == 0 and len(b) > 0:
+            self._validate_checksum()
+        return amount_read
+
     def _validate_checksum(self):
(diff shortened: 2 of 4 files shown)


ChrisFloofyKitsune:
Bumping this ancient PR because holy crap I am running into this issue in my day job.

I am using `StreamingBody` with `io.BufferedReader` and I am resorting to monkey patching the `readinto()` method onto the `StreamingBody` object itself before wrapping it with the `BufferedReader`.

khushail:
@aJanuary, this PR has gotten pretty old and needs a rebase. It would be helpful if you could rebase it and we could involve the team for further review. Thanks.

arandito (commented):
Thanks for this PR @aJanuary! The implementation looks to be mostly functionally correct. I've left some comments for some small changes. Here is some general feedback:
- Please rebase your branch onto `develop`
- Run `pre-commit` before making a commit to fix linting errors that are causing our linting workflows to fail

botocore/response.py
```diff
@@ -110,6 +110,22 @@ def read(self, amt=None):
...
+        """Read bytes into a pre-allocated, writable bytes-like object b, and return the number of bytes read."""
+        try:
+            amount_read = self._raw_stream.readinto(b)
+        except URLLib3ReadTimeoutError as e:
+            # TODO: the url will be None as urllib3 isn't setting it yet
+            raise ReadTimeoutError(endpoint_url=e.url, error=e)
+        except URLLib3ProtocolError as e:
+            raise ResponseStreamingError(error=e)
+        self._amount_read += amount_read
+        if amount_read == 0 and len(b) > 0:
```
aJanuary, line 123: I'm not hugely familiar with [byte-like objects](https://docs.python.org/3/glossary.html#term-bytes-like-object) that `readinto` takes. Does anyone know if we can use `len` on them like this?

ChrisFloofyKitsune: you can use `len` on them just fine.

it's required for anything that's `byte-like` to have a length

https://docs.python.org/3/glossary.html#term-bytes-like-object
https://docs.python.org/3/c-api/buffer.html#c.Py_buffer.len

botocore/response.py
```diff
@@ -110,6 +110,22 @@ def read(self, amt=None):
             self._verify_content_length()
         return chunk
 
+    def readinto(self, b):
```
SamRemis, line 113: Hi @aJanuary,

Sorry for doing this over two rounds of review a few weeks apart; this PR addressed all of the subclasses when you first made it, but a few months ago we added a rolling checksum to s3 streams in the `StreamingChecksumBody` class.  You can see the code for the read method [here](https://github.com/boto/botocore/blob/df53fce76a2e7b6e4fbc5d89024880761352d6fd/botocore/httpchecksum.py#L239-L244).

If we were to merge this today, anyone using `readinto` would not be using checksums by default.  Would you be open to adding an implementation and some tests in the subclass as well?  Or potentially taking a PR to your fork that will add support for this?

SamRemis: Looks good to me, just finishing up some manual testing now.  Unless I find any issues with this, we should be able to get it out in a release early next week.  

Thank you for your responsiveness here and your contribution

SamRemis (approved):

PR #2747, opened 2022-08-29
Added tests for new endpoints 2.0 files

jonathan343:
This PR includes new functional tests that validate the presence and content of new endpoint files such as `endpoint-rule-set.json`, `endpoint-tests.json`, and `partitions.json`.


tests/functional/test_model_completeness.py
```diff
@@ -46,3 +48,48 @@ def test_paginators_and_waiters_are_not_lost_in_new_version(
...
+
+
+@pytest.mark.parametrize(
+    "service_name, version",
+    _endpoint_rule_set_cases(),
+)
+def test_all_endpoint_rule_sets_exist(service_name, version):
+    """Tests the existence of endpoint-rule-set.json for each service
+    and verifies that content is present."""
+    loader = Loader()
```
jonemo, line 67: Nit: Since this test is effectively running in a loop, this creates a lot of `Loader` objects even though only a single one is needed. You can avoid that by just having a global (to this file) loader. We had a similar discussion in https://github.com/boto/botocore/pull/2737#discussion_r951764838. You won't get any caching benefits from consolidating into a single loader, but you'll save a few lines of code and a tiny amount of memory :)

tests/functional/test_model_completeness.py
```diff
@@ -46,3 +48,48 @@ def test_paginators_and_waiters_are_not_lost_in_new_version(
...
+    type_name = 'endpoint-rule-set'
+    loader.load_service_model(service_name, type_name, version)
+    full_path = os.path.join(service_name, version, type_name)
+    data = loader.load_data(full_path)
```
jonemo, line 71: I'm not sure I understand what is happening here. Isn't this basically doing the same thing twice and could instead be shrunk to 

```suggestion
    data = loader.load_service_model(service_name, type_name, version)
```

tests/functional/test_model_completeness.py
```diff
@@ -46,3 +48,48 @@ def test_paginators_and_waiters_are_not_lost_in_new_version(
...
+    assert len(data['rules']) >= 1
+
+
+test_data_dir = os.path.join(os.path.dirname(__file__), "endpoint-rules")
+
+
+def test_all_endpoint_tests_exist():
+    """Tests the existence of endpoint-tests.json for each service
+    and verifies that content is present."""
+    for service_name in Session().get_available_services():
```
jonemo, line 81: Could also consolidate this with the other copy of it, either in a module-scoped fixture or by making it a constant at the top of the file. This hits the file system, so minimizing how often it gets run speed things up.

jonemo (commented):
This is great. I've been burned by tests getting skipped silently due to missing test data files and this will help with that, so 👍 from me!

Do we already have a test elsewhere that asserts that `Session().get_available_services()` returns more than zero services? If not, that may be worth adding. Otherwise a completely empty data directory would result in these test passing.

I left a few inline comments inline regarding consolidating repeated code.

tests/functional/test_model_completeness.py
```diff
@@ -46,3 +48,48 @@ def test_paginators_and_waiters_are_not_lost_in_new_version(
...
+        with open(endpoint_tests_file) as f:
+            data = json.load(f)
+            assert len(data['testCases']) >= 1
+
+
+def test_partitions_exists():
+    """Tests the existence of partitions.json and verifies that content is present."""
+    loader = Loader()
+    data = loader.load_data('partitions')
+    assert len(data['partitions']) >= 1
```
nateprewitt, line 95: I think we know there will be at least 4 partitions. Let's set our lower bound for now to avoid regressing. `[aws, aws-cn, aws-iso, aws-iso-b]`

nateprewitt (changes requested):

tests/functional/test_model_completeness.py
```diff
@@ -10,16 +10,20 @@
...
+import json
+import os
 import pytest
 
 from botocore.exceptions import DataNotFoundError
 from botocore.loaders import Loader
-from botocore.session import Session
+
+loader = Loader()
+available_services = loader.list_available_services(type_name='service-2')
```
nateprewitt, line 21: Small nit; if variables are global constants you usually want to do the names in all caps (e.g. `LOADER`, `AVAILABLE_SERVICES`). Otherwise, you might look at using the [pytest fixture](https://docs.pytest.org/en/6.2.x/fixture.html) to adjust scoping if needed.

That would allow you to define something like:

```python
@pytest.fixture(scope="module")
def loader():
    return Loader()

@pytest.fixture(scope="module")
def available_services(loader):
    return loader.list_available_services(type_name='service-2')

[...]

 def test_all_endpoint_rule_sets_exist(loader, service_name, version):
     [...]
```

tests/functional/test_model_completeness.py
```diff
@@ -46,3 +49,47 @@ def test_paginators_and_waiters_are_not_lost_in_new_version(
...
+    """Tests the existence of endpoint-rule-set.json for each service
+    and verifies that content is present."""
+    type_name = 'endpoint-rule-set'
+    data = loader.load_service_model(service_name, type_name, version)
+    assert len(data['rules']) >= 1
+
+
+def _endpoint_tests_cases():
+    for service_name in available_services:
+        yield service_name
```
nateprewitt, line 75: This looks redundant, can we just use `AVAILABLE_SERVICES` as input to parametrize?

nateprewitt (changes requested):
Great, looking good! A couple more minor comments and I think we're close to done.

tests/functional/test_model_completeness.py
```diff
@@ -34,15 +38,53 @@ def test_paginators_and_waiters_are_not_lost_in_new_version(
...
+
+def _endpoint_rule_set_cases():
+    for service_name in AVAILABLE_SERVICES:
+        versions = LOADER.list_api_versions(service_name, 'service-2')
+        for version in versions:
+            yield service_name, version
+
+
+@pytest.mark.parametrize(
+    "service_name, version",
```
dlm6693, line 62: very minor nit and i don't think its causing any issues, but the docs don't have a space between the comma and next variable so it should probably be `"service_name,version"`

jonathan343: I did a quick search in the repo for `@pytest.mark.parametrize` and found that all cases with multiple parameters are formatted `"param1, param2, ..."`. Should we keep this consistency or follow the docs. Im fine with either, just wanted to bring that up.

dlm6693: I'd keep it as is then. Its more readable with the space and if its all over our repo then I'm confident it works fine.

tests/functional/test_model_completeness.py
```diff
@@ -34,15 +38,53 @@ def test_paginators_and_waiters_are_not_lost_in_new_version(
...
+)
+def test_all_endpoint_rule_sets_exist(service_name, version):
+    """Tests the existence of endpoint-rule-set.json for each service
+    and verifies that content is present."""
+    type_name = 'endpoint-rule-set'
+    data = LOADER.load_service_model(service_name, type_name, version)
+    assert len(data['rules']) >= 1
+
+
+test_data_dir = os.path.join(os.path.dirname(__file__), "endpoint-rules")
```
dlm6693, line 73: like the other module level constants this should be placed at the top of the file and written in all caps

tests/functional/test_model_completeness.py
```diff
@@ -34,15 +38,53 @@ def test_paginators_and_waiters_are_not_lost_in_new_version(
...
+    endpoint_tests_file = os.path.join(test_data_dir, service_name, file_name)
+    with open(endpoint_tests_file) as f:
+        data = json.load(f)
+        assert len(data['testCases']) >= 1
+
+
+def test_partitions_exists():
+    """Tests the existence of partitions.json and verifies that content is present."""
+    data = LOADER.load_data('partitions')
+    assert len(data['partitions']) >= 4
```
dlm6693, line 90: do we need to explicitly check the length or can we just check for the presence of the `partitions` key?

jonathan343: The length is important here since we know there will exist at least 4 partitions as Nate mentioned in [this](https://github.com/boto/botocore/pull/2747#discussion_r958674553) comment.

tests/functional/test_model_completeness.py
jonathan343: thanks! Will update👍

tests/functional/test_model_completeness.py
```diff
@@ -34,15 +38,53 @@ def test_paginators_and_waiters_are_not_lost_in_new_version(
...
+
+@pytest.mark.parametrize("service_name", AVAILABLE_SERVICES)
+def test_all_endpoint_tests_exist(service_name):
+    """Tests the existence of endpoint-tests.json for each service
+    and verifies that content is present."""
+    file_name = 'endpoint-tests.json'
+    endpoint_tests_file = os.path.join(test_data_dir, service_name, file_name)
+    with open(endpoint_tests_file) as f:
+        data = json.load(f)
+        assert len(data['testCases']) >= 1
```
dlm6693, line 84: not sure it matters, but may be better to run the assertion outside the context manager, so just unindent that line

tests/functional/test_model_completeness.py
dlm6693: got it thanks!

tests/functional/test_model_completeness.py
```diff
@@ -34,15 +39,48 @@ def test_paginators_and_waiters_are_not_lost_in_new_version(
...
+    assert len(data['rules']) >= 1
+
+@pytest.mark.parametrize("service_name", AVAILABLE_SERVICES)
+def test_all_endpoint_tests_exist(service_name):
+    """Tests the existence of endpoint-tests-1.json for each service
+    and verifies that content is present."""
+    endpoint_tests_file = os.path.join(TEST_DATA_DIR, service_name, 'endpoint-tests-1.json')
+    with open(endpoint_tests_file) as f:
+        data = json.load(f)
+    assert len(data['testCases']) >= 1
```
jonemo, line 80: This is now also covered in my branch (currently in WIP PR https://github.com/jonemo/botocore/pull/1) together with all other tests that need to access the `tests/functional/endpoint-rules` folder.

nateprewitt:
Resolving, this ended up addressed in a separate PR.

PR #2748, opened 2022-08-29
Convert dataclass to NamedTuple for backwards compatibility

nateprewitt:
Moving our new FrozenAuthToken class to a NamedTuple to provide temporary compatibility to some of our downstream users. This won't have any material change for users as the FrozenAuthToken is a private interface that's not intended for external consumption.


kyleknap (approved):
:ship:

PR #2750, opened 2022-08-31
Log InstanceMetadataProvider message at info level

pkuca:
Having used other providers in the past, I was looking for a "Credential found" log message from boto3 when verifying usage of the InstanceMetadataProvider.

I lost a little bit of time today wondering why I couldn't see the same verification while testing a boto3 client using the InstanceMetadataProvider, then was surprised to find that _just this_ provider logs the Credential confirmation at debug level.

This PR logs the "Credential found" message for the InstanceMetadataProvider at info to match the other providers with the same such message.


nateprewitt (approved):
This seems reasonable, thanks @pkuca!

PR #2751, opened 2022-09-01
Loader utilities for determining source of loaded data

jonemo:
Adds utilities to the `Loader` class to determine whether a file originates from within botocore's builtin data directory or not:
 * `load_data()` now accepts an optional argument `return_path`. If set to true, the path where the data is loaded from is returned as second return value
 * a new `is_builtin_path()` method that checks whether a given path is within the builtin data directory or not

Adds a new property `uses_builtin_data` to the `EndpointResolver` class. This property is set during component initialization using the helper methods from the `Loader` described above.

### DIFF
--- a/botocore/loaders.py
+++ b/botocore/loaders.py
@@ -488,9 +488,9 @@ class Loader:
                         yield full_path
 
     def is_builtin_path(self, path):
-        """Whether a given path is within the package's data directory
+        """Whether a given path is within the package's data directory.
 
-        This method can be used together with load_data(name, include_path=True)
+        This method can be used together with load_data_with_path(name)
         to determine if data has been loaded from a file bundled with the
         package, as opposed to a file in a separate location.
 
@@ -498,7 +498,6 @@ class Loader:
         :param path: The file path to check.
 
         :return: Whether the given path is within the package's data directory.
-
         """
         # normalize the path and resolve symlinks
         path = os.path.realpath(os.path.abspath(path))
--- a/botocore/loaders.py
+++ b/botocore/loaders.py
@@ -434,7 +434,8 @@ class Loader:
                 pass
 
     @instance_cache
-    def load_data(self, name):
+    def load_data(self, name, return_path=False):
+
         """Load data given a data path.
 
         This is a low level method that will search through the various
@@ -446,14 +447,23 @@ class Loader:
         :type name: str
         :param name: The data path, i.e ``ec2/2015-03-01/service-2``.
 
-        :return: The loaded data.  If no data could be found then
-            a DataNotFoundError is raised.
+        :type return_path: bool
+        :param return_path: Whether to return the path where the data was
+            loaded from as a second return value.
+
+        :return: The loaded data. If return_path is True, the path to the
+            data file is returned as string in the second return value. If
+            no data could be found then a DataNotFoundError is raised.
 
         """
         for possible_path in self._potential_locations(name):
             found = self.file_loader.load_file(possible_path)
             if found is not None:
-                return found
+                if return_path:
+                    return found, possible_path
+                else:
+                    return found
+
         # We didn't find anything that matched on any path.
         raise DataNotFoundError(data_path=name)
 
@@ -472,6 +482,23 @@ class Loader:
                         yield full_path
                     elif os.path.exists(full_path):
                         yield full_path
+
+    def is_builtin_path(self, path):
+        """Whether a given path is within the package's data directory
+
+        This method can be used together with load_data(name, include_path=True)
+        to determine if data has been loaded from a file bundled with the
+        package, as opposed to a file in a separate location.
+
+        :type path: str
+        :param path: The file path to check.
+
+        :return: Whether the given path is within the package's data directory.
+
+        """
+        # normalize and absolutize the path, self.BUILTIN_DATA_PATH already is
+        path = os.path.realpath(path)
+        return path.startswith(self.BUILTIN_DATA_PATH)
 
 
 class ExtrasProcessor:
--- a/botocore/loaders.py
+++ b/botocore/loaders.py
@@ -434,38 +434,42 @@ class Loader:
                 pass
 
     @instance_cache
-    def load_data(self, name, return_path=False):
+    def load_data_with_path(self, name):
+        """Same as ``load_data`` but returns file path as second return value.
 
+        :type name: str
+        :param name: The data path, i.e ``ec2/2015-03-01/service-2``.
+
+        :return: Tuple of the loaded data and the path to the data file
+            where the data was loaded from. If no data could be found then a
+            DataNotFoundError is raised.
+        """
+        for possible_path in self._potential_locations(name):
+            found = self.file_loader.load_file(possible_path)
+            if found is not None:
+                return found, possible_path
+
+        # We didn't find anything that matched on any path.
+        raise DataNotFoundError(data_path=name)
+
+    def load_data(self, name):
         """Load data given a data path.
 
         This is a low level method that will search through the various
         search paths until it's able to load a value.  This is typically
         only needed to load *non* model files (such as _endpoints and
         _retry).  If you need to load model files, you should prefer
-        ``load_service_model``.
+        ``load_service_model``.  Use ``load_data_with_path`` to get the
+        file system path of the data file as second return value.
 
         :type name: str
         :param name: The data path, i.e ``ec2/2015-03-01/service-2``.
 
-        :type return_path: bool
-        :param return_path: Whether to return the path where the data was
-            loaded from as a second return value.
-
-        :return: The loaded data. If return_path is True, the path to the
-            data file is returned as string in the second return value. If
-            no data could be found then a DataNotFoundError is raised.
-
+        :return: The loaded data. If no data could be found then
+            a DataNotFoundError is raised.
         """
-        for possible_path in self._potential_locations(name):
-            found = self.file_loader.load_file(possible_path)
-            if found is not None:
-                if return_path:
-                    return found, possible_path
-                else:
-                    return found
-
-        # We didn't find anything that matched on any path.
-        raise DataNotFoundError(data_path=name)
+        data, _ = self.load_data_with_path(name)
+        return data
 
     def _potential_locations(self, name=None, must_exist=False, is_dir=False):
         # Will give an iterator over the full path of potential locations
--- a/botocore/loaders.py
+++ b/botocore/loaders.py
@@ -496,8 +496,8 @@ class Loader:
         :return: Whether the given path is within the package's data directory.
 
         """
-        # normalize and absolutize the path, self.BUILTIN_DATA_PATH already is
-        path = os.path.realpath(path)
+        # normalize the path and resolve symlinks
+        path = os.path.realpath(os.path.abspath(path))
         return path.startswith(self.BUILTIN_DATA_PATH)
 
 
--- a/botocore/regions.py
+++ b/botocore/regions.py
@@ -98,7 +98,7 @@ class EndpointResolver(BaseEndpointResolver):
 
     def __init__(self, endpoint_data, uses_builtin_data=False):
         """
-        :type path: dict
+        :type endpoint_data: dict
         :param endpoint_data: A dict of partition data.
 
         :type uses_builtin_data: boolean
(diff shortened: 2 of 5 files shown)


tests/unit/test_loaders.py
```diff
@@ -128,6 +129,23 @@ def load_file(self, name):
...
+                expected_ending = os.path.join('bar', 'abc')
+                if name.endswith(expected_ending):
+                    return ['loaded data']
+
+        loader = Loader(
+            extra_search_paths=search_paths, file_loader=FakeLoader()
+        )
+        loaded, path = loader.load_data('abc', return_path=True)
+        self.assertEqual(loaded, ['loaded data'])
+        self.assertEqual(path, 'bar/abc')
```
dlm6693, line 147: you'll need to update this to `os.path.join('bar', 'abc')` or `f'bar{os.sep}abc'` to pass windows tests

botocore/loaders.py
```diff
@@ -473,6 +483,23 @@ def _potential_locations(self, name=None, must_exist=False, is_dir=False):
...
+        package, as opposed to a file in a separate location.
+
+        :type path: str
+        :param path: The file path to check.
+
+        :return: Whether the given path is within the package's data directory.
+
+        """
+        # normalize and absolutize the path, self.BUILTIN_DATA_PATH already is
+        path = os.path.realpath(path)
```
dlm6693, line 500: should we be using `abspath` instead here?

botocore/regions.py
```diff
@@ -96,13 +96,19 @@ class EndpointResolver(BaseEndpointResolver):
 
     _UNSUPPORTED_DUALSTACK_PARTITIONS = ['aws-iso', 'aws-iso-b']
 
-    def __init__(self, endpoint_data):
+    def __init__(self, endpoint_data, uses_builtin_data=False):
         """
+        :type path: dict
```
dlm6693, line 101: should be `endpoint_data` here

botocore/regions.py
```diff
@@ -96,13 +96,19 @@ class EndpointResolver(BaseEndpointResolver):
...
         :param endpoint_data: A dict of partition data.
+
+        :type uses_builtin_data: boolean
+        :param uses_builtin_data: Whether the endpoint data originates in the
+            package's data directory.
         """
         if 'partitions' not in endpoint_data:
             raise ValueError('Missing "partitions" in endpoint data')
```
dlm6693, line 109: maybe add `key` just to clarify that the its a dictionary key not just some string?

botocore/session.py
```diff
@@ -196,8 +196,9 @@ def _register_data_loader(self):
     def _register_endpoint_resolver(self):
         def create_default_resolver():
             loader = self.get_component('data_loader')
-            endpoints = loader.load_data('endpoints')
-            return EndpointResolver(endpoints)
+            endpoints, path = loader.load_data('endpoints', return_path=True)
+            builtin = loader.is_builtin_path(path)
```
dlm6693, line 200: nit: should be `uses_builtin` or `is_builtin`

tests/functional/test_regions.py
```diff
@@ -546,3 +546,17 @@ def test_unregionalized_client_with_unknown_region(self):
         self.assertTrue(
             stubber.requests[0].url.startswith('https://iam.amazonaws.com/')
         )
+
+
+@pytest.mark.parametrize("is_builtin", [(True,), (False,)])
```
dlm6693, line 551: pretty sure you can just pass as a list if there is just one parameter, it doesn't need to be tuples with a single value.

```suggestion
@pytest.mark.parametrize("is_builtin", [True, False])
```

tests/unit/test_loaders.py
```diff
@@ -128,6 +129,23 @@ def load_file(self, name):
         loaded = loader.load_data('baz')
         self.assertEqual(loaded, ['loaded data'])
 
+    @mock.patch('os.path.isdir', mock.Mock(return_value=True))
```
dlm6693, line 132: i don't think you need to create a Mock object here. can just supply the return value as a paremeter.

```suggestion
    @mock.patch('os.path.isdir', return_value=True)
```

botocore/loaders.py
```diff
@@ -446,14 +447,23 @@ def load_data(self, name):
...
         """
         for possible_path in self._potential_locations(name):
             found = self.file_loader.load_file(possible_path)
             if found is not None:
-                return found
+                if return_path:
+                    return found, possible_path
+                else:
+                    return found
+
```
dlm6693, line 466: i wonder if it would be better to create a new function that returns both the loaded data and the path rather than changing the data structure of a returnable given a certain condition

dlm6693 (changes requested):
Looks pretty good to me! Just have some minor suggestions and questions/comments. Let me know your thoughts.

tests/functional/test_regions.py
jonemo: Making that change makes the test fail with:

```
TypeError: TestLoader.test_load_data_optionally_returns_path() takes 1 positional argument but 2 were given
```

tests/unit/test_loaders.py
jonemo: Making that change results in this test failure:

```
TypeError: TestLoader.test_load_data_optionally_returns_path() takes 1 positional argument but 2 were given
```

tests/functional/test_regions.py
jonemo: Changed in 7aeb81341

botocore/regions.py
jonemo: Good catch. Fixed in 7aeb81341.

botocore/loaders.py
jonemo: I'm not going to remove `realpath()` because resolving symlinks is the primary thing I want from this function. But I do now realize that my inline comment sounds a lot more like `abspath()` and that won't hurt either so now it's both.

botocore/regions.py
jonemo: I see where you're coming from. But this line hasn't been changed in seven years and I see more potential problems caused by me changing it now than potential problems that would solve.

botocore/session.py
jonemo: Why? The function checks if the string in the argument is a builtin path or not: `myloader.is_builtin_path("/path/to/my/file")`. I think the current name is more descriptive and specific.

dlm6693: I disagree, just naming `builtin` implies that it is a built in and presents ambiguity. Including a verb in the name like `uses` or `is` indicates that it's a boolean. You included it in the function name after all 😉

tests/unit/test_loaders.py
dlm6693: whoops never mind then 😅

botocore/session.py
jonemo: Oh, you are talking about the variable name. I thought this is about the method name...

dlm6693: Ah sorry yeah the variable name!

botocore/loaders.py
nateprewitt: How about we make a new function that returns both indiscriminately. This function can start calling that new function and only return the data. That will give us the new API surface area, allows us to cache both at the lower function, and keep this method functionally the same.

botocore/loaders.py
```diff
@@ -434,28 +434,42 @@ def _find_extras(self, service_name, type_name, api_version):
...
     @instance_cache
+    def load_data_with_path(self, name):
+        """Same as ``load_data`` but returns file path as second return value.
+
+        :type name: str
+        :param name: The data path, i.e ``ec2/2015-03-01/service-2``.
+
+        :return: Tuple of the loaded data and the path to the data file
+            where the data was loaded from. If no data could be found then a
+            DataNotFoundError is raised.
```
dlm6693, line 445: i don't think the second sentence is necessarily needed here.

tests/functional/test_regions.py
```diff
@@ -546,3 +546,17 @@ def test_unregionalized_client_with_unknown_region(self):
...
+
+@pytest.mark.parametrize("is_builtin", [True, False])
+def test_endpoint_resolver_knows_its_datasource(is_builtin):
+    # The information whether or not the endpoints.json file was loaded from
+    # the builtin data directory or not should be passed from Loader to
+    # EndpointResolver.
+    session = _get_patched_session()
+    loader = session.get_component('data_loader')
+    with mock.patch.object(
+        loader, 'is_builtin_path', mock.Mock(return_value=is_builtin)
```
dlm6693, line 559: i don't thin you need to make a mock object and can just pass `return_value` as an arg to the mock patch context manager

tests/unit/test_loaders.py
```diff
@@ -210,6 +240,31 @@ def test_create_loader_parses_data_path(self):
...
+        path_elsewhere = __file__
+        self.assertTrue(loader.is_builtin_path(path_in_builtins))
+        self.assertFalse(loader.is_builtin_path(path_elsewhere))
+
+    @pytest.mark.skipif(
+        sys.platform == 'win32',
+        reason=(
+            'os.symlink() requires developer mode or elevated permissions'
+        ),
+    )
```
dlm6693, line 255: i think you said this was just effecting python 3.7 right? if so we should just skip tests for that version

dlm6693 (approved):
Looks good to me! A couple of comments but none of them are blockers. :shipit:

botocore/loaders.py
```diff
@@ -473,6 +487,23 @@ def _potential_locations(self, name=None, must_exist=False, is_dir=False):
...
+
+        This method can be used together with load_data(name, include_path=True)
+        to determine if data has been loaded from a file bundled with the
+        package, as opposed to a file in a separate location.
+
+        :type path: str
+        :param path: The file path to check.
+
+        :return: Whether the given path is within the package's data directory.
+
```
nateprewitt, line 501: nit;
```suggestion
```

botocore/loaders.py
```diff
@@ -473,6 +487,23 @@ def _potential_locations(self, name=None, must_exist=False, is_dir=False):
                     elif os.path.exists(full_path):
                         yield full_path
 
+    def is_builtin_path(self, path):
+        """Whether a given path is within the package's data directory
```
nateprewitt, line 491:
```suggestion
        """Whether a given path is within the package's data directory.
```

botocore/loaders.py
```diff
@@ -473,6 +487,23 @@ def _potential_locations(self, name=None, must_exist=False, is_dir=False):
...
+
+        This method can be used together with load_data(name, include_path=True)
+        to determine if data has been loaded from a file bundled with the
+        package, as opposed to a file in a separate location.
```
nateprewitt, line 495:
```suggestion
        This method can be used together with load_data_with_path(name)
        to determine if data has been loaded from a file bundled with the
        package, as opposed to a file in a separate location.
```

nateprewitt (commented):
Some minor tweaks to the docstrings, but otherwise I think this looks good. Thanks @jonemo!

tests/unit/test_loaders.py
jonemo: Creating symlinks on Windows requires one of the following two things:
 
* Run the script elevated as administrator. This is a temporary change, achieved for example by opening a command line window with right click -> run as administrator.
* Have developer mode enabled. This is a permanent setting admin-level users can make for their account in the Windows settings.

This applies to all languages and therefore all versions of Python. You can read more about this here: https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/

What I have previously mentioned that only affects Windows 3.7 is that `os.realpath()` does not resolve symlinks to their real path, at least not in my testing. This slightly reduces the functionality of the `is_builtin_path()` function for users of Python 3.7 on Windows. If a user on this version and platform decides to symlink the builtin data file inot their `AWS_DATA_PATH`, then it will not be recognized as builtin.

tests/functional/test_regions.py
```diff
@@ -546,3 +546,17 @@ def test_unregionalized_client_with_unknown_region(self):
...
+    ):
```
jonemo, line 560: Turning https://github.com/boto/botocore/pull/2751/files#r966037688 into suggestion for easy batching in Github UI:

```suggestion
    with mock.patch.object(loader, 'is_builtin_path', return_value=is_builtin):
```

botocore/loaders.py
jonemo: I decided to duplicate this from the `load_data` docstring for the convenience of the reader. It's not clear which one of the two is the "primary" function now given that `load_data` is the old one that is referenced everywhere else but `load_data_with_path` actually houses the functionality now. So someone looking for the docs could land in either place.

nateprewitt (approved):
⛵

PR #2754, opened 2022-09-02
Added retries to delete_objects; Removed redundant deletions

jonathan343:
Removed redundant calls to `delete_object` which should be taken care of by the clean up provided by the `teardown_module` method. Added retries for `delete_objects` to handle rare cases of failure.


dlm6693 (approved):
📦

Issue #2705, opened 2022-06-22
Deprecate Usage of `sslComonName` in Endpoint Creation


dlm6693:
## The Problem

Currently when creating a service client, an `sslCommonName` attribute may be used for endpoint construction in unique cases. The format of `sslCommonName` is typically `{region}.{service}.{dnsSuffix}`,  as opposed to the more common `{service}.{region}.{dnsSuffix}`. This usage originated from a time where Python versions (<2.7) didn't supply an SSL module, requiring specific certificate formats.

Now that the library only support Python 3.7+, we'll be deprecating the usage of `sslCommonName` to standardize Boto3 with all other AWS SDKs. This will also resolve long running issues of services such as SQS and GuardDuty being incompatible with certain VPC endpoint configurations.

## Required Actions

In the immediate term, we will start raising a deprecation warning when `sslCommonName` is used. This is to alert customers of the upcoming change and provide time to make any required changes.

For most users, this will not require any changes. The URL will automatically update when the next minor version (1.29.0) is released, and clients will continue to operate the same. For any users with strict network rules, explicitly allow listing domains, you will need to add support for `{service}.{region}.{dnsSuffix}` as demonstrated below:

**Old Format:** https://us-west-2.sqs.amazonaws.com
**New Format:** https://sqs.us-west-2.amazonaws.com

## Warning Mitigation Strategy

1. If you wish to ensure that your application does not use `sslCommonName` now or test the impending deprecation, we have created a new environment variable `BOTO_DISABLE_COMMONNAME`. Setting this to `true` will suppress the warning and convert to the new `hostname` format.
2. If you are concerned about this change causing disruptions, you can pin your version of `botocore` to `<1.29.0` until you are ready to migrate.
4. If you are only concerned about silencing the warning in your logs, use `warnings.filterwarnings` when instantiating a new service client.

```python
import warnings
warnings.filterwarnings('ignore', category=FutureWarning, module='botocore.client')
```

### Other Information

Endpoint Docs: https://docs.aws.amazon.com/general/latest/gr/rande.html
Related Issues: https://github.com/boto/botocore/issues/2376, https://github.com/boto/boto3/issues/1900, https://github.com/boto/boto3/issues/3311, https://github.com/boto/botocore/issues/2683

bblommers:
The EMR client now gives the following warning:

[...]

PR #2756, opened 2022-09-07
Don't raise warning if common name and hostname are the same

dlm6693:
There are some SSL common names that are identical to the hostname, which was raised in https://github.com/boto/botocore/issues/2705#issuecomment-1236390963. This will suppress warnings where this is true.


jonemo (approved):
LGTM

PR #2757, opened 2022-09-08
Remove symlink resolution in Loader.is_builtin_path

jonemo:
Followup to https://github.com/boto/botocore/pull/2751: This removes the call to `os.path.realpath()` and the corresponding test coverage related to symlinks.

Reasoning for this reversal at the last minute before releasing the #2751 PR:

Because the `Loader` class never sees the _filesystem_ path of the data file, only the _data_ path (which excludes the file extension), the call to `os.path.realpath()` in #2751 would have never resolved a symlink. Even if working correctly, this symlink resolution would only have taken effect for users who have a symlink in their data directory (`~/.aws/models`) pointing back to the botocore package's builtin file. After considering the benefit of covering this unlikely edge case and the effort involved in making the _filesystem_ path available to the `Loader` class, the decision is to remove symlink resolution from `is_builtin_path`.

### DIFF
--- a/botocore/loaders.py
+++ b/botocore/loaders.py
@@ -460,7 +460,7 @@ class Loader:
         only needed to load *non* model files (such as _endpoints and
         _retry).  If you need to load model files, you should prefer
         ``load_service_model``.  Use ``load_data_with_path`` to get the
-        file system path of the data file as second return value.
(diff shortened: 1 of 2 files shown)


nateprewitt (approved):
Looks good!

PR #2759, opened 2022-09-10
Endpoint resolution v2: builtin parameter values

jonemo:
This PR is part of the ongoing endpoint resolution v2 work, together with #2737, #2740, #2751, #2747. It contains the code that assigns values to a list of "builtin parameters" that may be referenced by endpoint rulesets. It's glue code between the existing `ClientCreator` and `ClientArgsCreator` and the future `EndpointProvider` (see #2737).

Q: By itself, this PR adds only unused code. Why is it a separate PR? — A: I decided to break this out of a larger PR with all the middleware for calling `EndpointProvider` to keep this a separate review/discussion. Even though the code is pretty simple, there could be a complex discussion hiding behind each parameter assignment and how it is tested. Let's get these debates out of the way first :)

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -468,7 +468,7 @@ class ClientArgsCreator:
         else:
             return val.lower() == 'true'
 
-    def _compute_endpoint_resolver_v2_builtin_defaults(
+    def compute_endpoint_resolver_builtin_defaults(
         self,
         region_name,
         service_name,
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -496,9 +496,11 @@ class ClientArgsCreator:
                 endpoint_bridge._resolve_endpoint_variant_config_var(
                     'use_fips_endpoint'
                 )
+                or False
             ),
             EPRBuiltins.AWS_USE_DUALSTACK: (
                 endpoint_bridge._resolve_use_dualstack_endpoint(service_name)
+                or False
             ),
             EPRBuiltins.AWS_STS_USE_GLOBAL_ENDPOINT: (
                 self._should_set_global_sts_endpoint(
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -26,6 +26,7 @@ import botocore.serialize
 import botocore.utils
 from botocore.config import Config
 from botocore.endpoint import EndpointCreator
+from botocore.regions import EndpointResolverBuiltins as EPRBuiltins
 from botocore.signers import RequestSigner
 
 logger = logging.getLogger(__name__)
@@ -344,8 +345,9 @@ class ClientArgsCreator:
     def _should_set_global_sts_endpoint(
         self, region_name, endpoint_url, endpoint_config
     ):
-        endpoint_variant_tags = endpoint_config['metadata'].get('tags')
-        if endpoint_url or endpoint_variant_tags:
+        if endpoint_url:
+            return False
+        if endpoint_config and endpoint_config.get('metadata', {}).get('tags'):
             return False
         return (
             self._get_sts_regional_endpoints_config() == 'legacy'
@@ -464,3 +466,63 @@ class ClientArgsCreator:
             return val
         else:
             return val.lower() == 'true'
+
+    def _compute_endpoint_resolver_v2_builtin_defaults(
+        self,
+        region_name,
+        service_name,
+        s3_config,
+        endpoint_bridge,
+        client_endpoint_url,
+        legacy_endpoint_url,
+    ):
+        # EndpointResolverv2 rulesets may accept an "SDK::Endpoint" as input.
+        # If the endpoint_url argument of create_client() is set, it always
+        # takes priority.
+        if client_endpoint_url:
+            given_endpoint = client_endpoint_url
+        # If an endpoints.json data file other than the one bundled within
+        # the botocore/data directory is used, the output of legacy
+        # endpoint resolution is provided to EndpointResolverv2.
+        elif not endpoint_bridge.endpoint_resolver.uses_builtin_data:
+            given_endpoint = legacy_endpoint_url
+        else:
+            given_endpoint = None
+
+        return {
+            EPRBuiltins.AWS_REGION: region_name,
+            EPRBuiltins.AWS_USE_FIPS: (
+                endpoint_bridge._resolve_endpoint_variant_config_var(
+                    'use_fips_endpoint'
+                )
+            ),
+            EPRBuiltins.AWS_USE_DUALSTACK: (
+                endpoint_bridge._resolve_use_dualstack_endpoint(service_name)
+            ),
+            EPRBuiltins.AWS_STS_USE_GLOBAL_ENDPOINT: (
+                self._should_set_global_sts_endpoint(
+                    region_name=region_name,
+                    endpoint_url=None,
+                    endpoint_config=None,
+                )
+            ),
+            EPRBuiltins.AWS_S3_USE_GLOBAL_ENDPOINT: (
+                self._should_force_s3_global(region_name, s3_config)
+            ),
+            EPRBuiltins.AWS_S3_ACCELERATE: s3_config.get(
+                'use_accelerate_endpoint', False
+            ),
+            EPRBuiltins.AWS_S3_FORCE_PATH_STYLE: (
+                s3_config.get('addressing_style') == 'path'
+            ),
+            EPRBuiltins.AWS_S3_USE_ARN_REGION: s3_config.get(
+                'use_arn_region', True
+            ),
+            EPRBuiltins.AWS_S3CONTROL_USE_ARN_REGION: s3_config.get(
+                'use_arn_region', False
+            ),
+            EPRBuiltins.AWS_S3_DISABLE_MRAP: s3_config.get(
+                's3_disable_multiregion_access_points', False
+            ),
+            EPRBuiltins.SDK_ENDPOINT: given_endpoint,
+        }
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -497,10 +497,12 @@ class ClientArgsCreator:
                     'use_fips_endpoint'
                 )
                 or False
+                and not given_endpoint
             ),
             EPRBuiltins.AWS_USE_DUALSTACK: (
                 endpoint_bridge._resolve_use_dualstack_endpoint(service_name)
                 or False
+                and not given_endpoint
             ),
             EPRBuiltins.AWS_STS_USE_GLOBAL_ENDPOINT: (
                 self._should_set_global_sts_endpoint(
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -345,9 +345,10 @@ class ClientArgsCreator:
     def _should_set_global_sts_endpoint(
         self, region_name, endpoint_url, endpoint_config
     ):
-        if endpoint_url:
-            return False
-        if endpoint_config and endpoint_config.get('metadata', {}).get('tags'):
+        has_variant_tags = endpoint_config and endpoint_config.get(
+            'metadata', {}
+        ).get('tags')
+        if endpoint_url or has_variant_tags:
             return False
         return (
             self._get_sts_regional_endpoints_config() == 'legacy'
--- a/botocore/regions.py
+++ b/botocore/regions.py
@@ -18,6 +18,7 @@ in a specific AWS partition.
 """
 import logging
 import re
+from enum import Enum
 
 from botocore.exceptions import (
     EndpointVariantError,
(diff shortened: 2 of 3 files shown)


botocore/regions.py
```diff
@@ -401,3 +402,34 @@ def _expand_template(
...
+    AWS_S3_ACCELERATE = "AWS::S3::Accelerate"
+    # Whether S3 Force Path Style has been enabled  (bool)
+    AWS_S3_FORCE_PATH_STYLE = "AWS::S3::ForcePathStyle"
+    # Whether to use the ARN region or raise an error when ARN and client
+    # region differ (for s3 service only)
+    AWS_S3_USE_ARN_REGION = "AWS::S3::UseArnRegion"
+    # Whether to use the ARN region or raise an error when ARN and client
+    # region differ (for s3-control service only).
+    AWS_S3CONTROL_USE_ARN_REGION = 'AWS::S3Control::UseArnRegion'
+    # Whether multi-region access points (MRAP) should be disabled.
```
nateprewitt, line 432: These are missing typing info in the comment. Given everything is either a string or bool, we could probably use inline typing for this class. We have precedent for this in the NamedTuples now.

botocore/regions.py
```diff
@@ -401,3 +402,34 @@ def _expand_template(
...
+    AWS_USE_DUALSTACK = "AWS::UseDualStack"
+    # Whether the global endpoint should be used, rather the the regional
+    # endpoint for us-east-1 (bool)
+    AWS_STS_USE_GLOBAL_ENDPOINT = "AWS::STS::UseGlobalEndpoint"
+    # Whether the global endpoint should be used, rather then the regional
+    # endpoint for us-east-1 (bool)
+    AWS_S3_USE_GLOBAL_ENDPOINT = "AWS::S3::UseGlobalEndpoint"
+    # Whether S3 Transfer Acceleration has been requested (bool)
+    AWS_S3_ACCELERATE = "AWS::S3::Accelerate"
+    # Whether S3 Force Path Style has been enabled  (bool)
```
nateprewitt, line 424:
```suggestion
    # Whether S3 Force Path Style has been enabled (bool)
```

botocore/regions.py
```diff
@@ -401,3 +402,34 @@ def _expand_template(
...
+    AWS_USE_FIPS = "AWS::UseFIPS"
+    # Whether the UseDualStackEndpoint configuration option has been enabled for
+    # the SDK client (bool)
+    AWS_USE_DUALSTACK = "AWS::UseDualStack"
+    # Whether the global endpoint should be used, rather the the regional
+    # endpoint for us-east-1 (bool)
+    AWS_STS_USE_GLOBAL_ENDPOINT = "AWS::STS::UseGlobalEndpoint"
+    # Whether the global endpoint should be used, rather then the regional
+    # endpoint for us-east-1 (bool)
+    AWS_S3_USE_GLOBAL_ENDPOINT = "AWS::S3::UseGlobalEndpoint"
```
nateprewitt, line 421: Maybe add sts/s3 to respective comments to differentiate.

botocore/args.py
```diff
@@ -464,3 +466,63 @@ def _ensure_boolean(self, val):
             return val
         else:
             return val.lower() == 'true'
+
+    def _compute_endpoint_resolver_v2_builtin_defaults(
```
nateprewitt, line 470: This is something I was going to adjust in David's PR, but I'm thinking we probably want to remove the "v2" designation from external functions. That will allow the interfaces to stay in place with future evolutions rather than requiring a v3 if there's ever a next iteration.

Maybe something like `_compute_endpoint_resolver_builtins`? Are they necessarily defaults, or the actual resolved values? I'm not sure if these are used elsewhere in future PRs.

tests/unit/test_args.py
```diff
@@ -459,3 +459,226 @@ def test_client_config_beats_config_store(self):
             client_config=Config(retries={'mode': 'standard'})
         )['client_config']
         self.assertEqual(config.retries['mode'], 'standard')
+
+    def test_does_set_global_sts_endpoint_if_in_legacy_region(self):
+        should = self.args_create._should_set_global_sts_endpoint(
+            region_name='us-west-1', endpoint_url=None, endpoint_config=None
+        )
+        self.assertTrue(should)
```
nateprewitt, line 467: Typically we won't test private methods directly, instead verifying the desired behavior at whatever the top level interface is. That allows us to make more substantial signature changes without having to update existing tests.

How are you intending to retrieve this from the args module? We may need to expose a "public" method even if it's not intended for use, similar to `compute_s3_config`. We can note support expectations in the docstring.

tests/unit/test_args.py
```diff
@@ -459,3 +459,226 @@ def test_client_config_beats_config_store(self):
...
+
+        def mock_cfgvar_fn(cfgvar):
+            return False if cfgvar == 'use_fips_endpoint' else 42
+
+        self.bridge._resolve_endpoint_variant_config_var = mock.Mock(
+            side_effect=mock_cfgvar_fn
+        )
+        self.bridge._resolve_use_dualstack_endpoint = mock.Mock(
+            return_value=False
+        )
```
nateprewitt, line 512: Maybe a quick comment on why we're doing this?

tests/unit/test_args.py
```diff
@@ -459,3 +459,226 @@ def test_client_config_beats_config_store(self):
...
+    def test_builtin_disable_mrap(self):
+        # This builtin has a default value of False. It will be True only if
+        # the ``s3_disable_multiregion_access_points`` field in the S3 config
+        # is True.
+        bins = self.call_compute_endpoint_resolver_v2_builtin_defaults(
+            s3_config={'s3_disable_multiregion_access_points': True},
+        )
+        self.assertEqual(bins['AWS::S3::DisableMultiRegionAccessPoints'], True)
+
+    def test_builtin_sdk_endpoint_1(self):
```
nateprewitt, line 652: Instead of using 1,2,3 can we make these test names a bit clearer?

e.g.
```
test_sdk_endpoint_both_inputs_set
test_sdk_endpoint_legacy_set_with_builtin_data
test_sdk_endpoint_legacy_set_without_builtin_data
```

nateprewitt: I'm also just now realizing that builtin is overloaded here. We might want to change the method name for detecting default endpoints.json data. `uses_default_endpoint_data`?

tests/unit/test_args.py
```diff
@@ -459,3 +459,226 @@ def test_client_config_beats_config_store(self):
...
+        )
+        bins = self.call_compute_endpoint_resolver_v2_builtin_defaults(
+            region_name='my-region-2',
+        )
+        self.assertEqual(bins['AWS::STS::UseGlobalEndpoint'], False)
+        self.args_create._should_set_global_sts_endpoint.assert_called_once_with(
+            region_name='my-region-2',
+            endpoint_url=None,
+            endpoint_config=None,
+        )
```
nateprewitt, line 597: This feels brittle if we add another arg to the method. That's probably solved if we focus on the output for the builtin.

nateprewitt (changes requested):
This is looking great. Left some feedback on testing and a couple questions on how we're setting defaults. Let me know if I can clarify anything.

botocore/args.py
jonemo: > Are they necessarily defaults, or the actual resolved values?

I added the `_defaults` suffix at the last minute here. In most cases, these are the actual values. The exception is the `AWS::S3::ForcePathStyle` builtin. To maintain full backwards compatibility, this will sometimes need to be changed from the default value based on operation-specific operation. 

Here is a specific example:

```python
import botocore.session

client = session.create_client(
    's3',
    config=Config(s3={addressing_style='path'}),
)

client.get_object_attributes(
    Bucket="arn:aws:s3:us-west-2:1234567890:accesspoint/my-bucket-access-point",
    Key="object.txt",
    ObjectAttributes=["ETag", "Checksum"],
)
```

If we set `AWS::S3::ForcePathStyle = True` purely based on the value of `addressing_style` in the config, then the ruleset will return this error:

```
Path-style addressing cannot be used with ARN buckets
```

Botocore's current behavior is to silently ignore the `addressing_style` config when an ARN is used. To maintain this behavior, we need to change the `AWS::S3::ForcePathStyle` to `False` when the bucket name is an ARN. The bucket name can be different for each operation. This is why I decided to call the builtin values assigned at client construction time "defaults". `AWS::S3::ForcePathStyle` is the only builtin where operation-time modifications are necessary, at least for now.

botocore/args.py
```diff
@@ -464,3 +466,63 @@ def _ensure_boolean(self, val):
...
+            ),
+            EPRBuiltins.AWS_S3_FORCE_PATH_STYLE: (
+                s3_config.get('addressing_style') == 'path'
+            ),
+            EPRBuiltins.AWS_S3_USE_ARN_REGION: s3_config.get(
+                'use_arn_region', True
+            ),
+            EPRBuiltins.AWS_S3CONTROL_USE_ARN_REGION: s3_config.get(
+                'use_arn_region', False
+            ),
```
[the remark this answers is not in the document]
jonemo: This is based on the following observations:

1. The botocore config variable `use_arn_region` defaults to `None`. 
2. In `utils.S3EndpointSetter`, the value of `use_arn_region` is consistently read with `.get('use_arn_region', True)`. 
3. Conversely, in `utils.S3ControlEndpointSetter`, the value is consistently read with `.get('use_arn_region', False)`.

botocore/args.py
```diff
@@ -464,3 +466,63 @@ def _ensure_boolean(self, val):
...
+                    'use_fips_endpoint'
+                )
+            ),
+            EPRBuiltins.AWS_USE_DUALSTACK: (
+                endpoint_bridge._resolve_use_dualstack_endpoint(service_name)
+            ),
+            EPRBuiltins.AWS_STS_USE_GLOBAL_ENDPOINT: (
+                self._should_set_global_sts_endpoint(
+                    region_name=region_name,
+                    endpoint_url=None,
```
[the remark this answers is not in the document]
jonemo: You are correct, the `endpoint_url` argument of `_should_set_global_sts_endpoint` normally gets populated with the `endpoint_url` client property. If this argument is set, `_should_set_global_sts_endpoint` always returns `False`. 

Going forward, this logic is delegated to EndpointResolverv2, which is why `endpoint_url` must always be set to `None` when calling `_should_set_global_sts_endpoint` for the purpose of computing builtin parameter values. This way, we can continue to use the other parts of `_should_set_global_sts_endpoint`. The alternative would have been to write a new function that duplicates half of `_should_set_global_sts_endpoint`.

tests/unit/test_args.py
jonemo: > How are you intending to retrieve this from the args module?

`_should_set_global_sts_endpoint` is a method in the class `ClientArgsCreator`. The assignment of builtin parameter values happens within the same class. 

This is, in fact, the reason why the `_compute_endpoint_resolver_v2_builtin_defaults` ended up in `ClientArgsCreator`: Most of the required information is available locally in this class.

tests/unit/test_args.py
jonemo: Tests are renamed in 87be3a7.

Re overloaded use of term "builtin": Unfortunately, both uses already have precedent. "Builtin data files" is consistent with `botocore.loaders. BUILTIN_DATA_PATH` and `botocore.loaders.BUILTIN_EXTRAS_TYPES`. (And of course the other use of "builtin" comes from the cross-SDK endpoints spec.)

tests/unit/test_args.py
jonemo: No longer applicable, this has been removed.

botocore/regions.py
```diff
@@ -401,3 +402,34 @@ def _expand_template(
         return template.format(
             service=service_name, region=endpoint_name, dnsSuffix=dnsSuffix
         )
+
+
+class EndpointResolverBuiltins(str, Enum):
+    # The AWS Region configured for the SDK client
+    AWS_REGION: str = "AWS::Region"
```
nateprewitt, line 409: I know I asked for this in the last review but I'm now realizing I was wrong. The values are all strings because they're the enums, I don't think the bool makes sense. The actual typing would be on the object returned from `compute_endpoint_resolver_builtin_defaults` but since it's just a dict, the comments here are fine. Sorry for the churn.

nateprewitt (changes requested):
Minor tweak due to a bad comment from me earlier. Otherwise, looks great, thanks!

botocore/regions.py
jonemo: Oops, these annotations looked so right I never thought twice about it. Good catch, removed in e0700d3.

nateprewitt (approved):
⛵

Issue #2752, opened 2022-09-01
Test failure


limburgher:
### Describe the bug

A test fails on 1.27.63 and up. Currently using Fedora 36-38, Python3.10 and 3.11.

### Expected Behavior

Test passes

### Current Behavior

tests/unit/test_client.py:1650: AssertionError
________________ TestClientErrors.test_BOTO_DISABLE_COMMONNAME _________________

self = <tests.unit.test_client.TestClientErrors testMethod=test_BOTO_DISABLE_COMMONNAME>

    @mock.patch.dict(os.environ, {'BOTO_DISABLE_COMMONNAME': 'true'})
    def test_BOTO_DISABLE_COMMONNAME(self):
        creator = self.create_client_creator()
        self.endpoint_data['sslCommonName'] = 'bar'
    
        with pytest.warns(None) as warning:
            creator.create_client(
                'myservice', 'us-west-2', credentials=self.credentials
            )
>           self.assertEqual(len(warning), 0)
E           AssertionError: 1 != 0

tests/unit/test_client.py:1650: AssertionError

### Reproduction Steps

pytest

### SDK version used

1.27.63-65

### Environment details (OS name and version, etc.)

Fedora 36-37-38

tim-finnigan:
Hi @limburgher thanks for reaching out. The unit test succeeded in all of the test builds so there may be something in your environment causing the test to fail. There may be another environment variable causing the issue. If you could share any relevant environment variables (with sensitive ones redacted) then that is something we could look more into.

tim-finnigan:
Hi @limburgher thanks for reaching out. The unit test succeeded in all of the test builds so there is likely something in your environment causing this. You could be getting an unrelated warning that is making the test fail. We’d need to know what the warning this which would require a further inspection of your code. Are you able to print the warning value?

limburgher:
[2507.patch.txt](https://github.com/boto/botocore/files/9478347/2507.patch.txt)
[build-1.27.65-1.fc38.log](https://github.com/boto/botocore/files/9478407/build-1.27.65-1.fc38.log)
Attaching the only patch we carry and the complete build log.

FYI this is failing during the RPM build, I'm one of the Fedora package maintainers.

glaubitz:
Tests are failing in openSUSE as well:

```
[  354s] =================================== FAILURES ===================================
[  354s] _____________ TestAutoGeneratedClient.test_BOTO_DISABLE_COMMONNAME _____________
[  354s] 
[...]

PR #2761, opened 2022-09-12
Add missing `super().tearDown()` for test classes with custom base classes

jonemo:
Addresses https://github.com/boto/botocore/issues/2752

The reported issue is a test failure of the new test `tests.unit.test_client.TestAutoGeneratedClient testMethod=test_BOTO_DISABLE_COMMONNAME`. This test failure is reproducible on all platforms by running the tests in `tests/functional/` before the failing test. Root cause was a test class in `tests/functional/test_mturk.py` that called `super().setUp()` instead of `super().tearDown()`. All tests run later in the same session ended up with a modified `os.environ`. 

This PR fixes the problematic test mentioned above. In addition, it adds a call to `super().tearDown()` to all `tearDown()` methods of test classes that have base class other than `unittest.TestCase` where this is not already in place.

### DIFF
--- a/tests/functional/test_mturk.py
+++ b/tests/functional/test_mturk.py
@@ -23,7 +23,7 @@ class TestMturk(BaseSessionTest):
         self.stubber.activate()
 
     def tearDown(self):
-        super().setUp()
+        super().tearDown()
         self.stubber.deactivate()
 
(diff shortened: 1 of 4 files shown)


nateprewitt (approved):
Thanks @jonemo, this all looks great!

PR #2762, opened 2022-09-16
aws_is_virtual_hostable_bucket endpoint provider standard library function

jonemo:
Implements the `aws.isVirtualHostableS3Bucket` standard library function for endpoint provider rulesets. This is a small addition to #2737.

The spec for the function is:

> Evaluates whether a string is a DNS compatible bucket name and can be used with virtual hosted style addressing. In addition to the restrictions defined in [RFC 1123](https://www.ietf.org/rfc/rfc1123.txt) and `isValidHostLabel`, the `isVirtualHostableS3Bucket` function must also validate that the bucket name is between [3,63] characters, does not contain upper case characters, and is not formatted as an IP address.

### DIFF
--- a/botocore/endpoint_provider.py
+++ b/botocore/endpoint_provider.py
@@ -394,9 +394,9 @@ class RuleSetStandardLibary:
         """
         if (
             value is None
+            or len(value) < 3
             or value.lower() != value
             or IPV4_RE.match(value) is not None
-            or (allow_subdomains is False and value.count(".") > 0)
         ):
             return False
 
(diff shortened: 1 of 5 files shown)


botocore/endpoint_provider.py
```diff
@@ -377,6 +377,39 @@ def _not(self, value):
...
+
+        if allow_subdomains is True:
+            return all(
+                self.aws_is_virtual_hostable_s3_bucket(label, False)
+                for label in value.split(".")
+            )
+
+        return 3 <= len(value) <= 63 and self.is_valid_host_label(
+            value, allow_subdomains=False
+        )
```
nateprewitt, line 411: We might be able to streamline this a bit more to rely on `is_valid_host_label`, but otherwise this looks great. Thanks!

```suggestion
        if (
            value is None
            or len(value) < 3
            or value.lower() != value
            or IPV4_RE.match(value) is not None
        ):
            return False

        if allow_subdomains is True:
            return all(
                self.aws_is_virtual_hostable_s3_bucket(label, False)
                for label in value.split(".")
            )

        return self.is_valid_host_label(value, allow_subdomains=False)
```

nateprewitt (approved):

Issue #2744, opened 2022-08-22
Newest version cause DeprecationWarning: 'urllib3.contrib.pyopenssl' module is deprecated


ddzialak:
### Describe the bug

Fresh installation cause warning:

```
DeprecationWarning: 'urllib3.contrib.pyopenssl' module is deprecated and will be removed in a future release of urllib3 2.x. Read more in this issue: https://github.com/urllib3/urllib3/issues/2680
```

That warning somehow cause, that `pytest` does not start tests.

### Expected Behavior

import of botocore should not issue that warning

### Current Behavior

`DeprecationWarning` is issued

### Reproduction Steps

after fresh installation:
```
(sf-new) [1]17:00:19:~/workspace/$ python -m botocore.httpsession
/home/darek/workspace/sf-new/lib/python3.10/site-packages/botocore/httpsession.py:41: DeprecationWarning: 'urllib3.contrib.pyopenssl' module is deprecated and will be removed in a future release of urllib3 2.x. Read more in this issue: https://github.com/urllib3/urllib3/issues/2680
  from urllib3.contrib.pyopenssl import orig_util_SSLContext as SSLContext
```

### Possible Solution

avoid import of `urllib3.contrib.pyopenssl`, for more see:
https://github.com/urllib3/urllib3/issues/2680

### Additional Information/Context

Workaround: enforcing version `urllib3==1.26.9` solve that issue

### SDK version used

Python 3.10.4
botocore 1.27.56

### Environment details (OS name and version, etc.)

Linux darek-dell 5.4.0-122-generic #138-Ubuntu SMP Wed Jun 22 15:00:31 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

nateprewitt:
Hi @ddzialak,

Botocore doesn't use pyopenssl, we have intentionally patched it to always use the standard library version of SSLContext. There's a lingering import from the pyopenssl module for backwards compatibility which appears to be triggering the warning.

If we were impacted we're already pinned to `urllib3<1.27` but this deprecation won't take effect until `urllib3==2.0`. We'll take a look at getting the warning suppressed, thanks for letting us know!

akx:
@ddzialak You likely have

```
filterwarnings =
    error
```

or similar in your Py.test configuration, causing deprecation warnings (such as this one) at test-dependency module import time to become errors.

You could configure that to something like

```
filterwarnings = 
    error
[...]

PR #2763, opened 2022-09-19
suppress pyopenssl deprecation warning with catch_warnings

dlm6693:
This addresses https://github.com/boto/botocore/issues/2744. It does so by suppressing all warnings of type `DeprecationWarning` when attempting to import from `pyOpenSSL`

### DIFF
--- a/botocore/httpsession.py
+++ b/botocore/httpsession.py
@@ -3,6 +3,7 @@ import os
 import os.path
 import socket
 import sys
+import warnings
 from base64 import b64encode
 
 from urllib3 import PoolManager, Timeout, proxy_from_url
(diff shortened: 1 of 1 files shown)


botocore/httpsession.py
```diff
@@ -37,8 +38,12 @@
     from ssl import OP_NO_TICKET, PROTOCOL_TLS_CLIENT
 
 try:
-    # Always import the original SSLContext, even if it has been patched
-    from urllib3.contrib.pyopenssl import orig_util_SSLContext as SSLContext
+    with warnings.catch_warnings():
```
nateprewitt, line 41:
```suggestion
    # pyopenssl will be removed in urllib3 2.0, we'll fall back to ssl_ at that point.
    # This can be removed once our urllib3 floor is raised to >=2.0.
    with warnings.catch_warnings():
```

nateprewitt (changes requested):
Quick change to clarify why this is being added, otherwise, looks good!

nateprewitt (approved):
⛵

PR #2768, opened 2022-09-28
Fetch query compatible error from header to support service

ziyanli-amazon:
### Motivation
As we decided to parse `AwsQuery` Compatible `Error` in the Http response header, the required changes need to take in place on both SDK client and Service. 
This will be a pre-requisite for migrating services from AWSQuery wire protocol to AWSJson.
### What changed
Error code can be fetched from response content (for example __type#fooException). This PR takes another scenario into consideration, of which when `x-amzn-query-error` header with valid value is found from the Http response. 
- The parsed error code remains unchanged if this header is missing, or contains null or invalid value.

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -683,6 +683,7 @@ class BaseJSONParser(ResponseParser):
     def _do_error_parse(self, response, shape):
         body = self._parse_body_as_json(response['body'])
         error = {"Error": {"Message": '', "Code": ''}, "ResponseMetadata": {}}
+        header = response['headers']
         # Error responses can have slightly different structures for json.
         # The basic structure is:
         #
@@ -694,10 +695,17 @@ class BaseJSONParser(ResponseParser):
         error['Error']['Message'] = body.get(
             'message', body.get('Message', '')
         )
+        # Error response may contain an x-amzn-query-error header for json
+        # we need to fetch the error code from this header in that case
+        query_error = header.get('x-amzn-query-error') if header else None
+        query_error_code = None
+        if query_error and query_error.split(';') and len(query_error.split(';')) == 2:
+            query_error_code = query_error.split(';')[0]
         # if the message did not contain an error code
         # include the response status code
         response_code = response.get('status_code')
-        code = body.get('__type', response_code and str(response_code))
+        code = query_error_code if query_error_code \
+            else body.get('__type', response_code and str(response_code))
         if code is not None:
             # code has a couple forms as well:
             # * "com.aws.dynamodb.vAPI#ProvisionedThroughputExceededException"
--- a/tests/unit/test_parsers.py
+++ b/tests/unit/test_parsers.py
@@ -1081,6 +1081,98 @@ class TestParseErrorResponses(unittest.TestCase):
         self.assertEqual(parsed['Error']['Message'], 'this is a message')
         self.assertEqual(parsed['Error']['Code'], 'ValidationException')
 
+    def test_response_with_query_error_for_json_protocol(self):
+        parser = parsers.JSONParser()
+        response = {
+            "body": b"""
+                {"__type":"amazon.foo.validate#ValidationException",
+                 "message":"this is a message"}
+                """,
+            "status_code": 400,
+            "headers": {
+                "x-amzn-requestid": "request-id",
+                "x-amzn-query-error": "AWS.SimpleQueueService.NonExistentQueue;Sender"
+            }
+        }
+        parsed = parser.parse(response, None)
+        # Even (especially) on an error condition, the
+        # ResponseMetadata should be populated.
+        self.assertIn('ResponseMetadata', parsed)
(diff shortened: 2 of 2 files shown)


botocore/parsers.py
```diff
@@ -694,10 +695,24 @@ def _do_error_parse(self, response, shape):
         error['Error']['Message'] = body.get(
             'message', body.get('Message', '')
         )
+        # Error response may contain an x-amzn-query-error header for json
+        # we need to fetch the error code from this header in that case
+        query_error = header.get('x-amzn-query-error') if header else None
```
nateprewitt, line 700: Are we trying to control for a case where header isn't a dict? Is that possible here? I'd expect something more idiomatic like this:
```suggestion
        query_error = headers.get('x-amzn-query-error', '')
```

tests/unit/test_parsers.py
```diff
@@ -1081,6 +1081,100 @@ def test_response_metadata_errors_alternate_form_json_protocol(self):
...
+                """,
+            "status_code": 400,
+            "headers": {
+                "x-amzn-requestid": "request-id",
+                "x-amzn-query-error": "AWS.SimpleQueueService.NonExistentQueue;Sender",
+            },
+        }
+        parsed = parser.parse(response, None)
+        # Even (especially) on an error condition, the
+        # ResponseMetadata should be populated.
```
nateprewitt, line 1099:
```suggestion
        # ResponseMetadata should always be populated.
```

botocore/parsers.py
```diff
@@ -683,6 +683,7 @@ def _handle_timestamp(self, shape, value):
     def _do_error_parse(self, response, shape):
         body = self._parse_body_as_json(response['body'])
         error = {"Error": {"Message": '', "Code": ''}, "ResponseMetadata": {}}
+        header = response['headers']
```
nateprewitt, line 686:
```suggestion
        headers = response['headers']
```

botocore/parsers.py
```diff
@@ -694,10 +695,24 @@ def _do_error_parse(self, response, shape):
...
+            query_error_code = query_error.split(';')[0]
         # if the message did not contain an error code
         # include the response status code
         response_code = response.get('status_code')
-        code = body.get('__type', response_code and str(response_code))
+        code = (
+            query_error_code
+            if query_error_code
+            else body.get('__type', response_code and str(response_code))
+        )
```
nateprewitt, line 715: We'll need to move the block above here so we have access to the response code. That will keep the logical grouping tighter as well.

tests/unit/test_parsers.py
```diff
@@ -1081,6 +1081,100 @@ def test_response_metadata_errors_alternate_form_json_protocol(self):
...
+                """,
+            "status_code": 400,
+            "headers": {
+                "x-amzn-requestid": "request-id",
+                "x-amzn-query-error": "AWS.SimpleQueueService.NonExistentQueue;sender;400",
+            },
+        }
+        parsed = parser.parse(response, None)
+        # Even (especially) on an error condition, the
+        # ResponseMetadata should be populated.
```
nateprewitt, line 1124: I'm not sure this comment is needed in every test. Was there something novel we were trying to note about this behavior?

botocore/parsers.py
```diff
@@ -694,10 +695,24 @@ def _do_error_parse(self, response, shape):
...
+        # Error response may contain an x-amzn-query-error header for json
+        # we need to fetch the error code from this header in that case
+        query_error = header.get('x-amzn-query-error') if header else None
+        query_error_code = None
+        if (
+            query_error
+            and query_error.split(';')
+            and len(query_error.split(';')) == 2
+        ):
+            query_error_code = query_error.split(';')[0]
```
nateprewitt, line 707: Let's avoid splitting this multiple times, we can simplify to just:

```suggestion
        query_error_components = query_error.split(';')
        code = None
        if len(query_error_components) == 2:
            code = query_error_components[0]
        if code is None:
            code =  body.get('__type', response_code and str(response_code))   
```

tests/unit/test_parsers.py
ziyanli-amazon: Thanks for catching that. This comment is indeed not needed for every test case. I will remove them accordingly and only leave one in the first test case as suggested

botocore/parsers.py
ziyanli-amazon: Thanks. Changes made accordingly.

botocore/parsers.py
ziyanli-amazon: Sounds good. Will move logic block together.

botocore/parsers.py
ziyanli-amazon: Thanks. Changes made accordingly.

ziyanli-amazon:
@nateprewitt Thanks so much for reviewing this PR and providing valuable comments. Commit with changes have been made accordingly.

nateprewitt (approved):
Looks good, thanks @ziyanli-amazon!

PR #2770, opened 2022-09-29
add unit test for endpoint resolution caching

dlm6693:
This PR adds a unit test for endpoint resolution caching, which the test suite did not previously include.

### DIFF
--- a/tests/unit/test_endpoint_provider.py
+++ b/tests/unit/test_endpoint_provider.py
@@ -14,6 +14,7 @@
 import json
 import logging
 import os
+from unittest.mock import Mock
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


nateprewitt (commented):
Looks like this is still lingering. Left a couple comments, let me know if we still needed this.

tests/unit/test_endpoint_provider.py
```diff
@@ -433,3 +434,24 @@ def test_auth_schemes_conversion_first_authtype_unknown(
...
+    request.cls.endpoint_provider = endpoint_provider
+
+
+@pytest.mark.usefixtures("test_endpoint_provider")
+class EndpointProviderCacheTestCase(unittest.TestCase):
+    @unittest.mock.patch("botocore.endpoint_provider.RuleSet.evaluate")
+    def test_endpoint_resolution_caches(self, mock_evaluate):
+
+        for _ in range(5):
+            self.endpoint_provider.resolve_endpoint(**{"Region": "us-east-1"})
```
jonemo, line 450: Why not the simpler:

```suggestion
            self.endpoint_provider.resolve_endpoint(Region="us-east-1")
```

(And the same a few lines below?)

tests/unit/test_endpoint_provider.py
```diff
@@ -343,3 +344,24 @@ def test_ruleset_unknown_parameter_type_raises(partitions):
...
     assert "Unknown parameter type: list." in str(exc_info.value)
+
+
+@pytest.fixture(scope="class")
+def test_endpoint_provider(endpoint_provider, request):
+    request.cls.endpoint_provider = endpoint_provider
+
+
+@pytest.mark.usefixtures("test_endpoint_provider")
+class EndpointProviderCacheTestCase(unittest.TestCase):
```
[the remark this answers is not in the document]
dlm6693: I think the reasoning behind this was because I was mixing/matching pytest and unittest. I see that pytest has the `monkeypatch` fixture. I'll try to incorporate that instead here.

tests/unit/test_endpoint_provider.py
dlm6693: Totally. Not sure why I did it like this. I tried to make the change to `assert_called_once_with` but it seems as though it expects it in the dict form

tests/unit/test_endpoint_provider.py
```diff
@@ -433,3 +434,22 @@ def test_auth_schemes_conversion_first_authtype_unknown(
...
+
+
+def test_endpoint_reevaluates_result(endpoint_provider, monkeypatch):
+    regions = ["us-east-1", "us-west-2"]
+    mock_evaluate = Mock()
+    monkeypatch.setattr(RuleSet, "evaluate", mock_evaluate)
+    for region in regions:
+        endpoint_provider.resolve_endpoint(Region=region)
+
+        # breakpoint()
```
jonemo, line 454: Debugging leftover.

jonemo (approved):
Looks like a debugging leftover snuck into the diff. Once that is removed this is good to merge 🚀

PR #2771, opened 2022-10-02
add tcp_keepalive to config_kwargs

miketheman:
The `tcp_keepalive` parameter was added to the Config object in #2766 When inspecting the return value of an instantiated Client's `meta` with the config, it is always `None`.

Does not handle resolving whether the value has been set in `scoped_config`.


dlm6693 (approved):
LGTM! Thanks for catching and submitting this so quickly @miketheman!

PR #2780, opened 2022-10-17
gzip rule set files for wheel

dlm6693:
This PR adds any service model file stored as a compressed GZIP JSON to be included in the wheel file that is pushed to PYPI as part of botocore's daily release cycle.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,6 +5,7 @@ include requirements-dev.txt
 include botocore/cacert.pem
 include botocore/vendored/requests/cacert.pem
 recursive-include botocore/data *.json
+recursive-include botocore/data *.json.gz
 graft docs
 prune docs/build
 graft tests


setup.py
```diff
@@ -38,7 +38,12 @@ def find_version(*file_paths):
     scripts=[],
     packages=find_packages(exclude=['tests*']),
     package_data={
-        'botocore': ['cacert.pem', 'data/*.json', 'data/*/*.json'],
+        'botocore': [
+            'cacert.pem',
+            'data/*.json',
+            'data/*/*.json',
```
dlm6693, line 44: I'm not exactly sure how this line has ever worked for including service model files because they're all one directory level down from what this indicates.

jonemo: Haven't tested this, but I suspect [this line in MANIFEST.in](https://github.com/boto/botocore/blob/develop/MANIFEST.in#L7) is responsible for the service model files. Both `package_data` and `MANIFEST.in` [are ok ways to define what files to include](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#data-files-support), but using them both in parallel _does_ seem a bit confusing...

nateprewitt: I think Jonas is right, having the recursive include is more complete and keeping the relevant files partitioned to the MANIFEST.in file may be better. Can we get this change migrated over to follow the pattern set already for *.json?

jonemo: +1, let's move this to MANIFEST.in.

Technically, we could also use globs in `setup.py` with something like:

```
...
'data/**/*.json',
'data/**/*.json.gz`,
...
```

But @nateprewitt made me go look up when that feature was added to setuptools and it turns out to be less than half a year old: https://setuptools.pypa.io/en/latest/history.html#v62-3-0 (i.e. a bit too young to be ok to use for us)

dlm6693: Ok thanks guys! I should be able to shoot up an update in the morning.

nateprewitt (changes requested):
It looks like you accidently included all of the files zipped and removed the s3/s3control files. Could you make sure those get fixed and then we'll be ready to merge.

nateprewitt (approved):

PR #2782, opened 2022-10-19
Move endpoint tests into correct position

nateprewitt:
Relocate s3/s3control endpoint test files into the testing directory instead of the base model directory for each service.


jonemo (approved):
LGTM

PR #2783, opened 2022-10-20
validation for bigdecimal type

irrelevelephant:
I have a use case for services with `bigdecimal` inputs. Adding validation for `bigdecimal` enables compatibility with the AWS CLI, which addresses a long-standing inconvenience.


irrelevelephant:
Discussed with maintainer, unlikely to be accepted.

PR #2788, opened 2022-10-21
Endpoint resolution v2 for Eventbridge

jonemo:
This PR is a small change stacked on top of https://github.com/boto/botocore/pull/2785. To see the diff between those two branches [click here](https://github.com/jonemo/botocore/compare/endpointsresolverv2-squash...jonemo:botocore:endpointsresolverv2-events).

https://github.com/boto/botocore/pull/2785 enables Endpoint resolution v2 for the `s3` and `s3control` clients. This PR adds the `events` (Eventbridge) client to the list. The PR bundles the relevant endpoint ruleset and endpoint tests data files. It also removes the code path for registering Eventbridge customizations which are completely replaced by the new endpoint ruleset.

After the code changes described above, seven test failures in the file `tests/functional/test_eventbridge.py` happened. These could be addressed by:

1. Removing assertions about the verbiage of exception messages. This is in line with the decision made for https://github.com/boto/botocore/pull/2785 to strive for consistent exception types but accept changes in text.
2. Add a trailing `/` to the expected URL in one assertion. This is a change in behavior (change from empty path to `/` path) but without significant effect. The new behavior is closer to [the documented example](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEvents.html#API_PutEvents_Examples) where `/` is used.

### DIFF
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -159,14 +159,6 @@ class ClientCreator:
         self._register_retries(service_client)
         self._register_s3_events(service_client, client_config, scoped_config)
         self._register_s3_control_events(service_client)
-
-        if client_args['endpoint_ruleset_resolver'] is None:
-            # When using the legacy endpoint resolver, several event handlers
-            # modify endpoint and request context.
-            self._register_eventbridge_events(
-                service_client, endpoint_bridge, endpoint_url
-            )
-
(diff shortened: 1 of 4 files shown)


nateprewitt (approved):
Looks good, thanks @jonemo!

PR #2791, opened 2022-10-22
Copy env vars when injecting fake creds for test

yan12125:
Without providing a copy of the current environment variables, the cmd-runner may not work as we are not proxying over environment variables that are needed to function correctly. For example, the environment can be setting the PYTHONPATH in order to find additional Python packages that are not installed in the default site packages.

The code is copied from https://github.com/aws/aws-cli/pull/6725

Fixes https://github.com/boto/botocore/issues/2676

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -165,9 +165,10 @@ class BaseClientDriverTest(unittest.TestCase):
 
     def setUp(self):
         self.driver = ClientDriver()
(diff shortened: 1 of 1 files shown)


yan12125:
Could someone have a look at this?

yan12125:
Alright, probably botocore devs are not interested in this.

PR #2792, opened 2022-10-24
Update S3 endpoint rules and tests

nateprewitt:
This PR will update our copies of S3's endpoint ruleset and tests to the latest version.


jonemo (approved):
LGTM

PR #2794, opened 2022-10-25
remove cgi module usage, which causes deprecation warnings on Python 3.11

wimglenn:
On Python 3.11, as soon as `botocore.utils` is imported there are [deprecation warnings](https://github.com/python/cpython/blob/v3.11.0/Lib/cgi.py#L57) due to the `import cgi`, a module which [is slated to be removed](https://github.com/python/cpython/issues/91217) from stdlib. [PEP 594](https://peps.python.org/pep-0594/#cgi) recommends replacing the `cgi.parse_header` with this `email.message` substitute.

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -12,8 +12,8 @@
 # language governing permissions and limitations under the License.
 import base64
 import binascii
-import cgi
 import datetime
+import email.message
 import functools
 import hashlib
 import io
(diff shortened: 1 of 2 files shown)


wimglenn:
@nateprewitt  Just saw you move CI to Python-3.11, could we get some eyes on this PR please?

nateprewitt:
Hi @wimglenn, thanks for checking in. It's on our radar but we're currently looking at other options than `email` to replace this behavior. Once we have a conclusion on that, we'll revisit here with a decision. For the time being, there's no immediate risk of removal until October 2023 at the earliest with the release of 3.12.

wimglenn:
Interesting, is there some problem with the stdlib email [recommendation](https://peps.python.org/pep-0594/#cgi) from the PEP?
I'm not concerned about the removal in 3.13, but about the warning itself (some of escalate warnings to errors in CI)

nateprewitt (approved):

PR #2796, opened 2022-10-26
Make Baseclient constructor signature backwards compatible with <1.28.0

jonemo:
#2785 (released in version 1.28.0) introduced a backwards-incompatible change to the signature of the constructor of `botocore.client.BaseClient`. This change modifies the signature to be backwards-compatible.


nateprewitt (approved):
Looks good, thanks @jonemo!

PR #2797, opened 2022-10-27
Support client context params for s3control

jonemo:
Context: #2740 introduced support for context params into Botocore. #2785 introduced a new way of resolving endpoints that utilizes context parameters. 

There are three types of context params: Static, dynamic, and client context params. Botocore currently intends to support _client_ context params only for the S3 services. The code in #2785 limits the use of client context params to `s3` only. This PR expands this to additionally include `s3control`. 

This PR also introduces functional test coverage that uses minimal fake models and rulesets to assert that all three types of context param are passed into the endpoint provider when expected, and not otherwise.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -239,7 +239,7 @@ class ClientArgsCreator:
         self._compute_connect_timeout(config_kwargs)
         s3_config = self.compute_s3_config(client_config)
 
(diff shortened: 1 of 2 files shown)


nateprewitt (approved):
Looks good, thanks @jonemo!

PR #2799, opened 2022-10-27
Fix for S3 pre-signed URLs with bucket names that contain dots

jonemo:
Addresses https://github.com/boto/botocore/issues/2798.

Version 1.28.0 erroneously introduced a change in behavior for S3 pre-signed URLs with bucket names that aren't valid host labels as defined in [RFC 1123](https://datatracker.ietf.org/doc/html/rfc1123#section-2). After the change, the resulting URLs no longer used the regional endpoint. 

This PR fixes the specific issue reported in https://github.com/boto/botocore/issues/2798 and adds test coverage for it. In addition, it reverts to pre-1.28.0 logic for whether to use a regional or global endpoint for certain S3 requests where either option is valid.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -46,6 +46,7 @@ from botocore.docs.utils import (
     AutoPopulatedParam,
     HideParamFromOperations,
 )
+from botocore.endpoint_provider import VALID_HOST_LABEL_RE
 from botocore.exceptions import (
     AliasConflictParameterError,
     ParamValidationError,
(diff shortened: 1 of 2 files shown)


botocore/handlers.py
```diff
@@ -1110,9 +1111,12 @@ def customize_endpoint_resolver_builtins(
...
-        context.get('is_presign_request')
-        and context.get('use_global_endpoint')
+        context.get('use_global_endpoint')
+        # don't use global endpoint when path-style addressing will be used
+        and (bucket_name is None or VALID_HOST_LABEL_RE.match(bucket_name))
         and not builtins[EndpointResolverBuiltins.AWS_S3_FORCE_PATH_STYLE]
+        # if the bucket name is an ARN, the ARN's region should be used in the
+        # endpoint
         and not bucket_is_arn
     ):
```
nateprewitt, line 1121:
```python
    # bucket names that are invalid host labels require path-style addressing
    # which is incompatible with the global endpoint for presigned urls.
    is_valid_host_label = bucket_name is None or VALID_HOST_LABEL_RE.match(bucket_name)

    # if the bucket name is an ARN, the ARN's region should be used in the endpoint
    not_bucket_arn = not bucket_is_arn

    if (
        context.get('use_global_endpoint')
        and is_valid_host_label
        and not builtins[EndpointResolverBuiltins.AWS_S3_FORCE_PATH_STYLE]
        and not_bucket_arn
    ):
```

nateprewitt (approved):

PR #2801, opened 2022-10-31
Add Endpoints 2.0 Rulesets

nateprewitt:
## Overview

This PR will add Endpoint Rulesets for all services as a follow up to #2785. These will provide deterministic rulesets for deriving endpoints based on user input, moving away from customized code for evaluating user input to resolve operation destination. This will improve inter-SDK compatibility for how endpoints are resolved, reducing a number of inconsistencies between different language implementations, as well as abstract the logic out of complex Botocore events.

This change will supersede the existing `endpoints.json` that many users are familiar with. To preserve backwards compatibility, we'll detect any deviations in our loaded `endpoints.json` file from what was distributed with the source code. If a variation **is** detected, we will use our old endpoints logic to load any customized endpoints.

## Botocore Size

With this change, we'll be adding a considerable number of additional models to Botocore. We're aware this is a common pain point for users given the size of Botocore. We've spent time this year to add optional support for gzipped models. 

Last week, we rolled out our first usage of this by compressing these rulesets for both S3 and S3Control for our [wheel](https://files.pythonhosted.org/packages/21/66/a34edc64b22cbc7a34da8c7ed53fc795204a0711778710878407ea405f98/botocore-1.28.4-py3-none-any.whl) distribution on PyPI. This initial test demonstrated an average space savings of 97.5% for our first two models. We'll be using this to minimize the initial impact of these model additions while we explore compressing the base `service-2.json` models in a future release. This has the potential to see a dramatic decrease in the overall size for Botocore. We'll track future work on this in the initial tracking issue (#2365).


tests/functional/test_model_completeness.py
```diff
@@ -34,15 +40,51 @@ def test_paginators_and_waiters_are_not_lost_in_new_version(
...
+@pytest.mark.parametrize("service_name", AVAILABLE_SERVICES)
+def test_all_endpoint_tests_exist(service_name):
+    """Tests the existence of endpoint-tests-1.json for each service
+    and verifies that content is present."""
+    endpoint_tests_file = os.path.join(
+        TEST_DATA_DIR, service_name, 'endpoint-tests-1.json'
+    )
+    with open(endpoint_tests_file) as f:
+        data = json.load(f)
+    assert len(data['testCases']) >= 1
```
jonemo, line 84: This test already exists in [`tests/functional/test_endpoint_rulesets.py:test_all_endpoint_tests_exist`](https://github.com/boto/botocore/blob/develop/tests/functional/test_endpoint_rulesets.py#L69-L74). In my opinion, it should stay there and be removed here because `test_endpoint_rulesets.py` includes other tests that use the contents of these files, all of which share a cache to reduce the number of times the `endpoint-tests-1.json` files need to be loaded.

jonemo:
It would probably make sense for this PR to also include some changes to the `tests/function/test_endpoint_rulesets.py` file. Currently, the constant `ALL_SERVICES` in this file is filtered by this conditional:

```python
if service_name in ENDPOINT_RESOLUTION_V2_SERVICES or FORCE_ENDPOINT_RESOLUTION_V2
```

After this PR, most (but not all) of the tests in `test_endpoint_rulesets.py` should run for all services, not the filtered list.

jonemo (approved):
LGTM

PR #2802, opened 2022-11-01
Enable endpoint resolution 2.0 for all services

jonemo:
Remove conditionals that limited ruleset based endpoint resolution to only s3, s3control, and eventbridge. After this PR, ruleset based endpoint resolution will be in use for all services/versions for which an `endpoint-rule-set-1.json` file exists in the service's `botocore/data/` subfolder.

Context:
* #2785 introduced logic for ruleset based endpoint resolution and included rulesets for s3 and s3control. 
* #2801 added rulesets for all services
* #2788 added eventbridge to the list of services for which ruleset based endpoint resolution is enabled.


botocore/client.py
```diff
@@ -117,16 +114,12 @@ def create_client(
...
-        ):
+        try:
             endpoints_ruleset_data = self._load_service_endpoints_ruleset(
                 service_name, api_version
             )
             partition_data = self._loader.load_data('partitions')
-        else:
+        except UnknownServiceError:
             endpoints_ruleset_data = None
             partition_data = None
```
nateprewitt, line 124: Can we do a quick info log here stating something like:

> Unknown service {service_name} detected, attempting to fallback to legacy endpoint routing.

tests/functional/test_endpoint_rulesets.py
```diff
@@ -189,7 +183,7 @@ def iter_e2e_test_cases_that_produce(endpoints=False, errors=False):
                     expected_endpoint,
                     marks=pytest.mark.skipif(
                         'sigv4a' in expected_authschemes,
-                        reason="Test case expects sigv4a which required CRT",
+                        reason="Test case expects sigv4a which requires CRT",
                     ),
```
nateprewitt, line 187: I just noticed this. I'd expect the conditional to be something closer to `'sigv4a' in expected_authschemes and not HAS_CRT`. Are these cases being run elsewhere or are we just dropping them?

nateprewitt (changes requested):
A quick logging assertion and question, otherwise looks good!

tests/functional/test_endpoint_rulesets.py
jonemo: That's an error, thank you for catching that. Fixed in 8188e27. This affected one test case. I confirmed that there are no other uses of `skipif` left that get this wrong.

botocore/client.py
jonemo: Done in 1cafc36.

nateprewitt (approved):
⛵ , thanks @jonemo!

PR #2804, opened 2022-11-01
Deprecate SSL common name

dlm6693:
This PR formally deprecates the usage of `sslCommonName` in endpoint resolution. This was a legacy codepath back when botocore supported python < 2.7. At this time python did not provide a SSL module, meaning certain certificate formats were needed. Now that new endpoint resolution via ruleset files has been released, this format is no longer necessary.

Please see https://github.com/boto/botocore/issues/2705 for additional details.


nateprewitt (commented):
Cool, I think this looks good! Let's add a release note entry for this, probably something like:

Enhancement
Endpoints
Discontinued use of sslCommonName hosts as detailed in 1.27.0 (old issue number here)

dlm6693:
@nateprewitt just pushed up the changelog. Also should mention I ran integration tests and didn't see any failures.

nateprewitt (approved):

PR #2805, opened 2022-11-01
Endpoint resolver 2.0: Use legacy resolver as source of default region name

jonemo:
Fall back to the legacy resolver as source of a default region name when a region name cannot be sourced from elsewhere.


nateprewitt (commented):
:shipit:

nateprewitt (approved):
Actually :shipit:

PR #2806, opened 2022-11-02
Remove the dependency on six.

mcepl:
Replaces #2512

Leaves vendored six.py and tests/functional/test_six_threading.py intact.

Otherwise references of six were removed from everywhere and the code depends on the standard library instead.


tests/unit/docs/bcdoc/test_style.py
```diff
@@ -294,7 +293,7 @@ def test_write_py_doc_string(self):
             'returns: None'
         )
         style.write_py_doc_string(docstring)
-        self.assertEqual(style.doc.getvalue(), six.b(docstring + '\n'))
+        self.assertEqual(style.doc.getvalue(), docstring + b'\n')
```
nateprewitt, line 296: We need to make 291-293 bytes if you're going to do this.

```
        docstring = (
            b'This describes a function\n'
            b':param foo: Describes foo\n'
            b'returns: None'
        )
```

tests/unit/auth/test_sigv4.py
```diff
@@ -55,21 +57,15 @@
...
-        # NO support
-        'get-header-key-duplicate',
-        'get-header-value-order',
-    ]
 
 log = logging.getLogger(__name__)
 
 
-class RawHTTPRequest(six.moves.BaseHTTPServer.BaseHTTPRequestHandler):
+class RawHTTPRequest(http.server.BaseHTTPRequestHandler):
```
nateprewitt, line 64: Rather than importing the whole `http.server` module can we do:

```
from http.server import BaseHTTPRequestHandler
```

tests/integration/test_client_http.py
```diff
@@ -1,6 +1,8 @@
 import contextlib
+import http.server
```
nateprewitt, line 2:
```suggestion
from http.server import BaseHTTPServer
```

botocore/httpsession.py
```diff
@@ -5,6 +5,7 @@
 import sys
 import warnings
 from base64 import b64encode
+from urllib.parse import unquote
```
nateprewitt, line 8: This is already in `botocore.compat` let's import from there.

botocore/docs/bcdoc/docstringparser.py
```diff
@@ -10,10 +10,10 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
-from botocore.compat import six
+import html.parser
```
nateprewitt, line 13:
```suggestion
from html.parser import HTMLParser
```

botocore/compat.py
```diff
@@ -17,6 +17,7 @@
 import inspect
 import warnings
 import hashlib
+import http.client
```
nateprewitt, line 20:
```suggestion
from http.client import HTTPMessage
```

nateprewitt (changes requested):
Thanks for the pared down PR, @mcepl! A couple minor comments, I believe the one around bytes should solve the test failures. Please let us know if you have any questions.

tests/unit/docs/bcdoc/test_style.py
mcepl: I don’t understand, what am I supposed to do?

```python
stitny~$ python3.11
Python 3.11.0 (main, Oct 25 2022, 08:54:39) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> docstring = (
...     b'This describes a function\n'
...     b':param foo: Describes foo\n'
...     b'returns: None'
... )
>>> len(docstring)
65
>>> type(docstring)
<class 'bytes'>
>>> 
```

mcepl: Tests are passing now. I guess, I got it, finally.

nateprewitt (approved):
Alright, it looks like this is set. Thanks @mcepl!

PR #2808, opened 2022-11-09
Revert signatures of _register_s3_events and _register_s3_control_events

jonemo:
Reverts the signatures of "private" functions `_register_s3_events` and `_register_s3_control_events` to what they were before botocore 1.28.0.


nateprewitt (approved):
Thanks, @jonemo!

PR #2809, opened 2022-11-10
Removed additional whitespace after hyperlinks for special cases in our docs

jonathan343:
### Problem
When auto-generating botocore documentation, an additional whitespace is hardcoded after every hyperlink is created. This may have been done under the assumption that a hyperlink would always be followed by a word but fails to consider cases where a space wouldn't be needed such as before the following characters `['.', ',', '?', '!', ':', ';']`.
**Examples:**
* https://docs.aws.amazon.com/cli/latest/reference/ec2/create-traffic-mirror-target.html
* https://docs.aws.amazon.com/cli/latest/reference/kms/generate-random.html

### Solution
This PR addresses this issue by providing a look-ahead (`next`) when possible, and if either of these characters is detected, a space is omitted. 

### Validation
These changes were validated by generating the documentation files before and after the change and comparing using 
```
diff -w <old_docs> <new_docs> | grep .rst
```
No changes other than the expected white space changes were detected.

### DIFF
--- a/botocore/docs/bcdoc/docstringparser.py
+++ b/botocore/docs/bcdoc/docstringparser.py
@@ -115,14 +115,13 @@ class StemNode(Node):
         self._write_children(doc)
 
     def _write_children(self, doc):
-        for index in range(len(self.children)):
-            if isinstance(self.children[index], TagNode):
-                next_child = None
-                if index + 1 < len(self.children):
-                    next_child = self.children[index + 1]
-                self.children[index].write(doc, next_child)
+        for index, child in enumerate(self.children):
+            next_child = None
+            if isinstance(child, TagNode) and index + 1 < len(self.children):
+                next_child = self.children[index + 1]
+                child.write(doc, next_child)
             else:
-                self.children[index].write(doc)
+                child.write(doc)
 
 
 class TagNode(StemNode):
@@ -135,21 +134,21 @@ class TagNode(StemNode):
         self.attrs = attrs
         self.tag = tag
 
-    def write(self, doc, next):
+    def write(self, doc, next_child=None):
         self._write_start(doc)
         self._write_children(doc)
-        self._write_end(doc, next)
+        self._write_end(doc, next_child)
 
     def _write_start(self, doc):
         handler_name = 'start_%s' % self.tag
         if hasattr(doc.style, handler_name):
             getattr(doc.style, handler_name)(self.attrs)
 
-    def _write_end(self, doc, next):
+    def _write_end(self, doc, next_child):
         handler_name = 'end_%s' % self.tag
         if hasattr(doc.style, handler_name):
             if handler_name == 'end_a':
-                getattr(doc.style, handler_name)(next)
+                getattr(doc.style, handler_name)(next_child)
             else:
                 getattr(doc.style, handler_name)()
 
@@ -158,9 +157,9 @@ class LineItemNode(TagNode):
     def __init__(self, attrs=None, parent=None):
         super().__init__('li', attrs, parent)
 
-    def write(self, doc, next):
+    def write(self, doc, next_child=None):
         self._lstrip(self)
-        super().write(doc, next)
+        super().write(doc, next_child)
 
     def _lstrip(self, node):
         """
--- a/botocore/docs/bcdoc/docstringparser.py
+++ b/botocore/docs/bcdoc/docstringparser.py
@@ -117,7 +117,7 @@ class StemNode(Node):
     def _write_children(self, doc):
         for index, child in enumerate(self.children):
             if isinstance(child, TagNode) and index + 1 < len(self.children):
-                # Provide a look ahead for TagNode's when one exists
+                # Provide a look ahead for TagNodes when one exists
                 next_child = self.children[index + 1]
                 child.write(doc, next_child)
             else:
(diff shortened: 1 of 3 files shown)


botocore/docs/bcdoc/docstringparser.py
```diff
@@ -115,8 +115,14 @@ def write(self, doc):
...
-        for child in self.children:
-            child.write(doc)
+        for index in range(len(self.children)):
+            if isinstance(self.children[index], TagNode):
+                next_child = None
+                if index + 1 < len(self.children):
+                    next_child = self.children[index + 1]
+                self.children[index].write(doc, next_child)
+            else:
+                self.children[index].write(doc)
```
nateprewitt, line 125: We might be able to simplify this logic a little, since we're already accessing the value at the target every time, we can use enumerate to grab it for us.

```suggestion
        for index, child in enumerate(self.children):
            next_child = None
            if isinstance(child, TagNode) and index + 1 < len(self.children):
                next_child = self.children[index + 1]
            child.write(doc, next_child)
```

botocore/docs/bcdoc/style.py
```diff
@@ -249,7 +249,13 @@ def end_a(self):
                 self.doc.hrefs[self.a_href] = self.a_href
                 self.doc.write('`__')
             self.a_href = None
-        self.doc.write(' ')
+
+        special_characters = ['.', ',', '?', '!', ':', ';']
```
nateprewitt, line 253: Rather than declaring this list every time it's called, let's move it to a module constant so it's only built once. We can also use a tuple instead of a list it make it immutable.

botocore/docs/bcdoc/style.py
```diff
@@ -227,7 +227,7 @@ def sphinx_reference_label(self, label, text=None):
         else:
             self.doc.write(text)
 
-    def end_a(self):
+    def end_a(self, next):
```
nateprewitt, line 230: Let's make next optional here:
```suggestion
    def end_a(self, next=None):
```

botocore/docs/bcdoc/docstringparser.py
```diff
@@ -129,29 +135,32 @@ def __init__(self, tag, attrs=None, parent=None):
...
         if hasattr(doc.style, handler_name):
             getattr(doc.style, handler_name)(self.attrs)
 
-    def _write_end(self, doc):
+    def _write_end(self, doc, next):
         handler_name = 'end_%s' % self.tag
         if hasattr(doc.style, handler_name):
-            getattr(doc.style, handler_name)()
+            if handler_name == 'end_a':
+                getattr(doc.style, handler_name)(next)
```
nateprewitt, line 152: Fact check me, because I may be wrong here. We should only be passing in a non-None value here when we've:

1.) Determined we're in a Tag node
2.) We have a look ahead character

Would it be fair to say we could pass next to any handler if the value of next is not None?

nateprewitt (changes requested):
Looking great! Left a few comments and can you check to see if we have any tests in place around this? It would be good if we can add some regression checks.

botocore/docs/bcdoc/docstringparser.py
```diff
@@ -129,29 +135,32 @@ def __init__(self, tag, attrs=None, parent=None):
...
-        self._write_end(doc)
+        self._write_end(doc, next)
 
     def _write_start(self, doc):
         handler_name = 'start_%s' % self.tag
         if hasattr(doc.style, handler_name):
             getattr(doc.style, handler_name)(self.attrs)
 
-    def _write_end(self, doc):
+    def _write_end(self, doc, next):
```
dlm6693, line 148: can you also rename `next` to something more detailed like `next_child` or something along those lines? `next` is a builtin function so we should avoid that namespace collision

botocore/docs/bcdoc/docstringparser.py
dlm6693: nit: set `index + 1` to a variable since we're using it multiple times. Maybe `next_child_index`?

jonathan343: This simplified logic would require adding an optional parameter to `DataNode.write(self, doc, next_child=None)` which would never be used since it currently only accepts `(self, doc)`. Is this a good tradeoff for a simplified logic?

nateprewitt: Yeah, if we're not overriding every function signature, let's reintroduce the if/else and call without `next_child` if it's not a TagNode.

nateprewitt: > nit: set index + 1 to a variable since we're using it multiple times. Maybe next_child_index?

Generally, we create variables for complex concepts like multi-part conditionals. That let's us easily understand what it's trying to express. For things like incrementing a counter, there's minimal performance hit in doing it repeatedly (in some cases it may take more time to allocate the variable than the temporary register value).

It's likely easier to read this way, so unless we have a compelling argument, it's probably fine to leave.

botocore/docs/bcdoc/docstringparser.py
jonathan343: I dont think this would work since it would allow a `next_child` to be passed for other tag node's such as `end_i`, `end_ul`, `end_li`... which currently don'g accent a `next_child` since its not needed. Only `end_a` accepts this parameter.

nateprewitt (changes requested):

nateprewitt (approved):
Minor grammar nit but otherwise looks good. Thanks @jonathan343!

PR #2811, opened 2022-11-14
Fix: S3 on Outpost requests miss x-amz-content-sha256 headers

jonemo:
Addresses an issue where requests for some S3 API operations that should include the `x-amz-content-sha256` header do not have this header for S3 on Outpost. This issue was introduced in version 1.28.0 of botocore.

### DIFF
--- a/tests/functional/test_s3.py
+++ b/tests/functional/test_s3.py
@@ -716,6 +716,9 @@ class TestAccesspointArn(BaseS3ClientConfigurationTest):
             "s3-outposts.us-west-2.amazonaws.com"
         )
         self.assert_endpoint(request, expected_endpoint)
+        sha_header = request.headers.get("x-amz-content-sha256")
+        self.assertIsNotNone(sha_header)
+        self.assertNotEqual(sha_header, b"UNSIGNED-PAYLOAD")
 
     def test_basic_outpost_arn_custom_endpoint(self):
         outpost_arn = (
--- a/tests/functional/test_s3.py
+++ b/tests/functional/test_s3.py
@@ -1405,6 +1405,10 @@ class TestS3SigV4(BaseS3OperationTest):
         self.assertIn("content-md5", self.get_sent_headers())
 
     def test_content_sha256_set_if_config_value_is_true(self):
+        # By default, put_object() does not include an x-amz-content-sha256
+        # header because it also includes a `Content-MD5`` header. The
+        # `payload_signing_enabled` config overrides this logic and forces the
+        # header.
         config = Config(
             signature_version="s3v4", s3={"payload_signing_enabled": True}
         )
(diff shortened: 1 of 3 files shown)


tests/functional/test_s3.py
```diff
@@ -1402,6 +1405,10 @@ def test_content_md5_set_empty_file(self):
         self.assertIn("content-md5", self.get_sent_headers())
 
     def test_content_sha256_set_if_config_value_is_true(self):
+        # By default, put_object() does not include an x-amz-content-sha256
+        # header because it also includes a `Content-MD5`` header. The
```
nateprewitt, line 1409:
```suggestion
        # header because it also includes a `Content-MD5` header. The
```

tests/functional/test_s3.py
```diff
@@ -1431,6 +1438,55 @@ def test_content_sha256_not_set_if_config_value_is_false(self):
...
         self.assertEqual(sha_header, b"UNSIGNED-PAYLOAD")
 
+    def test_content_sha256_set_if_config_value_not_set_put_object(self):
+        # The default behavior matches payload_signing_enabled=False. For
+        # operations where the `Content-MD5` is present this means that
+        # `x-amz-content-sha256` is present but not set.
+        config = Config(signature_version="s3v4")
+        self.client = self.session.create_client(
+            "s3", self.region, config=config
+        )
```
nateprewitt, line 1448: Did we intentionally set `s3v4` or was this carryover from other tests? If it's not proving something specific, maybe we want to leave this unconfigured to make sure s3 is choosing the right signer.

tests/functional/test_s3.py
```diff
@@ -1431,6 +1438,55 @@ def test_content_sha256_not_set_if_config_value_is_false(self):
...
+        self.http_stubber = ClientHTTPStubber(self.client)
+        self.http_stubber.add_response()
+        with self.http_stubber:
+            self.client.list_objects(Bucket="foo")
+        sent_headers = self.get_sent_headers()
+        sha_header = sent_headers.get("x-amz-content-sha256")
+        self.assertNotEqual(sha_header, b"UNSIGNED-PAYLOAD")
```
nateprewitt, line 1471: Do we want to also test this is not None? I don't think we're actually proving the header was set.

tests/functional/test_s3.py
```diff
@@ -1431,6 +1438,55 @@ def test_content_sha256_not_set_if_config_value_is_false(self):
...
+
+    def test_content_sha256_set_s3_on_outpost(self):
+        # S3 on Outpost bucket names should behave the same way.
+        config = Config(signature_version="s3v4")
```
nateprewitt, line 1475: Same note here, if we're setting the signature_version, I'm not sure this would have caught the original bug.

nateprewitt (changes requested):
Couple of comments on the tests, but otherwise looks good.

tests/functional/test_s3.py
jonemo: This was a carryover from the other tests. 

At a high level: 
* The only test that fails before the fix in this PR and passes afterwards is in `test_basic_outpost_arn` where this PR adds three lines (719-721). This is in commit 384f88641.
* All changes to the `TestS3SigV4` class are tangential and improve test coverage and add inline comments because those tests misled us during the investigation of this bug. These are in commit 2f2b1acc2.

tests/functional/test_s3.py
jonemo: Same response as above: Generally agree with you, but this is in a pre-existing test class named `TestS3SigV4`.

nateprewitt (approved):
:shipit:

PR #2812, opened 2022-11-15
Add support for sso-session credentials

nateprewitt:
This PR will add support for loading credentials via a configured `sso-session` in your AWS config file.

### DIFF
--- a/tests/functional/test_credentials.py
+++ b/tests/functional/test_credentials.py
@@ -10,6 +10,7 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import json
 import math
 import os
 import shutil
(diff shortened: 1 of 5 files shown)


tests/functional/test_credentials.py
```diff
@@ -929,3 +934,110 @@ def test_assume_role_web_identity_uses_same_region_as_client(self):
...
+
+    def tearDown(self):
+        shutil.rmtree(self.tempdir)
+        super().tearDown()
+
+    def write_config(self, config):
+        with open(self.config_file, 'w') as f:
+            f.write(config)
+
+    def test_token_chosen_from_provider(self):
```
kyleknap, line 986: It would be great if we added test cases for some of the error cases such as:
1. Mismatching start url and sso region
2. SSO session not being defined

I'm also wondering if we should add a case for having an expired SSO token in the cache and needing the token provider to refresh first. However, I'm not sure how straightforward that will be to add. Thoughts?

kyleknap (commented):
Looks fine. Just had comments on the tests.

tests/functional/test_credentials.py
nateprewitt: The first two are super reasonable, I attempted the last one earlier but the MockCache gets very complicated to do this without the ability to change inputs.

kyleknap (approved):
Looks good :ship:

PR #2814, opened 2022-11-17
Readd missing six import in compat

nateprewitt:
This fixes a miss from #2806 where six was incorrectly removed from the compat module. This is required for external test frameworks and is needed for backwards compatibility.


dlm6693 (approved):
🚀

PR #2817, opened 2022-11-21
Fixed doc issues around links and nested formatting

jonathan343:
This PR addresses issues with botocore docs which include:
* Missing or necessary white space which causes links to render incorrectly
* [Nested formatting](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#:~:text=No%20nested%20inline%20markup%3A%20Something%20like%20*see%20%3Afunc%3A%60foo%60*%20is%20not%20possible.) not being supported by Sphinx.

Examples:
* https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/kms.html#:~:text=https%3A//docs.aws.amazon.com/kms/latest/developerguide/%60.
* https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/iam.html#:~:text=parameter%20in%20the-,%60%60%20GetOrganizationsAccessReport%20%60%60,-operation%20to%20check

### DIFF
--- a/botocore/docs/bcdoc/style.py
+++ b/botocore/docs/bcdoc/style.py
@@ -67,7 +67,7 @@ class BaseStyle:
     def add_leading_space(self):
         # Adds a leading white space if none exists.
         last_write = self.doc.pop_write()
-        if last_write == None:
+        if last_write is None:
             last_write = ''
         if last_write != '' and last_write[-1] != ' ':
             last_write += ' '
@@ -249,7 +249,9 @@ class ReSTStyle(BaseStyle):
         # Pop till we reach the link start character to retrieve link text.
         last_write = doc.pop_write()[::-1]
         while last_write[-1] != '`':
-            last_write += doc.pop_write()[::-1]  # Reverse text to preserve order
+            last_write += doc.pop_write()[
+                ::-1
+            ]  # Reverse text to preserve order
         doc.push_write('`')
         last_write = last_write[:-1]
 
--- a/botocore/docs/bcdoc/style.py
+++ b/botocore/docs/bcdoc/style.py
@@ -64,6 +64,15 @@ class BaseStyle:
     def italics(self, s):
         return s
 
+    def add_leading_space(self):
+        # Adds a leading white space if none exists.
+        last_write = self.doc.pop_write()
+        if last_write == None:
+            last_write = ''
+        if last_write != '' and last_write[-1] != ' ':
+            last_write += ' '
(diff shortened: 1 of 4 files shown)


botocore/docs/bcdoc/style.py
```diff
@@ -229,9 +244,26 @@ def sphinx_reference_label(self, label, text=None):
...
+            last_write += doc.pop_write()[
+                ::-1
+            ]  # Reverse text to preserve order
+        doc.push_write('`')
+        last_write = last_write[:-1]
+
+        # Remove whitespace and reverse again to correct order, then push clean text.
+        last_write = last_write.rstrip(' ')[::-1]
+        if last_write != '':
+            doc.push_write(last_write)
```
jonemo, line 261: I don't understand what all this reversing is for here. Isn't this just the same as the following?

```suggestion
        last_write = doc.pop_write()
        while not last_write.startswith('`'):
            last_write = doc.pop_write() + last_write
        doc.push_write('`')
        last_write = last_write[1:].rstrip(' ')
        if last_write != '':
            doc.push_write(last_write)
```

jonemo (changes requested):
Looks good, but I had a few comments where the code seems unnecessarily complex.

botocore/docs/bcdoc/style.py
jonathan343: Yup, this should achieve the same thing. I think I just ran with the first solution that clicked for me. Only think needed for this simplification is an `lstrip` instead of `rstrip` since we're trying to remove the space in an example such as 
```
-- Incorrect --
` Link Text <some_link>`

-- Correct --
`Link Text <some_link>`
```
I can also provide said example in a comment to help future readers.

botocore/docs/bcdoc/style.py
```diff
@@ -64,6 +64,15 @@ def underline(self, s):
     def italics(self, s):
         return s
 
+    def add_leading_space(self):
+        # Adds a leading white space if none exists.
```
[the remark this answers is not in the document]
jonathan343: These writes can be any reStructuredText previously generated. This function pops whatever was written most recently and adds adds a space. We mainly want to prevent:
```
foo<code>bar</code>
``` 
from becoming 
```
foo`bar`
```
and instead want 
```
foo `bar`
```
since inline markup must be separated from surrounding text. I'll update function name and improve description.

jonemo (approved):
🚢 🇮🇹

PR #2818, opened 2022-11-21
Resolve endpoint with default partition when no region is set

jonemo:
Addresses https://github.com/aws/aws-cli/issues/7455.

Changes endpoint resolution to resolve the endpoint for the default AWS partition when no region is given.

### DIFF
--- a/tests/functional/test_sts.py
+++ b/tests/functional/test_sts.py
@@ -288,3 +288,29 @@ class TestSTSEndpoints(BaseSessionTest):
             expected_url='https://sts.not-real.amazonaws.com/',
             expected_signing_region='not-real',
         )
+
+
+def test_assume_role_with_saml_no_region_custom_endpoint(patched_session):
+    # When an endpoint_url and no region are given, AssumeRoleWithSAML should
+    # resolve to the endpoint_url and succeed, not fail in endpoint resolution:
+    # https://github.com/aws/aws-cli/issues/7455
+
+    client = patched_session.create_client(
+        'sts', region_name=None, endpoint_url="https://custom.endpoint.aws"
+    )
+    assert client.meta.region_name is None
+
+    mock_response_body = b"""\
+<AssumeRoleWithSAMLResponse xmlns="https://sts.amazonaws.com/doc/2011-06-15/">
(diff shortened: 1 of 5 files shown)


tests/functional/test_sts.py
```diff
@@ -288,3 +288,27 @@ def test_client_for_unknown_region_with_regional_configured(self):
...
+    # https://github.com/aws/aws-cli/issues/7455
+
+    client = patched_session.create_client(
+        'sts', region_name=None, endpoint_url="https://custom.endpoint.aws"
+    )
+    mock_response_body = b"""\
+<AssumeRoleWithSAMLResponse xmlns="https://sts.amazonaws.com/doc/2011-06-15/">
+    <AssumeRoleWithSAMLResult></AssumeRoleWithSAMLResult>
+</AssumeRoleWithSAMLResponse>
+"""
```
kyleknap, line 305: We should also probably add an assertion that the `client.meta.region_name` is `None`. Even though we specify `region_name=None`, we could potential inherit region from the session. Then given the `patched_session` seems like it will be widely used, we could potentially drift from this initial state if for example we decided to add a region to the `patched_session` fixture later on.

kyleknap (commented):
Looks good. I also confirmed this fixed the issue raised in this PR: https://github.com/aws/aws-cli/issues/7455. I just had one small comment on the tests.

.changes/next-release/bugfix-dynamodb-38777.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "dynamodb",
```
nateprewitt, line 3: I think we got a stray changelog entry

jonemo: Correct, will clean that up with the next push that's actually meant for this branch, to avoid unnecessary GH Actions runs.

tests/functional/test_sts.py
jonemo: Done in bbd870b

kyleknap (approved):
Looks good! :ship:

nateprewitt (approved):

PR #2819, opened 2022-11-21
Fix: S3 Object Lambda requests miss x-amz-content-sha256 headers

jonemo:
Addresses https://github.com/aws/aws-cli/issues/7456: Requests for some S3 API operations that should include the x-amz-content-sha256 header do not have this header for S3 Object Lambdas. This issue was introduced in version 1.28.0 of botocore.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -94,6 +94,9 @@ _OUTPOST_ARN = (
     r'[a-zA-Z0-9\-]{1,63}[/:]accesspoint[/:][a-zA-Z0-9\-]{1,63}$'
 )
 VALID_S3_ARN = re.compile('|'.join([_ACCESSPOINT_ARN, _OUTPOST_ARN]))
+# signing names used for the services s3 and s3-control, for example in
+# botocore/data/s3/2006-03-01/endpoints-rule-set-1.json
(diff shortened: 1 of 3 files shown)


botocore/handlers.py
```diff
@@ -218,8 +218,9 @@ def set_operation_specific_signer(context, signing_name, **kwargs):
         if auth_type == 'v4-unsigned-body':
             context['payload_signing_enabled'] = False
 
-        # s3 and s3-control have customized signers "s3v4" and "s3v4a".
-        if signing_name in ('s3', 's3-outposts'):
+        # Signing names used by s3 and s3-control use customized signers "s3v4"
+        # and "s3v4a".
+        if signing_name in ('s3', 's3-outposts', 's3-object-lambda'):
```
nateprewitt, line 223: Let's hoist this up to something like `S3_SIGNING_NAMES` at the top of the module.

nateprewitt (changes requested):
One minor change but otherwise looks good. Did we validate the test below works in 1.27.96? That's the only other thing I'd want to validate before we merge.

botocore/handlers.py
jonemo: Done in 1674a9d

nateprewitt (approved):

PR #2830, opened 2022-12-09
Optionally allow HTTP redirects

### LINKED ISSUE #2571: Support S3 HTTP redirects to non-Amazon URI's by using the location header
I'm a developer contributing to [Apache Ozone](https://hadoop.apache.org/ozone/), an S3 compatible object store.

I would like to use HTTP redirects to balance the load amongst Ozone-based s3 servers, but botocore doesn't currently support redirects to non-amazon URI's.

Typically, http clients implement redirects using the 'location' header in the response as the destination.  botocore uses the region instead of the location header .  [From that, it constructs an amazonaws.com based URI to redirect to.](https://github.com/boto/botocore/blob/develop/botocore/utils.py#L1373-L1391)  This means that botocore only ever redirects s3 requests to amazon URI's even if it made the original request to a non-Amazon server.

Allow botocore to support s3 redirection to non-Amazon URI's  using the location header if available.  This functionality would be controlled by a config variable or command line option, and would be off by default.

 It could look something like [this.](https://github.com/GeorgeJahad/botocore/compare/187f927efc95eae71551bb4a31e9ec0fdf5dd8c4..GeorgeJahad:allowNonAwsRedirects)

simontraill:
This adds (optional, client configured) support for the HTTP "location" header when processing HTTP redirects for S3 responses.

I'm a developer at NVIDIA; I wrote this on behalf of the [AIStore project](https://github.com/NVIDIA/aistore). We'd really like to use botocore (and boto3) for client access; AIStore (and some other systems - like Apache Ozone, [referenced by this issue](https://github.com/boto/botocore/issues/2571)) rely on standard HTTP redirects - using the "location" header - for load balancing. 

botocore [constructs redirection URLs using the region instead](https://github.com/boto/botocore/blob/a94d3dfaa74b20d4f1ea700e411224467e76c7da/botocore/utils.py#L1595-L1635), and won't allow redirections outside of known Amazon URIs. 

I've gated the change here behind a config setting; that said, other AWS SDKs (aws-sdk-go, for instance) do appear to support redirects as normal, so hopefully this behaviour is in line with other clients.

This change - 

 - Introduces `allow_http_redirects` config option (default: `False`).
 - If set to `True`, performs [HTTP redirection](https://www.rfc-editor.org/rfc/rfc7231#section-6.4) based on the `Location` header
 - Limits the number of redirections for a given request to three (I didn't feel adding another config option was worth it in this case).
 - Closes #2571.

```
import boto3
from botocore.config import Config as Config

config = Config(allow_http_redirects=True)
session = boto3.Session()
s3 = session.resource("s3", endpoint_url="http://127.0.0.1:8080/s3", config=config)
```

I've added unit and functional tests and tried to keep the style consistent with your existing work. If there's any demand for caching using the `Cache-Control` or `Expires` headers, I could certainly do it in the follow-up.

Thanks.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -234,6 +234,7 @@ class ClientArgsCreator:
                 client_cert=client_config.client_cert,
                 inject_host_prefix=client_config.inject_host_prefix,
                 tcp_keepalive=client_config.tcp_keepalive,
+                allow_http_redirects=client_config.allow_http_redirects,
             )
         self._compute_retry_config(config_kwargs)
         self._compute_connect_timeout(config_kwargs)
--- a/botocore/config.py
+++ b/botocore/config.py
@@ -189,6 +189,18 @@ class Config:
         creating new connections if set to True.
 
         Defaults to False.
+
+    :type allow_http_redirects: bool
+    :param allow_http_redirects: Whether to allow HTTP redirects using the
+        'location' header.
+
+        Setting this True will cause botocore to redirect requests to
+        the URL specified by the HTTP 'location' header, if any, on receiving
+        a HTTP 301, 302 or 307 from the upstream service.
+
+        HTTP redirections are limited to a maximum of three per request.
+
+        Defaults to False.
     """
 
     OPTION_DEFAULTS = OrderedDict(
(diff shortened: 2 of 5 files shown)


ptrxyz:
So this is not going to happen I suppose?

PR #2831, opened 2022-12-12
Update ARN delimiter parsing in endpoint provider

dlm6693:
This addresses a bug for `aws_parse_arn` in `endpoint_provider.RuleSetStandardLibrary`. Currently the `resourceId` portion of a parsed ARN was only thought to either have a `/` or `:` as a delimiter. However, in services such as `kinesis`, they can have both.

Current Behavior
`arn:aws:kinesis:1234567890:us-east-1:thing/foo:bar`  -> `{...'resourceId': ['thing/foo', 'bar']}`

New Behavior
`arn:aws:kinesis:1234567890:us-east-1:thing/foo:bar`  -> `{...'resourceId': ['thing', 'foo', 'bar']}`

### DIFF
--- a/botocore/endpoint_provider.py
+++ b/botocore/endpoint_provider.py
@@ -47,6 +47,7 @@ GET_ATTR_RE = re.compile(r"(\w+)\[(\d+)\]")
 VALID_HOST_LABEL_RE = re.compile(
     r"^(?!-)[a-zA-Z\d-]{1,63}(?<!-)$",
 )
+ARN_DELIMITER_RE = re.compile(r"/|:")
 CACHE_SIZE = 100
 ARN_PARSER = ArnParser()
 STRING_FORMATTER = Formatter()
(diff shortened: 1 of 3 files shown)


jonathan343 (approved):
Looks good to me!

botocore/endpoint_provider.py
```diff
@@ -236,8 +237,7 @@ def aws_parse_arn(self, value):
         arn_dict["accountId"] = arn_dict.pop("account")
 
         resource = arn_dict.pop("resource")
-        delimiter = ":" if ":" in resource else "/"
-        arn_dict["resourceId"] = resource.split(delimiter)
+        arn_dict["resourceId"] = re.split(ARN_DELIMITER_RE, resource)
```
jonemo, line 240: I'm not sure if there is a difference, but why not use split on the compiled object?

```suggestion
        arn_dict["resourceId"] = ARN_DELIMITER_RE.split(resource)
```

jonemo (commented):
As discussed elsewhere already, using the regex in this case is approximately 3x slower in my tests than the non-regex solution:

```python
"abc:def/ghi".replace(":", "/"). split("/")
['abc', 'def', 'ghi']
```

If you decide to stick with the regex implementation, a nit: `ARN_DELIMITER_RE` would be more precise if called `ARN_RESOURCE_DELIMITER_RE` because "ARN delimiter" usually refers to the `:` that separate the top-level parts of the ARN.

dlm6693:
@jonemo thanks for the feedback. I did some testing of my own and I do also see that the standard string splitting in Python is much faster than using regex. We needed to get this merged before today's release otherwise I would've waited for more feedback, but I'll put up a followup PR that we can merge later today or tomorrow.

PR #2836, opened 2022-12-12
Add Type field for query errors from JSON protocol

kyleknap:
By default, the query parser parses all fields from the error response and includes them in the parsed error dictionary. This included a `Type` key. However, JSON protocol based services never return a `Type` as part of the parsed error response. So for backwards compatibility purposes, the SDK needs to parse the `Type` key from the `x-amz-query-error` header and include it as part of the parsed error response when the header is found for JSON protocol based services.

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -706,6 +706,7 @@ class BaseJSONParser(ResponseParser):
         code = None
         if len(query_error_components) == 2 and query_error_components[0]:
             code = query_error_components[0]
+            error['Error']['Type'] = query_error_components[1]
         if code is None:
             code = body.get('__type', response_code and str(response_code))
         if code is not None:
--- a/tests/unit/test_parsers.py
+++ b/tests/unit/test_parsers.py
@@ -1104,6 +1104,7 @@ class TestParseErrorResponses(unittest.TestCase):
         self.assertEqual(
             parsed['Error']['Code'], 'AWS.SimpleQueueService.NonExistentQueue'
         )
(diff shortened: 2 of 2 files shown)


botocore/parsers.py
```diff
@@ -706,6 +706,7 @@ def _do_error_parse(self, response, shape):
         code = None
         if len(query_error_components) == 2 and query_error_components[0]:
             code = query_error_components[0]
+            error['Error']['Type'] = query_error_components[1]
```
ziyanli-amazon, line 709: nit: do you think we should also declare the error structure at line#686 as
```
error = {"Error": {"Message": '', "Code": '', "Type": ''}, "ResponseMetadata": {}}
```

kyleknap: No. In general for JSON protocols, errors returned do not have an equivalent `Type` field. So if the parsed error was initialized to always include a `Type` key, it will end up being an empty string for most services supported by the SDK.

Furthermore, it was a slight oversight that the parsed response for query protocol has an `Type` field in the first place (i.e. the SDK generally documents parsed error dictionaries has having a `Code` and `Message`). The shim of injecting the `Type` key is more for adding it back in case someone saw the field as part of the parsed response and added logic to utilize it. So I'm not sure if we want to expand its presence to other usages.

ziyanli-amazon: Well explained. thanks. The PR LGTM. Please ask the repo owner to give you a go.

jonemo (approved):
Good catch! 🦅 👀

PR #2839, opened 2022-12-19
Ignore response/request sections in docs if they're too large.

jonathan343:
## Summary
Services such as `quicksight` have recently become very large and contain several layers of nested shapes that result in doc pages which are basically unusable in some cases such as [this one](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/quicksight.html).

## Solution
The temporary solution to get these pages usable again is to ignore response/request sections that are too large, specifically here are the limits:
* **`Request/Response Syntax Examples:`** Limited to 1500 lines. If larger, the section will be replaced with a "too large" message and be provided a link to AWS API Documentation for the corresponding service/section.
* **`Descriptions/Params:`** Limited to 5000 lines. If larger, the section will be replaced with a "too large" message and be provided a link to AWS API Documentation for the corresponding service/section.

**Affected Service Pages:** # sections  (min_lines - max_lines)
* `ec2.html` - 3 sections (5,974 - 5,989)
* `lexv2-models.html` - 20 sections (1,778 - 18,487)
* `quicksight.html` - 18 sections (29,481 - 160,523)  # This was the worst one
* `sagemaker.html` - 4 sections (2,133 - 14,376)
*  `securityhub.html` - 13 sections (3979 -  23,728)
*  `wafv2.html` - 10 sections (2,036 - 17,672)

### DIFF
--- a/botocore/docs/bcdoc/restdoc.py
+++ b/botocore/docs/bcdoc/restdoc.py
@@ -16,11 +16,22 @@ from botocore.compat import OrderedDict
 from botocore.docs.bcdoc.docstringparser import DocStringParser
 from botocore.docs.bcdoc.style import ReSTStyle
 
+LARGE_SECTION_MESSAGE = (
+    '    ::\n\n        # This section is too large to render.'
+    '\n        # Please see the AWS API Documentation linked below.\n\n    '
+)
 LOG = logging.getLogger('bcdocs')
 SECTION_LINE_LIMITS = {
-    'example': 1500,
-    'description': 10000,
-    'request-params': 10000,
+    'request-example': 1500,
+    'response-example': 1500,
+    'description': 5000,
+    'request-params': 5000,
+}
+SECTION_NAME_MAPPING = {
+    'response-example': 'Response Syntax',
+    'description': 'Response Structure',
+    'request-example': 'Request Syntax',
+    'request-params': 'Parameters',
 }
 
 
@@ -195,7 +206,7 @@ class DocumentStructure(ReSTDocument):
         """Delete a section"""
         del self._structure[name]
 
-    def flush_structure(self):
+    def flush_structure(self, docs_link=None):
         """Flushes a doc structure to a ReSTructed string
 
         The document is flushed out in a DFS style where sections and their
@@ -210,11 +221,27 @@ class DocumentStructure(ReSTDocument):
                     self.style.link_target_definition(refname, link)
         value = self.getvalue()
         for name, section in self._structure.items():
-            value += section.flush_structure()
-        # Ignores response/request sections if the line number exceeds our limit.
-        line_count = value.decode('utf-8').count('\n')
+            # Checks is the AWS API Documentation link has been generated.
+            # If it has been generated, it gets passed as a the doc_link parameter.
+            if b'`AWS API Documentation <' in value:
+                start = value.find(b'`AWS API Documentation <')
+                end = value.find(b'>`_', start) + 3
+                docs_link = value[start:end] + b'\n\n'
+            value += section.flush_structure(docs_link)
+
+        # Replace response/request sections if the line number exceeds our limit.
+        # The section is replaced with a message linking to AWS API Documentation.
+        line_count = len(value.splitlines())
         line_limit = SECTION_LINE_LIMITS.get(self.name)
-        return b'' if line_limit and line_count > line_limit else value
+        if line_limit and line_count > line_limit:
+            print(f'Lines:{line_count}, Section:{self.name}')
+            value = (
+                f'\n\n    **{SECTION_NAME_MAPPING.get(self.name)}**'
+                f'\n{LARGE_SECTION_MESSAGE}'
+            ).encode()
+            if docs_link is not None:
+                value += docs_link
+        return value
 
     def getvalue(self):
         return ''.join(self._writes).encode('utf-8')
--- a/botocore/docs/bcdoc/restdoc.py
+++ b/botocore/docs/bcdoc/restdoc.py
@@ -234,7 +234,6 @@ class DocumentStructure(ReSTDocument):
         line_count = len(value.splitlines())
         line_limit = SECTION_LINE_LIMITS.get(self.name)
         if line_limit and line_count > line_limit:
-            print(f'Lines:{line_count}, Section:{self.name}')
             value = (
                 f'\n\n    **{SECTION_NAME_MAPPING.get(self.name)}**'
                 f'\n{LARGE_SECTION_MESSAGE}'
--- a/botocore/docs/bcdoc/restdoc.py
+++ b/botocore/docs/bcdoc/restdoc.py
@@ -17,6 +17,11 @@ from botocore.docs.bcdoc.docstringparser import DocStringParser
 from botocore.docs.bcdoc.style import ReSTStyle
 
 LOG = logging.getLogger('bcdocs')
+SECTION_LINE_LIMITS = {
+    'example': 1500,
+    'description': 10000,
+    'request-params': 10000,
+}
 
 
 class ReSTDocument:
@@ -206,7 +211,10 @@ class DocumentStructure(ReSTDocument):
         value = self.getvalue()
         for name, section in self._structure.items():
             value += section.flush_structure()
-        return value
+        # Ignores response/request sections if the line number exceeds our limit.
+        line_count = value.decode('utf-8').count('\n')
+        line_limit = SECTION_LINE_LIMITS.get(self.name)
+        return b'' if line_limit and line_count > line_limit else value
 
     def getvalue(self):
         return ''.join(self._writes).encode('utf-8')
--- a/botocore/docs/bcdoc/restdoc.py
+++ b/botocore/docs/bcdoc/restdoc.py
@@ -11,27 +11,38 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import logging
+import re
 
 from botocore.compat import OrderedDict
 from botocore.docs.bcdoc.docstringparser import DocStringParser
 from botocore.docs.bcdoc.style import ReSTStyle
 
-LARGE_SECTION_MESSAGE = (
-    '    ::\n\n        # This section is too large to render.'
-    '\n        # Please see the AWS API Documentation linked below.\n\n    '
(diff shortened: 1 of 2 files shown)


botocore/docs/bcdoc/restdoc.py
```diff
@@ -206,7 +211,10 @@ def flush_structure(self):
         value = self.getvalue()
         for name, section in self._structure.items():
             value += section.flush_structure()
-        return value
+        # Ignores response/request sections if the line number exceeds our limit.
+        line_count = value.decode('utf-8').count('\n')
```
dlm6693, line 215: Would `len(value.decode('utf-8').splitlines())` be a more accurate measure since that method splits on other newline characters besides `\n`?

dlm6693 (approved):
LGTM. Just had an alternative suggestion for calculating total lines, but not a blocker.

dlm6693:
@jonathan343 I'm assuming you tested these changes by rendering the `quicksight` page and maybe a few other services just to make sure they're readable?

botocore/docs/bcdoc/restdoc.py
jonathan343: After running the doc generation again with this change it only added 1 extra line to the count for each section. I'll add this change to future proof incase we ever have other newline characters to catch. Thanks!

botocore/docs/bcdoc/restdoc.py
```diff
@@ -16,7 +16,23 @@
 from botocore.docs.bcdoc.docstringparser import DocStringParser
 from botocore.docs.bcdoc.style import ReSTStyle
 
+LARGE_SECTION_MESSAGE = (
+    '    ::\n\n        # This section is too large to render.'
+    '\n        # Please see the AWS API Documentation linked below.\n\n    '
```
nateprewitt, line 21: nit; if this still renders correctly, it may be nicer formatting in the code. Also do we need the trailing spaces for something?

```suggestion
    '    ::\n\n'
    '        # This section is too large to render.\n'
    '        # Please see the AWS API Documentation linked below.\n\n'
```

botocore/docs/bcdoc/restdoc.py
```diff
@@ -205,7 +221,25 @@ def flush_structure(self):
...
+
+        # Replace response/request sections if the line number exceeds our limit.
+        # The section is replaced with a message linking to AWS API Documentation.
+        line_count = len(value.splitlines())
+        line_limit = SECTION_LINE_LIMITS.get(self.name)
+        if line_limit and line_count > line_limit:
+            value = (
+                f'\n\n    **{SECTION_NAME_MAPPING.get(self.name)}**'
+                f'\n{LARGE_SECTION_MESSAGE}'
+            ).encode()
```
nateprewitt, line 240: Let's make this a template and we can do something like:

```suggestion
            section_name = SECTION_NAME_MAPPING.get(self.name)
            value = TEMPLATE.format(section_name)
```

There's also a question of can we ever hit this with a section name that doesn't exist in the mapping? Do we just write `None`?

botocore/docs/bcdoc/restdoc.py
```diff
@@ -205,7 +221,25 @@ def flush_structure(self):
...
         value = self.getvalue()
         for name, section in self._structure.items():
-            value += section.flush_structure()
+            # Checks is the AWS API Documentation link has been generated.
+            # If it has been generated, it gets passed as a the doc_link parameter.
+            if b'`AWS API Documentation <' in value:
+                start = value.find(b'`AWS API Documentation <')
+                end = value.find(b'>`_', start) + 3
+                docs_link = value[start:end] + b'\n\n'
+            value += section.flush_structure(docs_link)
```
nateprewitt, line 230: What is the scope of `_structure` here, is it only a single operation or the whole doc model? The case I'd be worried about is this code sets `doc_link` as soon as `b'`AWS API Documentation <'` is encountered. That means every future item in the for loop will always have this value set, right? Unless we override it again, or set it back to None, are we potentially writing erroneous links?

botocore/docs/bcdoc/restdoc.py
```diff
@@ -205,7 +221,25 @@ def flush_structure(self):
                     self.style.link_target_definition(refname, link)
         value = self.getvalue()
         for name, section in self._structure.items():
-            value += section.flush_structure()
+            # Checks is the AWS API Documentation link has been generated.
+            # If it has been generated, it gets passed as a the doc_link parameter.
+            if b'`AWS API Documentation <' in value:
+                start = value.find(b'`AWS API Documentation <')
```
nateprewitt, line 227: Another general comment, I'm wondering if doing something with a [regex substitution](https://docs.python.org/3/library/re.html#re.sub) may be safer here?

botocore/docs/bcdoc/restdoc.py
nateprewitt: Generating some test samples, it doesn't appear this is happening, so the scoping for `_structure` may be limited enough to prevent this.

botocore/docs/bcdoc/restdoc.py
jonathan343: Thanks! I'll use this to have nicer formatting. The trailing spaces are needed so the link thats added later is indented the same amount as this code block.

dlm6693: Personal preference but I would recommend using triple quotes here so you can write the string literally without the new line characters.  You'll just probably have to use `textwrap.dedent` to remove any unwanted indentations.

botocore/docs/bcdoc/restdoc.py
dlm6693: +1

botocore/docs/bcdoc/restdoc.py
dlm6693: Yeah agree with Nate here. If there's an easy way to inject the link shortly after you replace the text with `LARGE_SECTION_MESSAGE` that would be ideal

botocore/docs/bcdoc/restdoc.py
dlm6693: Not currently, but it does leave the possibility if `SECTION_LINE_LIMITS` and `SECTION_NAME_MAPPING` don't have exactly matching keys for whatever reason. Maybe combining those into one dictionary would be a bit better. Something like:

```python
line_count = len(value.splitlines())
line_limit_config = SECTION_LINE_LIMIT_CONFIG.get(self.name)
if line_limit_config and line_count > line_limit_config['limit']:
    value = (
        f'\n\n.   **{line_limit_config['name']}**\n{LARGE_SECTION_MESSAGE}'
)

File: botocore/docs/method.py [hunk 6]
```diff
@@ -217,7 +217,7 @@ def document_model_driven_method(
         method_intro_section.writeln('')
 
     # Add the example section.
-    example_section = section.add_new_section('example')
+    example_section = section.add_new_section('request-example')
```
dlm6693, line 220: Just wondering why did this and response one below it need to change?

jonathan343: There was no easy way to determine which section was which in order to determine the section title. They were both called `example`. With the updated names i can map `request-example` to `Request Syntax`.

botocore/docs/bcdoc/restdoc.py
jonathan343: So whats happening here is we're recursively going through each service's sub-structures in a DFS style. An example path for quicksight's `cancel_ingestion` is `['quicksight', 'client-api', 'methods', 'cancel_ingestion']`. The `cancel_ingestion` section will then have the following available sub-sections sections `method-intro`, `request-example`, `request-params`, `return`, & `exceptions`. The documentation link is found in `method-intro` and then passed into the other sub-sections (happens in the for loop). Nate is right that a documentation link can wrongfully be passed into a subsequent method's traversal. I'll add a condition that resets the `docs_link` to `None` when `len(self.path) == 4` since this indicates a new method.

botocore/docs/bcdoc/restdoc.py
jonathan343: I updated this to use `re.search(<pattern>, <text>)`. See new changes for details.

botocore/docs/bcdoc/restdoc.py
```diff
@@ -11,27 +11,38 @@
...
 from botocore.docs.bcdoc.docstringparser import DocStringParser
 from botocore.docs.bcdoc.style import ReSTStyle
 
-LARGE_SECTION_MESSAGE = (
-    '    ::\n\n        # This section is too large to render.'
-    '\n        # Please see the AWS API Documentation linked below.\n\n    '
+DEFAULT_AWS_DOCS_LINK = 'https://docs.aws.amazon.com/index.html'
+DOCUMENTATION_LINK_REGEX = (
+    r'`AWS API Documentation '
+    r'<https://docs.aws.amazon.com/goto/WebAPI/[a-z0-9-.]*/[a-zA-Z]*>`_'
```
dlm6693, line 23: could you wrap this in `re.compile`?

botocore/docs/bcdoc/restdoc.py
```diff
@@ -214,32 +225,38 @@ def flush_structure(self, docs_link=None):
...
+            docs_link = None
         value = self.getvalue()
         for name, section in self._structure.items():
             # Checks is the AWS API Documentation link has been generated.
             # If it has been generated, it gets passed as a the doc_link parameter.
-            if b'`AWS API Documentation <' in value:
-                start = value.find(b'`AWS API Documentation <')
-                end = value.find(b'>`_', start) + 3
-                docs_link = value[start:end] + b'\n\n'
+            match = re.search(DOCUMENTATION_LINK_REGEX, value.decode())
```
dlm6693, line 241: After updating the regex string with `re.compile` to a regex object, you can update this to `DOCUMENTATION_LINK_REGEX.search(value.decode())`

botocore/docs/bcdoc/restdoc.py
```diff
@@ -214,32 +225,38 @@ def flush_structure(self, docs_link=None):
...
+            docs_link = (
+                (match.group(0) + '\n\n').encode() if match else docs_link
```
dlm6693, line 243: nit: use an f-string here instead: `f'{match.group(0)}\n\n'`

dlm6693 (changes requested):
I think we're almost there. Just a couple of minor things and this should be good.

dlm6693 (approved):
:shipit:

nateprewitt (approved):

PR #2847, opened 2023-01-11
Avoid BytesWarning on logging responses

ikonst:
`type(response['body'])` is `bytes`. Formatting it as `%s` results in a `BytesWarning` on Python 3.


nateprewitt (approved):
Thanks @ikonst! To clarify where you're hitting this, you're running Python with `-b` set and logging enabled? I wasn't able to produce this with Python normal warning flagging in any of our supported versions.

PR #2851, opened 2023-01-19
Plumb profile_name into SSOTokenProvider

hssyoo:
Fixes https://github.com/aws/aws-cli/issues/7496

When resolving credentials, `profile_name` should be plumbed into `SSOTokenProvider` so that if they need to be resolved recursively, the newly set `profile_name` gets used to load the SSO config rather than using the profile name defined in the session.

Example of a valid config that is currently failing to resolve:
```
[profile test]
role_arn = arn:aws:iam:XXXXXXXXXXXX:role/testrole-1
source_profile = AdministratorAccess-XXXXXXXXXXXX
region = us-west-2

[profile AdministratorAccess-XXXXXXXXXXXX]
sso_session = dev
sso_account_id = XXXXXXXXXXXX
sso_role_name = AdministratorAccess
region = us-west-2
output = json

[sso-session dev]
sso_start_url = https://d-XXXXXXXXX.awsapps.com/start
sso_region = us-west-2
sso_registration_scopes = sso:account:access
```

### DIFF
--- a/botocore/tokens.py
+++ b/botocore/tokens.py
@@ -183,7 +183,9 @@ class SSOTokenProvider:
     _GRANT_TYPE = "refresh_token"
     DEFAULT_CACHE_CLS = JSONFileCache
 
-    def __init__(self, session, cache=None, time_fetcher=_utc_now):
+    def __init__(
+        self, session, cache=None, time_fetcher=_utc_now, profile_name=None
+    ):
         self._session = session
         if cache is None:
             cache = self.DEFAULT_CACHE_CLS(
@@ -193,15 +195,17 @@ class SSOTokenProvider:
         self._now = time_fetcher
         self._cache = cache
         self._token_loader = SSOTokenLoader(cache=self._cache)
+        self._profile_name = (
+            profile_name
+            or self._session.get_config_variable("profile")
+            or 'default'
+        )
 
     def _load_sso_config(self):
         loaded_config = self._session.full_config
         profiles = loaded_config.get("profiles", {})
         sso_sessions = loaded_config.get("sso_sessions", {})
-        profile_name = self._session.get_config_variable("profile")
-        if not profile_name:
-            profile_name = "default"
-        profile_config = profiles.get(profile_name, {})
+        profile_config = profiles.get(self._profile_name, {})
 
         if "sso_session" not in profile_config:
             return
@@ -211,7 +215,7 @@ class SSOTokenProvider:
 
         if not sso_config:
             error_msg = (
-                f'The profile "{profile_name}" is configured to use the SSO '
+                f'The profile "{self._profile_name}" is configured to use the SSO '
                 f'token provider but the "{sso_session_name}" sso_session '
                 f"configuration does not exist."
             )
@@ -224,7 +228,7 @@ class SSOTokenProvider:
 
         if missing_configs:
             error_msg = (
-                f'The profile "{profile_name}" is configured to use the SSO '
+                f'The profile "{self._profile_name}" is configured to use the SSO '
                 f"token provider but is missing the following configuration: "
                 f"{missing_configs}."
             )
--- a/tests/functional/test_credentials.py
+++ b/tests/functional/test_credentials.py
@@ -252,9 +252,10 @@ class TestAssumeRole(BaseAssumeRoleTest):
         mock_instance.CANONICAL_NAME = provider_cls.CANONICAL_NAME
         return mock_instance
 
-    def create_session(self, profile=None):
+    def create_session(self, profile=None, sso_token_cache=None):
         session = StubbedSession(profile=profile)
-
+        if not sso_token_cache:
+            sso_token_cache = JSONFileCache(self.tempdir)
         # We have to set bogus credentials here or otherwise we'll trigger
         # an early credential chain resolution.
         sts = session.create_client(
@@ -277,7 +278,7 @@ class TestAssumeRole(BaseAssumeRoleTest):
             ),
             profile_provider_builder=ProfileProviderBuilder(
                 session,
-                sso_token_cache=JSONFileCache(self.tempdir),
+                sso_token_cache=sso_token_cache,
             ),
         )
         stubber = session.stub('sts')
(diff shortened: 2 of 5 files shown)


hssyoo:
No walrus for me 😞

botocore/tokens.py
```diff
@@ -324,3 +327,25 @@ def load_token(self):
...
+            profile_name = source_profile.get('source_profile')
+            source_profile = profiles.get(profile_name, {})
+            if profile_name in visited_profiles:
+                valid = self._has_static_credentials(source_profile)
+                if valid:
+                    break
+                raise InfiniteLoopConfigError(
+                    source_profile=profile_name,
+                    visited_profiles=visited_profiles,
+                )
```
kyleknap, line 345: Playing around with a bit, it does not seem this error will get called because the credential provider will be resolved before the auth token is resolved. Instead of raising an error here, it probably makes sense to return `None` and delegate any configuration errors related to the assume role credential provider class going forward instead of having it be done in this class which really only cares about `source_profile` so it works with the assume role provider

botocore/tokens.py
```diff
@@ -324,3 +327,25 @@ def load_token(self):
...
+            visited_profiles.add(profile_name)
+        return source_profile
+
+    def _has_static_credentials(self, profile_config):
```
kyleknap, line 349: We should also be able to remove this logic if we do not raise errors for infinite loops in this auth token class.

botocore/tokens.py
```diff
@@ -324,3 +327,25 @@ def load_token(self):
         return DeferredRefreshableToken(
             self.METHOD, self._refresher, time_fetcher=self._now
         )
+
+    def _resolve_source_profile(self, profiles, initial_profile_config):
+        profile_name = initial_profile_config.get('source_profile', None)
```
kyleknap, line 332: When recursing through source profiles in addition to checking that the profile has a `source_profile`, it probably makes sense to check that a `role_arn` is set in the profile as well. The reason being that it makes it more strict when it comes to deciding whether to recurse through profiles to find an SSO configuration and limit this behavior only from the context of assume role configurations.

Specifically with the way we have it now, it is possible to chain profiles together to resolve an sso session solely on `source_profile` when retrieving and using the SSO token directly for authentication, which is not an intended interface:
```
[profile test]
source_profile = AdministratorAccess-XXXXXXXXXXXX
region = us-west-2

[profile AdministratorAccess-XXXXXXXXXXXX]
sso_session = dev
region = us-west-2

[sso-session dev]
sso_start_url = https://d-XXXXXXXXX.awsapps.com/start
sso_region = us-west-2
sso_registration_scopes = sso:account:access
```
Specifically, this configuration will allow some commands to be able to specify `--profile test` assuming the command uses bearer authentication, which is not a usage pattern we intend to support (instead the `test` profile would have to specify a `sso_session`)

kyleknap (commented):
Looks good. I just had some smaller comments

botocore/tokens.py
kyleknap: Thinking about this more, it may even make sense to make the behavior of recursing through profiles to locate the sso session name something that is opt in at the initialization of the `SSOTokenProvider` class. We can then toggle it on when instantiating the `SSOTokenProvider` as part of the credential provider chain, which would negate any potential side effects in the default auth token provider chain. Thoughts? I think the main point is that we should try to isolate this recursing logic to when we are using it as part of the assume role provider.

botocore/tokens.py
```diff
@@ -183,7 +183,9 @@ class SSOTokenProvider:
     _GRANT_TYPE = "refresh_token"
     DEFAULT_CACHE_CLS = JSONFileCache
 
-    def __init__(self, session, cache=None, time_fetcher=_utc_now):
+    def __init__(
+        self, session, cache=None, profile_name=None, time_fetcher=_utc_now
```
kyleknap, line 187: Whenever adding new arguments to a signature, we should add it to the end. Otherwise, it would break backwards compatibility if it was called using purely positional arguments. While this class is not considered a public interface, it is still good practice to not break something unless there is a strong reason to.

botocore/tokens.py
```diff
@@ -248,6 +249,14 @@ def _client(self):
         )
         return self._session.create_client("sso-oidc", config=config)
 
+    @CachedProperty
+    def profile_name(self):
```
kyleknap, line 253: Let's avoid making this a public property on the class. In general, this token class needs to follow the interface of the base token class so that we can uphold the property that all token classes has a specific interface, which in the end simplifies downstream usages. Instead, I'd recommend just computing the profile name and setting it to `self._profile_name` as part of the initializer.

tests/functional/test_credentials.py
```diff
@@ -630,6 +631,69 @@ def test_sso_source_profile(self):
...
+            }
+        }
+        sso_stubber.add_response('get_role_credentials', sso_role_response)
+
+        expected_creds = self.create_random_credentials()
+        assume_role_response = self.create_assume_role_response(expected_creds)
+        sts_stubber.add_response('assume_role', assume_role_response)
+
+        actual_creds = session.get_credentials()
+        # breakpoint()
```
kyleknap, line 683: Let's remove this comment

tests/unit/test_tokens.py
```diff
@@ -99,15 +99,15 @@ def parametrize(cases):
...
+    mock_session.get_config_variable.return_value = "default"
     mock_session.full_config = config
     return mock_session
 
 
 @parametrize(sso_provider_resolution_cases)
 def test_sso_token_provider_resolution(test_case):
     mock_session = _create_mock_session(test_case["config"])
-    resolver = SSOTokenProvider(mock_session)
+    resolver = SSOTokenProvider(mock_session, profile_name='test')
```
kyleknap, line 110: Let's avoid updating existing test cases as they are covering that we can use the profile name from the session. Instead, let's get a new test case that is explicitly checking that we can override the profile name from what is on the session

kyleknap (commented):
Nice! This is looking much simpler. I just had some comments. It would also be great to check into why the tests on Windows are failing.

kyleknap:
We should also make sure there is a changelog entry for this. It looks like the previous one may have been wiped out

nateprewitt (approved):
Outside of Kyle's comment, I think this looks good. Once that's resolved feel free to merge.

kyleknap (approved):
Looks good! :ship:

PR #2854, opened 2023-01-23
route53: Marker is listed as required but is actually optional

mbacarella:
Despite being listed as required, the `Marker` field is often missing from Route53 responses.

E.g.
```xml
<?xml version="1.0"?>
<ListHostedZonesResponse xmlns="https://route53.amazonaws.com/doc/2013-04-01/">
  <HostedZones/>
  <IsTruncated>false</IsTruncated>
  <MaxItems>100</MaxItems>
</ListHostedZonesResponse>
```

This PR fixes it so that API generators that enforce "required" don't fail on these responses.

### DIFF
--- a/botocore/data/route53/2013-04-01/service-2.json
+++ b/botocore/data/route53/2013-04-01/service-2.json
@@ -4240,7 +4240,6 @@
       "type":"structure",
       "required":[
         "HealthChecks",
-        "Marker",
         "IsTruncated",
         "MaxItems"
       ],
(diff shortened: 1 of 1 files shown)


mbacarella:
Is there any way I can help get this merged?

tim-finnigan:
Hi @mbacarella thanks for reporting this issue. These models are owned by service teams and shared across SDKs, so we'll need to escalate this issue to the Route53 for further action. I created this issue for tracking going forward: https://github.com/aws/aws-sdk/issues/495. I'll reach out to the Route53 team and update that issue when we have more information. I'll go ahead and close this PR since the changes must be pulled in upstream rather than made directly in botocore, but thanks again for creating this and letting us know.

PR #2855, opened 2023-01-24
Docs: collapse HTML whitespace for correct RST rendering

jonemo:
HTML in docstrings often contains extraneous white-space around tags, for readability. Browsers would collapse this white-space before rendering. However, the botocore docs are generated by converting the HTML from upstream into RST. If the whitespace is not collapsed before conversion to RST where white-space is part of the syntax, for example for indentation, it can result in incorrect output.

This screenshot from [our docs page for KMS](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/kms.html) is an example of where this currently goes wrong. The text starting with "cross-account use" is not supposed to be part of the note.

Screen Shot 2023-01-11 at 2 12 26 PM

The HTML source for the example in the screenshot is [here](https://github.com/boto/botocore/blob/c2bc52dcf9d58d9699983b18d80f244511f07d36/botocore/data/kms/2014-11-01/service-2.json#L140).

This PR adds logic that mimics the whitespace collapsing logic implemented in browser, as described [here on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace) and [here by W3C](https://www.w3.org/TR/css-text-3/#white-space-rules).

After this change, the example screenshot above looks like this:

Screen Shot 2023-01-11 at 2 12 49 PM

Previously, we had a few instances of narrowly scoped code to remove extraneous whitespace, for example for list items or for hyperlinks `<code>` tags introduced recently in #2809. The new logic supersedes these.


botocore/docs/bcdoc/docstringparser.py
```diff
@@ -11,10 +11,12 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 from html.parser import HTMLParser
+from itertools import zip_longest
 
 PRIORITY_PARENT_TAGS = ('code', 'a')
 OMIT_NESTED_TAGS = ('span', 'i', 'code', 'a')
 OMIT_SELF_TAGS = ('i', 'b')
+HTML_BLOCK_DISPLAY_TAGS = ['p', 'note', 'ul', 'li']
```
jonathan343, line 19: nit; is there a reason for making this a list and not a tuple like the others? Do we expect this to be mutated?

jonathan343 (commented):
These changes and test coverage look great! 

Two main things to look into:
1. There are two tests failing in `tests/unit/test_client.py` which are probably quick fixes related to whitespace.
2. I generated html docs for a few random services (`kms`, `acm`, & `iam`) before and after the changes and compared them. As expected, most changes were related to issues like the one you mentioned in the  overview. I did notice while quickly scanning that there were some changes where `&nbsp;` are removed. The specific ones I saw were in IAM and didn't seem too important but I wonder if there are cases where we want to preserve these.

jonemo:
With #2866 and #2888 merged, it's time to carry this over the finish line. I rebased on `develop` and took care of all the failing tests. Where "take care of" means to carefully review that none of the required changes to assertions are whitespace-only and make no functional difference.

The original issue of additional getting pulled into indented notes still gets fixed. Here is a [Meld](https://meldmerge.org/) screenshot of one example:

Screenshot 2023-03-22 at 3 15 24 PM

... and visual before:

kms-client-create_key-before

... and after:

kms-client-create_key-after

All other changes I saw in the visual diff are removal of incorrect whitespace around inline tags. This page has a good assortment (annotation by [Uprightdiff](https://www.mediawiki.org/wiki/Uprightdiff)):

emr-client-put_auto_scaling_policy-upright

jonathan343:
I think this is ready to merge after boto3 tests are verified to be unaffected from the change.

jonathan343 (approved):
LGTM now that boto3 tests were fixed!🎉

PR #2857, opened 2023-01-29
Replace OrderedDict with dict

serhii73:


nateprewitt:
Hi @serhii73, thanks for your contribution! While this will work from a functional level, this is unfortunately breaking for anyone relying on the compat import. It also has potential flow implications for users extending the OPTION_DEFAULTS in custom configs. I don't think we'll be able to accept this at time.

PR #2859, opened 2023-01-30
Append `aws-chunked` to `Content-Encoding` header if present

dlm6693:
Primarily in various S3 operations, users may provide a `ContentEncoding` parameter. This tells the service how the content that is being sent/requested is encoded. An example of this would be `gzip`. However, for streaming AWS operations (i.e. ones that have trailers) where a `ChecksumAlgorithm` parameter is provided, the SDK is currently overriding any user input to `aws-chunked`. Instead this special content encoding should be appended to the user input.

### DIFF
--- a/tests/functional/test_s3.py
+++ b/tests/functional/test_s3.py
@@ -2141,6 +2141,36 @@ def test_checksums_included_in_expected_operations(
         assert "Content-MD5" in stub.requests[-1].headers
 
 
+@pytest.mark.parametrize(
+    "input_param, expected_header",
+    [("foo", "foo,aws-chunked"), (None, "aws-chunked")],
+)
+def test_checksum_content_encoding(input_param, expected_header):
+    environ = {}
+    op_kwargs = {
+        "Bucket": "mybucket",
+        "Key": "mykey",
+        "Body": b"foo",
+        "ChecksumAlgorithm": "sha256",
+    }
+    if input_param is not None:
+        op_kwargs["ContentEncoding"] = input_param
+    with mock.patch("os.environ", environ):
+        environ["AWS_ACCESS_KEY_ID"] = "access_key"
+        environ["AWS_SECRET_ACCESS_KEY"] = "secret_key"
+        environ["AWS_CONFIG_FILE"] = "no-exist-foo"
+        session = create_session()
+        session.config_filename = "no-exist-foo"
+        s3 = session.create_client("s3")
+        with ClientHTTPStubber(s3) as http_stubber:
+            http_stubber.add_response()
+            s3.put_object(**op_kwargs)
+            assert (
+                http_stubber.requests[-1].headers["Content-Encoding"].decode()
+                == expected_header
+            )
+
+
 def _s3_addressing_test_cases():
     # The default behavior for sigv2. DNS compatible buckets
     yield dict(
--- a/tests/functional/test_s3.py
+++ b/tests/functional/test_s3.py
@@ -2154,13 +2154,7 @@ def test_checksum_content_encoding(content_encoding, expected_header):
     }
     if content_encoding is not None:
         op_kwargs["ContentEncoding"] = content_encoding
-    s3 = _create_s3_client(
-        region="us-west-2",
-        is_secure=True,
-        s3_config={},
-        signature_version="s3v4",
-        endpoint_url=None,
-    )
+    s3 = _create_s3_client()
     with ClientHTTPStubber(s3) as http_stubber:
         http_stubber.add_response()
         s3.put_object(**op_kwargs)
--- a/tests/functional/test_s3.py
+++ b/tests/functional/test_s3.py
@@ -2142,33 +2142,32 @@ def test_checksums_included_in_expected_operations(
 
 
 @pytest.mark.parametrize(
-    "input_param, expected_header",
+    "content_encoding, expected_header",
     [("foo", "foo,aws-chunked"), (None, "aws-chunked")],
 )
-def test_checksum_content_encoding(input_param, expected_header):
-    environ = {}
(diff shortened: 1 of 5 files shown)


tests/unit/test_httpchecksum.py
```diff
@@ -301,6 +301,34 @@ def test_apply_request_checksum_flex_header_trailer_explicit_digest(self):
         # The body should not have been wrapped
         self.assertIsInstance(request["body"], bytes)
 
+    def test_apply_request_checksum_content_encoding_preset(self):
```
nateprewitt, line 304: We may want to either do a one-off manual test or add a functional test to ensure operations that bind input to the `Content-Encoding` header aren't affected when a checksum is used. S3 PutObject would be the best place to verify that.

tests/unit/test_httpchecksum.py
```diff
@@ -301,6 +301,34 @@ def test_apply_request_checksum_flex_header_trailer_explicit_digest(self):
...
+        request["context"]["checksum"] = {
+            "request_algorithm": {
+                "in": "trailer",
+                "algorithm": "crc32",
+                "name": "x-amz-checksum-crc32",
+            }
+        }
+        request["headers"]["Content-Encoding"] = "foo"
+        apply_request_checksum(request)
+        # The content encoding should not have been modified
```
nateprewitt, line 315: This may be more accurate. We are modifying it, but we don't want to overwrite it.
```suggestion
        # The content encoding should only have been appended
```

tests/unit/test_httpchecksum.py
dlm6693: Oh sorry I didn't mention in the PR that I did some manual testing making actual `put_object` requests to a test bucket. I did think this may be worthy of a new functional/integration test but wasn't sure.

Checking a stubbed request context/headers should be simple enough to add. I'll update in the morning. I'll also add a changelog unless you don't think it necessary?

tests/functional/test_s3.py
```diff
@@ -2148,6 +2148,36 @@ def test_checksums_included_in_expected_operations(
...
+        session = create_session()
+        session.config_filename = "no-exist-foo"
+        s3 = session.create_client("s3")
+        with ClientHTTPStubber(s3) as http_stubber:
+            http_stubber.add_response()
+            s3.put_object(**op_kwargs)
+            assert (
+                http_stubber.requests[-1].headers["Content-Encoding"].decode()
+                == expected_header
+            )
```
nateprewitt, line 2178: We can probably clean this up by using existing functions instead of rewriting the client creation function.
```suggestion
    op_kwargs = {
        "Bucket": "mybucket",
        "Key": "mykey",
        "Body": b"foo",
        "ChecksumAlgorithm": "sha256",
    }
    if input_param is not None:
        op_kwargs["ContentEncoding"] = content_encoding
    s3 = _create_s3_client(
        region='us-west-2', is_secure=True, s3_config={}, signature_version='s3v4'
    )
    with ClientHTTPStubber(s3) as http_stubber:
        http_stubber.add_response()
        s3.put_object(**op_kwargs)
        request_headers =  http_stubber.requests[-1].headers
        assert request_headers["Content-Encoding"] == expected_header
```

tests/functional/test_s3.py
```diff
@@ -2148,6 +2148,36 @@ def test_checksums_included_in_expected_operations(
             assert "Content-MD5" in stub.requests[-1].headers
 
 
+@pytest.mark.parametrize(
+    "input_param, expected_header",
```
nateprewitt, line 2152:
```suggestion
    "content_encoding, expected_header",
```

tests/functional/test_s3.py
dlm6693: I thought about this but opted for consistency with the other checksum tests right above this especially since the test has nothing to with a secure request, custom config or signature version. I don't feel strongly about this though.

nateprewitt: Yeah, I think the existing checksum test case is the outlier rather than the rule. All other tests use `BaseS3OperationTest` or `_create_s3_client`. We should avoid continuing a bad pattern.

The args aren't super important but we can add sensible defaults to the `_create_s3_function` if needed like we're doing here. I'd say save that for a separate PR though to avoid conflating the two.

dlm6693: Yeah totally I agree.

dlm6693:
@nateprewitt I'll wait until #2860 is merged, then I'll update the functional test to rely on the defaults in the client constructor.

tests/functional/test_s3.py
```diff
@@ -2141,6 +2141,29 @@ def test_checksums_included_in_expected_operations(
...
+    if content_encoding is not None:
+        op_kwargs["ContentEncoding"] = content_encoding
+    s3 = _create_s3_client()
+    with ClientHTTPStubber(s3) as http_stubber:
+        http_stubber.add_response()
+        s3.put_object(**op_kwargs)
+        assert (
+            http_stubber.requests[-1].headers["Content-Encoding"].decode()
+            == expected_header
+        )
```
nateprewitt, line 2164: We probably still want to clean this up. It's going to be difficult to understand the failure because we've put multiple failure cases onto a single line.

```suggestion
        request_headers = http_stubber.requests[-1].headers
        assert request_headers["Content-Encoding"] == expected_header
```

nateprewitt: And just in case it's missed, we need the expected_header to be bytes.

dlm6693: Ok I'll shoot that up now.

nateprewitt (approved):
:shipit: Looks good once the tests wrap

PR #2880, opened 2023-03-06
Sphinx domain fix

jonathan343:
## Overview
We previously had all class and method documentation on the same service page, however, we implemented nested docs to decrease page size. In doing so, we broke the way sphinx handles cross-references since methods on individual pages (not in the `py:class` body) should have the full class name attached as explained [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#:~:text=Methods%20and%20attributes%20belonging%20to%20the%20class%20should%20be%20placed%20in%20this%20directive%E2%80%99s%20body.%20If%20they%20are%20placed%20outside%2C%20the%20supplied%20name%20should%20contain%20the%20class%20name%20so%20that%20cross%2Dreferences%20still%20work.).

## Examples
Before Nested Docs:
```
# Page: <service>.html
.. py:class:: Foo
   .. py:method:: bar()
```

Current:
```
# Page: <service>.html
.. py:class:: Foo.Client

# Page: <method>.html
.. py:method:: bar()
```

New:
```
# Page: <service>.html
.. py:class:: Foo.Client

# Page: <method>.html
.. py:method:: Foo.Client.bar()
```

### DIFF
--- a/botocore/docs/client.py
+++ b/botocore/docs/client.py
@@ -148,26 +148,17 @@ class ClientDocumenter:
         if self._is_custom_method(method_name):
             self._add_custom_method(
                 method_section,
-                method_name,
+                full_method_name,
                 method,
-                full_method_name=full_method_name,
             )
         else:
-            self._add_model_driven_method(
-                method_section, method_name, full_method_name=full_method_name
-            )
+            self._add_model_driven_method(method_section, full_method_name)
 
     def _is_custom_method(self, method_name):
         return method_name not in self._client.meta.method_to_api_mapping
 
-    def _add_custom_method(
-        self, section, method_name, method, full_method_name=None
-    ):
-        if full_method_name is None:
-            full_method_name = method_name
-        document_custom_method(
-            section, full_method_name, method, full_method_name=None
-        )
+    def _add_custom_method(self, section, method_name, method):
+        document_custom_method(section, method_name, method)
 
     def _add_method_exceptions_list(self, section, operation_model):
         error_section = section.add_new_section('exceptions')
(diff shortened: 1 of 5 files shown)


nateprewitt (approved):

tests/unit/docs/test_client.py
```diff
@@ -51,26 +51,26 @@ def test_document_client(self):
             ]
         )
         self.assert_contains_lines_in_order(
-            ['.. py:method:: can_paginate(operation_name)'],
+            ['.. py:method:: MyService.Client.can_paginate(operation_name)'],
```
nateprewitt, line 54: This looks a little strange to me without seeing the context around it. Which page would we be generating methods like this on?

botocore/docs/client.py
```diff
@@ -143,9 +143,17 @@ def _add_client_methods(self, client_methods):
 
     def _add_client_method(self, section, method_name, method):
         section.add_title_section(method_name)
-        method_section = section.add_new_section(method_name)
+        full_method_name = f'{self._client_class_name}.Client.{method_name}'
+        method_section = section.add_new_section(
+            method_name,
+            context={'full_method_name': full_method_name},
```
nateprewitt, line 149: I'm wondering if we pass something more flexible like "qualifier" that can be joined with the method name as needed? I'm not sure how much that expands the PR, just a thought to avoid the redundant name handling.

jonathan343: Added the suggested changes in a new commit.

tests/functional/docs/__init__.py
```diff
@@ -76,9 +77,11 @@ def get_title_section_for(self, service_name):
 
     def get_method_document_block(self, operation_name, contents):
         contents = contents.decode('utf-8')
-        start_method_document = '.. py:method:: %s(' % operation_name
-        start_index = contents.find(start_method_document)
-        self.assertNotEqual(start_index, -1, 'Method is not found in contents')
+        regex = rf'.. py:method:: ([a-zA-Z0-9]*\.)*{operation_name}\('
+        match = re.search(regex, contents)
+        self.assertNotEqual(match, None, 'Method is not found in contents')
```
jonemo, line 82: Nit:

```suggestion
        self.assertIsNotNone(match, 'Method is not found in contents')
```

jonemo (approved):

potiuk:
:eyes:

PR #2881, opened 2023-03-08
get_available_regions: try service name first, then endpoint prefix

jonemo:
The `botocore.sessions.Session.get_available_regions` method is already documented as not always returning the complete set of regions where a service is available (at least [in the boto3 docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.get_available_regions), botocore's docs don't cover the `session` module). This PR fixes the opposite problem: In some rare situations the method returns too many regions.

Consider the case of the `neptune` service: [It uses the endpoint prefix `rds`](https://github.com/boto/botocore/blob/develop/botocore/data/neptune/2014-10-31/service-2.json#L5). As a result, `get_available_regions('neptune')` will return the available regions for RDS, not Neptune. For example, `me-central-1` is listed as available region for Neptune when in fact it is not, as per [Endpoints and Quotas docs for the Neptune service](https://docs.aws.amazon.com/general/latest/gr/neptune.html).

Any service that shares an endpoint prefix with another service would be affected by this. In a quick search I found `neptune` and `docdb` as sharing `rds`. There may be other examples.

This PR solves this by making `get_available_regions` try to use the given service name first, then fall back to looking up the endpoint prefix. There was a test case specifically asserting against doing this which I changed.

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -1081,18 +1081,27 @@ class Session:
         :return: Returns a list of endpoint names (e.g., ["us-east-1"]).
         """
         resolver = self._get_internal_component('endpoint_resolver')
-        results = []
+        # Attempt to use the given service_name verbatim first. This is
+        # important for services that share an endpoint prefix with another
+        # service, for example "neptune" which has endpoint prefix "rds".
+        regions_by_name = resolver.get_available_endpoints(
+            service_name, partition_name, allow_non_regional
+        )
+        if regions_by_name:
+            return regions_by_name
+        # If the given service does not resolve, use the endpoint prefix of the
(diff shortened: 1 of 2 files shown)


jonemo:
After offline discussion, we decided to not merge this. It introduces a small but non-zero risk of a future reverse name clash where one service's name is another service's endpoint prefix. 

For anyone who comes across this via Google search: Consider using the AWS SSM Parameter Store API introduced [in this blog post](https://aws.amazon.com/blogs/aws/new-query-for-aws-regions-endpoints-and-more-using-aws-systems-manager-parameter-store/) as an alternative source of region information.

PR #2889, opened 2023-03-16
Replace `lru_cache` with `instance_cache`

ron8mcr:
`lru_cache` depends on `self` argument.
As `lru_cache` is global is not tied to instance, this causes heavy RAM usage.

Fixes: https://github.com/boto/boto3/issues/3614

### DIFF
--- a/botocore/endpoint_provider.py
+++ b/botocore/endpoint_provider.py
@@ -24,7 +24,6 @@ or you can look at the test files in /tests/unit/data/endpoints/valid-rules/
 import logging
 import re
 from enum import Enum
-from functools import lru_cache
 from string import Formatter
 from typing import NamedTuple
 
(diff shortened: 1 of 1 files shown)


jonemo:
`instance_cache` has no `max_size` and could therefore grow unbounded. Realistic situation where this happens are very long-lived sessions that need to resolve endpoints for many different parameters. One example is a session which is used to call S3 operations for many different buckets (the bucket name is an input for endpoint resolution).

While creating many sessions (the root cause of the problem this PR aims to fix) _is not_ recommended, having long-lived sessions _is_. We'd either need a way to mitigate this problem, or find an alternative solution for use cases with many sessions.

jonemo:
We decided to go for a slightly different solution, see https://github.com/boto/botocore/pull/2934 for details. 

Thank you again @ron8mcr for reporting the issue and proposing this solution!

PR #2896, opened 2023-03-27
Fix links and code edge cases in docs generation

jonemo:
Fixes two classes of errors that were introduced in https://github.com/boto/botocore/pull/2855:

* When link text starts with a space, the link no longer gets rendered correctly. This one was discovered by @nateprewitt 
* Preformatted and code snippets that contain stylized quotation marks (`“`) instead of plain `"`.

As part of investigating this, I found that the special treatment for trailing whitespace in data nodes (where one whitespace is always left behind) is another special case whitespace treatment longer needed and removed it.

#### Examples

Before

Screenshot 2023-03-27 at 10 02 19 AM

After

Screenshot 2023-03-27 at 10 02 35 AM

Before

Screenshot 2023-03-27 at 9 04 57 AM

After

Screenshot 2023-03-27 at 9 05 04 AM

Before

Screenshot 2023-03-27 at 9 16 53 AM

After

Screenshot 2023-03-27 at 9 16 57 AM

#### Testing

Spot checked HTML of ~200 pages with diffs from the changes in this PR. All of them are either fixes of the issues described above, or further removals of unnecessary whitespace.

### DIFF
--- a/botocore/docs/bcdoc/docstringparser.py
+++ b/botocore/docs/bcdoc/docstringparser.py
@@ -296,15 +296,10 @@ class DataNode(Node):
             self.rstrip()
 
     def rstrip(self):
-        # If there is no content, consolidate all white-space on the right.
-        if self._stripped_data == '' and self._leading_whitespace != '':
-            self._trailing_whitespace = (
-                f"{self._leading_whitespace}{self._trailing_whitespace}"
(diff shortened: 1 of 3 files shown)


nateprewitt (approved):
⛵

PR #2903, opened 2023-04-05
fix: check datetime string format pattern before parse timestamp with tzinfo

### LINKED ISSUE #2902: S3 raises Invalid timestamp for string format time
### Describe the bug

When I try to give a Head Object of an object in S3 sometimes it returns LastModified in the format "Mon, 05 Apr 2023 14:30:00 EDT". 

When trying to convert it raises an error 
```console
ValueError: Invalid timestamp "Mon, 05 Apr 2023 14:30:00 EDT": Unknown string format: Mon, 05 Apr 2023 14:30:00 EDT
```

### Expected Behavior

Returns a datetime

### Current Behavior

Raises ValueError

### Reproduction Steps

Try to give a Head Object of an object in S3 that the LastModified is in the format "Mon, 05 Apr 2023 14:30:00 EDT".

### Possible Solution

`botocore/utils.py` : 908
```python
def _parse_timestamp_with_tzinfo(value, tzinfo):
    """Parse timestamp with pluggable tzinfo options."""
    if isinstance(value, (int, float)):
        # Possibly an epoch time.
        return datetime.datetime.fromtimestamp(value, tzinfo())
    else:
        try:
            return datetime.datetime.fromtimestamp(float(value), tzinfo())
        except (TypeError, ValueError):
            pass
    try:
        # enforce that GMT == UTC.
        STRING_FORMAT_PATTERN = r'^[a-zA-Z]{3}, \d{2} [a-zA-Z]{3} \d{4} \d{2}:\d{2}:\d{2} [A-Z]{3}$'
        if re.match(STRING_FORMAT_PATTERN, value):
            return datetime.datetime.strptime(value, '%a, %d %b %Y %H:%M:%S %Z')

        raise ValueError(f'Invalid timestamp "{value}": {e}')
```

_No response_

### SDK version used

1.29.106

Darwin Satheler 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:04:44 PST 2022; root:xnu-8792.61.2~4/RELEASE_ARM64_T8103 arm64

satheler:
fixes #2902

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -919,6 +919,12 @@ def _parse_timestamp_with_tzinfo(value, tzinfo):
         # In certain cases, a timestamp marked with GMT can be parsed into a
         # different time zone, so here we provide a context which will
         # enforce that GMT == UTC.
+        STRING_FORMAT_PATTERN = r'^[a-zA-Z]{3}, \d{2} [a-zA-Z]{3} \d{4} \d{2}:\d{2}:\d{2} [A-Z]{3}$'
+        if re.match(STRING_FORMAT_PATTERN, value):
+            return datetime.datetime.strptime(
+                value, '%a, %d %b %Y %H:%M:%S %Z'
+            ).replace(tzinfo=tzutc())
+
         return dateutil.parser.parse(value, tzinfos={'GMT': tzutc()})
     except (TypeError, ValueError) as e:
         raise ValueError(f'Invalid timestamp "{value}": {e}')
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,3 +9,6 @@ pytest-cov==2.12.1
 pytest-xdist==2.4.0
 atomicwrites>=1.0 # Windows requirement
 colorama>0.3.0 # Windows requirement
+
+# Pre commit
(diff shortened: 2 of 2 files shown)


jonemo:
Hi @satheler, thank you for opening a PR for this. I just re-read the issue and still don't quite understand what is causing the problem you see. Could you please share additional details in the issue discussion, including the debug level log output and full exception message with stack trace?

Based on the issue description and the solution you propose in this PR, I understand that a timestamp of the format `Mon, 05 Apr 2023 14:30:00 EDT` gets returned by an S3 API. This value will then get passed into `_parse_timestamp_with_tzinfo` when the API response gets parsed. Based on the solution proposed in this PR, I assume you are seeing the exception on this line?

```
```

In my testing, I can't get this line to fail in the way you describe with the input value mentioned above.

I don't think that adding the regex and `strptime` call here to effectively duplicate the functionality we expect from python-dateutil is the way forward. However, to fully understand, please provide more details on the problem. Alternatively or additionally, you could also add a test case to this PR that fails before the fix and passes afterwards.

satheler:
My mistake, I found the error. Thank you!

PR #2905, opened 2023-04-10
add custom error message for using crc32c without crt

dlm6693:
Adds a custom error message for customers trying to use `CRC32C` checksum algorithm, but don't have `awscrt` installed.

### DIFF
--- a/botocore/httpchecksum.py
+++ b/botocore/httpchecksum.py
@@ -250,7 +250,14 @@ def resolve_request_checksum_algorithm(
             supported_algorithms = _SUPPORTED_CHECKSUM_ALGORITHMS
 
         algorithm_name = params[algorithm_member].lower()
-        if algorithm_name not in supported_algorithms:
+        if algorithm_name == "crc32c" and not HAS_CRT:
+            raise FlexibleChecksumError(
+                error_msg=(
+                    "CRC32C is not supported without the CRT. To use "
+                    "CRC32C, please install the `awscrt` package."
+                )
+            )
+        elif algorithm_name not in supported_algorithms:
             raise FlexibleChecksumError(
                 error_msg="Unsupported checksum algorithm: %s" % algorithm_name
             )
--- a/botocore/httpchecksum.py
+++ b/botocore/httpchecksum.py
@@ -253,8 +253,8 @@ def resolve_request_checksum_algorithm(
         if algorithm_name == "crc32c" and not HAS_CRT:
             raise FlexibleChecksumError(
                 error_msg=(
-                    "CRC32C is not supported without the CRT. To use "
-                    "CRC32C, please install the `awscrt` package."
+                    "Using CRC32C requires an additional dependency. You will "
+                    "need to pip install botocore[crt] before proceeding."
                 )
             )
         elif algorithm_name not in supported_algorithms:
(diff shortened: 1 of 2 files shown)


botocore/httpchecksum.py
```diff
@@ -250,7 +250,14 @@ def resolve_request_checksum_algorithm(
             supported_algorithms = _SUPPORTED_CHECKSUM_ALGORITHMS
 
         algorithm_name = params[algorithm_member].lower()
-        if algorithm_name not in supported_algorithms:
+        if algorithm_name == "crc32c" and not HAS_CRT:
+            raise FlexibleChecksumError(
+                error_msg=(
+                    "CRC32C is not supported without the CRT. To use "
+                    "CRC32C, please install the `awscrt` package."
```
dlm6693, line 257: Should we explicitly say the command here? `pip install awscrt`

jonathan343: In other places we've used messages such as:
```python
"Using S3 with an MRAP arn requires an additional "
"dependency. You will need to pip install "
"botocore[crt] before proceeding."
```
Maybe we can do something similar here?
```python
"Using CRC32C requires an additional "
"dependency. You will need to pip install "
"botocore[crt] before proceeding."
```
Source: https://github.com/boto/botocore/blob/develop/botocore/utils.py#LL2206C17-L2210C18

jonathan343: Im not extremely familiar with using CRT but it might be safer to have users install CRT through `botocore[crt]` since this ensures they install a compatible version. We're currently pinning to `awscrt==0.16.9`.

https://github.com/boto/botocore/blob/develop/setup.cfg#LL11C1-L11C21

dlm6693: Yep you're right on both counts. I'll update shortly.

jonathan343 (approved):
Looks good to me! 🚀

botocore/httpchecksum.py
```diff
@@ -250,7 +250,14 @@ def resolve_request_checksum_algorithm(
...
 
         algorithm_name = params[algorithm_member].lower()
-        if algorithm_name not in supported_algorithms:
+        if algorithm_name == "crc32c" and not HAS_CRT:
+            raise FlexibleChecksumError(
+                error_msg=(
+                    "Using CRC32C requires an additional dependency. You will "
+                    "need to pip install botocore[crt] before proceeding."
+                )
+            )
```
nateprewitt, line 259: This should be a MissingDependencyError not a FlexibleChecksumError to stay consistent with our CRT exception messaging. See [here](https://github.com/boto/botocore/blob/develop/botocore/utils.py#L2206-L2210) for an example.

We also want to avoid hardcoding this for one algorithm. We should be able to have a defined list of algorithms supported by the CRT and check against that.

```python
if algorithm_name not in supported_algorithms:
    if not HAS_CRT and algorithm_name in crt_algorithms:
        raise MissingDependencyError(
            error_msg=(
                f"Using {algorithm_name} requires an additional dependency. "
                f"You will need to pip install botocore[crt] before proceeding."
            )
        )  
    raise FlexibleChecksumError(
        error_msg="Unsupported checksum algorithm: %s" % algorithm_name
    )

dlm6693: Ok I can open a followup PR. My only question is should `CRC32` be included in `crt_algorithms`? Wondering because we have an implementation with and without the CRT.

nateprewitt: Everything we support in the CRT can be in the list. We should assert that the dictionary keys we generate [here](https://github.com/boto/botocore/blob/develop/botocore/httpchecksum.py#L466-L470) are all in the list to keep everything in sync.

PR #2906, opened 2023-04-11
change error for missing crt. use list of crt checksums instead

dlm6693:
Follow up to https://github.com/boto/botocore/pull/2905. Use `MissingDependencyException` and check against extendible list of CRT checksums rather than single strings. Added test to validate CRT checksum list is in sync with checksum class dictionary keys.

### DIFF
--- a/botocore/httpchecksum.py
+++ b/botocore/httpchecksum.py
@@ -476,6 +476,8 @@ if HAS_CRT:
     }
     _CHECKSUM_CLS.update(_CRT_CHECKSUM_CLS)
     # Validate this list isn't out of sync with _CRT_CHECKSUM_CLS keys
-    assert all(name in _CRT_CHECKSUM_ALGORITHMS for name in _CRT_CHECKSUM_CLS)
+    assert all(
+        name in _CRT_CHECKSUM_ALGORITHMS for name in _CRT_CHECKSUM_CLS.keys()
+    )
 _SUPPORTED_CHECKSUM_ALGORITHMS = list(_CHECKSUM_CLS.keys())
 _ALGORITHMS_PRIORITY_LIST = ['crc32c', 'crc32', 'sha1', 'sha256']
--- a/botocore/httpchecksum.py
+++ b/botocore/httpchecksum.py
@@ -244,7 +244,6 @@ def resolve_request_checksum_algorithm(
     operation_model,
     params,
     supported_algorithms=None,
-    crt_supported_algorithms=None,
 ):
     http_checksum = operation_model.http_checksum
     algorithm_member = http_checksum.get("requestAlgorithmMember")
@@ -253,16 +252,15 @@ def resolve_request_checksum_algorithm(
         # request supports it, use that instead of checksum required
         if supported_algorithms is None:
             supported_algorithms = _SUPPORTED_CHECKSUM_ALGORITHMS
-        if crt_supported_algorithms is None:
-            crt_supported_algorithms = _CRT_SUPPORTED_CHECKSUM_ALGORITHMS
 
         algorithm_name = params[algorithm_member].lower()
         if algorithm_name not in supported_algorithms:
-            if not HAS_CRT and algorithm_name in crt_supported_algorithms:
+            if not HAS_CRT and algorithm_name in _CRT_CHECKSUM_ALGORITHMS:
                 raise MissingDependencyException(
                     msg=(
-                        "Using CRC32C requires an additional dependency. You will "
-                        "need to pip install botocore[crt] before proceeding."
+                        f"Using {algorithm_name.upper()} requires an "
+                        "additional dependency. You will need to pip install "
+                        "botocore[crt] before proceeding."
                     )
                 )
             raise FlexibleChecksumError(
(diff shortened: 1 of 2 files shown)


botocore/httpchecksum.py
```diff
@@ -462,12 +469,16 @@ def _handle_bytes_response(http_response, response, algorithm):
...
 
 if HAS_CRT:
     # Use CRT checksum implementations if available
-    _CHECKSUM_CLS.update(
-        {"crc32": CrtCrc32Checksum, "crc32c": CrtCrc32cChecksum}
-    )
+    _CRT_CHECKSUM_CLS = {
+        "crc32": CrtCrc32Checksum,
+        "crc32c": CrtCrc32cChecksum,
+    }
```
dlm6693, line 479: Saving this to a variable instead lets us validate the keys match the hardcoded list when the CRT is present.

We can't take the list of keys when the CRT is present because it will be empty and is needed for the custom exception.

botocore/httpchecksum.py
```diff
@@ -462,12 +469,16 @@ def _handle_bytes_response(http_response, response, algorithm):
...
+    _CHECKSUM_CLS.update(_CRT_CHECKSUM_CLS)
```
dlm6693, line 480: We could move this line outside of the if block

botocore/httpchecksum.py
```diff
@@ -248,16 +253,18 @@ def resolve_request_checksum_algorithm(
         # request supports it, use that instead of checksum required
         if supported_algorithms is None:
             supported_algorithms = _SUPPORTED_CHECKSUM_ALGORITHMS
+        if crt_supported_algorithms is None:
+            crt_supported_algorithms = _CRT_SUPPORTED_CHECKSUM_ALGORITHMS
```
nateprewitt, line 257: I don't think this needs to be passed in, we can use the static list. There isn't a strong reason for it to be dynamic/adjustable per invocation like supported checksums.

botocore/httpchecksum.py
```diff
@@ -248,16 +253,18 @@ def resolve_request_checksum_algorithm(
...
-        if algorithm_name == "crc32c" and not HAS_CRT:
-            raise FlexibleChecksumError(
-                error_msg=(
-                    "Using CRC32C requires an additional dependency. You will "
-                    "need to pip install botocore[crt] before proceeding."
+        if algorithm_name not in supported_algorithms:
+            if not HAS_CRT and algorithm_name in crt_supported_algorithms:
+                raise MissingDependencyException(
+                    msg=(
+                        "Using CRC32C requires an additional dependency. You will "
```
nateprewitt, line 264:
```suggestion
                        f"Using {algorithm_name} requires an additional dependency. You will "
```

botocore/httpchecksum.py
```diff
@@ -462,12 +469,16 @@ def _handle_bytes_response(http_response, response, algorithm):
     "sha1": Sha1Checksum,
     "sha256": Sha256Checksum,
 }
+_CRT_CHECKSUM_CLS = {}
```
nateprewitt, line 472: Why does this need to be declared outside of the HAS_CRT conditional?

nateprewitt (changes requested):

botocore/httpchecksum.py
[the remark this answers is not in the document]
dlm6693: Per your comment below, I think I was under the impression that we never want to make assertions at runtime, just in our test suite.

tests/unit/test_httpchecksum.py
```diff
@@ -536,6 +546,13 @@ def test_handle_checksum_body_checksum_skip_streaming(self):
         body = response_dict["body"]
         self.assertEqual(body.read(), b"hello world")
 
+    @requires_crt()
+    def test_checksum_cls_crt_supported_algorithms_in_sync(self):
+        self.assertEqual(
+            sorted(_CRT_SUPPORTED_CHECKSUM_ALGORITHMS),
+            sorted(list(_CRT_CHECKSUM_CLS.keys())),
+        )
+
```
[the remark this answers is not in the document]
dlm6693: Just curious, can you explain the preference of calling the assertion at runtime instead of exposing a private variable in the test suite? Is this just a design principle or is there potential risk to exposing private attributes outside of runtime?

tests/unit/test_httpchecksum.py
```diff
@@ -166,7 +170,7 @@ def test_request_checksum_algorithm_model_unsupported_algorithm(self):
         operation_model = self._make_operation_model(
             http_checksum={"requestAlgorithmMember": "Algorithm"},
         )
-        params = {"Algorithm": "crc32"}
+        params = {"Algorithm": "sh256"}
```
dlm6693, line 173: When an empty list is passed to `supported_algorithms`, the custom missing dependency error will be raised if the provided algorithm is `crc32`. This isn't technically true so is this acceptable?

nateprewitt: This is only ever used in testing, so it won't matter to the end user.

Is `sh256` a typo here?

nateprewitt (approved):

PR #2908, opened 2023-04-13
Use codecov-action instead of pip installing.

jonathan343:
## Overview
Codecov has removed their package from PyPi ([see message](https://about.codecov.io/blog/message-regarding-the-pypi-package/)) which has caused our HitHub workflows to fail since we `pip install codecov`. This PR uses their GitHub Action ([codecov-action](https://github.com/codecov/codecov-action)) instead.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -22,13 +22,11 @@ jobs:
           python-version: '${{ matrix.python-version }}'
       - name: Install dependencies
         run: |
-          pip install codecov
           python scripts/ci/install
       - name: Run tests
         run: |
           python scripts/ci/run-tests --with-cov --with-xdist
       - name: Run codecov
-        run: |
-          rm tests/coverage.xml
-          mv tests/.coverage ./
-          codecov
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -22,13 +22,11 @@ jobs:
...
           python scripts/ci/install
       - name: Run tests
         run: |
           python scripts/ci/run-tests --with-cov --with-xdist
       - name: Run codecov
-        run: |
-          rm tests/coverage.xml
-          mv tests/.coverage ./
-          codecov
+        uses: codecov/codecov-action@v2
```
dlm6693, line 30: The current major version is v3. I think we should be using that instead here.

dlm6693 (changes requested):

.github/workflows/run-tests.yml
jonathan343: Fixed in the most recent commit. Thanks!

dlm6693 (approved):
📦

PR #2917, opened 2023-04-25
Remove XML example files for SQS

jonemo:
In preparation for support of thw AWS JSON protocol in SQS, this removes the example/test files that contain XML responses for SQS.


nateprewitt (approved):
:shipit:

PR #2920, opened 2023-04-26
Include params set in provide-client-param event handlers in dynamic context params for endpoint resolution

jonemo:
This PR moves the `provide-client-params.*` and `before-parameter-build.*` from immediately after endpoint resolution to immediately before endpoint resolution.

This addresses the fact that for endpoint rulesets that accept dynamic context parameters (currently `s3`, `s3-control`, `kinesis`, and `events`) any parameters changed or updated in an event handler for the `provide-client-params.*` were not accounted for in endpoint resolution because the event was only emitted _after_ endpoint resolution.

### DIFF
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -819,7 +819,6 @@ class ClientEndpointBridge:
 
 
 class BaseClient:
-
     # This is actually reassigned with the py->op_name mapping
     # when the client creator creates the subclass.  This value is used
     # because calls such as client.get_paginator('list_objects') use the
@@ -913,6 +912,11 @@ class BaseClient:
             'has_streaming_input': operation_model.has_streaming_input,
             'auth_type': operation_model.auth_type,
         }
+        api_params = self._emit_api_params(
+            api_params=api_params,
+            operation_model=operation_model,
+            context=request_context,
+        )
         endpoint_url, additional_headers = self._resolve_endpoint_ruleset(
             operation_model, api_params, request_context
         )
@@ -984,9 +988,6 @@ class BaseClient:
         headers=None,
         set_user_agent_header=True,
     ):
-        api_params = self._emit_api_params(
-            api_params, operation_model, context
-        )
         request_dict = self._serializer.serialize_to_request(
             api_params, operation_model
         )
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -663,7 +663,11 @@ def generate_presigned_url(
         context,
         ignore_signing_region=(not bucket_is_arn),
     )
-
+    params = self._emit_api_params(
+        api_params=params,
+        operation_model=operation_model,
+        context=context,
+    )
     request_dict = self._convert_to_request_dict(
         api_params=params,
         operation_model=operation_model,
@@ -786,7 +790,11 @@ def generate_presigned_post(
         context,
         ignore_signing_region=(not bucket_is_arn),
     )
-
+    params = self._emit_api_params(
+        api_params=params,
+        operation_model=operation_model,
+        context=context,
+    )
     request_dict = self._convert_to_request_dict(
         api_params=params,
         operation_model=operation_model,
--- a/tests/functional/test_context_params.py
+++ b/tests/functional/test_context_params.py
@@ -343,7 +343,7 @@ def test_client_context_param_sent_to_endpoint_resolver(
         ),
     )
 
-    # Stub client to prevent a request from getting sent and asceertain that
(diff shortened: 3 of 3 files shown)


botocore/client.py
```diff
@@ -819,7 +819,6 @@ def _resolve_signature_version(self, service_name, resolved):
 
 
 class BaseClient:
-
```
dlm6693, line 822: Was this intentional?

botocore/client.py
```diff
@@ -913,6 +912,11 @@ def _make_api_call(self, operation_name, api_params):
             'has_streaming_input': operation_model.has_streaming_input,
             'auth_type': operation_model.auth_type,
         }
+        api_params = self._emit_api_params(
+            api_params=api_params,
+            operation_model=operation_model,
+            context=request_context,
+        )
```
dlm6693, line 919: Maybe this could be called from within `_resolve_endpoint_ruleset` at the very start of it? I'm thinking in purely organizational terms just because `_make_api_call` is a rather long, complex function already.

dlm6693 (approved):
Just a couple of stylistic comments, but otherwise LGTM!

botocore/signers.py
```diff
@@ -663,7 +663,11 @@ def generate_presigned_url(
         context,
         ignore_signing_region=(not bucket_is_arn),
     )
-
+    params = self._emit_api_params(
```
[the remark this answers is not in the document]
dlm6693: I think using my suggestion of putting this in `_resolve_endpoint_ruleset` would accomplish this no?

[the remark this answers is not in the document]
dlm6693: Or does this still need to happen immediately after endpoint resolution for presigned URLs?

botocore/client.py
jonemo: I didn't add it intentionally. But when I saw and understood it, I intentionally did not revert it.

This is a change applied by the black formatter. black makes this change only in versions [23.1.0](https://black.readthedocs.io/en/stable/change_log.html#id8) and higher. 23.1.0 is the version that adheres to black's [2023 stable style](https://github.com/psf/black/issues/3407). I suspect but did not validate that [this change to empty line handling](https://github.com/psf/black/pull/3302) which is included in the 2023 stable style causes this change to be applied. botocore specifies the black version used for CI in [the precommit config file](https://github.com/boto/botocore/blob/a91a7bf/.pre-commit-config.yaml#L37). 

I decided to leave the change in for two reasons:
1. More and more contributors will update their local black version to >=23.1.0 and this change will continue sneaking into diffs (and then require unnecessary attention).
2. Versions of black <23.1.0 do not revert this change, so there is downside to keeping it.

jonemo:
@dlm6693 @nateprewitt Added two commits that consolidate the previously duplicated code. This comes at the cost of a change to the API of the `_convert_to_request_dict` method. The signature remains backwards compatible, but the `endpoint_url` argument is no longer used inside the function. There is a small chance that this will break anyone who relies on this "private" internal interface.

While making this change I also fixed the presign functions which were actually still affected by the original bug and added test coverage for this oversight.

botocore/client.py
dlm6693: Ack. If black did it and you reversed the change, it probably would've happened the next time this file was updated. Thanks for the explanation!

dlm6693 (approved):

jonemo:
After reviewing potential impact of the API change contained in bd44524, @nateprewitt and I decided to revert to the less well organized state that has been discussed above. Note that f8e01a6 is _not_ an exact line-for-line revert because it keeps the bugfix described in https://github.com/boto/botocore/pull/2920#issuecomment-1539183503.

botocore/signers.py
```diff
@@ -656,14 +656,18 @@ def generate_presigned_url(
...
+        context=context,
+    )
     bucket_is_arn = ArnParser.is_arn(params.get('Bucket', ''))
     endpoint_url, additional_headers = self._resolve_endpoint_ruleset(
         operation_model,
         params,
         context,
         ignore_signing_region=(not bucket_is_arn),
     )
-
```
dlm6693, line 666: Was this also black?

dlm6693 (approved):

botocore/signers.py
jonemo: No. Brought the two empty lines back in d8ed89d.

dlm6693:
@jonemo still LGTM after the latest commit 👍

nateprewitt (approved):
:shipit:

PR #2928, opened 2023-05-01
Addresses admonition accessibility issue

jonathan343:
## Overview
This PR addresses an accessibility issue related to the titles of what sphinx calls [paragraph-level directives](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#paragraph-level-markup) (also know as notes/warnings/admonitions). This currently generates both the title and message content using the HTML `<p>` tag. This causes issues in the following categories:
1. **Visual Hierarchy** - "Making texts larger helps guide the eye around the page." Using headings and making them visually apparent is especially helpful for users with cognitive disabilities.". By not providing an apparent visual difference between the title and message content, we're providing certain users with a poor user experience.
2. **Screen Reader** - " screen reader users can also benefit from headings. Screen reader users can navigate a page according to its headings, listen to a list of all headings, and skip to a desired heading to begin reading at that point. Screen reader users can use headings to skip the repeated blocks of content like headers, menus, and sidebars, for example." By not having the titles as a heading, we're providing screen reader users with a bad user experience.

By inheriting from the sphinx  [HTML5Translator](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/writers/html5.py) class, we can define a custom translator that takes uses `<h3>` tags where we want.
Cases:
1. For every admonition we'll use `<h3>` for the title instead of the default `<p>`.
2. For a specific list of sub-headings we currently generate as strong text (bold).

Without customization, a `.. note::` generates something like:
```
<div class="admonition note">
<p class="admonition-title">Note</p>
...
</div>
```

After the customization we get something like
```
<div class="admonition note">
<h2 class="admonition-title">Note</h2>
...
</div>
```

## Results
**Before:**
Screen Shot 2023-05-04 at 11 06 35 AM

**After:**
Screen Shot 2023-05-04 at 11 05 59 AM
Corresponding HTML:

## Additional Note
Custom CSS was applied to make the change function correctly.

## References
https://usability.yale.edu/web-accessibility/articles/headings


jonemo (commented):
Assuming that we aren't missing a clever way to configure the HTML tag for admonition headings in docutils, this does seem like the best way to do this. Two small comments/questions from me, but otherwise this looks good.

tests/unit/docs/test_client.py
```diff
@@ -144,18 +148,22 @@ def test_modeled_exceptions(self):
...
                 '{',
                 "'Message': 'string',",
                 "'Error': {",
                 "'Code': 'string',",
                 "'Message': 'string'",
                 '}',
                 '}',
-                '**Structure**',
+                '    .. raw:: html',
+                '      <h3>Structure</h3>',
```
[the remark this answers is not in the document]
jonathan343: * Additional indentation - This is just me applying stricter validation that not only validates content but its indentation as well. This assertion checks if a provided string is a substring of a pages content so if the true value of a line in our docs was `    ..rst:: html` and we made an unintentional update that changed it to ` ..rst:: html` it would continue to pass if we were validating with ` ..rst:: html` but fail with `    ..rst:: html` since it expects more indentation. This is good for catching future unintended whitespace issues but can also be a pain if we make an intentional whitespace change.

* Indentation Variance - Some lines may have more indentation than others since they're more nested than other headings. If you look at this [example page](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/exceptions/NoSuchBucket.html), you'll see that `Syntax` and `Structure` are more nested than `Examples`.

tests/unit/docs/bcdoc/test_style.py
```diff
@@ -128,19 +128,44 @@ def test_important(self):
...
+                b'\n\n.. raw:: html\n\n  </div>\n\n'
+            ),
+        )
 
     def test_danger(self):
         style = ReSTStyle(ReSTDocument())
         style.start_danger()
         style.end_danger()
-        self.assertEqual(style.doc.getvalue(), b'\n\n.. danger::\n\n  \n\n')
+        print(style.doc.getvalue())
```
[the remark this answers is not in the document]
jonathan343: I'll removed this. Thanks for catching!

jonemo (approved):

jonemo (commented):
Me, four days ago:

> Assuming that we aren't missing a clever way to configure the HTML tag for admonition headings in docutils

Thanks for finding this alternative approach and proving my assumption wrong! This is much better.

One small comment regarding an unlikely but possible failure mode.

docs/source/conf.py
```diff
@@ -283,3 +285,53 @@
...
+        'Response Structure',
+        'Response Syntax',
+        'Structure',
+        'Syntax'
+    ]
+
+    def visit_admonition(self, node, name=''):
+        """Uses the h3 tag for admonition titles instead of the p tag"""
+        self.body.append(self.starttag(
+            node, 'div', CLASS=('admonition ' + name)))
```
dlm6693, line 307: nit: I don't like this formatting 😄

docs/source/conf.py
```diff
@@ -283,3 +285,53 @@
...
+        if name:
+            title = (
+                f"<h3 class='admonition-title'>"
```
dlm6693, line 310:
```suggestion
                "<h3 class='admonition-title'>"
```

docs/source/conf.py
```diff
@@ -283,3 +285,53 @@
 
 # How to display URL addresses: 'footnote', 'no', or 'inline'.
 #texinfo_show_urls = 'footnote'
+
+class HTML5Translator(SphinxHTML5Translator):
+    """
+    This is our custom HTML5Translator which extends the one provided by Sphinx.
+    """
```
dlm6693, line 292: This shouldn't be conversational. Also its convention to start the doc string on the same line as the triple quote and if the doc string is multiple lines the end triple quote should be on a new line. If its not long enough then it should be on the same line.

```suggestion
    """Extension of Sphinx's ``HTML5Translator for Botocore documentation."""
```

dlm6693: Let's also rename to something more specific like `BotocoreHTML5Translator` or `BotoHTML5Translator`

docs/source/conf.py
```diff
@@ -283,3 +285,53 @@
...
+        'Exceptions',
+        'Request Syntax',
+        'Response Structure',
+        'Response Syntax',
+        'Structure',
+        'Syntax'
+    ]
+
+    def visit_admonition(self, node, name=''):
+        """Uses the h3 tag for admonition titles instead of the p tag"""
```
dlm6693, line 305:
```suggestion
        """Uses the h3 tag for admonition titles instead of the p tag."""
```

docs/source/conf.py
dlm6693: Also we're mixing double and single quotes in this file. I know there's some mixing in other places, but can we try to stick to one in this PR? I'd vote for double quotes.

docs/source/conf.py
```diff
@@ -283,3 +285,53 @@
...
+                f"<h3 class='admonition-title'>"
+                f"{admonitionlabels[name]}</h3>"
+            )
+            self.body.append(title)
+
+    def visit_strong(self, node):
+        """
+        Opens the h3 tag for a specific set of words/phrases and opens the
+        strong tag for all others.
+        """
```
dlm6693, line 319: Another doc string convention in Python is to have a brief sentence on one line, then a line break, then some more details if needed.

I'm not sure if this first sentence is exactly accurate, but something along these lines:

```suggestion
        """Visit a strong HTML element.
        
        Opens the h3 tag for a specific set of words/phrases and opens the
        strong tag for all others.
        """
```

docs/source/conf.py
```diff
@@ -283,3 +285,53 @@
...
+        if len(node) > 0 and node[0] in self.STRONG_TO_H3_HEADINGS:
+            self.body.append(self.starttag(node, 'h3', ''))
+        else:
+            self.body.append(self.starttag(node, 'strong', ''))
+
+    def depart_strong(self, node):
+        """
+        Closes the h3 tag for a specific set of words/phrases and closes the
+        strong tag for all others.
+        """
```
dlm6693, line 329:
```suggestion
        """Depart strong HTML element.
        
        Closes the h3 tag for a specific set of words/phrases and closes the
        strong tag for all others.
        """
```

jonathan343:
> Just some notes on doc strings. Not blockers though.

Thanks for the feedback! I addressed all issues/suggestions in the most recent commit. Lmk if there is anything I missed or something else you'd like to see in this PR.

dlm6693 (approved):
:shipit:

PR #2931, opened 2023-05-05
Revert SQS protocol changes

nateprewitt:
Roll back recent SQS changes.

### DIFF
--- a/.changes/next-release/bugfix-sqs-16349.json
+++ b/.changes/next-release/bugfix-sqs-16349.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "``sqs``",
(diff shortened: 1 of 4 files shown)


kyleknap (approved):
:ship:

bblommers:
Hi @nateprewitt, I was just looking into fixing this for `Moto`.
Will the JSON format be reintroduced at some point in the future, or is the SQS-service going to stick with XML for now?

bblommers:
Appreciate the info @nateprewitt - having a timeline/estimate would be very useful.

Do you know if other XML-based services, like EC2 or S3, will get the same treatment?

bblommers:
Awesome, thanks @nateprewitt!

PR #2934, opened 2023-05-08
Reduce memory usage for scripts with many Session objects

jonemo:
Addresses https://github.com/boto/boto3/issues/3614
Alternative to https://github.com/boto/botocore/pull/2889

This retains the cache on `botocore.endpoint_provider.EndpointProvider.resolve_endpoint` while mitigating the effect on memory usage reported in https://github.com/boto/boto3/issues/3614. 

**Problem description**

The current code results in excessive memory usage when a large number of botocore `Session` objects is created in a script. Previous to botocore version 1.29.0, Python's garbage collector would have cleaned these objects up in most situations. The introduction of the `EndpointProvider` and the use of `lru_cache` within now results in up to 100 `Session` objects staying memory while they are referenced in the cache.

**Solution in this PR:**

This PR proposes a solution where Python's `weakref` module is used to temporarily replace the full reference with a weak reference during the cacheing process. This way, the cache no longer interferes with the garbage collector but otherwise performs the same function. This comes at the expense of a small computational overhead for creating (always) and resolving (for cache misses) the weak reference.

**Other solutions considered:**

* Remove the cache entirely: This is the best solution for the use case with many sessions. However, creating many sessions is only required in rare circumstances. Session object reuse is the recommended best practice and more common use case. For long-lived sessions where only a few services and operations are called, the cache has positive impact on runtime that we want to retain.
* Use botocore's `instance_cache` decorator (https://github.com/boto/botocore/pull/2889): For long-lived sessions where each operation call results in a cache miss, this can lead to indefinitely growing memory usage. This is because `instance_cache` has no `maxsize` parameter.
* Use `lru_cache` in the constructor (https://github.com/boto/boto3/issues/3614#issuecomment-1469172535): This solution addresses the problem and performs well on both memory and runtime metrics on all Python versions we tested. However, it relies on an undocumented way of using the `lru_cache` decorate (namely: not as a decorator) and could therefore result in unexpected behavior changes in future Python versions.
* Keep `lru_cache` in place but reduce `maxsize` to 10: This avoids the excessive memory usage but still uses on the order of MBs more than the solution in this PR with a maxsize of 100.

Benchmarking results:

multiplot

multiplot

multiplot

Each line represents a single run.


nateprewitt (commented):
⛵ Thanks @jonemo. Awesome write up!

nateprewitt (approved):

nateprewitt (changes requested):
Can we get a changelog entry too before we release?

nateprewitt (approved):
:shipit:

PR #2936, opened 2023-05-11
Fix .git-blame-ignore-revs hash

nateprewitt:
The merge commit changed the commit ID for the recent black updates (459a6a370ea1a4ab52b792f3d6e8311bc774b583).


jonemo (approved):
I don't quite understand what the 9faeaa commit is/was, but 459a6a is the right one.

nateprewitt:
459a6a370ea1a4ab52b792f3d6e8311bc774b583 was 9faeaa but I amended the commit to add the .git-blame-ignore-revs entry 🤦

Issue #2938, opened 2023-05-12
virtual-host addressing style not working over http with short subdomain fragments


MartinNowak:
### Describe the bug

```py
client = boto3.client('s3', endpoint='http://s3.amazonaws.com', config=Config(s3=dict(addressing_style='virtual')))
client.list_objects_v2(Bucket='failing.ab.example.com', MaxKeys=1) # Endpoint provider result: http://s3.amazonaws.com/failing.ab.example.com (path-style)
client.list_objects_v2(Bucket='working.abc.example.com', MaxKeys=1) # Endpoint provider result: http://working.abc.example.com.s3.amazonaws.com (virtual-host-style)
```

Due to recursively calling aws_is_virtual_hostable_s3_bucket on each subdomain fragment
https://github.com/boto/botocore/blob/2091c2629f320c2dd16c9e29a0690d493375c8a7/botocore/endpoint_provider.py#L402-L405
and requiring each to be longer than 2 chars
https://github.com/boto/botocore/blob/2091c2629f320c2dd16c9e29a0690d493375c8a7/botocore/endpoint_provider.py#L395

### Expected Behavior

Should address subdomain buckets via virtual-host addressing style so one can properly filter egress.

### Current Behavior

Subdomain buckets with a short fragment (< 3 chars) are always forced to use path-style addressing.

### Reproduction Steps

See above.

### Possible Solution

Validate subdomain fragments without length requirement.

### Additional Information/Context

> Important
When you're using virtual-hosted–style buckets with SSL, the SSL wildcard certificate matches only buckets that do not contain dots (.). To work around this limitation, use HTTP or write your own certificate-verification logic. For more information, see [Amazon S3 Path Deprecation Plan](http://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/) on the AWS News Blog.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html

### SDK version used

botocore==1.29.105

### Environment details (OS name and version, etc.)

na

nateprewitt:
Hi @MartinNowak, thanks for the report! This does in fact appear to be a bug. We're intending to validate that the entirety of the bucket name is >= 3 characters, but this is also being incorrectly applied to each subcomponent. We'll take a look at putting together a fix and respond here once we have something ready.

nateprewitt:
[...]

PR #2941, opened 2023-05-12
Address virtual host style addressing bugs for short subdomains provided in ``endpoint_url``

dlm6693:
This addresses https://github.com/boto/botocore/issues/2938. Currently `RuleSetStandardLibrary.aws_is_virtual_hostable_s3_bucket` is performing the exact same validations on the entire input value as it does with individual subdomains within the input value. This is incorrect because only the entire value needs to be at least 3 characters, not each individual component. As described in the issue, this caused a bug when enabling virtual style addressing and providing an `endpoint_url` to an S3 client that is insecure (http not https). For bucket names that look like URLs containing short subdomain fragments, path style is being used instead which is incorrect. 

Additionally, we only need to perform validation that the input value is not an IPV4 address, None and all lowercase at the top level so the recursive check has been removed.

### DIFF
--- a/.changes/next-release/bugfix-EndpointProvider-38157.json
+++ b/.changes/next-release/bugfix-EndpointProvider-38157.json
@@ -1,5 +1,5 @@
 {
   "type": "bugfix",
   "category": "EndpointProvider",
-  "description": "Addresses issue `#2938 <https://github.com/boto/botocore/issues/2938>`__"
+  "description": "Fixed bug in virtual addressing for S3 Buckets `#2938 <https://github.com/boto/botocore/issues/2938>`__"
 }
--- a/.changes/next-release/bugfix-EndpointProvider-38157.json
+++ b/.changes/next-release/bugfix-EndpointProvider-38157.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "EndpointProvider",
+  "description": "Addresses issue `#2938 <https://github.com/boto/botocore/issues/2938>`__"
+}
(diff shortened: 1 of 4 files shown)


.changes/next-release/bugfix-EndpointProvider-38157.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "EndpointProvider",
+  "description": "Addresses issue `#2938 <https://github.com/boto/botocore/issues/2938>`__"
```
nateprewitt, line 4: Lets provide a clear message on the issue and then we can link.

```suggestion
  "description": "Fixed bug in virtual addressing for S3 Buckets `#2938 <https://github.com/boto/botocore/issues/2938>`__"
```

tests/functional/test_s3.py
```diff
@@ -3487,6 +3487,38 @@ def _addressing_for_presigned_url_test_cases():
...
+        customer_provided_endpoint="https://s3.us-west-2.amazonaws.com",
+        expected_url="https://s3.us-west-2.amazonaws.com/foo.b.biz/key",
+    )
+    yield dict(
+        region="us-west-1",
+        bucket="foo.b.biz",
+        key="key",
+        s3_config={"addressing_style": "virtual"},
+        expected_url="https://s3.us-west-1.amazonaws.com/foo.b.biz/key",
+    )
```
nateprewitt, line 3521: What are we testing between these two? Region shouldn't be related to virtual construction unless we're looking at us-east-1 which I think is already covered above.

nateprewitt (changes requested):

tests/functional/test_s3.py
dlm6693: Sure. I'll remove.

nateprewitt (approved):
⛵

PR #2942, opened 2023-05-16
Support for sdk_ua_app_id config setting

jonemo:
This PR adds support for the `AWS_SDK_UA_APP_ID` environment variable and `sdk_ua_app_id` setting in the [shared configuration file](https://docs.aws.amazon.com/sdkref/latest/guide/file-format.html). This is already supported by some other AWS SDKs today and will in future be the recommended way for third party apps that use botocore to inject an identifier into the User-Agent header. Botocore will also continue supporting the `user_agent` and `user_agent_extra` config settings for backwards compatibility.

For test coverage, this PR introduces new functional testing for all user agent header related config settings. While there are existing tests for user agent settings in [unit/test_client.py](https://github.com/boto/botocore/blob/develop/tests/unit/test_client.py), [unit/test_session.py](https://github.com/boto/botocore/blob/develop/tests/unit/test_session.py), and [unit/test_session_legacy.py](https://github.com/boto/botocore/blob/develop/tests/unit/test_session_legacy.py), they either test session-level settings (which get overridden or modified by client-level config) or mock the legacy endpoint resolver.

Open questions where for which I'd be particularly interested to hear feedback:

1. Naming of the config variable. Right now I made the config variable name similar to existing config variables but different from the config _file_ setting name.
2. Test fixtures: I wonder if there is an easier way to stub the requests that requires less setup and simply captures all outgoing requests. Probably not because it would require somehow aborting the request before the response is parsed...

### DIFF
--- a/botocore/config.py
+++ b/botocore/config.py
@@ -38,6 +38,12 @@ class Config:
     :param user_agent_extra: The value to append to the current User-Agent
         header value.
 
+    :type user_agent_appid: str
+    :param user_agent_appid: A value that gets included in the User-Agent
+        string in the format "app/<user_agent_appid>" which is consistent
+        across all AWS SDKs. Allowed characters are ASCII alphanumerics and
+        ``!$%&'*+-.^_`|~``. All other characters will be replaced by a ``-``.
+
     :type connect_timeout: float or int
     :param connect_timeout: The time in seconds till a timeout exception is
         thrown when attempting to make a connection. The default is 60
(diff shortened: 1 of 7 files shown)


botocore/config.py
```diff
@@ -38,6 +38,12 @@ class Config:
     :param user_agent_extra: The value to append to the current User-Agent
         header value.
 
+    :type user_agent_appid: str
+    :param user_agent_appid: A value that gets included in the User-Agent
+        string in the format "app/<user_agent_appid>" which is consistent
+        across all AWS SDKs. Allowed characters are ASCII alphanumerics and
+        ``!$%&'*+-.^_`|~``. All other characters will be replaced by a ``-``.
```
nateprewitt, line 45: Where are we enforcing this currently?

jonemo:
Closing this in favor of a larger PR with more user agent header changes. Just too many interconnected pieces to cleanly separate this into smaller pieces...

PR #2944, opened 2023-05-17
Presigning content type bug

dlm6693:
This addresses a bug in URL presigning. For certain operations, like Polly's `SynthesizeSpeech`, the URL used for presigning does not match what is modeled for the request. Typically it is sent as a `POST` with API parameters attached to a body, but through presigning, it is a `GET` with parameters attached in the URL query. 

This issue arises in the `RestJson` serializers where the input params are still treated as a body, so a `Content-Type` header is injected, causing a signature mismatch in the presigned URL that is returned.

This PR generally removes this header for presigning from requests using methods where bodies should not be present or should have no meaning.

UPDATE: After some issues were discovered when running presigning for the JSON-RPC protocol, we've decided to scale back this solution to only be applied as a customization for Polly's `synthesize_speech` when generating a presigned URL.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -1137,7 +1137,7 @@ def customize_endpoint_resolver_builtins(
         builtins[EndpointResolverBuiltins.AWS_S3_USE_GLOBAL_ENDPOINT] = True
 
 
-def remove_content_type_header_for_polly_presigning(request, **kwargs):
+def remove_content_type_header_for_presigning(request, **kwargs):
     if (
         request.context.get('is_presign_request') is True
         and 'Content-Type' in request.headers
@@ -1250,7 +1250,7 @@ BUILTIN_HANDLERS = [
     ('before-sign.s3', remove_arn_from_signing_path),
     (
         'before-sign.polly.SynthesizeSpeech',
-        remove_content_type_header_for_polly_presigning,
+        remove_content_type_header_for_presigning,
     ),
     ('after-call.s3.ListObjects', decode_list_object),
     ('after-call.s3.ListObjectsV2', decode_list_object_v2),
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -1137,6 +1137,14 @@ def customize_endpoint_resolver_builtins(
         builtins[EndpointResolverBuiltins.AWS_S3_USE_GLOBAL_ENDPOINT] = True
 
 
+def remove_content_type_header_for_polly_presigning(request, **kwargs):
+    if (
+        request.context.get('is_presign_request') is True
+        and 'Content-Type' in request.headers
+    ):
+        del request.headers['Content-Type']
+
+
 # This is a list of (event_name, handler).
 # When a Session is created, everything in this list will be
 # automatically registered with that Session.
@@ -1240,6 +1248,10 @@ BUILTIN_HANDLERS = [
     ('before-parameter-build.route53', fix_route53_ids),
     ('before-parameter-build.glacier', inject_account_id),
     ('before-sign.s3', remove_arn_from_signing_path),
+    (
+        'before-sign.polly.SynthesizeSpeech',
+        remove_content_type_header_for_polly_presigning,
+    ),
     ('after-call.s3.ListObjects', decode_list_object),
     ('after-call.s3.ListObjectsV2', decode_list_object_v2),
     ('after-call.s3.ListObjectVersions', decode_list_object_versions),
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -682,13 +682,13 @@ def generate_presigned_url(
     if http_method is not None:
         request_dict['method'] = http_method
 
-    # If we're transforming to a method without formal body
-    # semantics we need to remove the Content-Type header.
-    if (
-        request_dict['method'].lower() in ('get', 'head', 'options')
-        and 'Content-Type' in request_dict['headers']
-    ):
-        del request_dict['headers']['Content-Type']
+        # If we're transforming to a method without formal body
+        # semantics we need to remove the Content-Type header.
+        if (
+            request_dict['method'].lower() in ('get', 'head', 'options')
+            and 'Content-Type' in request_dict['headers']
+        ):
+            del request_dict['headers']['Content-Type']
 
     # Generate the presigned url.
     return request_signer.generate_presigned_url(
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -682,6 +682,14 @@ def generate_presigned_url(
     if http_method is not None:
         request_dict['method'] = http_method
 
+    # Requests that are bodiless should not have the `Content-Type` header`
+    if (
+        request_dict['method'].lower()
+        in ('get', 'head', 'delete', 'options', 'trace')
+        and 'Content-Type' in request_dict['headers']
+    ):
+        del request_dict['headers']['Content-Type']
+
     # Generate the presigned url.
     return request_signer.generate_presigned_url(
         request_dict=request_dict,
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -682,7 +682,7 @@ def generate_presigned_url(
     if http_method is not None:
         request_dict['method'] = http_method
 
-    # Requests that are bodiless should not have the `Content-Type` header`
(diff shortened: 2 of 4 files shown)


tests/unit/test_signers.py
```diff
@@ -1168,3 +1175,37 @@ def test_custom_region(self):
...
+            'OutputFormat': 'mp3',
+            'Text': 'Hello world!',
+            'VoiceId': 'Joanna',
+        },
+        HttpMethod=request_method,
+    )
+    assert (
+        'Content-Type'
+        in generate_presigned_url_mock.call_args[1]['request_dict']['headers']
+    ) == content_type_present
```
nateprewitt, line 1211: Do we need a mock for this? It should just be in the final URL and that's the part we're actually concerned about.

botocore/signers.py
```diff
@@ -682,6 +682,13 @@ def generate_presigned_url(
     if http_method is not None:
         request_dict['method'] = http_method
 
+    # Requests that are bodiless should not have the `Content-Type` header.
+    if (
+        request_dict['method'].lower() in ('get', 'head', 'options')
+        and 'Content-Type' in request_dict['headers']
+    ):
+        del request_dict['headers']['Content-Type']
+
```
nateprewitt, line 691:
```suggestion
    if http_method is not None:
        request_dict['method'] = http_method

        # If we're transforming to a method without formal body
        # semantics we need to remove the Content-Type header.
        if (
            http_method.lower() in ('get', 'head', 'options')
            and 'Content-Type' in request_dict['headers']
        ):
            del request_dict['headers']['Content-Type']

```

nateprewitt (changes requested):

tests/unit/test_signers.py
dlm6693: Yeah tests were failing earlier since the signer needs to use credentials. I initially was checking the URL, but I chose to mimic this test suite which uses a mock. I'll see if I can just provide the client with dummy credentials instead.

dlm6693: Yeah tests were failing earlier since the signer needs to use credentials. I initially was checking the URL, but I chose to mimic this test suite which uses a mock. I'll see if I can just provide the client with dummy credentials instead.

nateprewitt: I think we already have [credentials set on the test class](https://github.com/boto/botocore/pull/2944/files#diff-c93558930e421bfe0fd7767dad14f92702b68bf878b550b3d231ccc09e7d47e8R51) in this test suite. We can do the same thing for your client fixture.

tests/unit/test_signers.py
```diff
@@ -1168,3 +1176,32 @@ def test_custom_region(self):
...
+    url = polly_client.generate_presigned_url(
+        'synthesize_speech',
+        Params={
+            'OutputFormat': 'mp3',
+            'Text': 'Hello world!',
+            'VoiceId': 'Joanna',
+        },
+        HttpMethod=request_method,
+    )
+    assert ('content-type' in url) == content_type_present
```
nateprewitt, line 1207: I was going to ask if we could be more specific with this assertion, but it turns out we only add this to the SignedHeaders section and not the query parameters themselves. That means anyone using the URL needs to also have a client to pass the headers. That's likely expected when a body is present, but we should validate this against a couple other SDKs first.

Let's see what Java v2 and JS v3 do in these cases.

nateprewitt (changes requested):

tests/unit/test_signers.py
dlm6693: Here's a `PutObject` example for Java v2. It looks like its expected to manually set headers after generating the presigned URL:
```Java
/ Upload content to the Amazon S3 bucket by using this URL.
            URL url = presignedRequest.url();

            // Create the connection and use it to upload the new object.
            HttpURLConnection connection = (HttpURLConnection) url.openConnection();
            connection.setDoOutput(true);
            connection.setRequestProperty("Content-Type","text/plain");
            connection.setRequestProperty("x-amz-meta-author","Mary Doe");
            connection.setRequestProperty("x-amz-meta-version","1.0.0.0");
```

dlm6693: I'm also seeing manually providing headers in the `PutObject` example in this blog post for the JS v3: https://aws.amazon.com/blogs/developer/generate-presigned-url-modular-aws-sdk-javascript/

```javascript
const response = await fetch(presignedUrl, {
    method: "PUT",
    body: payload,
    headers: {
         "Content-Length": statSync(filePath).size
    }
});
```

I can try to look for other examples if you like.

tests/unit/test_signers.py
```diff
@@ -39,6 +40,13 @@
 from tests import assert_url_equal, mock, unittest
 
 
+@pytest.fixture(scope='function')
```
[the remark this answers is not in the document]
dlm6693: I've removed it in the latest commit.

tests/unit/test_signers.py
nateprewitt: I think we're mostly concerned about the content of the string.

dlm6693: Sure. Here's a redacted example from JS v3. Note that `content-length` is in `SignedHeaders`:

```zsh
https://bar.s3.us-east-1.amazonaws.com/baz?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Content-Sha256=UNSIGNED-PAYLOAD&X-Amz-Credential=xxx%2F20230518%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20230518T212834Z&X-Amz-Expires=3600&X-Amz-Security-Token=xxx&X-Amz-Signature=xxx&X-Amz-SignedHeaders=content-length%3Bhost&x-id=PutObject

Comment (dlm6693) [reply tests/unit/test_signers.py:1207]: Sorry this took awhile, but here's a redacted Java v2 example.

```zsh
https://foo.s3.amazonaws.com/bar?X-Amz-Security-Token=xxx&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-
Date=20230518T223024Z&X-Amz-SignedHeaders=host&X-Amz-Expires=600&X-Amz-
Credential=xxx%2F20230518%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Signature=xxx

File: botocore/handlers.py [hunk 5]
```diff
@@ -1137,6 +1137,14 @@ def customize_endpoint_resolver_builtins(
         builtins[EndpointResolverBuiltins.AWS_S3_USE_GLOBAL_ENDPOINT] = True
 
 
+def remove_content_type_header_for_polly_presigning(request, **kwargs):
+    if (
+        request.context.get('is_presign_request') is True
+        and 'Content-Type' in request.headers
+    ):
+        del request.headers['Content-Type']
```
nateprewitt, line 1145:
```suggestion
def remove_content_type_header_for_presigning(request, **kwargs):
    if (
        request.context.get('is_presign_request') is True
        and 'Content-Type' in request.headers
    ):
        del request.headers['Content-Type']
```

botocore/handlers.py
```diff
@@ -1240,6 +1248,10 @@ def customize_endpoint_resolver_builtins(
     ('before-parameter-build.route53', fix_route53_ids),
     ('before-parameter-build.glacier', inject_account_id),
     ('before-sign.s3', remove_arn_from_signing_path),
+    (
+        'before-sign.polly.SynthesizeSpeech',
+        remove_content_type_header_for_polly_presigning,
```
nateprewitt, line 1253:
```suggestion
        remove_content_type_header_for_presigning,
```

tests/unit/test_signers.py
```diff
@@ -1168,3 +1176,23 @@ def test_custom_region(self):
...
+    url = polly_client.generate_presigned_url(
+        'synthesize_speech',
+        Params={
+            'OutputFormat': 'mp3',
+            'Text': 'Hello world!',
+            'VoiceId': 'Joanna',
+        },
+        HttpMethod=request_method,
+    )
+    assert 'content-type' not in url
```
nateprewitt, line 1198: We may run `.lower()` on the URL to remove any concerns about casing.

nateprewitt (changes requested):

nateprewitt (approved):
🚀

Issue #2626, opened 2022-02-28
Credentials logged in plain text


dmerz75:
The debug logging reports all of the response content with the access key revealed in plain text. Should the logger just report the response code, or perhaps the xml could be parsed to exclude the access key?

- Confirmed behavior in 1.23.51 and 1.24.8.

```
2022-02-28 13:22:03|DEBUG|endpoint:236|_do_get_response(): Sending http request:
```
This logger prints the AccessKeyId:
```
logger.debug("Sending http request: %s", request)
```
Output:
```
Sending http request: <AWSPreparedRequest stream_output=False, method=POST, url=https://sts.amazonaws.com/, headers={'Content-Type': b'application/x-www-form-urlencoded; charset=utf-8', 'User-Agent': b'Boto3/1.21.8 Python/3.7.11 Windows/10 Botocore/1.24.8', 'X-Amz-Date': b'20220228T182203Z', 'Authorization': b'AWS4-HMAC-SHA256 Credential=***ACTUAL_KEY***
```

```
2022-02-28 13:22:03|DEBUG|parsers:234|parse():
```
This logger prints the AccessKeyId as well:
```
LOG.debug('Response body:\n%s', response['body'])
```
Output:
```
Response body: <GetSessionTokenResult>\n    <Credentials>\n      <AccessKeyId>***ACTUAL_KEY***
```

nateprewitt:
Hi @dmerz75,

Debug logging in Botocore produces wire level logs for http which we have called out in the [documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/boto3.html). These are intended to be left untouched as raw bytes off the wire.

> ### Warning
> Be aware that when logging anything from 'botocore' the full wire trace will appear in your logs. If your payloads contain sensitive data this should not be used in production.

This is done prior to any parsing on the response, so we have no way to know what is "sensitive" at that point. This has been raised in a few issues previously, but we've arrived at keeping the current functionality as it's a requirement for many use cases. We could implement a general scrubber that tries to redact things that _look_ sensitive but this is imprecise, leading to obscuring relevant information.

dmerz75:
That's fine! Thanks - I think I'll end up using a decorator too, to prevent logging below INFO level. Users won't have access to this authorization portion. Resolved.
```
def info_logging_only(func):
    """Decorator that uses info level logging only."""

    def wrapper(*args, **kwargs):
        logger.info(f"Current log level: {logger.root.level}")
[...]

Issue #1211, opened 2017-06-01
Request: redact plaintext in kms calls in debug logging


2rs2ts:
Since kms can be used to encrypt sensitive data, leaking that data to disk is quite possible if users are not careful when enabling debug logging. It might be prudent to treat kms as a special case and make sure that plaintext is encrypted before being logged.

stealthycoin:
KMS is specifically server side encryption so it isn't going to get encrypted before being logged. Redacted possibly, but not encrypted since that would be client side encryption.

Marking this as a feature request.

2rs2ts:
Thanks, I misspoke when I said 

>that plaintext is encrypted before being logged

I totally meant redacted. Thanks a bunch for the response!

felipeochoa:
I'm using the following filter to achieve this. It redacts more than just the sensitive KMS data, but it may be a useful starting point for others:

```python
def redact_kms_logs(record: logging.LogRecord):
    "Redact the plaintext entries in KMS log records."
    if record.levelno > logging.DEBUG:
        return 1
    if 'body' in record.msg or 'headers' in record.msg:
        if isinstance(record.args, tuple):
            record.args = ('REDACTED',) * len(record.args)
        elif isinstance(record.args, dict):
            record.args = {key: 'REDACTED' for key in record.args}
        else:
            record.args = 'REDACTED'
    return 1


logging.getLogger('botocore.parsers').addFilter(redact_kms_logs)
```

PR #2945, opened 2023-05-18
Redact sensitive headers in AWSPreparedRequest repr method

jamesroberts:
Hi there,

This is a small change that redacts the information in two headers that contain credentials or authorization information in the [AWSPreparedRequest](https://github.com/boto/botocore/blob/97ba59024bdc313df6d18865f9ef233d24d60f95/botocore/awsrequest.py#L486) class. 

This object is [frequently logged](https://github.com/boto/botocore/blob/97ba59024bdc313df6d18865f9ef233d24d60f95/botocore/endpoint.py#L265) when sending requests to AWS services, and the `repr` method is returning a string that shows the contents of the request headers in plaintext. These headers contain customer signatures and credential information, which, ideally should not be logged as its an attack vector if logs are mismanaged. 

I understand that the [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/boto3.html#boto3.set_stream_logger) call out the use of DEBUG level logging and that wire level information that gets logged may contain sensitive data. I agree that it may be difficult to determine what is sensitive info vs what is not in all logs, making redacting information a hard problem. 

However, not having protection for fields that are generally known to contain sensitive customer information is a significant security hole.

Relying on a notice in the documentation to be wary of using debug mode only helps so much. If customers mistakenly or unknowingly set their log level to debug, Botocore should still offer some level of protection against logging customer secrets, signatures or other sensitive information. 

Issues I found related to or similar to logging sensitive info:
- #2626
- #1211
- https://github.com/boto/boto3/issues/2292
- https://github.com/boto/boto3/issues/1203

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -508,7 +508,8 @@ class AWSPreparedRequest:
             '<AWSPreparedRequest stream_output=%s, method=%s, url=%s, '
             'headers=%s>'
         )
-        return fmt % (self.stream_output, self.method, self.url, self.headers)
+        headers = self._redact_headers()
+        return fmt % (self.stream_output, self.method, self.url, headers)
 
     def reset_stream(self):
         """Resets the streaming body to it's initial position.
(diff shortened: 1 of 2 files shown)


nateprewitt:
Hi @jamesroberts, thanks for reaching out about this. Both the `Authorization` header and `X-Amz-Security-Token` are _not_ considered sensitive information. They are used for validation of the signature in the receiving service and included in plain-text with all presigned URLs that customers are free to share publicly without security concerns.

Please reach out through our other channels if you have any specific security concerns you'd like to discuss in more detail. Thanks!

PR #2946, opened 2023-05-19
fix: random crashes in credentials.py

markste-in:
os.path.sep and os.sep are supposed to do the same but os.path.sep can lead to random crashes in newer python versions with the following error

_AttributeError: module 'posixpath' has no attribute 'sep'_

This is also known to other projects and is caused by some broken python versions

https://github.com/pytest-dev/pytest/issues/10547
https://github.com/Toblerity/Fiona/issues/1143
https://github.com/pallets/jinja/issues/1697

The small proposed fix will avoid the issue altogether, even with broken python versions

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -651,7 +651,7 @@ class CachedCredentialFetcher:
 
     def _make_file_safe(self, filename):
         # Replace :, path sep, and / to make it the string filename safe.
-        filename = filename.replace(':', '_').replace(os.path.sep, '_')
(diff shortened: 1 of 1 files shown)


nateprewitt:
Thanks for the report, @markste-in. It's not clear for the other issues _why_ this is happening. All the explanation appear vague. I don't think this is an unreasonable change though. I'm going to close/open the issue to get CI running since it appears to have not run when you opened the PR. We'll take a closer look after that finishes.

nateprewitt (approved):
Thanks, @markste-in!

PR #2948, opened 2023-05-19
Fix unconditional key removal from `request.headers`

thejcannon:
This code mirrors https://github.com/boto/botocore/blob/97ba59024bdc313df6d18865f9ef233d24d60f95/botocore/auth.py#L191 and https://github.com/boto/botocore/blob/97ba59024bdc313df6d18865f9ef233d24d60f95/botocore/auth.py#L449

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -98,6 +98,8 @@ def _get_body_as_dict(request):
         data = json.loads(data)
     return data
 
+def _replace_header(headers, key, value):
+    
 
 class BaseSigner:
     REQUIRES_REGION = False
@@ -815,7 +817,8 @@ class HmacV1Auth(BaseSigner):
         self, method, split, headers, expires=None, auth_path=None
     ):
         if self.credentials.token:
-            del headers['x-amz-security-token']
+            if 'X-Amz-Security-Token' in request.headers:
+                del request.headers['X-Amz-Security-Token']
             headers['x-amz-security-token'] = self.credentials.token
         string_to_sign = self.canonical_string(
             method, split, headers, auth_path=auth_path


botocore/auth.py
```diff
@@ -98,6 +98,8 @@ def _get_body_as_dict(request):
         data = json.loads(data)
     return data
 
+def _replace_header(headers, key, value):
+
```
thejcannon, line 102:
```suggestion
```

botocore/auth.py
```diff
@@ -815,7 +815,8 @@ def get_signature(
         self, method, split, headers, expires=None, auth_path=None
     ):
         if self.credentials.token:
-            del headers['x-amz-security-token']
+            if 'X-Amz-Security-Token' in headers:
+                del headers['X-Amz-Security-Token']
```
kaos, line 819: or
```suggestion
            headers.pop('x-amz-security-token', None)
```

thejcannon: Nope: https://github.com/boto/botocore/blob/97ba59024bdc313df6d18865f9ef233d24d60f95/botocore/auth.py#L841

Welcome to botocore :stuck_out_tongue:

kaos: ???? nvm then.. 😮

jonemo:
This appears correct at first glance. Can you share how you discovered this problem? I am surprised that this code has existed for 9 years. Having some details on how to reproduce it will help find any related issues and create test coverage.

kaos:
In the linked issue (https://github.com/pantsbuild/pants/pull/19056) it is mentioned this was only an issue when using token-based auth.

thejcannon:
Admittedly, my usage was using a plain dict instead of the botocore types because we're already doing a bit of monkeypatching. I wanted to reduce how much from botocore we were using.

When I hit this behavior I figured the N other places something similar is being done (delete before replacement) meant this code was just either forgotten about or came later. IIUC the behavior is likely possible? It seems like the other locations are to avoid the multiple-values-for-one-header behavior.

After this issue, I found out that botocore is using the email message header class from stdlib and just switched to that. So our specific use case is OK. I suspect there is a still a bug with multiple headers though (likely only through the programmatic API).

PR #2950, opened 2023-05-22
Keep overrides when deep copying config value store

kdaily:
Config providers were updated for smart default support to be deep copy-able. This PR fixes a bug when deep copying the config value store. Previously, override values were not preserved. The fix adds the override values to the new copy of the value store.

A test was added to confirm that the overrides are preserved in the copy.

### DIFF
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -402,7 +402,11 @@ class ConfigValueStore:
                 self.set_config_provider(logical_name, provider)
 
     def __deepcopy__(self, memo):
-        return ConfigValueStore(copy.deepcopy(self._mapping, memo))
+        config_store = ConfigValueStore(copy.deepcopy(self._mapping, memo))
+        for logical_name, override_value in self._overrides.items():
+            config_store.set_config_variable(logical_name, override_value)
+
+        return config_store
 
     def get_config_variable(self, logical_name):
         """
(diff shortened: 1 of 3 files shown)


tests/unit/test_config_provider.py
```diff
@@ -363,6 +363,20 @@ def test_can_get_config_provider_non_chain_provider(self):
         self.assertIsInstance(provider, ConstantProvider)
         self.assertEqual(value, 'bar')
 
+    def test_deepcopy_preserves_overrides(self):
+        provider = ConstantProvider(100)
+        config_store = ConfigValueStore(mapping={'fake_variable': provider})
+        config_store.set_config_variable('fake_variable', 'foo')
```
kyleknap, line 369: Let's make the value more self-descriptive here to make the test easier to follow. I'd recommend just making the value `'override-value'` instead of `'foo'`

tests/unit/test_config_provider.py
```diff
@@ -363,6 +363,20 @@ def test_can_get_config_provider_non_chain_provider(self):
...
+
+        config_store_deepcopy = copy.deepcopy(config_store)
+
+        self.assertDictEqual(
+            {'fake_variable': 'foo'}, config_store_deepcopy._overrides
+        )
```
kyleknap, line 375: Let's avoid using internal members in the tests as they are not public interfaces. I'd say we should just remove this line and rely on the `get_config_variable` interface.

botocore/configprovider.py
```diff
@@ -402,7 +402,11 @@ def __init__(self, mapping=None):
                 self.set_config_provider(logical_name, provider)
```
kyleknap, line 402: Let's also make sure to add a `bugfix` changelog for this.

kyleknap (commented):
Looks good. I just had a few small comments.

kyleknap (approved):
Looks good :ship:

PR #2953, opened 2023-05-23
Shallow copy config value store and providers

kdaily:
Instead of deep copying, use a shallow copy of the config value store when using it for creating a new client from a session.

Update behavior of the smart defaults update functions to use a copy of the existing config providers in all cases. Update the logic for determining which new provider to create.

Add a `__copy__` method to all config providers that need it, or for uniformity.

Move unit tests for config provider smart defaults to functional tests and rework them to better test behavior. A mocked defaults  data file is used to confirm that the correct values are loaded to protect against changes to the live data file.

### DIFF
--- a/.changes/next-release/enhancement-configprovider-27540.json
+++ b/.changes/next-release/enhancement-configprovider-27540.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "configprovider",
+  "description": "Always use shallow copy of session config value store for clients"
+}
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -408,6 +408,13 @@ class ConfigValueStore:
 
         return config_store
 
+    def __copy__(self):
+        config_store = ConfigValueStore(copy.copy(self._mapping))
+        for logical_name, override_value in self._overrides.items():
+            config_store.set_config_variable(logical_name, override_value)
+
+        return config_store
+
     def get_config_variable(self, logical_name):
         """
         Retrieve the value associeated with the specified logical_name
@@ -546,24 +553,28 @@ class SmartDefaultsConfigStoreFactory:
         return 'standard'
 
     def _update_provider(self, config_store, variable, value):
-        provider = config_store.get_config_provider(variable)
+        original_provider = config_store.get_config_provider(variable)
         default_provider = ConstantProvider(value)
-        if isinstance(provider, ChainProvider):
-            provider.set_default_provider(default_provider)
-            return
-        elif isinstance(provider, BaseProvider):
+        if isinstance(original_provider, ChainProvider):
+            chain_provider_copy = copy.deepcopy(original_provider)
+            chain_provider_copy.set_default_provider(default_provider)
+            default_provider = chain_provider_copy
+        elif isinstance(original_provider, BaseProvider):
             default_provider = ChainProvider(
-                providers=[provider, default_provider]
+                providers=[original_provider, default_provider]
             )
         config_store.set_config_provider(variable, default_provider)
 
     def _update_section_provider(
         self, config_store, section_name, variable, value
     ):
-        section_provider = config_store.get_config_provider(section_name)
-        section_provider.set_default_provider(
+        section_provider_copy = copy.deepcopy(
+            config_store.get_config_provider(section_name)
+        )
+        section_provider_copy.set_default_provider(
             variable, ConstantProvider(value)
         )
+        config_store.set_config_provider(section_name, section_provider_copy)
 
     def _set_retryMode(self, config_store, value):
         self._update_provider(config_store, 'retry_mode', value)
--- a/tests/functional/test_config_provider.py
+++ b/tests/functional/test_config_provider.py
@@ -10,8 +10,11 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+from pathlib import Path
+
 import pytest
 
+import botocore.exceptions
 from botocore.config import Config
 from botocore.session import get_session
 
@@ -26,6 +29,13 @@ _SDK_DEFAULT_CONFIGURATION_VALUES_ALLOWLIST = (
 session = get_session()
 loader = session.get_component('data_loader')
 sdk_default_configuration = loader.load_data('sdk-default-configuration')
+
+
+def assert_client_uses_standard_defaults(client):
+    assert client.meta.config.s3['us_east_1_regional_endpoint'] == 'regional'
+    assert client.meta.config.connect_timeout == 3.1
+    assert client.meta.endpoint_url == 'https://sts.us-west-2.amazonaws.com'
+    assert client.meta.config.retries['mode'] == 'standard'
 
 
 @pytest.mark.parametrize("mode", sdk_default_configuration['base'])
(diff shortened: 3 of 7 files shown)


tests/functional/test_config_provider.py
```diff
@@ -49,3 +53,68 @@ def test_default_configurations_resolve_correctly():
...
+    session = botocore.session.Session()
+    session.register_component('data_loader', loader)
+    return session
+
+
+def config_store_updated(original_config_store, updated_config_store):
+    return original_config_store is not updated_config_store
+
+
+class TestSmartDefaultsConfigStoreFactory:
```
kyleknap, line 77: The name of this class is too specific to the implementation details. Let's rename it to `TestConfigurationDefaults` to emphasize we are testing end to end configuration default resolution.

tests/functional/test_config_provider.py
```diff
@@ -49,3 +53,68 @@ def test_default_configurations_resolve_correctly():
...
+    def test_defaults_mode_resolved_invalid_mode_exception(self, session):
+        with pytest.raises(botocore.exceptions.InvalidDefaultsMode):
+            config = Config(defaults_mode='foo')
+            session.create_client('sts', 'us-west-2', config=config)
+
+    @mock.patch(
+        'botocore.session.SmartDefaultsConfigStoreFactory.merge_smart_defaults'
+    )
+    def test_defaults_mode_resolved_legacy(
+        self, merge_smart_defaults, session
```
kyleknap, line 117: We should also have a functional test that after creating a client who was configured to use the `standard` defaults that it will not affect the config provider on the session.

botocore/configprovider.py
```diff
@@ -546,24 +553,26 @@ def resolve_auto_mode(self, region_name):
...
+        elif isinstance(provider, ChainProvider):
             provider.set_default_provider(default_provider)
-            return
         elif isinstance(provider, BaseProvider):
-            default_provider = ChainProvider(
-                providers=[provider, default_provider]
-            )
-        config_store.set_config_provider(variable, default_provider)
+            provider = ChainProvider(providers=[provider, default_provider])
+        config_store.set_config_provider(variable, provider)
```
kyleknap, line 564: We should make sure to add unit tests for the smart default factory that we do not mutate the chain providers nor the section providers (nor their sub-chain providers) as well.

kyleknap (commented):
Looks like a good start. I think we may need to tweak our approach slightly to do deep copies on the merge logic instead of shallow copies, while keeping the shallow copy on the config provider. Otherwise, most of my comments were about tests.

tests/unit/test_config_provider.py
```diff
@@ -373,6 +373,36 @@ def test_deepcopy_preserves_overrides(self):
         value = config_store_deepcopy.get_config_variable('fake_variable')
         self.assertEqual(value, 'override-value')
 
+    def test_can_copy(self):
```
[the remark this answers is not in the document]
kdaily: Mutating the copy of the config providers would need to be done through private methods or attributes, is that OK?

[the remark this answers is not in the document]
kdaily: The test should be used for the config value store.

tests/functional/test_config_provider.py
kdaily: Added this as `test_no_mutate_session_provider`

tests/unit/test_config_provider.py
```diff
@@ -667,11 +717,71 @@ def _create_os_environ_patcher(self):
...
+        )
+        assert sts_regional_endpoints_before != sts_regional_endpoints_after
+
+        assert s3_section_provider.provide() == s3_section_before
+        assert s3_section_before != s3_section_after
+
+        assert retry_mode_provider.provide() == retry_mode_before
+        assert retry_mode_before != retry_mode_after
+
+    def test_config_store_providers_not_added_after_copy(self):
```
kyleknap, line 773: Do we need this test case? It seems like it is already covered here: https://github.com/boto/botocore/blob/1e85087e96651b413f624821ac9226faf21f4547/tests/unit/test_config_provider.py#L405

tests/unit/test_config_provider.py
```diff
@@ -667,11 +717,71 @@ def _create_os_environ_patcher(self):
...
+            'constant_value_copy', ConstantProvider('123')
+        )
+        assert config_store.get_config_variable('constant_value_copy') is None
+        assert config_store_copy.get_config_variable('constant_value') == 'ABC'
+
+    @pytest.fixture
+    def config_value_store(self):
+        return self._create_config_value_store()
+
+    def test_config_store_providers_not_mutated_after_merge(
```
kyleknap, line 734: It might be worth seeing if we can parameterize this by the smart config variable name to make it easier to read and potentially add to in the future.

kyleknap (commented):
Looks good. Most of my comments were around how we can make the tests a bit easier to read.

tests/unit/test_config_provider.py
kdaily: Sure, it was testing adding a different one to another variable, but essentially its the same thing. Removing.

tests/unit/test_config_provider.py
```diff
@@ -643,17 +691,19 @@ def fake_session(self):
...
         # Testing with three different providers to validate
         # SmartDefaultsConfigStoreFactory._get_new_chain_provider
         mapping = {
-            'sts_regional_endpoints': ChainProvider(providers=[provider_foo]),
-            'retry_mode': ChainProvider(providers=[environment_provider_foo]),
+            'sts_regional_endpoints': ChainProvider(
+                providers=[constant_provider]
+            ),
+            'retry_mode': ChainProvider(providers=[environment_provider]),
             's3': SectionConfigProvider('s3', fake_session, s3_mapping),
```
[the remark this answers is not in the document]
kdaily: I think I took care of this but needed to update one of my assertions that the value of the provider had changed, because if you set a value in the mapping to a provider, updating the section provider adds it to the end, so you still get the same value.

.changes/next-release/enhancement-configprovider-27540.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "configprovider",
+  "description": "Use shallow copy instead of deep copy for config value store"
```
kyleknap, line 4: Let's rephrase this to be more specific on where we are introducing the shallow copies e.g. `Always use shallow copy of session config value store for clients. This ensures that updates to the session's config value store cannot change client configuration after client instantiation.`

tests/unit/test_config_provider.py
```diff
@@ -373,6 +373,60 @@ def test_deepcopy_preserves_overrides(self):
...
 
+    def test_copy_preserves_provider_identities(self):
+        fake_variable_provider = ConstantProvider(100)
+        another_variable_provider = EnvironmentProvider(
+            name='AWS_ANOTHER_VARIABLE', env=dict(AWS_ANOTHER_VARIABLE='123')
+        )
+        config_store = ConfigValueStore(
+            mapping={
+                'fake_variable': fake_variable_provider,
+                'another_variable': another_variable_provider,
```
kyleknap, line 384: Can we remove the `another_variable` from the `ConfigValueStore`? It does not appear to be used or asserted against and makes the test longer than it needs to be.

tests/unit/test_config_provider.py
```diff
@@ -373,6 +373,60 @@ def test_deepcopy_preserves_overrides(self):
...
+
+        another_variable_provider = ConstantProvider('ABC')
+
+        config_store_copy.set_config_provider(
+            'fake_variable', another_variable_provider
+        )
+
+        assert (
+            config_store.get_config_provider('fake_variable')
+            is fake_variable_provider
```
kyleknap, line 423: For these assertions, I'd expect we would be getting the values for those configuration variables. Mainly I want to make sure we are returning the expected values back to make sure the input/output of the two config stores are exactly what we expect.

tests/unit/test_config_provider.py
```diff
@@ -667,11 +723,102 @@ def _create_os_environ_patcher(self):
...
-        config_store_copy.set_config_provider('fizz', ConstantProvider('buzz'))
-        assert config_store.get_config_variable('fizz') is None
-        assert config_store_copy.get_config_variable('foo') == 'bar'
+        config_store_copy.set_config_provider(
+            'constant_value_copy', ConstantProvider('123')
+        )
+        assert config_store.get_config_variable('constant_value_copy') is None
+        assert config_store_copy.get_config_variable('constant_value') == 'ABC'
+
+    # @pytest.fixture
```
kyleknap, line 736: Let's remove this instead of commenting it out.

tests/unit/test_config_provider.py
```diff
@@ -667,11 +723,102 @@ def _create_os_environ_patcher(self):
...
+    # def config_value_store(self):
+    #     return self._create_config_value_store()
+
+    @pytest.mark.parametrize(
+        'defaults_mode',
+        ['standard', 'in-region', 'cross-region', 'mobile', 'auto'],
```
kyleknap, line 742: I think I may have not been clear here by my previous comment. I meant parameterize by config store parameter name (e.g. `retry_mode`, `s3`, `sts_regional_endpoints` ). There is a bunch of duplicate code here around retrieving config providers and making assertions and parameterization will allow us to make the test body much simpler.

tests/unit/test_config_provider.py
```diff
@@ -667,11 +723,102 @@ def _create_os_environ_patcher(self):
...
+            == sts_regional_endpoints_before
+        )
+        assert sts_regional_endpoints_before != sts_regional_endpoints_after
+
+        assert s3_section_provider.provide() == s3_section_before
+
+        assert retry_mode_provider.provide() == retry_mode_before
+        assert retry_mode_before != retry_mode_after
+
+    def test_copy_update_does_not_mutate_source_config_store(self):
```
kyleknap, line 796: I thought we were planning to remove this test in favor of the one in `TestConfigValueStore`?

tests/unit/test_config_provider.py
```diff
@@ -667,11 +723,102 @@ def _create_os_environ_patcher(self):
...
+        )
+
+        sts_regional_endpoints_after = config_value_store.get_config_variable(
+            'sts_regional_endpoints'
+        )
+        retry_mode_after = config_value_store.get_config_variable('retry_mode')
+
+        assert (
+            sts_regional_endpoints_provider.provide()
+            == sts_regional_endpoints_before
```
kyleknap, line 787: And when we parameterize the config variable names, let's also add explicit parameter values for what we expect the before and after values are expected to be. When we generalize it like this it make it easy to cover up issues in the test where the values can match our expectations (e.g. the before and after values are not equal) but the values themselves are not as expected (e.g. one of the values is `None` instead of a string value).

kyleknap (commented):
It's looking better. I just had some more comments on the tests

.changes/next-release/enhancement-configprovider-27540.json
kdaily: Agree for more details, but is that completely accurate? Wouldn't it be:

> This ensures that updates to a config value store used for client instantiation cannot change the session's config value store or subsequent client instantiations.

kyleknap: That is true as well but that would seem to indicate the change was more to fix a bug. I'm also fine with just removing the second sentence and leaving it with the first one

kyleknap (approved):
Looks good! :ship:

PR #2955, opened 2023-05-26
User Agent 2.0

jonemo:
This PR implements an updated User-Agent header format that is standardized across AWS SDKs. For example, the AWS Ruby SDK's equivalent PR is https://github.com/aws/aws-sdk-ruby/pull/2854.

An example User-Agent header value before the change is:

```
Boto3/1.20.32 Python/3.9.13 Darwin/22.4.0 exec-env/AWS_Lambda_python3.7 Botocore/1.23.32
```

An example User-Agent header value after the change is:

```
Boto3/1.20.32 md/Botocore#1.23.32 ua/2.0 lang/python#3.7.16 os/linux#4.14.255-301-238.520.AMZN2.X86_64 exec-env/AWS_Lambda_python3.7
```

Various notes:

* This PR includes the formerly separate (but then closed) change to add the `AWS_SDK_UA_APP_ID` config: https://github.com/boto/botocore/pull/2942
* Users who use the `Config.user_agent` will not see a change in their User-Agent header string
* Both `Session.user_agent_extra` and `Config.user_agent_extra` continue to be appear at the end of the User-Agent header
* `Session.user_agent()` is no longer used internally in Botocore but remains in place and unchanged for any third party users
* In contrast to the Ruby SDK PR linked above, this PR does not include feature-related or any other request-level information in the User-Agent string.


botocore/useragent.py
```diff
@@ -0,0 +1,421 @@
...
+        )
+        clean_name = sanitize_user_agent_string_component(
+            self.name, allow_hash=False
+        )
+        if self.value is None or self.value == '':
+            return f'{clean_prefix}/{clean_name}'
+        clean_value = sanitize_user_agent_string_component(
+            self.value, allow_hash=True
+        )
+        return f'{clean_prefix}/{clean_name}#{clean_value}'
```
nateprewitt, line 84: Two quick notes here:

1. Are these only ever called once or do we recompute per request? If these are immutable, it seems like we could cache this work. Either by doing this during initialization or some other means of memoizing the work since these are immutable.

2. We're adding `to_string` in a few places instead of `__str__`. Is there a reason we want to keep those separate or is this a Java-ism creeping in from the design?

botocore/useragent.py
```diff
@@ -0,0 +1,421 @@
...
+    """
+
+    def __init__(
+        self,
+        platform_name,
+        platform_version,
+        python_version,
+        python_implementation,
+        execution_env,
+        crt_version=None,
```
nateprewitt, line 139: It's interesting that the CRT is a top level argument. Do we envision adding more attributes here as other core features are added or do we need something more general to expand metadata/feature info?

botocore/useragent.py
```diff
@@ -0,0 +1,421 @@
...
+        information from the start of the string, Botocore's name and version
+        are included as a separate field with "md" prefix.
+        """
+        if (
+            self._session_user_agent_name
+            and self._session_user_agent_version
+            and (
+                self._session_user_agent_name != _USERAGENT_SDK_NAME
+                or self._session_user_agent_version != botocore_version
+            )
```
nateprewitt, line 261: When conditionals start to get more complex, I'm usually in favor of breaking them into discrete variables. Not required, just a note.

botocore/useragent.py
```diff
@@ -0,0 +1,421 @@
...
+        of standardized OS names, a single component with prefix "os" is
+        returned. Otherwise, one component "os/other" is returned and a second
+        with prefix "md" and the raw platform name.
+
+        String representations of example return values:
+         * ``os/macos#10.13.6``
+         * ``os/linux``
+         * ``os/other``
+         * ``os/other md/foobar#1.2.3``
+         * ``os/linux md/arch#x86_64``
```
nateprewitt, line 289: We'd talked about this elsewhere, but this still strikes me as somewhat odd. Is `md/` that follows a more specific prefix always related or do you need to have implicit knowledge to associate entries?

Edit: Actually looking at the code below, how does the `arch` case work?

We either return one entry that's `os/{os_family}#{self._platform_version}` or `os/other md/{...}`. I don't see how we'd get linux + the md but maybe I'm missing it 😅

botocore/useragent.py
```diff
@@ -0,0 +1,421 @@
...
+    def _build_language_metadata(self):
+        """
+        Build the language components of the User-Agent header string.
+
+        Returns the Python version in a component with prefix "lang" and name
+        "python". The Python implementation (e.g. CPython, PyPy) is returned as
+        separate metadata component with prefix "md" and name "pyimpl".
+
+        String representation of an example return value:
+        ``lang/python/3.10.4 md/pyimpl/Cpython``
```
nateprewitt, line 323:
```suggestion
        ``lang/python/3.10.4 md/pyimpl/CPython``
```

nateprewitt: I find this choice interesting when you look at Cython, Jython, IronPython, etc where the version may not match a Python version. This brings up a question about how robust this needs to be. Do we have enough encapsulation that the client will never fail during user_agent construction? If so, what's our fallback?

That may be answered below so feel free to ignore (or I should come back and delete this).

botocore/useragent.py
```diff
@@ -0,0 +1,421 @@
...
+        from the environment variable AWS_EXECUTION_ENV.
+        """
+        if self._execution_env:
+            return [UserAgentComponent('exec-env', self._execution_env)]
+        else:
+            return []
+
+    def _build_feature_metadata(self):
+        """
+        Build the features componentes of the User-Agent header string.
```
nateprewitt, line 348:
```suggestion
        Build the features components of the User-Agent header string.
```

botocore/useragent.py
```diff
@@ -0,0 +1,421 @@
...
+
+        Returns a list of components with prefix "cfg" followed by the config
+        setting name and its value. Tracked configuration settings may be
+        added or removed in future versions.
+        """
+        if not self._client_config or not self._client_config.retries:
+            return []
+        retry_mode = self._client_config.retries.get('mode')
+        cfg_md = [UserAgentComponent('cfg', 'retry-mode', retry_mode)]
+        if self._client_config.endpoint_discovery_enabled:
```
nateprewitt, line 367: This seems unfortunately manual. It's a one off for now, but hopefully we have a more robust way of handling in the future. We don't need to engineer it now but if we expand it may be worth wiring something in the ConfigProvider itself.

botocore/utils.py
```diff
@@ -3315,3 +3315,12 @@ def _serialize_if_needed(self, value, iso=False):
...
         return value
+
+
+def get_crt_version():
+    try:
+        import awscrt
+
+        return awscrt.__version__
+    except AttributeError:
+        return None
```
nateprewitt, line 3326: Ideally this would remain private but I don't know if we can do that cleanly while reusing. Perhaps we at least leave a note indicating it's not for public use and subject to change.

tests/functional/test_useragent.py
```diff
@@ -0,0 +1,330 @@
...
+
+import pytest
+
+from botocore import __version__ as botocore_version
+from botocore.config import Config
+from botocore.stub import Stubber
+
+
+@contextmanager
+def uacap_client(client):
```
nateprewitt, line 26: Does this give us a testing efficiency over the HTTPStubber?

tests/functional/test_useragent.py
```diff
@@ -0,0 +1,330 @@
...
+    uafields = cap_client.captured_ua_string.split(' ')
+    assert 'app/123456' in uafields
+
+
+def test_user_agent_long_appid_yields_warning(
+    patched_session, stubbed_list_buckets, caplog
+):
+    # user_agent_appid config values longer than 50 characters should result
+    # in a warning
+    sixtychars = '000000000011111111112222222222333333333344444444445555555555'
```
nateprewitt, line 221: nit; we could assert the length on this

tests/functional/test_useragent.py
```diff
@@ -0,0 +1,330 @@
...
+    client_s3 = patched_session.create_client('s3', config=client_cfg)
+    with uacap_client(client_s3) as cap_client:
+        with stubbed_list_buckets(cap_client):
+            with caplog.at_level(logging.INFO):
+                cap_client.list_buckets()
+
+    assert (
+        'The configured value for user_agent_appid exceeds the maximum length'
+        in caplog.text
+    )
```
nateprewitt, line 232: Where is the truncation actually being done, is this deferred to server side? Does this provide a DOS vector with arbitrarily large values?

If it's client side, we may add that assertion to this/a test.

tests/functional/test_useragent.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+def test_user_agent_appid_gets_sanitized(
+    patched_session, stubbed_list_buckets, caplog
+):
+    # Parentheses and the copyright symbol are not valid characters in the user
```
nateprewitt, line 238: Was © removed or is that just an example? I only see a paren in the badchars string.

tests/functional/test_useragent.py
```diff
@@ -0,0 +1,330 @@
...
+        with stubbed_list_buckets(cap_client):
+            with caplog.at_level(logging.INFO):
+                cap_client.list_buckets()
+
+    # given string should be truncated to 50 characters
+    uafields = cap_client.captured_ua_string.split(' ')
+    assert 'app/1234-' in uafields
+
+
+def test_boto3_behavior(patched_session, stubbed_list_buckets):
```
nateprewitt, line 253: nit; We may want to be more specific than behavior.
```suggestion
def test_boto3_user_agent(patched_session, stubbed_list_buckets):
```

nateprewitt (changes requested):

botocore/useragent.py
```diff
@@ -0,0 +1,421 @@
...
+    request context. ``UserAgentString`` provides methods for collecting the
+    information and ``to_string`` for assembling it into the standardized
+    string format.
+
+    Example usage:
+
+        ua_session = UserAgentString.from_environment()
+        ua_session.set_session_config(...)
+        ua_client = ua_session.with_client_config(Config(...))
+        ua_string = ua_request.to_string()
```
jonathan343, line 117: There seems to be inconsistent naming here. Both should be `ua_client` or `ua_request`.

tests/functional/test_useragent.py
```diff
@@ -0,0 +1,330 @@
...
+def uacap_client(client):
+    """Contextmanager yielding client object that captures the last UA header
+
+    Usage:
+
+        with uacap_client(myclient) as cap_client:
+            myclient.operation()
+        print(cap_client.captured_ua_string)
+    """
+    client.captured_ua_string = None
```
jonathan343, line 35: Would there be any value to add or replace this with a list that captures sequential `ua_string`'s?

tests/functional/test_useragent.py
jonemo: The example in the spec changed and I failed to update the comment. As a supporter of inline comments I should really be better at keeping them in sync 😞

botocore/config.py
```diff
@@ -38,6 +38,12 @@ class Config:
     :param user_agent_extra: The value to append to the current User-Agent
         header value.
 
+    :type user_agent_appid: str
+    :param user_agent_appid: A value that gets included in the User-Agent
+        string in the format "app/<user_agent_appid>" which is consistent
```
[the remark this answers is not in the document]
jonemo: Good point. Dropped the reference in 34edd4b.

botocore/useragent.py
```diff
@@ -0,0 +1,421 @@
...
+
+
+def sanitize_user_agent_string_component(raw_str, allow_hash):
+    """Replaces all not allowed characters in the string with a dash ("-").
+
+    Allowed characters are ASCII alphanumerics and ``!$%&'*+-.^_`|~``. If
+    ``allow_hash`` is ``True``, "#"``" is also allowed.
+
+    :type allow_hash: bool
+    :param allow_hash: Whether "#" is considered an allowed character.
```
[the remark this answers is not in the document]
jonemo: No reason. Fixed in 34edd4b.

tests/functional/test_useragent.py
jonemo: I had initially coded up truncation but then realized that the spec only asks for a warning to be raised. So I dropped truncation (and had missed one mention of truncation in a comment that is now removed in 34edd4b).

tests/functional/test_useragent.py
jonemo: Done in 34edd4b24

botocore/args.py
```diff
@@ -646,3 +650,19 @@ def compute_endpoint_resolver_builtin_defaults(
...
+            )
+        if (
+            user_agent_appid is not None
+            and len(user_agent_appid) > USERAGENT_APPID_MAXLEN
+        ):
+            logger.warning(
+                'The configured value for user_agent_appid exceeds the '
+                f'maximum length of {USERAGENT_APPID_MAXLEN} characters.'
+            )
+        config_kwargs['user_agent_appid'] = user_agent_appid
```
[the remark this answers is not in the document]
jonemo: Same answer as https://github.com/boto/botocore/pull/2955#discussion_r1220383128:

> I had initially coded up truncation but then realized that the spec only asks for a warning to be raised. So I dropped truncation (and had missed one mention of truncation in a comment that is now removed in https://github.com/boto/botocore/commit/34edd4b24c5bf072edd8d6292dbe2ee23562a0ac).

botocore/args.py
```diff
@@ -249,7 +254,6 @@ def compute_client_args(
         return {
             'service_name': service_name,
             'parameter_validation': parameter_validation,
-            'user_agent': user_agent,
```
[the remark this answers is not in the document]
jonemo: We can bring back this line together with [lines 196-203 above](https://github.com/boto/botocore/pull/2955/files#diff-019afcd91cca6d46dd1046132b61e78198063592925d01e9a4db1fd9f1bf8c3fL196-L203) where its value gets computed using the legacy method. There is no harm in returning the `user_agent` key in the dict, it simply won't get used by botocore code. Do you think that's preferable? I was worried the confusion caused by the dead code existing and an unused value being returned might outweigh any benefit to any third-party that may be relying on the `user_agent` dict key being present.

botocore/useragent.py
jonemo: I don't disagree with the naming being somewhat confusing. But I don't think your suggestions would clarify things. In fact your suggestions might be a sign that I successfully confused you!

* `ua_session` is a `UserAgentString` object with session-level information only.
* `ua_client` is a `UserAgentString` object with session-level and client-level information.
* `ua_string` is a `str`.

botocore/useragent.py
jonemo: The only assumption we make is that we get a string back from [platform.python_implementation](https://docs.python.org/3.8/library/platform.html#platform.python_implementation). I believe that can be relied on.

botocore/args.py
```diff
@@ -66,13 +69,14 @@ def __init__(
         loader,
         exceptions_factory,
         config_store,
+        user_agent_creator,
```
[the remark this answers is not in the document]
jonemo: bda23a95a is a refactor that prioritizes backwards compatible interfaces over everything (especially over avoiding duplication).

botocore/useragent.py
jonemo: e19e237 addresses the fact that `crt_version` wasn't actually getting used yet. Including it at all, and as a top level "environment" property, is to maintain [current botocore behavior](https://github.com/boto/botocore/blob/develop/botocore/session.py#L554). The spec doesn't mention the CRT version at all.

botocore/useragent.py
jonemo: > Edit: Actually looking at the code below, how does the arch case work?

Fixed in e19e237

> Is md/ that follows a more specific prefix always related or do you need to have implicit knowledge to associate entries?

The idea is that they are related. Or at least that's my understanding of how the spec came about. I agree that this example of `arch` following `os` seems like a tenuous relationship.

botocore/useragent.py
```diff
@@ -0,0 +1,421 @@
...
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import os
```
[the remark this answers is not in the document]
jonemo: Done in f08429b

botocore/utils.py
jonemo: I found a way to keep _get_crt_version private 66ad86a, at the cost of one cross-module call to a private function.

botocore/useragent.py
jonemo: Agreed, once there is a longer list of config settings to track, a more generic solution to this would be desirable. Our organically grown collection of interfaces won't make this easier because some things others might consider "configuration" comes from a mix of sources in Botocore including the config provider, config store, session object properties, create_client() arguments, and so on...

tests/functional/test_useragent.py
jonemo: Done in eb7ded6

tests/functional/test_useragent.py
jonemo: I ended up replacing this entire fixture in eb7ded6 to address another comment above. To answer the question though: I didn't see a reason to do more than one request in a single test because each request's UA is independent of the others.

botocore/useragent.py
```diff
@@ -0,0 +1,483 @@
...
+_USERAGENT_ALLOWED_OS_NAMES = [
+    'windows',
+    'linux',
+    'macos',
+    'android',
+    'ios',
+    'watchos',
+    'tvos',
+    'other',
+]
```
nateprewitt, line 44: nit; constants are ideally tuples for fixed memory size and immutability.

nateprewitt (commented):
2 minor comments, otherwise I think this looks ready to go. Thanks @jonemo!

botocore/useragent.py
jonemo: Done in b0a29f8

nateprewitt (approved):
:shipit:

PR #2958, opened 2023-05-26
Implement configured endpoint url resolver

kdaily:
Adds a config provider to resolve the endpoint URL provided in an environment variable or shared configuration file. It resolves the endpoint in the following manner:

1. The value provided through the `endpoint_url` parameter provided to the client constructor.
2. The value provided by a service-specific environment variable.
3. The value provided by the global endpoint environment variable (`AWS_ENDPOINT_URL`).
4. The value provided by a service-specific parameter from a services definition section in the shared configuration file.
5. The value provided by the global parameter from a services definition section in the shared configuration file.
6. The value resolved when no configured endpoint URL is provided.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -107,7 +107,7 @@ class ClientArgsCreator:
         s3_config = final_args['s3_config']
         partition = endpoint_config['metadata'].get('partition', None)
         socket_options = final_args['socket_options']
-
+        configured_endpoint_url = final_args['configured_endpoint_url']
         signing_region = endpoint_config['signing_region']
         endpoint_region_name = endpoint_config['region_name']
 
@@ -152,7 +152,7 @@ class ClientArgsCreator:
             service_model,
             endpoint_region_name,
             region_name,
-            endpoint_url,
+            configured_endpoint_url,
             endpoint,
             is_secure,
             endpoint_bridge,
@@ -202,10 +202,16 @@ class ClientArgsCreator:
                 user_agent += ' %s' % client_config.user_agent_extra
 
         s3_config = self.compute_s3_config(client_config)
+
+        configured_endpoint_url = self._compute_configured_endpoint_url(
+            client_config=client_config,
+            endpoint_url=endpoint_url,
+        )
+
         endpoint_config = self._compute_endpoint_config(
             service_name=service_name,
             region_name=region_name,
-            endpoint_url=endpoint_url,
+            endpoint_url=configured_endpoint_url,
             is_secure=is_secure,
             endpoint_bridge=endpoint_bridge,
             s3_config=s3_config,
@@ -250,6 +256,7 @@ class ClientArgsCreator:
             'service_name': service_name,
             'parameter_validation': parameter_validation,
             'user_agent': user_agent,
+            'configured_endpoint_url': configured_endpoint_url,
             'endpoint_config': endpoint_config,
             'protocol': protocol,
             'config_kwargs': config_kwargs,
@@ -258,6 +265,25 @@ class ClientArgsCreator:
                 scoped_config, client_config
             ),
         }
+
+    def _compute_configured_endpoint_url(self, client_config, endpoint_url):
+        if endpoint_url is not None or self._ignore_configured_endpoint_urls(
+            client_config
+        ):
+            return endpoint_url
+
+        return self._config_store.get_config_variable('endpoint_url')
+
+    def _ignore_configured_endpoint_urls(self, client_config):
+        if (
+            client_config
+            and client_config.ignore_configured_endpoint_urls is not None
+        ):
+            return client_config.ignore_configured_endpoint_urls
+
+        return self._config_store.get_config_variable(
+            'ignore_configured_endpoint_urls'
+        )
 
     def compute_s3_config(self, client_config):
         s3_configuration = self._config_store.get_config_variable('s3')
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -18,6 +18,7 @@ import logging
 import os
 
 from botocore import utils
+from botocore.exceptions import InvalidConfigError
 
 logger = logging.getLogger(__name__)
 
@@ -105,6 +106,12 @@ BOTOCORE_DEFAUT_SESSION_VARIABLES = {
     'use_fips_endpoint': (
         'use_fips_endpoint',
         'AWS_USE_FIPS_ENDPOINT',
+        None,
+        utils.ensure_boolean,
+    ),
+    'ignore_configured_endpoint_urls': (
+        'ignore_configured_endpoint_urls',
+        'AWS_IGNORE_CONFIGURED_ENDPOINT_URLS',
         None,
         utils.ensure_boolean,
     ),
@@ -851,3 +858,159 @@ class ConstantProvider(BaseProvider):
 
     def __repr__(self):
         return 'ConstantProvider(value=%s)' % self._value
+
+
+class ConfiguredEndpointProvider(BaseProvider):
+    """Lookup an endpoint URL from environment variable or shared cong file.
+
+    NOTE: This class is considered private and is subject to abrupt breaking
+    changes or removal without prior announcement. Please do not use it
+    directly.
+    """
+
+    _ENDPOINT_URL_LOOKUP_ORDER = [
+        'environment_service',
+        'environment_global',
+        'config_service',
+        'config_global',
+    ]
+
+    def __init__(
+        self,
+        full_config,
+        scoped_config,
+        client_name,
+        environ=None,
+    ):
+        """Initialize a ConfiguredEndpointProviderChain.
+
+        :type full_config: dict
+        :param full_config: This is the dict representing the full
+            configuration file.
+
+        :type scoped_config: dict
+        :param scoped_config: This is the dict representing the configuration
+            for the current profile for the session.
+
+        :type client_name: str
+        :param client_name: The name used to instantiate a client using
+            botocore.session.Session.create_client.
+
+        :type environ: dict
+        :param environ: A mapping to use for environment variables. If this
+            is not provided it will default to use os.environ.
+        """
+        self._full_config = full_config
+        self._scoped_config = scoped_config
+        self._client_name = client_name
+
+        if environ is None:
+            environ = os.environ
+        self._environ = environ
+
+    def provide(self):
+        """Lookup the configured endpoint URL.
+
+        The order is:
+
+        1. The value provided by a service-specific environment variable.
+        2. The value provided by the global endpoint environment variable
+           (AWS_ENDPOINT_URL).
+        3. The value provided by a service-specific parameter from a services
+           definition section in the shared configuration file.
+        4. The value provided by the global parameter from a services
+           definition section in the shared configuration file.
+        """
+        for location in self._ENDPOINT_URL_LOOKUP_ORDER:
+            logger.debug(
+                'Looking for endpoint for %s via: %s',
+                self._client_name,
+                location,
+            )
+
+            endpoint_url = getattr(self, f'_get_endpoint_url_{location}')()
+
+            if endpoint_url:
+                logger.info(
+                    'Found endpoint for %s via: %s.',
+                    self._client_name,
+                    location,
+                )
+                return endpoint_url
+
+        logger.debug('No configured endpoint found.')
+        return None
+
+    def _get_snake_case_service_id(self, client_name):
+        # Use lookups to get the serice ID without loading the service data
+        # file. `client_name`` refers to the name used to instantiate a client
+        # through botocore.session.Session.create_client (parameter name there
+        # is `service_name`). This client name is generally the hyphenized
+        # service ID.  We need to look up services that have been renamed
+        # (SERVICE_NAME_ALIASES) as well as services that do not use
+        # the service ID as their data directory name
+        # (CLIENT_NAME_TO_HYPHENIZED_SERVICE_ID_OVERRIDES).
+        client_name = utils.SERVICE_NAME_ALIASES.get(client_name, client_name)
+        hyphenized_service_id = (
+            utils.CLIENT_NAME_TO_HYPHENIZED_SERVICE_ID_OVERRIDES.get(
+                client_name, client_name
+            )
+        )
+        return hyphenized_service_id.replace('-', '_')
+
+    def _get_service_env_var_name(self):
+        transformed_service_id_env = self._get_snake_case_service_id(
+            self._client_name
+        ).upper()
+        return f'AWS_ENDPOINT_URL_{transformed_service_id_env}'
+
+    def _get_services_config(self):
+        if 'services' not in self._scoped_config:
+            return {}
+
+        section_name = self._scoped_config['services']
+        services_section = self._full_config.get('services', {}).get(
+            section_name
+        )
+
+        if not services_section:
+            error_msg = (
+                f'The profile is configured to use the services '
+                f'section but the "{section_name}" services '
+                f'configuration does not exist.'
+            )
+            raise InvalidConfigError(error_msg=error_msg)
+
+        return services_section
+
+    def _get_endpoint_url_config_service(self):
+        snakecase_service_id = self._get_snake_case_service_id(
+            self._client_name
+        ).lower()
+        return (
+            self._get_services_config()
+            .get(snakecase_service_id, {})
+            .get('endpoint_url')
+        )
+
+    def _get_endpoint_url_config_global(self):
+        return self._scoped_config.get('endpoint_url')
+
+    def _get_endpoint_url_environment_service(self):
+        return EnvironmentProvider(
+            name=self._get_service_env_var_name(), env=self._environ
+        ).provide()
+
+    def _get_endpoint_url_environment_global(self):
+        return EnvironmentProvider(
+            name='AWS_ENDPOINT_URL', env=self._environ
+        ).provide()
+
+    def __repr__(self):
+        return (
+            f'ConfiguredEndpointProvider('
+            f'full_config={self._full_config}, '
+            f'scoped_config={self._scoped_config}, '
+            f'client_name={self._client_name}, '
+            f'environ={self._environ})'
+        )
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -73,6 +73,7 @@ from botocore.compat import MD5_AVAILABLE  # noqa
 from botocore.exceptions import MissingServiceIdError  # noqa
 from botocore.utils import hyphenize_service_id  # noqa
 from botocore.utils import is_global_accesspoint  # noqa
+from botocore.utils import SERVICE_NAME_ALIASES  # noqa
 
 
 logger = logging.getLogger(__name__)
@@ -98,8 +99,6 @@ VALID_S3_ARN = re.compile('|'.join([_ACCESSPOINT_ARN, _OUTPOST_ARN]))
 # botocore/data/s3/2006-03-01/endpoints-rule-set-1.json
 S3_SIGNING_NAMES = ('s3', 's3-outposts', 's3-object-lambda')
 VERSION_ID_SUFFIX = re.compile(r'\?versionId=[^\s]+$')
-
-SERVICE_NAME_ALIASES = {'runtime.sagemaker': 'sagemaker-runtime'}
 
 
 def handle_service_name_alias(service_name, **kwargs):
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -42,6 +42,7 @@ from botocore.compat import HAS_CRT, MutableMapping
 from botocore.configprovider import (
     BOTOCORE_DEFAUT_SESSION_VARIABLES,
     ConfigChainFactory,
+    ConfiguredEndpointProvider,
     ConfigValueStore,
     DefaultConfigResolver,
     SmartDefaultsConfigStoreFactory,
@@ -962,6 +963,12 @@ class Session:
             smart_defaults_factory.merge_smart_defaults(
                 config_store, defaults_mode, region_name
             )
+
+        self._add_configured_endpoint_provider(
+            client_name=service_name,
+            config_store=config_store,
+        )
+
         client_creator = botocore.client.ClientCreator(
             loader,
             endpoint_resolver,
@@ -1029,6 +1036,17 @@ class Session:
             )
 
         return lmode
+
+    def _add_configured_endpoint_provider(self, client_name, config_store):
+        chain = ConfiguredEndpointProvider(
+            full_config=self.full_config,
+            scoped_config=self.get_scoped_config(),
+            client_name=client_name,
+        )
+        config_store.set_config_provider(
+            logical_name='endpoint_url',
+            provider=chain,
+        )
 
     def _missing_cred_vars(self, access_key, secret_key):
         if access_key is not None and secret_key is None:
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -3315,3 +3315,79 @@ class JSONFileCache:
                 return value.isoformat()
             return value.strftime('%Y-%m-%dT%H:%M:%S%Z')
         return value
+
+
+# This parameter is not part of the public interface and is subject to abrupt
+# breaking changes or removal without prior announcement.
+# Mapping of services that have been renamed for backwards compatibility reasons.
+# Keys are the previous name that should be allowed, values are the documented
+# and preferred client name.
+SERVICE_NAME_ALIASES = {'runtime.sagemaker': 'sagemaker-runtime'}
+
+
+# This parameter is not part of the public interface and is subject to abrupt
+# breaking changes or removal without prior announcement.
+# Mapping to determine the service ID for services that do not use it as the
+# model data directory name. The keys are the data directory name and the
+# values are the transformed service IDs (lower case and hyphenated).
+CLIENT_NAME_TO_HYPHENIZED_SERVICE_ID_OVERRIDES = {
+    # Actual service name we use -> Allowed computed service name.
+    'alexaforbusiness': 'alexa-for-business',
+    'apigateway': 'api-gateway',
+    'application-autoscaling': 'application-auto-scaling',
+    'appmesh': 'app-mesh',
+    'autoscaling': 'auto-scaling',
+    'autoscaling-plans': 'auto-scaling-plans',
+    'ce': 'cost-explorer',
+    'cloudhsmv2': 'cloudhsm-v2',
+    'cloudsearchdomain': 'cloudsearch-domain',
+    'cognito-idp': 'cognito-identity-provider',
+    'config': 'config-service',
+    'cur': 'cost-and-usage-report-service',
+    'datapipeline': 'data-pipeline',
+    'directconnect': 'direct-connect',
+    'devicefarm': 'device-farm',
+    'discovery': 'application-discovery-service',
+    'dms': 'database-migration-service',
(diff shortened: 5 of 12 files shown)


botocore/configprovider.py
```diff
@@ -510,6 +517,116 @@ def set_config_provider(self, logical_name, provider):
         self._mapping[logical_name] = provider
 
 
+class ConfiguredEndpointProviderChain:
```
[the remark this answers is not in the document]
kdaily: Added `__deepcopy__` method, but `__repr__` is tricky. Other config providers have succinct string-based parameters; the init params for `ConfiguredEndpointProvider` are all potentially large dictionaries.

botocore/args.py
```diff
@@ -259,6 +265,27 @@ def compute_client_args(
...
+
+    def _ignore_configured_endpoint_urls(self, client_config):
+        if (
+            client_config
+            and client_config.ignore_configured_endpoint_urls is not None
+        ):
+            return client_config.ignore_configured_endpoint_urls
+
+        return self._config_store.get_config_variable(
+            'ignore_configured__endpoint_urls'
```
kdaily, line 286: Fix typo here.

tests/unit/test_config_provider.py
```diff
@@ -942,3 +943,174 @@ def test_resolve_auto_mode_imds_region_provider_connect_timeout(self):
...
+            full_config_map=full_config_map,
+            scoped_config_map=scoped_config_map,
+            expected_value='global-from-config',
+        )
+
+
+@pytest.mark.parametrize(
+    "service_name,service_id", _known_service_names_and_ids()
+)
+def test_service_env_var_name_is_correct(service_name, service_id):
```
kdaily, line 1106: Move this to a functional test and use a session, etc. to test it end-to-end with all services.

botocore/configprovider.py
```diff
@@ -851,3 +858,145 @@ def provide(self):
...
+
+        return services_section
+
+    def __deepcopy__(self, memo):
+        return ConfiguredEndpointProvider(
+            full_config=copy.deepcopy(self._full_config, memo),
+            scoped_config=copy.deepcopy(self._scoped_config, memo),
+            client_name=copy.deepcopy(self._client_name),
+            environ=copy.deepcopy(self._environ),
+        )
```
dlm6693, line 1002: A couple of things here. Do we need a `deepcopy` implementation? I'm pretty sure all of these values are native Python data types so the custom logic isn't necessary. If we do though, I have a couple of suggestions.

1. `client_name` is a string and therefore immutable. Copying it is pointless and unnecessary.
2. We may want to account for recursive calls here by doing some work with `memo` instead of just handing it off to `copy.deepcopy`. Something like the following could be added:

```python
if memo is None:
    memo = {}

if id(self) in memo:
    return memo[id(self)]

new_instance = ConfiguredEndpointProvider(...)
memo[id(new_instance)] = new_instance
return new_instance

Comment (dlm6693) [reply botocore/configprovider.py:1002]: An additional nit: constructing the copy with `self.__class__` is probably a bit more maintainable.

File: botocore/session.py [hunk 5]
```diff
@@ -962,6 +963,12 @@ def create_client(
             smart_defaults_factory.merge_smart_defaults(
                 config_store, defaults_mode, region_name
             )
+
+        self._add_configured_endpoint_provider(
+            client_name=service_name,
+            config_store=config_store,
+        )
+
```
dlm6693, line 971: This doesn't seem right to me. At the very least you should be able to manually add a new section to the config mapping in `create_botocore_default_config_mapping`.  Is there any reason in particular you didn't go that route? We should avoid mutating the config store after initialization and instead make sure to include this new provider along with all the others in `_register_config_store`.

botocore/utils.py
```diff
@@ -3315,3 +3315,76 @@ def _serialize_if_needed(self, value, iso=False):
...
+
+
+def handle_service_name_alias(service_name, **kwargs):
+    return SERVICE_NAME_ALIASES.get(service_name, service_name)
+
+
+# Mapping to determine the service ID for services that do not use it as the
+# model data directory name. The keys are the data directory name and the
+# values are the service IDs.
+CLIENT_NAME_TO_HYPHENIZED_SERVICE_ID_OVERRIDES = {
```
dlm6693, line 3330: To be clear, are you saying that the values are how customers would refer to the service in the config file and/or the endpoint prefix for the service? If so can you make that a bit more obvious here 😸

tests/unit/cfg/aws_services_config
```diff
@@ -0,0 +1,10 @@
+[default]
+endpoint_url = https://localhost:1234/
+services = my-services
+
+[services my-services]
+s3 =
+  endpoint_url = https://localhost:5678/
+  addressing_style = path
+dynamodb =
```
dlm6693, line 9: Should we add a test case for one of the services in the list where its being renamed?

tests/unit/test_config_provider.py
```diff
@@ -942,3 +943,172 @@ def test_resolve_auto_mode_imds_region_provider_connect_timeout(self):
...
+        self.assertEqual(value, expected_value)
+
+    def test_does_provide_env_var(self):
+        self.assert_does_provide(
+            service="batch",
+            environ_map={'AWS_ENDPOINT_URL': 'from-env'},
+            full_config_map={},
+            scoped_config_map={},
+            expected_value='from-env',
+        )
```
dlm6693, line 985: It looks like you're running this function a lot. We should have one test function that just uses `pytest.mark.parametrize` instead. It should probably also exist outside of a test class and just be a standalone function.

tests/unit/test_config_provider.py
```diff
@@ -942,3 +943,172 @@ def test_resolve_auto_mode_imds_region_provider_connect_timeout(self):
...
+def _known_service_names_and_ids():
+    my_session = session.get_session()
+    loader = my_session.get_component('data_loader')
+    available_services = loader.list_available_services('service-2')
+
+    result = set()
+    for service_name in available_services:
+        model = my_session.get_service_model(service_name)
+        result.add((model.service_name, model.service_id))
+    return sorted(result)
```
dlm6693, line 957: Why does this need to be sorted? If it isn't necessary we could get rid of the set and yield instead:
```python
yield (model.service_name, model.service_id)

File: tests/unit/test_session_legacy.py [hunk 10]
```diff
@@ -708,13 +708,19 @@ def test_create_client_verify_param_overrides_all(self, client_creator):
...
         self.session.create_client('myservice', 'us-west-2')
         call_kwargs = client_creator.return_value.create_client.call_args[1]
         self.assertEqual(call_kwargs['api_version'], None)
 
     @mock.patch('botocore.client.ClientCreator')
-    def test_create_client_uses_api_version_from_config(self, client_creator):
+    def test_create_client_uses_api_version_from_config(
+        self,
+        client_creator,
+    ):
```
dlm6693, line 723: Why are all of these changes here? Can you remove? I'm guessing your local formatter is a bit different than our.

dlm6693 (changes requested):
This is looking pretty good so far. I'd like to see some organizational changes and have some questions about the underlying logic in a couple of places, but we're definitely on the right track here.

tests/unit/test_config_provider.py
kdaily: The tests fail with this error if they are not sorted:

https://github.com/pytest-dev/pytest-xdist/issues/432

tests/unit/test_session_legacy.py
kdaily: Hmm. I'm only using the pre-commit hook, but I think I touched these in a previous commit that was rebased and reverted. Will fix it though!

tests/unit/test_config_provider.py
dlm6693: Hmm that says its specifically for running multiple tests in parallel. Are you still seeing that error without passing `-n`? I just find this very strange since we have many of these types of functions across our test suite where we don't need to sort.

botocore/configprovider.py
```diff
@@ -851,3 +858,145 @@ def provide(self):
...
+        return EnvironmentProvider(
+            name=self._get_service_env_var_name(), env=self._environ
+        ).provide()
+
+    def _get_endpoint_url_environment_global(self):
+        return EnvironmentProvider(
+            name="AWS_ENDPOINT_URL", env=self._environ
+        ).provide()
+
+    def unhyphenize_service_name(self, service_name):
```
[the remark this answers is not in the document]
kdaily: There's an analogous `hyphenize_service_id` in `utils.py` thats just as short. I moved it to `configprovider.py` since thats the only place using it, but I would be fine to remove it. The service name/service ID usage is really confusing so I thought being explicit with the intention was helpful.

botocore/configprovider.py
```diff
@@ -851,3 +858,145 @@ def provide(self):
...
+        return f"AWS_ENDPOINT_URL_{transformed_service_id_env}"
+
+    def _get_services_config(self):
+        if "services" not in self._scoped_config:
+            return {}
+
+        section_name = self._scoped_config["services"]
+        services_section = self._full_config.get("services", {}).get(
+            section_name
+        )
```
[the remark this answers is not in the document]
kdaily: The scoped config does not have the contents of the services section.

Similar logic:

https://github.com/boto/botocore/blob/develop/botocore/tokens.py#L204

botocore/configprovider.py
```diff
@@ -851,3 +858,145 @@ def provide(self):
...
+        :type full_config: dict
+        :param full_config: This is the dict representing the full
+            configuration file.
+
+        :type scoped_config: dict
+        :param scoped_config: This is the dict representing the configuration
+            for the current profile for the session.
+
+        :type service_name: str
+        :param service_name: service name.
```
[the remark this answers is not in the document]
kdaily: Yep!

botocore/configprovider.py
```diff
@@ -851,3 +858,145 @@ def provide(self):
...
+
+        :type service_id: str
+        :param service_id: service id.
```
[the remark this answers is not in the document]
kdaily: Thanks! Forgot to remove.

botocore/configprovider.py
kdaily: Existing config providers that have constant `str` values (and other native Python datatypes) also explicitly deep copy them. Example:

https://github.com/boto/botocore/blob/a447b7a9e849e216e31ca4aa616fba6e70b2b988/botocore/configprovider.py#L831

botocore/session.py
kdaily: This is modeled after the smart defaults behavior in order to have access to all of the data needed. If we did this in the default mapping, we would need to know every service that could possibly exist and add logic for it; this is not known at that time (need a session for that, which introduces a circular dependency).

botocore/utils.py
kdaily: This MUST be marked as non-public which I neglected to do!

I don't think customers ever need to refer to any of the values of this dictionary, they would only need the keys.

tests/unit/cfg/aws_services_config
kdaily: The functional tests (both the one I modified in `test_endpoints.py` and the ones for the config provider) enumerates all available services, which should address this?

botocore/configprovider.py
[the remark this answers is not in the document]
dlm6693: Why don't we keep it to make it explicit, but privatize it since it isn't intended to be called anywhere outside of this class.

botocore/configprovider.py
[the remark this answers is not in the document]
dlm6693: Got it thanks!

tests/unit/test_config_provider.py
kdaily: These are how the other config provider test cases are structured; as I interpreted it to provide readable names for the specific cases that would otherwise need to be reverse engineered from the contents (since the testing is often to confirm that something is done in the right order of precedence).

The functional tests (in a separate PR) parameterize over all possible combinations of values provided for setting the endpoint URL.

botocore/configprovider.py
dlm6693: That doesn't mean they're right 😃 I would still remove the copy from the string since it isn't doing anything.

tests/unit/cfg/aws_services_config
dlm6693: I just think it could be beneficial to have one in here too. Not necessary though.

botocore/utils.py
dlm6693: maybe lets make it a private constant as well as indicating in a comment this is not intended for public use

A-Levin:
any news, guys?)

tests/unit/test_config_provider.py
dlm6693: I wouldn't use precedence in our test suite as a reason to structure tests this way. In pretty much every case when we're writing an entire new group of tests, we use newer functional patterns with pytest and use parametrize to remove redundancy. There isn't a need to use individual test functions as a means to document behavior of the new provider, so long as the feature is well-documented.

botocore/configprovider.py
kdaily: OK! I tested out that the class is deep-copyable without this implementation, so I removed it and added a unit test to prevent regression.

botocore/args.py
```diff
@@ -139,7 +139,6 @@ def get_client_args(
             client_cert=new_config.client_cert,
             proxies_config=new_config.proxies_config,
         )
-
```
dlm6693, line 142: nit: add this line back in.

botocore/configprovider.py
```diff
@@ -851,3 +858,149 @@ def provide(self):
...
+        section_name = self._scoped_config['services']
+        services_section = self._full_config.get('services', {}).get(
+            section_name
+        )
+
+        if not services_section:
+            error_msg = (
+                f'The profile is configured to use the services '
+                f'section but the "{section_name}" services '
+                f'configuration does not exist.'
```
dlm6693, line 979:
```suggestion
                'configuration does not exist.'
```

botocore/configprovider.py
```diff
@@ -851,3 +858,149 @@ def provide(self):
...
+        :type full_config: dict
+        :param full_config: This is the dict representing the full
+            configuration file.
+
+        :type scoped_config: dict
+        :param scoped_config: This is the dict representing the configuration
+            for the current profile for the session.
+
+        :type client_name: str
+        :param client_name: client name.
```
dlm6693, line 896: Could we get a little more detail here?

botocore/configprovider.py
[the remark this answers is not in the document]
kyleknap: Looking over this, I do agree we should not make this a public method. I think @dlm6693 was just suggesting just have the method start with an underscore to explicitly note that the method is private (e.g. `_unhyphenize_service_name`). That being said I'd honestly prefer us to remove this method. See comment later on related to simplifying the logic.

botocore/configprovider.py
```diff
@@ -851,3 +858,149 @@ def provide(self):
...
+        client_name = (
+            utils._CLIENT_NAME_TO_HYPHENIZED_SERVICE_ID_OVERRIDES.get(
+                client_name, client_name
+            )
+        )
+
+        return self.unhyphenize_service_name(client_name)
+
+    def _snakecase_service_id(self, service_id):
+        return service_id.replace(' ', '_')
```
kyleknap, line 958: Let's just combine these two methods to be:
```python
def _get_snake_case_service_id(self):
    client_name = utils.handle_service_name_alias(client_name)
    client_name = (
        utils._CLIENT_NAME_TO_HYPHENIZED_SERVICE_ID_OVERRIDES.get(
            self._client_name, self._client_name
        )
    )
    return client_name.replace('-', '_')
```
Within this function, it would then be good to add a comment about how we are supposed to be using the service id but instead deriving it from the client name and what assumptions/properties we are expecting in order to do this.

The main reason that I suggest this is that the logic seems fairly complicated now as there are three or four different methods now in how we are transforming the service id; when in reality, we only just need shared logic for getting the snake cased service id and from there can call upper and lower on it.

botocore/session.py
```diff
@@ -1030,6 +1037,17 @@ def _resolve_defaults_mode(self, client_config, config_store):
...
         return lmode
 
+    def _add_configured_endpoint_provider(self, client_name, config_store):
+        chain = ConfiguredEndpointProvider(
+            full_config=self.full_config,
+            scoped_config=self.get_scoped_config(),
+            client_name=client_name,
+        )
+        config_store.set_config_provider(
+            logical_name='configured_endpoint_url',
```
kyleknap, line 1047: Is there a particular reason we are not using `endpoint_url` here as the config store key? We use the `endpoint_url` naming convention everywhere else for the parameter name to override the endpoint url so it probably makes sense to stay consistent with that name in the `config_store`.

botocore/utils.py
```diff
@@ -3315,3 +3315,77 @@ def _serialize_if_needed(self, value, iso=False):
...
+
+def handle_service_name_alias(service_name, **kwargs):
+    return SERVICE_NAME_ALIASES.get(service_name, service_name)
+
+
+# Mapping to determine the service ID for services that do not use it as the
+# model data directory name. The keys are the data directory name and the
+# values are the service IDs. This parameter is considered private and is
+# subject to abrupt breaking changes or removal without prior announcement.
+_CLIENT_NAME_TO_HYPHENIZED_SERVICE_ID_OVERRIDES = {
```
kyleknap, line 3331: Can we instead just document that this map is private as part of the comment and remove the leading underscore? It may also help making the note that this is private as the very first thing said in the comment so it is obvious and maybe change the wording from it being "private" to it being "not a public interface" to clarify that "we" can use this map but people consuming botocore should not.

In general, we should not be marking this private with an underscore as that would indicate that it can only be used within this classes/functions of this `utils.py` file, which is not the case. We use the map in the config provider.

botocore/handlers.py
```diff
@@ -99,12 +101,6 @@
 S3_SIGNING_NAMES = ('s3', 's3-outposts', 's3-object-lambda')
 VERSION_ID_SUFFIX = re.compile(r'\?versionId=[^\s]+$')
 
-SERVICE_NAME_ALIASES = {'runtime.sagemaker': 'sagemaker-runtime'}
-
-
-def handle_service_name_alias(service_name, **kwargs):
-    return SERVICE_NAME_ALIASES.get(service_name, service_name)
-
```
kyleknap, line 107: For reusing the map, I would recommend:

1. Keep with the change of moving `SERVICE_NAME_ALIASES` to `utils`
2. In `utils` document it similar to the `CLIENT_NAME_TO_HYPHENIZED_SERVICE_ID_OVERRIDES` map where it is not part of botocore's public interface and also document what exactly it is supposed to represent for clarity.
3. Do not move `handle_service_name_alias` to the `utils` module and keep it defined here.
4. In the config provider, just use the `SERVICE_NAME_ALIASES` map from `utils`.

The main reason I'm suggesting this is that it does not make a lot of sense to move the `handle_service_name_alias` to `utils` as we are not really generalizing it to be a public utility and I don't think we should be making it a public utility as there is not a strong reason to do so. Referencing from the alias map is straightforward enough and the `handle_service_name_alias` only exists so we encapsulate the logic in the format of a botocore event handler.

tests/unit/docs/__init__.py
```diff
@@ -89,13 +89,13 @@ def setup_client(self):
             ],
         }
 
-        def load_service_mock(*args, **kwargs):
```
kyleknap, line 92: Why did this test need to be updated?

tests/unit/test_config_provider.py
```diff
@@ -942,3 +943,187 @@ def test_resolve_auto_mode_imds_region_provider_connect_timeout(self):
...
+        assert provider is not provider_deepcopy
+        assert provider.provide() == 'https://endpoint-override'
+        assert provider_deepcopy.provide() == 'https://endpoint-override'
+
+        env['AWS_ENDPOINT_URL_BATCH'] = 'https://another-new-endpoint-override'
+        assert provider.provide() == 'https://another-new-endpoint-override'
+        assert provider_deepcopy.provide() == 'https://endpoint-override'
+
+
+@pytest.mark.parametrize(
```
kyleknap, line 1117: Let's shelf this test for the functional test PR. In general, unit tests should not be dependent on service models. That is reserved more for functional tests. I think when we do port this to be a functional test we would also use the client interface so that we can have a close to an end to end guarantee as we can get.

We should also make sure to include any service aliases in this test (e.g. `sagemaker.runtime`)

tests/unit/test_session.py
```diff
@@ -697,7 +697,10 @@ def test_create_client_verify_param_overrides_all(self, client_creator):
             self.assertEqual(call_kwargs['verify'], 'verify-certs.pem')
 
     @mock.patch('botocore.client.ClientCreator')
-    def test_create_client_use_no_api_version_by_default(self, client_creator):
+    def test_create_client_use_no_api_version_by_default(
```
kyleknap, line 700: Why did this file change? Was this just never run through the formatter before? We should make sure we are not making updates if there is no changes to the actual logic in order to help cut down the noise in reviews.

tests/unit/test_session_legacy.py
```diff
@@ -708,7 +708,10 @@ def test_create_client_verify_param_overrides_all(self, client_creator):
             self.assertEqual(call_kwargs['verify'], 'verify-certs.pem')
 
     @mock.patch('botocore.client.ClientCreator')
-    def test_create_client_use_no_api_version_by_default(self, client_creator):
+    def test_create_client_use_no_api_version_by_default(
```
kyleknap, line 711: Yeah same thing here. I would have not expected this file to change unless we were updating logic in the tests.

kyleknap (commented):
Looks good. I just had feedback around organizations and some of the refactors that were done. I do think we are close though to get this one merged so that we can move on to the functional test PR.

botocore/configprovider.py
kdaily: I made changes like this previously and it was suggested that for uniformity (and clarity that its an f-string) to use the `f` syntax on all string sections.

botocore/session.py
kdaily: A vestige of previous refactoring from `session.py`, will rename.

tests/unit/test_session.py
kdaily: From previous refactors this line was touched and then `pre-commit` cleaned it up. Will revert it.

tests/unit/test_session_legacy.py
kdaily: From previous refactors this line was touched and then `pre-commit` cleaned it up. Will revert it.

botocore/configprovider.py
dlm6693: When was that suggested? I've never heard of doing that before. If you've gotten that feedback, keep it, but I find it weird the need to clarify an f string on every line when its abundantly obvious where the formatting is happening, particularly in modern IDEs with good syntax highlighting.

botocore/configprovider.py
```diff
@@ -851,3 +858,147 @@ def provide(self):
...
+    def _get_snake_case_service_id(self, client_name):
+        # Use lookups to get the serice ID without loading the service data
+        # file. `client_name`` refers to the name used to instantiate a client
+        # through botocore.session.Session.create_client (parameter name there
+        # is `service_name`).  We need to look up services that have been
+        # renamed (SERVICE_NAME_ALIASES) as well as services that do not use
+        # the service ID as their data directory name
+        # (CLIENT_NAME_TO_HYPHENIZED_SERVICE_ID_OVERRIDES).
+        client_name = utils.SERVICE_NAME_ALIASES.get(client_name, client_name)
+        service_id = utils.CLIENT_NAME_TO_HYPHENIZED_SERVICE_ID_OVERRIDES.get(
```
kyleknap, line 953: Nitpick but let's rename this to `hyphenized_service_id` to be clear this is not the actual service id but that transform of it.

botocore/configprovider.py
```diff
@@ -851,3 +858,147 @@ def provide(self):
...
+                    location,
+                )
+                return endpoint_url
+
+        logger.debug('No configured endpoint found.')
+        return None
+
+    def _get_snake_case_service_id(self, client_name):
+        # Use lookups to get the serice ID without loading the service data
+        # file. `client_name`` refers to the name used to instantiate a client
```
kyleknap, line 946: We should also put a statement in this description that states generally the client name will be the hyphenized transformation of the service id to be explicit that we are making this assumption to derive what the service id is. Otherwise, I'm not sure if this is still clear.

botocore/utils.py
```diff
@@ -3315,3 +3315,79 @@ def _serialize_if_needed(self, value, iso=False):
                 return value.isoformat()
             return value.strftime('%Y-%m-%dT%H:%M:%S%Z')
         return value
+
+
+# This parameter is not part of the public interface and is subject to abrupt
+# breaking changes or removal without prior announcement.
+# Mapping of services that have been renamed for backwards compatibility reasons.
+# Keys are the previous name that should be allowed, values are the transformed
```
kyleknap, line 3323: I'm not sure if we should include service id in this explanation. We should be able to shorten it to something like: "values are the documented/preferred client name" to simplify it down.

tests/unit/test_config_provider.py
```diff
@@ -942,3 +943,142 @@ def test_resolve_auto_mode_imds_region_provider_connect_timeout(self):
...
+            full_config_map={
+                'profiles': {
+                    'default': {
+                        'services': 'my-services',
+                        'endpoint_url': 'global-from-config',
+                    }
+                },
+                'services': {
+                    'my-services': {
+                        'batch': {'endpoint_url': "service-from-config"}
```
kyleknap, line 1037: We should probably use the variable `service` here instead of `batch`

tests/unit/test_config_provider.py
```diff
@@ -942,3 +943,142 @@ def test_resolve_auto_mode_imds_region_provider_connect_timeout(self):
...
+                    }
+                },
+            },
+            expected_value='global-from-config',
+        ),
```
kyleknap, line 1042: As a sanity check, it may be worth adding two more test cases to this in order to make sure we capture the mapping logic we are doing in the provider:

1. Service alias (e.g. using `runtime.sagemaker` as the name of the client)
2. A service where the client name does not map to the hyphenized service id (e.g. `apigateway`). We only need to do one of these to just make sure our mapping works correctly

kyleknap (commented):
Looks good. I just had some small comments, but we should be set after.

tests/unit/test_config_provider.py
```diff
@@ -942,3 +943,178 @@ def test_resolve_auto_mode_imds_region_provider_connect_timeout(self):
...
+                    }
+                },
+                'services': {
+                    'my-services': {
+                        service: {'endpoint_url': "service-from-config"}
+                    }
+                },
+            },
+            expected_value='service-from-config',
+        ),
```
kyleknap, line 1024: One more case that I did not catch we were missing from the last review. Let's make sure the env vars override what is in the config as well. We should be good to just add two more cases:

1. Make sure the `AWS_ENDPOINT_URL` overrides `endpoint_url` in config
2. Make sure the `AWS_ENDPOINT_URL_{service.upper()}` overrides `endpoint_url` in the config

kyleknap (commented):
Looks good. Just one last comment that I missed from the last review.

kyleknap (approved):
Looks good! :ship: Feel free to squash the commits and merge.

PR #2959, opened 2023-05-26
request compression

dlm6693:
This pull request enables request compression in the AWS Python SDK. It is on by default, but only used if the `requestcompression` trait is attached to the operation being called. At the moment, the only compression algorithm supported is `gzip`, but this is subject to change in the future. This also addresses a few typos I discovered while working on this project. Below is a summary of the new features in this PR.

* Add `request_min_compression_size_bytes` setting to `Config` object and config file, and support for `AWS_REQUEST_MIN_COMPRESSION_SIZE_BYTES` environment variable.
    * If not provided by a user, the value will default to `10240`. The value MUST be a parseable integer between 0 and 1048576 (1024**2).
* Add `disable_request_compression` setting to `Config` object and config file, and support for `AWS_DISABLE_REQUEST_COMPRESSION` environment variable.
    * If not provided by a user, the value will default to `False`.  If not a parseable boolean, the value will be coerced to `False`, similar to other config settings that are boolean types in botocore.
* Add `request_compression` property to `OperationModel`. This will allow the SDK to check for the presence of this operation trait and run compression if so. 
* Add the `compress.py` file. This file contains logic that is actually responsible for compressing requests. It has one public API:
    * `maybe_compress_request`: This checks whether compression should be performed, actually does compression if so and injects the corresponding `Content-Encoding` header if needed.
* Compression is supported for all types of bodies the SDK already supports: `str`, `bytes`, `bytearray`, and file-like objects like `io.StringIO` and `io.BytesIO` for example.
* Compression is called just before flexible checksums are calculated in `BaseClient._make_api_call`
* Besides the above new settings, compression will also not occur if a customer has configured `signature_version` to be `v2`
* In order to support compression for services using the `query` protocol, dictionary bodies are serialized to bytes if the operation supports compression and the customer hasn't explicitly disabled compression.


botocore/awsrequest.py
```diff
@@ -534,6 +557,84 @@ def reset_stream(self):
...
+        return body
+
+    def _compress_body(self, body, encoder):
+        if isinstance(body, str):
+            return encoder(body.encode('utf-8'))
+        elif isinstance(body, (bytes, bytearray)):
+            return encoder(body)
+        elif hasattr(body, 'read'):
+            return encoder(b''.join(self._read_body(body)))
+        raise TypeError('Unable to compress body of type: %s' % type(body))
```
dlm6693, line 591: Maybe an f string here instead.

botocore/endpoint.py
```diff
@@ -121,6 +121,12 @@ def make_request(self, operation_model, request_dict):
     def create_request(self, params, operation_model=None):
         request = create_request_object(params)
         if operation_model:
+            if operation_model.request_compression is not None:
+                compress_aws_request(
+                    params['context']['client_config'],
```
dlm6693, line 126: It shouldn't ever be in practice, but maybe should check if config is None here before performing compression.

botocore/awsrequest.py
```diff
@@ -337,6 +340,26 @@ def _urljoin(endpoint_url, url_path, host_prefix):
...
+        operation_model=operation_model,
+        body=body,
+    ):
+        # AWSRequest.data is the raw input for the request body. This is the parameter
+        # that should be updated to properly construct an AWSPreparedRequest object.
+        request.data = compressor.compress(
+            body=body,
+            headers=request.headers,
+            encodings=operation_model.request_compression['encodings'],
+        )
```
jonemo, line 360: `compressor.compress` mixes in-place modification (of the `request.headers` dict) and returning a modified value (of `request.body`). Mixing the two styles is surprising by itself in Python, but for two properties of the same object?

Is there anything that can be done to make this behavior less surprising?

botocore/awsrequest.py
```diff
@@ -534,6 +557,84 @@ def reset_stream(self):
...
+            else:
+                logger.debug('Unsupported compression encoding: %s' % encoding)
+        return body
+
+    def _compress_body(self, body, encoder):
+        if isinstance(body, str):
+            return encoder(body.encode('utf-8'))
+        elif isinstance(body, (bytes, bytearray)):
+            return encoder(body)
+        elif hasattr(body, 'read'):
```
jonemo, line 589: What's the motivation for this specific check? How does it compare to, for example, `isinstance(io.IOBase)` ([IOBase docs](https://docs.python.org/3/library/io.html#io.IOBase))? (Not asking for a change, just trying to understand the reasoning.)

botocore/endpoint.py
jonemo: Do we have any tests that assert (or assume) that client_config is set here? If so, no need to worry about the check. Otherwise, I would add it.

botocore/exceptions.py
```diff
@@ -814,3 +814,18 @@ class EndpointResolutionError(EndpointProviderError):
 
 class UnknownEndpointResolutionBuiltInName(EndpointProviderError):
     fmt = 'Unknown builtin variable name: {name}'
+
+
+class InvalidRequestMinCompressionSizeError(BotoCoreError):
+    fmt = (
+        'Invalid value provided to "request_min_compression_size_bytes": '
+        '{input_value} must be an integer greater than or equal to 0 and '
+        'less than or equal to 10485760.'
```
jonemo, line 823: Can we move the 10485760 into a constant so that we can be sure to always use the same number in the check and in the exception message? How do other exception messages in this file deal with similar constant values?

botocore/utils.py
```diff
@@ -3315,3 +3317,29 @@ def _serialize_if_needed(self, value, iso=False):
...
+def validate_compression_config(config):
+    disable_request_compression = config.disable_request_compression
+    if (
+        disable_request_compression is not None
+        and not isinstance(disable_request_compression, bool)
+        and not (
+            isinstance(disable_request_compression, str)
+            and disable_request_compression.lower() in ('true', 'false')
+        )
+    ):
```
jonemo, line 3331: Adding to my sequence of comments about using "conversion_func" in `configparser.py`. I don't think we have equivalent code for any other config value and think we should follow the existing pattern.

jonemo (commented):
I focused on how config values are handled and the `AWSRequestCompressor` class so far. Taking a break now and will continue with the other parts in the next review session.

botocore/awsrequest.py
```diff
@@ -534,6 +557,84 @@ def reset_stream(self):
...
+        # compression must be explicitly disabled
+        if (
+            botocore.utils.ensure_boolean(config.disable_request_compression)
+            is not True
+        ):
+            min_size = config.request_min_compression_size_bytes
+            if min_size is None:
+                min_size = 10240
+            else:
+                min_size = int(min_size)
```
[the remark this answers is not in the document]
dlm6693: I should have mentioned it in the PR description, but this was a very intentional part of the design. I chose not to use `ensure_boolean` and `int` in `conversion_func` because the former coerces any non-boolean value that isn't equal to `true` to `False` and the latter would provide a relatively opaque error message to the user. 

I made the tradeoff of being more closely aligned with the spec than how other config values are handled in botocore:
```
This setting is a boolean with values true or false (case-sensitive). The SDK MUST provide an error for any other value. The SDK MUST use false as the default value.
```

botocore/awsrequest.py
```diff
@@ -337,6 +340,26 @@ def _urljoin(endpoint_url, url_path, host_prefix):
     return reconstructed
 
 
+def compress_aws_request(config, request, operation_model):
+    botocore.utils.validate_compression_config(config)
+    compressor = AWSRequestCompressor()
```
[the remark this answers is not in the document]
dlm6693: Definitely can be reused. I can add this as an instance variable on the `Endpoint` class instead.

botocore/awsrequest.py
dlm6693: Sure I can separate these into two separate methods or just modify the headers in the top-level function.

dlm6693: Sure I can separate these into two separate methods or just modify the headers in the top-level function.

botocore/awsrequest.py
```diff
@@ -337,6 +340,26 @@ def _urljoin(endpoint_url, url_path, host_prefix):
...
+    # ``AWSRequest.body`` is a property that will compute a URL encoded value
+    # if its a dictionary. So we shouild call it once hear to minimize recalculation.
```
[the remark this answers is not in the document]
dlm6693: Are there multiple typos? I think the only I see is using `its` instead of `it's`. I'll rephrase to be clearer.

botocore/awsrequest.py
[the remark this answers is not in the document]
dlm6693: To follow up on this: Can these checks and conversions actually be dropped even if we use `conversion_func`? That wouldn't cover when customers use their own `Config` object.

botocore/awsrequest.py
```diff
@@ -534,6 +557,84 @@ def reset_stream(self):
...
+                min_size = 10240
+            else:
+                min_size = int(min_size)
+            # Request is compressed no matter the content length if it has a streaming input.
+            # However, if the stream has the `requiresLength` trait it is NOT compressed.
+            return (
+                operation_model.has_streaming_input
+                and 'requiresLength'
+                not in operation_model.get_streaming_input().metadata
+            ) or min_size <= self._get_body_size(body)
```
dlm6693, line 624: This is slightly wrong. If the operation has streaming input and the `requiresLength` trait it should not compress, not fall back to checking the length.

botocore/awsrequest.py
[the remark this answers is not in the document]
nateprewitt: We shouldn't be changing behavior from how all other options in the SDK work. If a non-True value is passed we default to False. We can go back and have the wording in the design adjusted but if there isn't a strong argument for this requirement, it's going to be an unexpected idiosyncrasy of this feature.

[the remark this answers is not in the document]
dlm6693: Thanks Nate. I'll remove all of this custom code and just default to how other config settings are handled in the SDK.

botocore/exceptions.py
nateprewitt: Arguably this value shouldn't even be hardcoded. We can format the min/max in the exception.

botocore/awsrequest.py
```diff
@@ -534,6 +557,84 @@ def reset_stream(self):
...
+        if hasattr(body, 'seek'):
+            body.seek(0)
+
+    def should_compress_request(self, config, body, operation_model):
+        """Check if an AWSRequest object's body should be compressed."""
+        # compression must be explicitly disabled
+        if (
+            botocore.utils.ensure_boolean(config.disable_request_compression)
+            is not True
+        ):
```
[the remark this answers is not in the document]
dlm6693: It wasn't previously since I intentionally chose not to coerce these settings in the config provider, but that's changed and as a result the `ensure_boolean` call is redundant here and will be removed in a future commit.

botocore/awsrequest.py
```diff
@@ -534,6 +557,84 @@ def reset_stream(self):
...
+            min_size = config.request_min_compression_size_bytes
+            if min_size is None:
```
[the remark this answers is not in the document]
dlm6693: Same answer. It should be no 😄

botocore/awsrequest.py
```diff
@@ -534,6 +557,84 @@ def reset_stream(self):
             raise UnseekableStreamError(stream_object=self.body)
 
 
+class AWSRequestCompressor:
```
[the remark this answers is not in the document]
dlm6693: The spec describes this concept as "request compression" and the trait is known as `requestcompression` so it seemed logical to associate this with the entire object of `AWSRequest` rather than a specific component of it.

botocore/awsrequest.py
```diff
@@ -337,6 +340,26 @@ def _urljoin(endpoint_url, url_path, host_prefix):
...
+    botocore.utils.validate_compression_config(config)
+    compressor = AWSRequestCompressor()
+    # ``AWSRequest.body`` is a property that will compute a URL encoded value
+    # if its a dictionary. So we shouild call it once hear to minimize recalculation.
+    body = request.body
+    if compressor.should_compress_request(
+        config=config,
+        operation_model=operation_model,
+        body=body,
+    ):
```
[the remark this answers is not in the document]
dlm6693: I'm updating this to a private method that is called within a function called `maybe_compress`. The reasoning before was to be explicit about compression possibly not happening.

botocore/awsrequest.py
```diff
@@ -534,6 +557,84 @@ def reset_stream(self):
...
+                if 'Content-Encoding' not in headers:
+                    headers['Content-Encoding'] = encoding
+                elif encoding not in headers['Content-Encoding']:
+                    headers.replace_header(
+                        'Content-Encoding',
+                        f'{headers["Content-Encoding"]},{encoding}',
+                    )
+                return self._compress_body(body, encoder)
+            else:
+                logger.debug('Unsupported compression encoding: %s' % encoding)
```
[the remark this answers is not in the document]
dlm6693: Are you suggesting raising an error then? Because that seems awfully risky if a broken model somehow made it through to production and we elect to error out instead of letting requests proceed with uncompressed bodies.

botocore/awsrequest.py
```diff
@@ -534,6 +557,84 @@ def reset_stream(self):
...
+            return encoder(body)
+        elif hasattr(body, 'read'):
+            return encoder(b''.join(self._read_body(body)))
+        raise TypeError('Unable to compress body of type: %s' % type(body))
+
+    def _read_body(self, body):
+        while True:
+            # urllib3 and by extension the http stdlib uses a chunksize of 8192
+            # so we do the same here.
+            chunk = body.read(8192)
```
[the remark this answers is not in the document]
dlm6693: Let me take a look at this again. I'm not sure you can do that like you can for a `sha256` object as an example.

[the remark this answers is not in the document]
dlm6693: We can use a `GzipFile` object to update as we read. I'll work on adding that here.

botocore/awsrequest.py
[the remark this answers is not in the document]
dlm6693: I didn't set this up quite right go. We need to check if the operation model has streaming input, then return if `requiresLength` is in the streaming input metadata. So we can't couple those checks into one variable. Did you still want to see this change?

botocore/exceptions.py
```diff
@@ -814,3 +814,18 @@ class EndpointResolutionError(EndpointProviderError):
...
+    )
+
+
+class InvalidDisableRequestCompressionError(BotoCoreError):
+    fmt = (
+        'Invalid value provided to "disable_request_compression": '
+        '{input_value} must be a boolean.'
+    )
```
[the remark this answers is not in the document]
dlm6693: There are a ton of specific configuration errors in this file. The only seemingly generic one is `InvalidConfigError` and that's only ever used in `credentials.py`. Only one error inherits from this and all of the other configuration related ones seem to inherit directly from `BotoCoreError`.

botocore/model.py
```diff
@@ -87,6 +87,7 @@ class Shape:
         'union',
         'contextParam',
         'clientContextParams',
+        'requiresLength',
```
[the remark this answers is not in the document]
dlm6693: It doesn't appear we've really used it anywhere, but yes it exists.
https://github.com/boto/botocore/blob/8d665304dea8608d0c4b9ccef2c88cfeeac85863/botocore/data/lookoutvision/2020-11-20/service-2.json#L2047
https://github.com/boto/botocore/blob/8d665304dea8608d0c4b9ccef2c88cfeeac85863/botocore/data/omics/2022-11-28/service-2.json#L5946

botocore/utils.py
```diff
@@ -3315,3 +3317,29 @@ def _serialize_if_needed(self, value, iso=False):
...
+
+def validate_compression_config(config):
+    disable_request_compression = config.disable_request_compression
+    if (
+        disable_request_compression is not None
+        and not isinstance(disable_request_compression, bool)
+        and not (
+            isinstance(disable_request_compression, str)
+            and disable_request_compression.lower() in ('true', 'false')
+        )
```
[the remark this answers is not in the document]
dlm6693: This is going away, but I opted for explicit checks because `ensure_boolean` coerces bad inputs to `False` which I wanted to avoid here/raise an error instead.

botocore/awsrequest.py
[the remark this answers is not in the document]
nateprewitt: I wouldn't expect the compression utility to be public facing. We should know it's a conditional compression but if not, you would generally make that an argument on the function that's akin to `strict`.

botocore/awsrequest.py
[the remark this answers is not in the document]
nateprewitt: Whatever concepts make sense. What's here now is a bit difficult to visually parse.

botocore/exceptions.py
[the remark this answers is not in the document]
nateprewitt: > There are a ton of specific configuration errors in this file. 

Most of these are for specific code paths for specific features, we're failing in a very common code path here. At the leas, these should subclass `InvalidConfigError`. It's debatable you'd be able to do any reasonable flow control with these though.

What would you expect the user to be doing when these are raised?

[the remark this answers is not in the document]
nateprewitt: > What would you expect the user to be doing when these are raised?

Sorry, I was asking this non-rhetorically. What would my code look like if this was raised as opposed to the general InvalidConfigError which appears to be used in both `tokens.py` and `credentials.py` for similar use cases.

botocore/model.py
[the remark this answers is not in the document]
dlm6693: Here is more info on the trait

https://smithy.io/1.0/spec/core/stream-traits.html#requireslength-trait

botocore/args.py
```diff
@@ -543,6 +550,47 @@ def _compute_connect_timeout(self, config_kwargs):
...
+            disabled = self._config_store.get_config_variable(
+                'disable_request_compression'
+            )
+        else:
+            # if the user provided a value we must check if its a boolean
+            disabled = ensure_boolean(disabled)
+        config_kwargs['disable_request_compression'] = disabled
+
+    def _validate_min_compression_size(self, min_size):
+        if min_size is not None:
```
dlm6693, line 575: This is a bit of a quirk in the current implementation. This value should never be `None` since the config provider provides a default value. Without it, however, a number of test failures happen in `test_client` that do not explicitly set this variable on the config store. To avoid this, we may consider moving this function to `utils.py`, use it as the `conversion_func` in the config provider and call it in `_compute_request_compression_config` if set on the client config.

botocore/compress.py
```diff
@@ -0,0 +1,112 @@
...
+            )
+        return False
+
+    @staticmethod
+    def _get_body_size(body):
+        size = determine_content_length(body)
+        if size is None:
+            logger.debug(
+                'Unable to get length of the request body: %s. Not compressing.'
+                % body
```
nateprewitt, line 109: You don't need to use `%` with logger statements. Alternatively, we could use f-strings.
```suggestion
                'Unable to get length of the request body: %s. Skipping compression.', body
```

botocore/compress.py
```diff
@@ -0,0 +1,112 @@
...
+import gzip
+import io
+import logging
+
+from botocore.utils import determine_content_length
+
+logger = logging.getLogger(__name__)
+
+
+class RequestCompressor:
```
nateprewitt, line 23: What's our reasoning for the class here given we have no constants or shared state? It looks like we're only exposing a single method (`compress`).

botocore/compress.py
```diff
@@ -0,0 +1,112 @@
...
+    """A class that can compress a request body."""
+
+    @classmethod
+    def compress(cls, config, request_dict, operation_model):
+        """Compresses the request body using the specified encodings.
```
nateprewitt, line 28:
```suggestion
        """Attempt to compress the request body using the modeled encodings.
```

botocore/compress.py
```diff
@@ -0,0 +1,112 @@
...
+
+        Check if the request should be compressed based on the contents of its
+        body and config settings. Set or append the `Content-Encoding` header
+        with the matched encoding if not present.
```
nateprewitt, line 32:
```suggestion
```

botocore/compress.py
```diff
@@ -0,0 +1,112 @@
...
+                    elif encoding not in ce_header.split(','):
+                        headers['Content-Encoding'] = f'{ce_header},{encoding}'
+                    logger.debug(
+                        'Compressing request with %s encoding', encoding
+                    )
+                    request_dict['body'] = encoder(body)
+                else:
+                    logger.debug(
+                        'Unsupported compression encoding: %s' % encoding
+                    )
```
nateprewitt, line 53: I think we discussed this earlier, but doesn't this implementation violate the design? You're encoding the body with _all_ encodings that match an implemented function. Was there a reason we didn't fix this? We should be selecting the first supported and then moving on.

```suggestion
        if not cls._should_compress_request(config, body, operation_model):
            return
        encodings = operation_model.request_compression['encodings']
        encoding = _choose_encoding(encodings)
        if encoding is not None:
            encoder = encoding_map.get(encoding)
            logger.debug('Compressing request with %s encoding', encoding)
            request_dict['body'] = encoder(body)
            _set_compression_header(request_dict['headers'], encoding)
        
    def _set_compression_header(headers, encoding):
        content_encoding = headers.get('Content-Encoding')
        if content_encoding is None:
            headers['Content-Encoding'] = encoding
        elif encoding not in ce_header.split(','):
            headers['Content-Encoding'] = f'{ce_header},{encoding}'
                
    def _choose_encoding(encodings):
        for encoding in encodings:
            if encoding in encoding_map:
                return encoding
            else:
                logger.debug('Unsupported compression encoding: %s', encoding)
```

botocore/config.py
```diff
@@ -206,6 +206,21 @@ class Config:
         creating new connections if set to True.
 
         Defaults to False.
+
+    :type request_min_compression_size_bytes: int
+    :param request_min_compression_bytes: The minimum size in bytes that a
+    request body should be to trigger compression. A request is compressed
+    regardless of the size of the body if the operation has streaming input,
+    but if the operation also contains the `requiresLength` trait, the request
+    is not compressed.
```
nateprewitt, line 215: > A request is compressed regardless of the size of the body if the operation has streaming input,
    but if the operation also contains the `requiresLength` trait, the request
    is not compressed.

We have this written in a few places but it seems incorrect. We don't compress streaming input regardless of body size. We will compress all streaming input that doesn't require length.

botocore/config.py
```diff
@@ -206,6 +206,21 @@ class Config:
...
+
+        Defaults to None.
+
+    :type disable_request_compression: bool
+    :param disable_request_compression: Disables the request body compression
+    if set to True.
```
nateprewitt, line 221:
```suggestion
    :param disable_request_compression: Disables request body compression
    if set to True.
```

botocore/configprovider.py
```diff
@@ -147,6 +147,20 @@
     # whatever the defaults are in _retry.json.
     'max_attempts': ('max_attempts', 'AWS_MAX_ATTEMPTS', None, int),
     'user_agent_appid': ('sdk_ua_app_id', 'AWS_SDK_UA_APP_ID', None, None),
+    # This must be a parsable integer between 0 and 1048576, but validation
+    # is performed during client initialization instead of here.
```
nateprewitt, line 151: We can leave this to the validation and comment in the client to avoid needing to update this in many places.

nateprewitt (changes requested):

botocore/compress.py
dlm6693: I thought it was encouraged to use %s strings in logging statements for optimization reasons? Its pretty much the only place I ever use them. https://docs.python.org/3/howto/logging.html#optimization

botocore/compress.py
dlm6693: Ah so sorry yes this was a miss on my part. There is a missing `return` statement after `request_dict['body'] = encoder(body)`. I'll add a test to ensure that compression isn't attempted more than once when there is more than one compression encoding available in the operation model.

botocore/args.py
dlm6693: About 3/4 tests in `tests/unit/test_client.py`. The error is pretty much the same for all of them:

```
E           TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

The issue is that the test client is using an empty `ConfigValueStore` object: https://github.com/boto/botocore/blob/9ef2d48793a564d5a637d67f3af817116bfe1aff/tests/unit/test_client.py#L151
so there's nothing to fall back to. We could add defaults here as the config provider normally does, but it didn't seem wise to me to mess with something that is foundational to hundreds of tests.

The exact tests that fail are any in `TestAutoGeneratedClient` and `TestClientErrors` that don't explicitly set `request_min_compression_size_bytes` in either the config store or on a client config object.

botocore/args.py
```diff
@@ -543,6 +550,47 @@ def _compute_connect_timeout(self, config_kwargs):
...
+            error_msg_base = (
+                f'Invalid value "{min_size}" for '
+                'request_min_compression_size_bytes. '
+            )
```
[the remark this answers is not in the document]
dlm6693: Sure I can change that.

botocore/compress.py
nateprewitt: Note the difference between:

```python
'Unable to get length of the request body: %s. Not compressing.' % body
```

and

```python
'Unable to get length of the request body: %s. Skipping compression.', body
```

With the former syntax, you're circumventing the logger optimization. We don't want to do that.

botocore/args.py
```diff
@@ -543,6 +550,47 @@ def _compute_connect_timeout(self, config_kwargs):
...
+                raise botocore.exceptions.InvalidConfigError(
+                    error_msg=f'{error_msg_base}Value must be an integer.'
+                )
+            if not 0 <= min_size <= 1048576:
+                raise botocore.exceptions.InvalidConfigError(
+                    error_msg=(
+                        f'{error_msg_base}Value must '
+                        'be between 0 and 1048576.'
+                    )
+                )
```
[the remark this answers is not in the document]
dlm6693: Sure I'll put at the top of the file just beneath `USERAGENT_APPID_MAXLEN`. The value is contained within `error_msg_base`.

botocore/compress.py
dlm6693: This is discussed in the design. Its for organization and extensibility purposes. While it may not happen any time soon, additional compression encodings could be modeled in the future and require the SDK's support.

botocore/compress.py
dlm6693: If you think its a bit easier to read/digest by yanking the selection of an encoding into a helper function, I can certainly do that.

botocore/compress.py
dlm6693: Ah got it thank you for clarifying!

botocore/compress.py
dlm6693: I elected to not use a `_choose_encoding` helper method and instead include a `return` statement in the for loop if the encoder isn't None.  This makes it easier to log without passing the encoding name around.

botocore/args.py
jonemo: That explanation makes sense to me. But why does this only surface now, and not for example in #2955 where I added new config variables?

tests/functional/test_neptune.py
```diff
@@ -20,8 +20,8 @@ def setUp(self):
         self.http_stubber = ClientHTTPStubber(self.client)
 
     def assert_presigned_url_injected_in_request(self, body):
-        self.assertIn('PreSignedUrl', body)
-        self.assertNotIn('SourceRegion', body)
+        self.assertIn(b'PreSignedUrl', body)
+        self.assertNotIn(b'SourceRegion', body)
```
jonemo, line 24: What's causing these changes? Were these `body`s not also `bytes` before?

tests/unit/test_utils.py
```diff
@@ -3476,3 +3477,29 @@ def cached_fn(self, a, b):
...
+
+
+@pytest.mark.parametrize(
+    'request_dict, protocol, signature_version, expected_body',
+    [
+        ({'body': {'foo': 'bar'}}, 'query', 'v4', b'foo=bar'),
+        ({'body': b''}, 'query', 'v4', b''),
+        ({'body': {'foo': 'bar'}}, 'json', 'v4', {'foo': 'bar'}),
+        ({'body': {'foo': 'bar'}}, 'query', 'v2', {'foo': 'bar'}),
+        ({'body': 'foo=bar'}, 'query', 'v4', 'foo=bar'),
```
jonemo, line 3494: At the risk of revealing my ignorance here: Why is the expected body here a `str` whereas in line 3490 is is `bytes`?

tests/unit/test_compress.py
```diff
@@ -0,0 +1,274 @@
...
+def _op_with_compression():
+    op = mock.Mock()
+    op.request_compression = {'encodings': ['gzip']}
+    op.has_streaming_input = False
+    return op
+
+
+def _op_with_multiple_compressions():
+    op = mock.Mock()
+    op.request_compression = {'encodings': ['gzip', 'gzip']}
```
jonemo, line 32: Nit: Could leave a `todo` here to explain that you had to use the same value twice because it's the only valid one, and how in if there's ever another one this can be changed to two different ones. Otherwise someone five years from now will spend a day researching if it's safe to make that change :lolsob:

tests/unit/test_compress.py
```diff
@@ -0,0 +1,274 @@
...
+        'body': REQUEST_BODY,
+        'headers': {'foo': 'bar'},
+    }
+
+
+def request_dict_with_content_encoding_header():
+    return {
+        'body': REQUEST_BODY,
+        'headers': {'foo': 'bar', 'Content-Encoding': 'identity'},
+    }
```
jonemo, line 102: Nit: I think these can simply be constants instead of functions.

botocore/compress.py
```diff
@@ -0,0 +1,105 @@
...
+import gzip
+import io
+import logging
+
+from botocore.utils import determine_content_length
+
+logger = logging.getLogger(__name__)
+
+
+def compress_request(config, request_dict, operation_model):
```
jonemo, line 23: A bit late in the game to bring this up, but an idea after seeing all the examples in the tests where this doesn't compress: Would it be more clear to call this `maybe_compress_request`? It's a naming pattern I've first encountered in [Smithy repos](https://github.com/smithy-lang) and found very useful for understanding that a function only conditionally does what its name suggest.

nateprewitt: Yep, I think that's reasonable. I'll throw `attempt_request_compression` as an alternative but I think they both work well.

botocore/compress.py
dlm6693: Per offline discussion I've updated to a functional implementation.

tests/functional/test_neptune.py
dlm6693: Yeah the request's `body` value was a string before. Doing some investigation for why that was the case now.

dlm6693: This is the relevant bit of code: https://github.com/boto/botocore/blob/develop/botocore/awsrequest.py#L418-L420

It url encodes the body dictionary to a string, but doesn't then subsequently encode the string to bytes. It seems as though for `query` requests that gets handed off to urllib3.

tests/unit/test_compress.py
dlm6693: Good call! I can add a comment for future readers.

tests/unit/test_compress.py
dlm6693: I initially thought so too, but the changes persist between each test case. I think that would only work if these were immutable unfortunately.

botocore/compress.py
dlm6693: I think I actually called this `maybe_compress_request` in an earlier iteration so I'd vote to change it back to that.

botocore/args.py
dlm6693: I believe it's because no default is expected from the config provider for `sdk_ua_app_id` whereas for `request_compression_min_size_bytes` there is.

botocore/config.py
dlm6693: Per discussion offline updated wording in the latest commit.

botocore/args.py
```diff
@@ -59,6 +59,10 @@
 # Maximum allowed length of the ``user_agent_appid`` config field. Longer
 # values result in a warning-level log message.
 USERAGENT_APPID_MAXLEN = 50
+# Minimum and maximum allowed value of the `request_min_compression_size_bytes`
+# config field. Lower or higher values result in an InvalidConfigError raised.
```
nateprewitt, line 63: There's a question here of if these should be at module scope. That probably also should have been caught for `USERAGENT_APPID_MAXLEN`. They're now importable and not explicitly private. If they're only used in the single function can we just define them to that scope? We can always hoist higher later if needed, but this may be premature.

tests/functional/test_neptune.py
nateprewitt: I'm not sure I understand this in response to Jonas' question. We've changed the type of the body for these unrelated tests because of a new change we added. That seems to be because we're now forcing these to bytes prematurely when we're encoding everything in the [new query transform](https://github.com/boto/botocore/pull/2959/files#diff-d5dc8bb337890042943977b62e8bb2e7db71415d35bb17a41ad4bf4780fb3617R3435-R3445). That's done indiscriminately regardless of compression.

Do we have a reason we need to do that there? This feels like a very likely case we break something.

nateprewitt: Yeah, it looks like we broke most of these tests and then updated them to match the breakage. Removing [this line](https://github.com/boto/botocore/pull/2959/files#diff-d5dc8bb337890042943977b62e8bb2e7db71415d35bb17a41ad4bf4780fb3617R3445) that was added avoids this issue and doesn't seem to affect any of the tests.

Can we take another look at this on Monday? If there's not something I'm missing let's revert it.

tests/unit/test_utils.py
nateprewitt: They should both be strings I believe. Related issue in [this comment](https://github.com/boto/botocore/pull/2959#discussion_r1264245663) would make the tests consistent.

tests/functional/test_neptune.py
dlm6693: Thanks for catching this!

I wouldn't say the tests are unrelated necessarily. There is a section in the design doc that discusses the handlers that generate the presigned URLs which get injected into the bodies of some operations, and these are the ones in question. 

However, I don't see any harm in delaying encoding to bytes until later in the operation call if we're worried about potential breakages. I'll take another look Monday, but it should be fine to revert it.

botocore/args.py
dlm6693: Yeah that's totally fair. There's no need to expose these at all. I'll remove the comments as well since the uses should be more inherently obvious when defined in the function.

tests/unit/test_args.py
```diff
@@ -529,6 +529,90 @@ def test_doesnt_create_ruleset_resolver_if_not_given_data(self):
...
+        with self.assertRaises(exceptions.InvalidConfigError):
+            self.call_get_client_args()
+
+    def test_bad_value_disable_request_compression(self):
+        config = self.call_get_client_args(
+            client_config=Config(
+                disable_request_compression='foo',
+            ),
+        )['client_config']
+        self.assertFalse(config.disable_request_compression)
```
dlm6693, line 614: This test suite also uses an empty `ConfigValueStore` object. Because of that, this test would not use the `ensure_boolean` conversion function defined in the config provider. Maybe as a followup we should look into running these tests as well as the ones in `test_client.TestAutoGeneratedClient` using a config value store that contains the default config mapping.

tests/unit/test_compress.py
```diff
@@ -0,0 +1,276 @@
...
+    # is supported.
+    op.request_compression = {'encodings': ['gzip', 'gzip']}
+    op.has_streaming_input = False
+    return op
+
+
+def _op_unknown_compression():
+    op = mock.Mock()
+    op.request_compression = {'encodings': ['foo']}
+    op.has_streaming_input = None
```
nateprewitt, line 42: Why is this one `None` while the others are `False`?

tests/unit/test_compress.py
```diff
@@ -0,0 +1,276 @@
...
+    op = _streaming_op_with_compression()
+    streaming_shape = mock.Mock()
+    streaming_shape.metadata = {'requiresLength': True}
+    op.get_streaming_input.return_value = streaming_shape
+    return op
+
+
+OP_WITH_COMPRESSION = _op_with_compression()
+OP_UNKNOWN_COMPRESSION = _op_unknown_compression()
+STREAMING_OP_WITH_COMPRESSION = _streaming_op_with_compression()
```
nateprewitt, line 72: These feel a bit unnecessary, especially since you're assigning them all to constants. Could we simplify this with something like:

```suggestion

def _make_op(request_compression=None, streaming=None, metadata=None):
    op = mock.Mock()
    op.request_compression = request_compression
    op.has_streaming_input = streaming_input
    if metadata is not None:
        streaming_shape = mock.Mock()
        streaming_shape.metadata = metadata
        op.get_streaming_input.return_value = streaming_shape
    return op


OP_WITH_COMPRESSION = _make_op({'encodings': ['gzip']}, False)
OP_WITHOUT_COMPRESSION = _make_op(None, False)
OP_WITH_MULTIPLE_COMPRESSION = _make_op({'encodings': ['gzip', 'gzip']}, False)
OP_UNKNOWN_COMPRESSION = _make_op({'encodings': ['foo']}, None)
STREAMING_OP_WITH_COMPRESSION = _make_op({'encodings': ['gzip']}, True, {})
STREAMING_OP_WITH_REQUIRES_LENGTH = _make_op({'encodings': ['gzip']}, True, {'requiresLength': True})

```

botocore/args.py
```diff
@@ -543,6 +550,52 @@ def _compute_connect_timeout(self, config_kwargs):
...
+            disabled = self._config_store.get_config_variable(
+                'disable_request_compression'
+            )
+        else:
+            # if the user provided a value we must check if it's a boolean
+            disabled = ensure_boolean(disabled)
+        config_kwargs['disable_request_compression'] = disabled
+
+    def _validate_min_compression_size(self, min_size):
+        min_allowed_min_size = 0
```
nateprewitt, line 575: What was the final decision on this? Are we actually creating larger bodies for empty bodies?

dlm6693: I got push back when I asked about this offline. I'm double checking with the owner of the spec for confirmation.  However, it does look like the Ruby SDK, which released compression a few weeks ago, has the absolute minimum at 0:

https://github.com/aws/aws-sdk-ruby/blob/ee2f883b8766aded0b86b3ee2b27575c90d9f3ac/gems/aws-sdk-core/lib/aws-sdk-core/plugins/request_compression.rb#L50

tests/unit/test_compress.py
```diff
@@ -0,0 +1,276 @@
...
+    op.request_compression = {'encodings': ['gzip']}
+    op.has_streaming_input = False
+    return op
+
+
+def _op_with_multiple_compressions():
+    op = mock.Mock()
+    # TODO: Update second value to a different encoding. Currently only gzip
+    # is supported.
+    op.request_compression = {'encodings': ['gzip', 'gzip']}
```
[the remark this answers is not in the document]
dlm6693: This test case is to ensure that compression doesn't run twice even when more than one encoding is available so we need real encodings that have existing methods in the SDK to properly validate this. Since `gzip` is currently the only one, we have to use it twice.

tests/functional/test_compress.py
```diff
@@ -0,0 +1,136 @@
...
+        region_name="us-west-2",
+        config=Config(request_min_compression_size_bytes=100),
+    )
+    with ClientHTTPStubber(client, strict=True) as http_stubber:
+        http_stubber.add_response(
+            status=200,
+            body=(
+                b"<response><status>success</status><message>"
+                b"Request processed successfully</message></response>"
+            ),
```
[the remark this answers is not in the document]
dlm6693: Yeah using the default empty bytes I get this
```
E           botocore.parsers.ResponseParserError: Unable to parse response (no element found: line 1, column 0), invalid XML received. Further retries may succeed:
E           b''
```
I could use a different protocol, but given the changes I thought using `query` made the most sense for a functional test.

tests/unit/test_compress.py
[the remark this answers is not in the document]
nateprewitt: > so we need real encodings that have existing methods in the SDK to properly validate this.

What's prompting that requirement? If you configure this as `["gzip", "foo"]`, you're testing that only the first is used. If we reach foo, we've failed your test case. If they're both `gzip` and you assert `gzip` was used, we're adding ambiguity on which was selected.

tests/functional/test_context_params.py
```diff
@@ -226,19 +226,6 @@
...
-    def mock_load_service_model(service_name, type_name, api_version=None):
-        if type_name == 'service-2':
-            return service_model_json
-        if type_name == 'endpoint-rule-set-1':
-            return ruleset_json
-
-    loader = session.get_component('data_loader')
-    monkeypatch.setattr(loader, 'load_service_model', mock_load_service_model)
-
-
```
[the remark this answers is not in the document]
dlm6693: I needed this function for the functional tests. It didn't really make sense to import it directly from `test_context_params`.  Would you expect it to be defined twice instead, then cleaned up in a follow up PR?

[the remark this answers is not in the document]
nateprewitt: No, but you could create a PR that defines these in `tests/__init__.py` and then imports them into the appropriate locations, right? Is any of this specific to request compression?

tests/unit/test_compress.py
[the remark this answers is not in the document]
dlm6693: `_assert_compression` will only ever decompress once so if it did use gzip twice the test would fail. To your point though, there is still ambiguity on which gzip in the list was selected. However, `foo` will never compress since its not a supported encoding. How do we validate that `foo` was never reached? I don't really see a clear way of doing that.

tests/functional/test_context_params.py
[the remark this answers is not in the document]
dlm6693: Sure I can do that.

tests/unit/test_compress.py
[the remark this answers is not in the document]
dlm6693: On second thought, I'll change to a mocked value like "foo". Since multiple compressions aren't supported yet, we don't necessarily need to test for this behavior, just that compression is only run once. If and when we do, we can add a test case that explicitly tests an operation with multiple supported compression encodings.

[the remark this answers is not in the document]
nateprewitt: I think there's a lot of ways we can ensure this is checked. A simple option would be to do something like:

```
def bad_compression():
    raise ValueError("Reached unintended compression algorithm 'foo'")
    
MOCK_COMPRESSION = {"foo": bad_compression}.update(COMPRESSION_MAPPING)
with mock.patch('botocore.compress.COMPRESSION_MAPPING', MOCK_COMPRESSION):

File: tests/functional/test_context_params.py (hunk 43 above)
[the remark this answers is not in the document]
Comment (dlm6693) [reply tests/functional/test_context_params.py:241]: https://github.com/boto/botocore/pull/2986

File: tests/unit/test_compress.py (hunk 41 above)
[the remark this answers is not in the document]
Comment (dlm6693) [reply tests/unit/test_compress.py:34]: Thanks! I guess I was worried about messing with internals and using mocking, but this seems like a decent option.

File: tests/functional/test_compress.py (hunk 42 above)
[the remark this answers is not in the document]
Comment (nateprewitt) [reply tests/functional/test_compress.py:121]: Would something like what we use in other XML based tests not work?

```python
 http_stubber.add_response(body=b'<response/>')
 ```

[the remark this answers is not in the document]
Comment (dlm6693) [reply tests/functional/test_compress.py:121]: Sorry I thought you were referring to defining the body at all. And yeah that should definitely work.

File: tests/unit/test_compress.py (hunk 38 above)
Comment (dlm6693) [reply tests/unit/test_compress.py:42]: I think this must've been a mistake. It'll be cleaned up in the next commit.

File: botocore/args.py (hunk 40 above)
Comment (dlm6693) [reply botocore/args.py:575]: Seems as though we're updating the spec for the minimum to be at least 1. I'll update this accordingly

File: botocore/compress.py [hunk 44]
```diff
@@ -0,0 +1,119 @@
...
+        if hasattr(body, 'seek') and hasattr(body, 'tell'):
+            current_position = body.tell()
+            compressed_obj = _gzip_compress_fileobj(body)
+            body.seek(current_position)
+            return compressed_obj
+        return _gzip_compress_fileobj(body)
+    elif isinstance(body, dict):
+        return gzip.compress(
+            urlencode(body, doseq=True, encoding='utf-8').encode('utf-8')
+        )
```
nateprewitt, line 94:
```suggestion
```

botocore/compress.py
```diff
@@ -0,0 +1,119 @@
...
+            # length. If compression should occur, the encoded body will be saved
+            # to the request dictionary so it doesn't need to be encoded again.
+            body = request_dict['body']
+            if isinstance(body, dict):
+                body = urlencode(body, doseq=True, encoding='utf-8')
+            config_min = config.request_min_compression_size_bytes
+            should_compress = config_min <= _get_body_size(body)
+            if should_compress:
+                request_dict['body'] = body
+            return should_compress
```
nateprewitt, line 63:
```suggestion
        if not _attempt_type_normalization(request_dict):
            body_type = type(request_dict['body'])
            logger.debug("Body type {body_type} does not support compression.")
            return False
        if operation_model.has_streaming_input:
            streaming_metadata = operation_model.get_streaming_input().metadata
            return 'requiresLength' not in streaming_metadata

        body_size = _get_body_size(request_dict['body'])
        should_compress = config.request_min_compression_size_bytes <= body_size
        return should_compress

def _attempt_type_normalization(request_dict):
    """Attempt to coerce request body into format for compression."""
    if isinstance(request_dict['body'], dict):
        body = urlencode(request_dict['body'], doseq=True, encoding='utf-8')
        request_dict['body'] = body
    is_supported_type = isinstance(request_dict['body'], [str, bytes, bytesarray])
    return is_supported_type or hasattr(body, 'read')
```

nateprewitt (commented):
This is a rough idea but let's try to avoid having to do this kind of normalization with every new encoder we add. Ideally we have a rough type interface for what we'll support compressing. I proposed we do `bytes`, `bytesarray`, `str` and objects that support `read`.

botocore/compress.py
dlm6693: Thanks so much @nateprewitt! Taking a look at this now, but I think broad strokes this proposal makes sense.

dlm6693: The only functional difference I made is doing the entire encoding (i.e. string to bytes) when coercing the dict into a compressible type. I also changed the name of the function to `_is_compressible_type` since that more closely reflects what the function is returning, while adding a more specific comment instead of a doc string.

tests/unit/test_compress.py
```diff
@@ -0,0 +1,382 @@
...
+    ],
+)
+def test_compression_non_seekable_streams(
+    config, request_dict, operation_model, encoding
+):
+    # since the body can't be reset, we must make a copy
+    # of the original body to test against
+    original_body = deepcopy(request_dict['body'])
+    maybe_compress_request(config, request_dict, operation_model)
+    assert_compression(original_body, request_dict, encoding)
```
dlm6693, line 260: This can definitely be rolled up into the normal `test_compression` test cases, but it didn't seem right to make copies of the body when it isn't necessary.

botocore/args.py
```diff
@@ -260,10 +260,17 @@ def compute_client_args(
                 tcp_keepalive=client_config.tcp_keepalive,
                 user_agent_extra=client_config.user_agent_extra,
                 user_agent_appid=client_config.user_agent_appid,
+                request_min_compression_size_bytes=(
+                    client_config.request_min_compression_size_bytes
+                ),
+                disable_request_compression=(
+                    client_config.disable_request_compression
+                ),
```
dlm6693, line 268: For formatting, we could alternatively do the following. The names would need to be very short for it to work with black.
```suggestion
         min_comp_size =  client_config.request_min_compression_size_bytes
         disable_comp =  client_config.disable_request_compression
         ...
                request_min_compression_size_bytes=min_comp_size,
                disable_request_compression=disable_comp,
```

tests/unit/test_compress.py
```diff
@@ -0,0 +1,382 @@
...
+    def __init__(self, buffer):
+        self._buffer = buffer
+
+    def read(self, size=None):
+        return self._buffer.read(size)
+
+
+def _request_dict(body=None, headers=None):
+    if body is None:
+        body = b''
```
nateprewitt, line 86: The reason we often use `None` over a more reasonable default is because you can't use an empty list or dict in function signatures. Otherwise, it's evaluated at runtime before any code is executed and assigned to a singleton that carries state between invocations.

e.g.
```python
def test(i, data={}):
   data[i] = 'hello'
   print(data)

for i in range(5):
    test(i)

{0: 'hello'}
{0: 'hello', 1: 'hello'}
{0: 'hello', 1: 'hello', 2: 'hello'}
{0: 'hello', 1: 'hello', 2: 'hello', 3: 'hello'}
{0: 'hello', 1: 'hello', 2: 'hello', 3: 'hello', 4: 'hello'}
```

If we're just declaring something like an empty byte-string we can just do:
`def _request_dict(body=b"", headers=None):`

tests/functional/test_compress.py
```diff
@@ -0,0 +1,128 @@
...
+        ]
+        client.mock_operation(MockOpParamList=params_list)
+        param_template = (
+            "MockOpParamList.member.{i}.MockOpParam=MockOpParamValue{i}"
+        )
+        serialized_body = "&".join(
+            param_template.format(i=i) for i in range(1, 21)
+        )
+        actual_body = gzip.decompress(http_stubber.requests[0].body)
+        assert serialized_body.encode('utf-8') in actual_body
```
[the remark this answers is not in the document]
dlm6693: The body includes the name of the operation and the API version.

```
b'Action=MockOperation&Version=2020-02-02&MockOpParamList.member.1.MockOpParam=MockOpParamValue1&...
```
I can definitely add this to the test and assert an exact match instead. Is that preferable?

tests/unit/test_compress.py
```diff
@@ -0,0 +1,382 @@
...
+    b'&MetricData.member.1.MetricName=metric&MetricData.member.1.Unit=Bytes'
+    b'&MetricData.member.1.Value=128'
+)
+
+REQUEST_BODY_STRING = REQUEST_BODY.decode('utf-8')
+
+DEFAULT_COMPRESSION_CONFIG = Config(
+    disable_request_compression=False,
+    request_min_compression_size_bytes=10420,
+)
```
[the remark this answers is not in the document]
dlm6693: Its labeled 'default' because these are what the values will default to from the config provider if not provided by a customer. However, since we are not initializing a client here, we need to write them ourselves. To prevent the actual defaults and these tests from ever getting out of sync, I could do something like this.

```
from botocore.configprovider import BOTOCORE_DEFAUT_SESSION_VARIABLES

default_disable_request_compression = BOTOCORE_DEFAUT_SESSION_VARIABLES['disable_request_compression'[2]
default_request_min_compression_size_bytes = BOTOCORE_DEFAULT_SESSION_VARIABLES['request_min_compression_size_bytes'][2]

DEFAULT_COMPRESSION_CONFIG = Config(
    disable_request_compression=default_disable_request_compression,
    request_min_compression_size_bytes=default_request_min_compression_size_bytes,
)

File: tests/unit/test_compress.py (hunk 48 above)
Comment (dlm6693) [reply tests/unit/test_compress.py:86]: Ah yeah this was a mistake. I should know that its only mutable defaults that are concerning. I'll fix.

File: tests/unit/test_compress.py [hunk 51]
```diff
@@ -0,0 +1,382 @@
...
+COMPRESSION_CONFIG_128_BYTES = Config(
+    disable_request_compression=False,
+    request_min_compression_size_bytes=128,
+)
+COMPRESSION_CONFIG_1_BYTE = Config(
+    disable_request_compression=False,
+    request_min_compression_size_bytes=1,
+)
```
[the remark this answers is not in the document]
dlm6693: 128 comes from test cases outlined in the spec. I believe that number was chosen because its a sensible value that is less than the length of `REQUEST_BODY`, which is also used in the test cases in the spec. So its a useful way to outline test cases using this body derived from the `PutMetricData` operation where we're validating whether or not compression is enabled since we know its of the requisite length.

1 could be removed, but I chose it for a few reasons. First I've defined a request dict using a dict body that is rather short `{'foo': 'bar'}`. For this case specifically, I wanted to ensure that compression was enabled and able to parse/compress a dictionary body successfully. However, I could parse `REQUEST_BODY` into a dictionary instead. I also used it for test cases where we expect that the length of the body couldn't be obtained (incompatible types and nonseekable streams whose operations don't have streaming input. These are essentially testing `_get_body_size` is allowing us to disable compression for certain body types even when it is at its absolute lowest.

10240 is the default value. When using it with `REQUEST_BODY`, it is a convenient way to test inverse cases that we 're doing for 128. We know that this number is quite a lot bigger than the length of the body, so we can use it to check when we expect compression to be enabled/disabled. For example, if an operation has the compression trait and streaming input, we know that compression should occur regardless of the length of the body.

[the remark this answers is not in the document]
dlm6693: I believe we can safely migrate away from 1 without changing any of the test cases, except for the dictionary. Let me know if you think I should go ahead and do that.

tests/functional/test_compress.py
```diff
@@ -0,0 +1,128 @@
...
+# language governing permissions and limitations under the License.
+
+import gzip
+
+import pytest
+
+from botocore.config import Config
+from tests import ALL_SERVICES, ClientHTTPStubber, patch_load_service_model
+
+KNOWN_COMPRESSION_ENCODINGS = ("gzip",)
```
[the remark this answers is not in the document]
dlm6693: That's a great question. This was a miss and I should be importing it instead. Will fix.

tests/unit/test_compress.py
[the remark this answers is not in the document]
dlm6693: I'm also realizing this constant is misspelled. Should we fix this in a separate PR?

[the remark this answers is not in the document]
dlm6693: This should also have been 10240 not 10420 🤦 . It's fixed in the latest commit

tests/functional/test_compress.py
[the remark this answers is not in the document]
nateprewitt: Yeah, if it's deterministic we should do something like:

```suggestion
        body_params = f'SomeTemplate{serialized_body}'
        assert body_params.encode('utf-8') == actual_body
```

nateprewitt:
One minor comment, but once tests pass, I think this should be set.

nateprewitt (approved):
Looks good pending passing tests, thanks @dlm6693!

PR #2964, opened 2023-06-05
Configured endpoint url resolver functional tests

kdaily:
Add functional testing for use of a configured endpoint url through the shared config file and environment variables.

First, the tests use a data file to load tests that enumerate the combinatorial ways that the endpoint URL can be defined (shared configuration file and environment variable for individual service and global/all services), and asserts that the correct endpoint is returned.

The tests also assert that the correct config section name and environment variable name are used to configure and override the endpoint URL for all services.

Lastly, the tests check that a operations for a subset of services use the endpoint override in a request.

### DIFF
--- a/tests/functional/configured_endpoint_urls/__init__.py
+++ b/tests/functional/configured_endpoint_urls/__init__.py
@@ -1,0 +1,12 @@
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
--- a/tests/functional/configured_endpoint_urls/profile-tests.json
+++ b/tests/functional/configured_endpoint_urls/profile-tests.json
@@ -1,0 +1,478 @@
+{
+  "description": [
+    "These are test descriptions that describe how specific data should be loaded from a profile file based on a ",
+    "profile name."
+  ],
+
+  "testSuites": [
+    {
+      "profiles": {
+        "default": {
+          "aws_access_key_id": "123",
+          "aws_secret_access_key": "456",
+          "region": "fake-region-10"
+        },
+        "service_localhost_global_only": {
+          "aws_access_key_id": "123",
+          "aws_secret_access_key": "456",
+          "region": "fake-region-10",
+          "endpoint_url": "http://localhost:1234"
+        },
+        "service_global_only": {
+          "aws_access_key_id": "123",
+          "aws_secret_access_key": "456",
+          "region": "fake-region-10",
+          "endpoint_url": "https://global.endpoint.aws"
+        },
+        "service_specific_s3": {
+          "aws_access_key_id": "123",
+          "aws_secret_access_key": "456",
+          "services": "service_specific_s3",
+          "region": "fake-region-10"
+        },
+        "global_and_service_specific_s3": {
+          "aws_access_key_id": "123",
+          "aws_secret_access_key": "456",
+          "endpoint_url": "https://global.endpoint.aws",
+          "services": "service_specific_s3",
+          "region": "fake-region-10"
+        },
+        "ignore_global_and_service_specific_s3": {
+          "aws_access_key_id": "123",
+          "aws_secret_access_key": "456",
+          "endpoint_url": "https://global.endpoint.aws",
+          "services": "service_specific_s3",
+          "region": "fake-region-10",
+          "ignore_configured_endpoint_urls": "true"
+        },
+        "service_specific_dynamodb_and_s3": {
+          "aws_access_key_id": "123",
+          "aws_secret_access_key": "456",
+          "services": "service_specific_dynamodb_and_s3",
+          "region": "fake-region-10"
+        }
+      },
+
+      "services": {
+        "service_specific_s3": {
+          "s3": {
+            "endpoint_url": "https://s3.endpoint.aws"
+          }
+        },
+        "service_specific_dynamodb_and_s3": {
+          "dynamodb": {
+            "endpoint_url": "https://dynamodb.endpoint.aws"
+          },
+          "s3": {
+            "endpoint_url": "https://s3.endpoint.aws"
+          }
+        }
+      },
+
+      "client_configs": {
+        "default": {},
+        "endpoint_url_provided":{
+          "endpoint_url": "https://client-config.endpoint.aws"
+        }
+      },
+
+      "environments": {
+        "default": {},
+        "global_only": {
+          "AWS_ENDPOINT_URL": "https://global-from-envvar.endpoint.aws"
+        },
+        "service_specific_s3": {
+          "AWS_ENDPOINT_URL_S3": "https://s3-from-envvar.endpoint.aws"
+        },
+        "global_and_service_specific_s3": {
+          "AWS_ENDPOINT_URL": "https://global-from-envvar.endpoint.aws",
+          "AWS_ENDPOINT_URL_S3": "https://s3-from-envvar.endpoint.aws"
+
+        },
+        "ignore_global_and_service_specific_s3": {
+          "AWS_ENDPOINT_URL": "https://global-from-envvar.endpoint.aws",
+          "AWS_ENDPOINT_URL_S3": "https://s3-from-envvar.endpoint.aws",
+          "AWS_IGNORE_CONFIGURED_ENDPOINT_URLS": "true"
+        },
+        "service_specific_dynamodb_and_s3": {
+          "AWS_ENDPOINT_URL_DYNAMODB": "https://dynamodb-from-envvar.endpoint.aws",
+          "AWS_ENDPOINT_URL_S3": "https://s3-from-envvar.endpoint.aws"
+        }
+      },
+
+    "endpointUrlTests": [
+      {
+        "name": "Global endpoint url is read from services section and used for an S3 client.",
+        "profile": "service_global_only",
+        "client_config": "default",
+        "environment": "default",
+        "service": "s3",
+        "output": {
+          "endpointUrl": "https://global.endpoint.aws"
+        }
+      },
+      {
+        "name": "Service specific endpoint url is read from services section and used for an S3 client.",
+        "profile": "service_specific_s3",
+        "client_config": "default",
+        "environment": "default",
+        "service": "s3",
+        "output": {
+          "endpointUrl": "https://s3.endpoint.aws"
+        }
+      },
+      {
+        "name": "S3 Service-specific endpoint URL from configuration file takes precedence over global endpoint URL from configuration file.",
+        "profile": "global_and_service_specific_s3",
+        "client_config": "default",
+        "environment": "default",
+        "service": "s3",
+        "output": {
+          "endpointUrl": "https://s3.endpoint.aws"
+        }
+      },
+      {
+        "name": "Global endpoint url environment variable takes precedence over the value resolved by the SDK.",
+        "profile": "default",
+        "client_config": "default",
+        "environment": "global_only",
+        "service": "s3",
+        "output": {
+          "endpointUrl": "https://global-from-envvar.endpoint.aws"
+        }
+      },
+      {
+        "name": "Global endpoint url environment variable takes precendence over global endpoint configuration option.",
+        "profile": "service_global_only",
+        "client_config": "default",
+        "environment": "global_only",
+        "service": "s3",
+        "output": {
+          "endpointUrl": "https://global-from-envvar.endpoint.aws"
+        }
+      },
+      {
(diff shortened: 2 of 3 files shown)


tests/functional/test_configured_endpoint_url.py
```diff
@@ -106,6 +110,7 @@ def setup_configuration_environment(test_case):
     test_case['augmented_data']['profile_string'] = (
         profile_str + service_section_str
     )
+    print(f"{test_case['augmented_data']['profile_string']}")
```
kdaily, line 113: Remove this.

tests/functional/test_configured_endpoint_url.py
```diff
@@ -145,3 +150,92 @@ def test_resolve_configured_endpoint_url(test_case):
...
+    for service_name, service_id in _known_service_names_and_ids():
+        config_transformed_service_id = service_id.replace(" ", "_").upper()
+        test_case = _generate_test_data_for_service_env_var(
+            service_name, config_transformed_service_id
+        )
+        setup_configuration_environment(test_case)
+        yield test_case
+
+
+def _generate_test_data_for_service_shared_config(service_name, service_id):
```
kdaily, line 187: rename to `_generate_test_case...`

tests/functional/test_configured_endpoint_url.py
```diff
@@ -145,3 +150,92 @@ def test_resolve_configured_endpoint_url(test_case):
...
+                "services": "service_specific",
+            },
+            "environment": {},
+            "client_config": {},
+            'services': {service_id: {"endpoint_url": ENDPOINT_OVERRIDE}},
+        },
+    }
+
+
+def _generate_test_data_for_service_env_var(service_name, service_id):
```
kdaily, line 205: rename to `_generate_test_case...`

botocore/utils.py
```diff
@@ -3315,3 +3325,77 @@ def _serialize_if_needed(self, value, iso=False):
...
+
+
+SERVICE_NAME_ALIASES = {'runtime.sagemaker': 'sagemaker-runtime'}
+
+
+def handle_service_name_alias(service_name, **kwargs):
+    return SERVICE_NAME_ALIASES.get(service_name, service_name)
+
+
+SERVICE_RENAMES = {
```
nateprewitt, line 3337: All of this should be clearly private, I don't think we want anything being built on this externally.

tests/unit/cfg/aws_services_config
```diff
@@ -0,0 +1,10 @@
+[default]
+endpoint_url = https://localhost:1234/
+services = my-services
+
+[services my-services]
+s3 =
+  endpoint_url = https://localhost:5678/
+  addressing_style = path
```
nateprewitt, line 8: How does this reconcile against defining this in the default profile? I may have missed where that logic is handled.

botocore/configprovider.py
```diff
@@ -851,3 +858,136 @@ def provide(self):
...
+        return self._scoped_config.get("endpoint_url")
+
+    def _snakecase_service_id(self, service_id):
+        return service_id.replace(" ", "_")
+
+    def _get_service_env_var_name(self):
+        transformed_service_id_env = self._snakecase_service_id(
+            self._service_id
+        ).upper()
+        return f"AWS_ENDPOINT_URL_{transformed_service_id_env}"
```
nateprewitt, line 965: How are we expecting customers to derive these?

botocore/utils.py
```diff
@@ -1533,6 +1533,16 @@ def hyphenize_service_id(service_id):
...
 
 
+def unhyphenize_service_name(service_name):
+    """Translate from the service name used internally back to its original.
+
+    Note: does not change capitalization.
+
+    :param service_name: The service name to convert.
+    """
+    return service_name.replace('-', ' ')
```
nateprewitt, line 1543: Is this right? Do we know that no services names will have a dash in them?

tests/unit/cfg/aws_services_config
kdaily: Nope shouldn't be there, need to remove that. Thanks!

tests/functional/test_configured_endpoint_url.py
```diff
@@ -0,0 +1,283 @@
...
+    service_section_str = dict_to_ini_section(
+        test_case['augmented_data']['services'],
+        section_header=f'services {services_section_name}',
+    )
+    test_case['augmented_data']['profile_string'] = (
+        profile_str + service_section_str
+    )
+
+
+def get_env_config_file_patched_client(session, test_case):
```
kyleknap, line 116: Let's make the following updates to this function:
1. Rename it to `create_client()`. The added descriptions to the function name make it hard to follow.
2. Turn the `create_client()` function into a pytest fixture that yields back a factory to create clients. This will allow us to keep the temporary file and environment patch around for the length of the test instead of have it end when the client is returned
3. Create the session as part of the factory. This will allow us to avoid patching the `configprovider` module

So in the end, I'd imagine it would look roughly like this:
```python
@pytest.fixture
def create_client(tmp_path):
    tmp_config_file = tmp_path / config
    environ = {'AWS_CONFIG_FILE': tmp_config_file}
    def _do_create_client(service, profile, client_args=None, config_file_contents=None, environment=None):
         environ.update(environment)
         # Write contents to tmp config file here
         session = botocore.session.Session(profile=profile)
         return client = session.create_client(service, **client_args)

    with mock.patch('os.envrion', environ)
        yield _do_create_client
```

tests/functional/test_configured_endpoint_url.py
```diff
@@ -0,0 +1,283 @@
...
+@pytest.mark.parametrize(
+    'service_name,service_model', _known_service_names_and_ids()
+)
+def test_service_env_var_name_is_correct(
+    service_name, service_model, botocore_session
+):
+    test_case = create_service_env_test_case(service_name, service_model)
+    setup_configuration_environment(test_case)
+
+    client = get_env_config_file_patched_client(botocore_session, test_case)
```
kyleknap, line 246: It would be interesting to see if we can simplify these test cases down. Mainly it seems the complexity is in we are trying to reuse the test case schema in order to reuse the scaffolding and I think if we can move away from building these test cases using the test schema it can help simplifying the tests. I think with some of the suggestions that I had early should help with this. So for example for this test it would be:
```python
@pytest.mark.parametrize(
    'service_name,service_model', _known_service_names_and_ids()
)
def test_service_env_var_name_is_respected(
    service_name, service_model, create_client
):
    expected_endpoint_url_env_var = (
        ''AWS_ENDPOINT_URL_{service_model.service_id.replace(' ', '_').upper()}'
    )
    client = create_client(
        service_name,
        environment={
           expected_endpoint_url_env_var: 'https://endpoint-override'
        }
    assert_endpoint_url_used(
        client=client, expected_endpoint_url='https://endpoint-override'
    )
```

tests/functional/test_configured_endpoint_url.py
```diff
@@ -0,0 +1,283 @@
...
+        yield new_test
+
+
+def _normalize_endpoint(url):
+    split_endpoint = urlsplit(url)
+    actual_endpoint = f"{split_endpoint.scheme}://{split_endpoint.netloc}"
+    return actual_endpoint
+
+
+def assert_endpoint_url_used(client, expected_endpoint_url):
```
kyleknap, line 75: Let's rename this to something more explicit to make it more clear on the type of assertion we are doing. Mainly based on the assertion name here, I'd assume it would be making an assertion similar to `assert_endpoint_url_used_for_operation`. Maybe we can instead call it `assert_client_endpoint_url` to indicate we are asserting the endpoint_url on the client?

tests/functional/test_configured_endpoint_url.py
```diff
@@ -0,0 +1,283 @@
...
+    assert client.meta.endpoint_url == expected_endpoint_url
+
+
+def assert_endpoint_url_used_for_operation(
+    client, expected_endpoint_url, operation=None, params=None
```
kyleknap, line 80: For `operation` and `params` we should probably make it required.

kyleknap (commented):
Looks like a good start. For this initial pass, I focused on seeing how we can trim the tests down to make it easier to walk through. In the next pass, we can focus more on the details on what cases we are testing

tests/functional/test_configured_endpoint_url.py
kdaily: I'm pretty sure I still need to provide a profile here (even a default one).

tests/functional/test_configured_endpoint_url.py
```diff
@@ -0,0 +1,279 @@
...
+    )
+
+
+@pytest.mark.parametrize(
+    'service_name,service_model', _known_service_names_and_models()
+)
+def test_service_env_var_name_is_correct(
+    service_name, service_model, client_creator
+):
+    test_case = create_service_env_test_case(service_name, service_model)
```
kyleknap, line 243: I have the same comment as the shared config version of this test. Let's remove this helper test function and have the inputs to `create_client()` explicitly defined in the test case body to make the test easier to follow.

tests/functional/test_configured_endpoint_url.py
```diff
@@ -0,0 +1,279 @@
...
+        config_file_contents=None,
+        environment=None,
+    ):
+        environ.update(environment)
+        tmp_config_file = open(tmp_config_file_path, 'w')
+        tmp_config_file.write(config_file_contents)
+        tmp_config_file.flush()
+
+        session = botocore.session.Session(profile=profile)
+        session.set_config_variable('profile', profile)
```
kyleknap, line 100: Do we need to call `set_config_variable` if it is provided on the `Session` object?

tests/functional/test_configured_endpoint_url.py
```diff
@@ -0,0 +1,279 @@
...
+        service,
+        profile,
+        client_args=None,
+        config_file_contents=None,
+        environment=None,
+    ):
+        environ.update(environment)
+        tmp_config_file = open(tmp_config_file_path, 'w')
+        tmp_config_file.write(config_file_contents)
+        tmp_config_file.flush()
```
kyleknap, line 97: Let's use a context manager to make sure we close the file after writing it:
```python
with open(tmp_config_file_path, 'w') as f:
    f.write(config_file_contents)
    f.flush()
```

tests/functional/test_configured_endpoint_url.py
```diff
@@ -0,0 +1,279 @@
...
+        == expected_endpoint_url
+    )
+
+
+def _known_service_names_and_models():
+    my_session = botocore.session.get_session()
+    loader = my_session.get_component('data_loader')
+    available_services = loader.list_available_services('service-2')
+
+    result = set()
```
kyleknap, line 139: Is there a reason a set is used here? I think if we use a list instead it should be able to keep a stable order.

tests/functional/data/configured-endpoint-urls/test-cases/profile-tests.json
```diff
@@ -0,0 +1,426 @@
...
+        "client_config": "default",
+        "environment": "ignore_global_and_service_specific_s3",
+        "service": "s3",
+        "output": {
+          "endpointUrl": "https://s3.fake-region-10.amazonaws.com"
+        }
+      },
+      {
+        "name": "All configured endpoints ignored due to shared config variable.",
+        "profile": "ignore_global_and_service_specific_s3",
```
kyleknap, line 413: Do we have a case when the ignore configured endpoints is set and we provide an endpoint_url explicitly? If not, it would probably be worth adding one.

kyleknap (commented):
It's looking better. I just had some more comments related to cases we are testing and structuring the tests.

tests/functional/data/configured-endpoint-urls/test-cases/profile-tests.json
```diff
@@ -0,0 +1,426 @@
+{
```
kyleknap, line 1: Another thing I noticed... Instead of making a new `data` directory in the functional test suite, let's just make a new directory for `configured-endpoint-urls` and make this JSON file as a direct child in that directory. So the end directory structure would be:
```
tests/functional/configured-endpoint-urls/
    __init__.py
   test_configured_endpoint_url.py
   test-cases.json
```
Mainly I don't think we should be creating a `data` directory which conflates with the model directory naming pattern at the root of the functional test suite just for this set of tests.

tests/functional/data/configured-endpoint-urls/test-cases/profile-tests.json
```diff
@@ -0,0 +1,426 @@
...
+      "services": {
+        "service_specific_s3": {
+          "s3": {
+            "endpoint_url": "https://s3.endpoint.aws"
+          }
+        },
+        "service_specific_dynamodb": {
+          "s3": {
+            "endpoint_url": "https://s3.endpoint.aws"
+          }
```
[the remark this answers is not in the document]
kdaily: I'll remove it.

The idea was to have a test that set `service=dynamodb` and then assert that an endpoint set for a different service was not used. The difficulty there was to be able to define exactly what the expected URL would be for the default. However in retrospect even in that case this services section isn't needed, because I could write the test with one of the existing service sections for S3 and then set the service to DynamoDB.

tests/functional/data/configured-endpoint-urls/test-cases/profile-tests.json
kdaily: We do now!

kyleknap (approved):
Cool this looks good assuming CI passes. :ship: Let's also make sure to squash this before merging it.

PR #2967, opened 2023-06-08
Start testing 3.12 pre-releases

nateprewitt:
This PR will start running our test suite against Python 3.12 beta builds to hopefully catch any issues early before the final release later this year.


requirements-dev.txt
```diff
@@ -2,6 +2,7 @@ wheel==0.38.1
 behave==1.2.5
 jsonschema==2.5.1
 coverage==5.5
+setuptools==67.8.0;python_version>="3.12"
```
dlm6693, line 5: What happens if we don't pin `setuptools` to this version for 3.12?

setup.py
```diff
@@ -52,11 +52,13 @@ def find_version(*file_paths):
         'Natural Language :: English',
         'License :: OSI Approved :: Apache Software License',
         'Programming Language :: Python',
+        "Programming Language :: Python :: 3 :: Only",
```
dlm6693: Just curious why add this now?

jonathan343, line 55: 1. nit - We should use single quotes to keep it consistent with the other classifiers.
2. Should we remove the existing `'Programming Language :: Python :: 3'`? What is the reason for keeping it in addition to the new `"Programming Language :: Python :: 3 :: Only"`?

scripts/ci/install-dev-deps
```diff
@@ -25,4 +26,7 @@ def run(command):
 
 if __name__ == "__main__":
     with cd(REPO_ROOT):
-        run("pip install -r requirements-dev-lock.txt")
+        if sys.version_info[:2] == (3, 12):
+            run("pip install setuptools")
+
+    run("pip install -r requirements-dev-lock.txt")
```
jonathan343: Why does this line need to be run outside the context manager now?

dlm6693: I see that because we specify the python version for `setuptools` in `requirements-dev.txt`, the dependency isn't added to the lock file. So I understand the need to be calling it separately here. However, should we be maintaining this behavior in two different places? Is there any way we can get around this? If not at the very least I think we should be pinning the version to `67.8.0` here as well.

Also any particular reason we're moving the older pip install command outside of the context manager?

setup.py
nateprewitt: Yep, fair enough, good catch. This got copy pasted and we don't normalize quotes with black.

scripts/ci/install-dev-deps
nateprewitt: Should already fixed in [3b493d0](https://github.com/boto/botocore/pull/2967/commits/3b493d03cb2e0a78fc008790c22dc02fe1841771).

setup.py
nateprewitt: [7518e68](https://github.com/boto/botocore/pull/2967/commits/7518e68f073c86c3dc3d5ee48bbae4e3f4ee6d17)

dlm6693 (commented):
Had a couple of comments/questions, but overall I think this looks fine. One thing I do find pretty concerning is the amount of time these tests take to run. It seems as though the 3.12 tests take 3-4x longer than 3.7. Not sure if there's anything we can do about that, but did raise my eyebrows a bit. Also it appears as though there are hundreds of new warnings being raised in the test suite. Also not something we need to do anything about, but might want to clean up down the road.

requirements-dev.txt
nateprewitt: `setuptools` is no longer provided with Python starting in Python 3.12. We need to explicitly add it as a dependency.

setup.py
nateprewitt: It was missed previously.

scripts/ci/install-dev-deps
nateprewitt: The version doesn't matter for CI and we generate the lock file for 3.7 currently, soon 3.8, which ignores the other rule. We specifically don't want to maintain the version in two places which is why it's not included here. The context manager issue was already fixed in [3b493d0](https://github.com/boto/botocore/pull/2967/commits/3b493d03cb2e0a78fc008790c22dc02fe1841771).

setup.py
dlm6693: This isn't included in any of the other repos we own. If we add it here, we should add it to the others as well.

dlm6693 (approved):

PR #2973, opened 2023-06-20
Configured endpoint URL resolution

kdaily:
This PR implements the proposal for configuring endpoints per service:

https://github.com/aws/aws-sdk/pull/230

Adds a config provider to resolve the endpoint URL provided in an environment variable or shared configuration file. It resolves the endpoint in the following manner:

1. The value provided through the `endpoint_url` parameter provided to the client constructor.
2. The value provided by a service-specific environment variable.
3. The value provided by the global endpoint environment variable (`AWS_ENDPOINT_URL`).
4. The value provided by a service-specific parameter from a services definition section in the shared configuration file.
5. The value provided by the global parameter from a services definition section in the shared configuration file.
6. The value resolved when no configured endpoint URL is provided.

The endpoint config provider uses the client name (name used to instantiate a client object) for construction and add to the config value store. This uses multiple client/service name lookups to handle service name changes for backwards compatibility.

### DIFF
--- a/.changes/next-release/bugfix-configprovider-46546.json
+++ b/.changes/next-release/bugfix-configprovider-46546.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "configprovider",
+  "description": "Fix bug when deep copying config value store where overrides were not preserved"
+}
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -287,9 +287,11 @@ class ClientArgsCreator:
         }
 
     def _compute_configured_endpoint_url(self, client_config, endpoint_url):
-        if endpoint_url is not None or self._ignore_configured_endpoint_urls(
-            client_config
-        ):
+        if endpoint_url is not None:
+            return endpoint_url
+
+        if self._ignore_configured_endpoint_urls(client_config):
+            logger.debug("Ignoring configured endpoint URLs.")
             return endpoint_url
 
         return self._config_store.get_config_variable('endpoint_url')
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -115,7 +115,7 @@ class ClientArgsCreator:
         s3_config = final_args['s3_config']
         partition = endpoint_config['metadata'].get('partition', None)
         socket_options = final_args['socket_options']
-
+        configured_endpoint_url = final_args['configured_endpoint_url']
         signing_region = endpoint_config['signing_region']
         endpoint_region_name = endpoint_config['region_name']
 
@@ -160,7 +160,7 @@ class ClientArgsCreator:
             service_model,
             endpoint_region_name,
             region_name,
-            endpoint_url,
+            configured_endpoint_url,
             endpoint,
             is_secure,
             endpoint_bridge,
@@ -210,10 +210,16 @@ class ClientArgsCreator:
                 parameter_validation = ensure_boolean(raw_value)
 
         s3_config = self.compute_s3_config(client_config)
+
+        configured_endpoint_url = self._compute_configured_endpoint_url(
+            client_config=client_config,
+            endpoint_url=endpoint_url,
+        )
+
         endpoint_config = self._compute_endpoint_config(
             service_name=service_name,
             region_name=region_name,
-            endpoint_url=endpoint_url,
+            endpoint_url=configured_endpoint_url,
             is_secure=is_secure,
             endpoint_bridge=endpoint_bridge,
             s3_config=s3_config,
@@ -270,6 +276,7 @@ class ClientArgsCreator:
         return {
             'service_name': service_name,
             'parameter_validation': parameter_validation,
+            'configured_endpoint_url': configured_endpoint_url,
             'endpoint_config': endpoint_config,
             'protocol': protocol,
             'config_kwargs': config_kwargs,
@@ -278,6 +285,25 @@ class ClientArgsCreator:
                 scoped_config, client_config
             ),
         }
+
+    def _compute_configured_endpoint_url(self, client_config, endpoint_url):
+        if endpoint_url is not None or self._ignore_configured_endpoint_urls(
+            client_config
+        ):
+            return endpoint_url
+
+        return self._config_store.get_config_variable('endpoint_url')
+
+    def _ignore_configured_endpoint_urls(self, client_config):
+        if (
+            client_config
+            and client_config.ignore_configured_endpoint_urls is not None
+        ):
+            return client_config.ignore_configured_endpoint_urls
+
+        return self._config_store.get_config_variable(
+            'ignore_configured_endpoint_urls'
+        )
 
     def compute_s3_config(self, client_config):
         s3_configuration = self._config_store.get_config_variable('s3')
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -862,7 +862,7 @@ class ConstantProvider(BaseProvider):
 
 
 class ConfiguredEndpointProvider(BaseProvider):
-    """Lookup an endpoint URL from environment variable or shared cong file.
+    """Lookup an endpoint URL from environment variable or shared config file.
 
     NOTE: This class is considered private and is subject to abrupt breaking
     changes or removal without prior announcement. Please do not use it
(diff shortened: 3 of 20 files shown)


botocore/args.py
```diff
@@ -259,6 +266,27 @@ def compute_client_args(
...
+
+        if endpoint_url is not None:
+            return endpoint_url
+
+        return self._config_store.get_config_variable('endpoint_url')
+
+    def _ignore_configured_endpoint_urls(self, client_config):
+        if (
+            client_config
+            and client_config.ignore_configured_endpoint_urls is not None
```
kyleknap, line 282: We should make sure there are functional tests that ensure we obey the `ignore_configured_endpoint_urls` from the `client_config`. I forgot that this was an available interface when going over the the functional test PR

kdaily: https://github.com/boto/botocore/pull/2974

botocore/configprovider.py
```diff
@@ -836,3 +858,159 @@ def provide(self):
 
     def __repr__(self):
         return 'ConstantProvider(value=%s)' % self._value
+
+
+class ConfiguredEndpointProvider(BaseProvider):
+    """Lookup an endpoint URL from environment variable or shared cong file.
```
nateprewitt, line 864:
```suggestion
    """Lookup an endpoint URL from environment variable or shared config file.
```

botocore/configprovider.py
```diff
@@ -836,3 +858,159 @@ def provide(self):
...
+                    self._client_name,
+                    location,
+                )
+                return endpoint_url
+
+        logger.debug('No configured endpoint found.')
+        return None
+
+    def _get_snake_case_service_id(self, client_name):
+        # Use lookups to get the serice ID without loading the service data
```
nateprewitt, line 945:
```suggestion
        # Use lookups to get the service ID without loading the service data
```

nateprewitt, line 945: nit; we may be able to convey similar information in a more succinct comment. Describing that we're resolving any aliases and standardizing names with hyphens may be enough without the other context.

botocore/configprovider.py
```diff
@@ -836,3 +858,159 @@ def provide(self):
...
+            utils.CLIENT_NAME_TO_HYPHENIZED_SERVICE_ID_OVERRIDES.get(
+                client_name, client_name
+            )
+        )
+        return hyphenized_service_id.replace('-', '_')
+
+    def _get_service_env_var_name(self):
+        transformed_service_id_env = self._get_snake_case_service_id(
+            self._client_name
+        ).upper()
```
nateprewitt, line 964: Should we just resolve this once at initialization?

botocore/configprovider.py
```diff
@@ -836,3 +858,159 @@ def provide(self):
...
+            name='AWS_ENDPOINT_URL', env=self._environ
+        ).provide()
+
+    def __repr__(self):
+        return (
+            f'ConfiguredEndpointProvider('
+            f'full_config={self._full_config}, '
+            f'scoped_config={self._scoped_config}, '
+            f'client_name={self._client_name}, '
+            f'environ={self._environ})'
```
nateprewitt, line 1015: This seems like it could be _really_ big. Do we want need all config and environ variables in the repr?

nateprewitt (changes requested):

botocore/configprovider.py
kdaily: If we did, would we also resolve the services section name once too?

https://github.com/boto/botocore/pull/2973/files/f2d4384ff9213d732bcb88dc4c8f3bf292e7e76f#diff-1818b8b628e5570486a0305e24ca6036a980fae9b2b73125b3298c7d02163313R987

nateprewitt: Yep, that was my thought. The casing can be done inside the function. The actual value lookup should be static for a given _client_name though, right?

.changes/next-release/bugfix-configprovider-46546.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "configprovider",
+  "description": "Fix bug when deep copying config value store where overrides were not preserved"
```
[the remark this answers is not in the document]
kyleknap: For this bug fix one, yes it technically is. Specifically the `config_store` is a public interface via the `Session.get_component()` API. So a user could deepcopy the original `config_store` on the session, update it, and reattach it to the session. The bug is that the deep copy would not have the same default overrides that were provided to the original config store set here: https://github.com/boto/botocore/blob/a587e9261d11473e10491a1b09a55a0148951334/botocore/session.py#L255




For the `enhancement` one, we don't believe it will impact end users. When creating a client, the `config_store` is only used at instantiation of the client so providing a copy of it instead to the client should not impact an end to end user (i.e. it is not possible to update the `config_store` on the session and expected the change to permeate to a preexisting client).

The purpose of adding a changelog entry for it was more for awareness/acknowledge that we did make this subtle change if you are more knowledgeable on the architecture of botocore. It was labeled as an enhancement because there are not really any other categories it fits into. The argument for calling it an enhancement was that the client's config store is now more inline with the client's event emitters where they are snapshot copies of the original config_store/event_emitter from the session. I can go either way on a changelog for this or not.

botocore/configprovider.py
kyleknap: I think was done to match parity with the other config providers (e.g. it is done here for the `EnvironmentProvider`: https://github.com/boto/botocore/blob/f2d4384ff9213d732bcb88dc4c8f3bf292e7e76f/botocore/configprovider.py#L781-L782). Otherwise, I don't think there was any other reason.

Thinking about it more @kdaily do you know if the repr gets logged anywhere in the debug logs? It could be problematic if say the env var or the config had credentials in them as we would not want those logged.

botocore/configprovider.py
kdaily: Yes.

botocore/configprovider.py
kdaily: The only debug logs added for this feature are to output one of the [names of the providers](https://github.com/boto/botocore/pull/2973/files#diff-1818b8b628e5570486a0305e24ca6036a980fae9b2b73125b3298c7d02163313R871-R876) and the client name.

- looking for an endpoint from a provider [here](https://github.com/boto/botocore/pull/2973/files#diff-1818b8b628e5570486a0305e24ca6036a980fae9b2b73125b3298c7d02163313R925)
- found an endpoint from a provider [here](https://github.com/boto/botocore/pull/2973/files#diff-1818b8b628e5570486a0305e24ca6036a980fae9b2b73125b3298c7d02163313R934)

kdaily: It does feel that it would be risky to print the environment or config representation by default for that reason.

kdaily: Removed the repr.

botocore/args.py
kdaily: Merged #2974.

botocore/configprovider.py
kdaily: Moved this to the init.

botocore/configprovider.py
kdaily: Updated the comment

kdaily: Fixed!

botocore/configprovider.py
kdaily: Fixed!

kyleknap (approved):
Looks good :ship:

nateprewitt (approved):
:shipit:

larroy:
Will this work for setting the S3 endpoint url?

A-Levin:
when will you merge it?

tests/functional/configured_endpoint_urls/test_configured_endpoint_url.py
```diff
@@ -0,0 +1,249 @@
...
+    # Call an operation on the client
+    getattr(client, operation)(**params)
+
+    assert (
+        _normalize_endpoint(http_stubber.requests[0].url)
+        == expected_endpoint_url
+    )
+
+
+def _known_service_names_and_models():
```
kyleknap, line 161: Let's rename this to `_known_service_names_and_ids()` to be consistent with the return value

kyleknap (commented):
Looks good just had a small comment on the test update

kyleknap (approved):
Looks good! :ship:

PR #2974, opened 2023-06-20
Add client config ignore endpoint tests

kdaily:
Add tests to check that setting the client creation parameter to ignore the configured endpoint URLs is respected.

### DIFF
--- a/tests/functional/configured_endpoint_urls/profile-tests.json
+++ b/tests/functional/configured_endpoint_urls/profile-tests.json
@@ -73,6 +73,13 @@
         "default": {},
         "endpoint_url_provided":{
           "endpoint_url": "https://client-config.endpoint.aws"
+        },
+        "ignore_configured_endpoint_urls": {
+          "ignore_configured_endpoint_urls": true
+        },
+        "provide_and_ignore_configured_endpoint_urls": {
+          "ignore_configured_endpoint_urls": true,
+          "endpoint_url": "https://client-config.endpoint.aws"
         }
       },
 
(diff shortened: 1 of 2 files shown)


tests/functional/configured_endpoint_urls/profile-tests.json
```diff
@@ -73,6 +73,13 @@
         "default": {},
         "endpoint_url_provided":{
           "endpoint_url": "https://client-config.endpoint.aws"
+        },
+        "ignore_configured_endpoint_urls": {
+          "ignore_configured_endpoint_urls": "true"
```
kyleknap, line 78: Should these booleans instead of the string "true"?

tests/functional/configured_endpoint_urls/test_configured_endpoint_url.py
```diff
@@ -63,6 +66,22 @@ def create_cases():
         )
 
 
+def get_create_client_args(client_config):
+    create_client_args = {}
+
+    if client_config.get('endpoint_url', None):
```
kyleknap, line 72: For these checks, let's just do an `in` comparator (e.g. `if 'endpoint_url' in client_config:`). Mainly if the value in the test case is falsey then it would skip the check which could be possible with the `ignore_configured_endpoint_urls` configuration value.

kyleknap (commented):
Looks fine. I just had a few comments

tests/functional/configured_endpoint_urls/profile-tests.json
kdaily: I used string "true" previously, but I suppose that's because it was for contents of the config file or the environment variable value. Boolean `true` should work here.

Example:

https://github.com/boto/botocore/blob/132f004f6d84ccee1aef91990a0a03a99dfb1927/tests/functional/configured_endpoint_urls/profile-tests.json#L46

kyleknap: I think more of my point is that the expected `Config` interface is to use a boolean and not the string (e.g. `Config(ignore_configured_endpoint_urls=True`) and so it makes more sense to have the value as the expected type here to avoid having to do a type conversion of the string `"true"`. I also suspect other projects using this file would want to avoid doing the conversion as well.

tests/functional/configured_endpoint_urls/test_configured_endpoint_url.py
```diff
@@ -63,6 +66,22 @@ def create_cases():
         )
 
 
+def get_test_case_client_config(client_config):
```
kyleknap, line 69: Oh I meant changing the parameter name `client_config` to `test_case_client_config` and not the test name.

kyleknap (approved):
:ship:

PR #2977, opened 2023-06-28
User Agent 2.0, second attempt

jonemo:
#2955 was rolled back in #2976 because some libraries, including our very own s3transfer, expect `client.meta.config.user_agent` to contain the resolved user agent string (not the user agent string provided by the user in the original `Config` object). To minimize disruption to third party libraries, this PR updates the code formerly merged in #2955 to maintain this behavior. 

Only the last two commits on this branch are new. All others have previously been part of #2955. You can use [this link](https://github.com/jonemo/botocore/compare/d33a106beb2c17326432dbff821c8db368fb9efc...user-agent-2) to only see the newly added changes.


nateprewitt (approved):
⛵

Issue #2926, opened 2023-04-29
Allow urllib3 2.0.0 to be installed


epicserve:
### Describe the feature

There is a new version of urllib3 (v2.0.0) that can't be installed in projects that use botocore because, in the setup.py file, it has the version restriction `urllib3>=1.25.4,<1.27`.

### Use Case

So other projects that want to upgrade to urllib3 2.0.0 can without being blocked by the botocore library.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

latest

### Environment details (OS name and version, etc.)

all

nateprewitt:
Hi @epicserve, thanks for reaching out. The pin is currently intentional while we wait for the major version bump to stabilize. We intend to move the pin shortly after Requests does to ensure we're not negatively impacting users.

We'll keep this issue updated and feel free to follow along in psf/requests#6432 as well if you like.

jplock:
I started getting this error message in my Lambda logs after doing a recent deploy which I believe is pulling in urllib3 >=2.0.0

```
Runtime.ImportModuleError: Unable to import module 'app.lambda_handler': cannot import name 'DEFAULT_CIPHERS' from 'urllib3.util.ssl_' (/opt/python/urllib3/util/ssl_.py)
```

which is coming from https://github.com/boto/botocore/blob/master/botocore/httpsession.py#L52

```python
try:
    from urllib3.util.ssl_ import DEFAULT_CIPHERS
except ImportError:
    # Defer to system configuration starting with
    # urllib3 2.0. This will choose the ciphers provided by
    # Openssl 1.1.1+ or secure system defaults.
    DEFAULT_CIPHERS = None
```

which is concerning because urllib3 removed `DEFAULT_CIPHERS` in https://github.com/urllib3/urllib3/pull/2705 which is part of v2.0.0

What I'm not understanding is why the `ImportError` isn't being caught?

nateprewitt:
Hi @jplock, this was already resolved in https://github.com/boto/botocore/pull/2924. You'll likely want to ensure you're not upgrading to urllib3 2.0 for the time being or add a more recent version of Botocore to your Lambda as detailed [here](https://docs.aws.amazon.com/lambda/latest/dg/python-package.html).

pistolilla:
Same as @jplock here. I started seeing this error today on a lambda after I made deployment. Tried rolling back but error remained.

nateprewitt:
[...]

PR #2984, opened 2023-07-18
Handle 100 Continue in urllib3>=2.

tjni:
This maintains compatibility from urllib3 1.25.4 to urllib3 2.0.3 at least by moving the code in _send_request to request. In urllib3>=2, the _send_request function of http.client.HTTPConnection is skipped in favor of a custom implementation.

Related to https://github.com/boto/botocore/issues/2926.

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -84,16 +84,14 @@ class AWSConnection:
         self._expect_header_set = False
         self.response_class = self._original_response_cls
 
-    def _send_request(self, method, url, body, headers, *args, **kwargs):
+    def request(self, method, url, body=None, headers={}, *args, **kwargs):
         self._response_received = False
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @tjni, thanks for this PR! We're working on the fix internally at the moment but this appears to be on the right track. We will be moving to the `request` wrapper interface provided in CPython's HTTPConnection class but the way urllib3 wrote this creates some performance issues when the body is dropped from a non-100 response.

I'll have a patch to put up once we've resolved those issues. I can either attempt to rebase my changes onto this branch or will post a new PR. Please let me know if you have a preference.

tjni:
No preference, except that you do whatever is easiest for you. Thank you for taking a look.

nateprewitt:
Resolving in favor of #2990. Thanks again, @tjni!

Issue #1246, opened 2017-07-18
Clarify multithreading documentation


dfee:
The [documentation for `boto3`](http://boto3.readthedocs.io/en/latest/guide/resources.html?highlight=multithreading#multithreading-multiprocessing) states that:
> It is recommended to create a **resource instance for each thread** / process in a multithreaded or multiprocess application rather than sharing a single instance among the threads / processes.(emphasis mine)

The documentation than goes on to show a code example where a `session` is created per thread, not a `resource`.

Reading through previous github issues, I see a [note](https://github.com/boto/botocore/issues/895#issuecomment-216972651) that we should create a separate session per thread. The comment that immediately follows says "resource", however.

So, do we need one session per thread, or are sessions thread safe, but not resources? Is there a 1:1 mapping between the thread safety of a `resource` and a `client`?

As a bonus question, how expensive / wasteful is it to create new clients (or sessions, as above) on-demand per executor thread...?

JordonPhillips:
Thanks for the feedback! I agree that it's a little ambiguous, we'll make sure to get that updated.

dfee:
In the meantime (before you update the documentation), can you provide some insight to those questions? Thanks :)

joethompsonduo:
An update is still needed.

shadycuz:
This ^

joethompsonduo:
Once again, echoing the sentiment that some sort of response is needed. Can you acknowledge that the maintainers are at least receiving these communications?

jweinst1:
Question, for firing large amounts of s3 restore_object requests, across multiple threads, what is the safest approach to take?

tuukkamustonen:
Also wondering how this is. I've been carelessly sharing session and client instances over threads for years without (perceived) problems. Not saying it's the way to go, but I wonder what are the cases where it might break?

rahulr3:
Can someone comment ? I'm working on a multithreaded application and each thread is creating it's own session/ resource. Are sessions/resource thread safe ? How expensive it is to create session, resource, clients at scale ?

zoresvit:
@JordonPhillips 

Guys, it's really not cool to promise the update, but then *completely ignore* the issue, the update and all the questions for more than a year.

lhufnagel:
[...]

PR #2985, opened 2023-07-18
Avoid a KeyError when a ComponentLocator is being called concurrently

akx:
The symptom is e.g. `KeyError: 'credential_provider'`, detailed in e.g. https://github.com/boto/botocore/issues/1246#issuecomment-632444392

As detailed in the comment, even if the component factory gets called twice, cleaning the deferred up should not cause a KeyError.

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -1141,7 +1141,15 @@ class ComponentLocator:
             # Only delete the component from the deferred dict after
             # successfully creating the object from the factory as well as
             # injecting the instantiated value into the _components dict.
-            del self._deferred[name]
+            try:
+                del self._deferred[name]
+            except KeyError:
+                # If we get here, it's likely that get_component was called
+                # concurrently from multiple threads, and another thread
+                # already deleted the entry. This means the factory was
+                # probably called twice, but cleaning up the deferred entry
+                # should not crash outright.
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @akx, thanks for the PR. This seems like a reasonable thing to fix, but we should caveat that this case should never be hit under documented usage. Sessions are not thread safe and client creation shouldn't be done in parallel using the same session. We have the details on usage of Sessions around parallelization primitives [here](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/session.html#multithreading-or-multiprocessing-with-sessions).

akx:
@nateprewitt I kind of understand that, but it really isn't very obvious that you should not use  `boto3.client()` (that uses the implicit shared default session) from more than one thread. 

Maybe the default session should be thread-local or plain refuse to work between threads if it's not actually supported?

nateprewitt:
I think we cover this to some extent in the Session section and we note that boto3.client uses the default session in the docstring, but I agree it could be easier to find. I'll open an issue for that and merge this to help in the cases this is missed.

nateprewitt (approved):

nateprewitt:
#2991 created for reference.

PR #2989, opened 2023-07-21
Update GHA configurations

nateprewitt:
This PR adds:
* Explicit scoping for token usage in our existing workflows.
* Updates close-stale-issues to only run once a day to save worker resources.
* Adds CodeQL to run static analysis over code base.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,35 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: ["develop"]
+  pull_request:
+    branches: ["develop"]
+  schedule:
+    - cron: "0 0 * * 5"
+
+permissions: "read-all"
+
+jobs:
+  analyze:
+    name: "Analyze"
+    runs-on: "ubuntu-latest"
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+    steps:
+    - name: "Checkout repository"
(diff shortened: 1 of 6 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,35 @@
...
+  analyze:
+    name: "Analyze"
+    runs-on: "ubuntu-latest"
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+    steps:
+    - name: "Checkout repository"
+      uses: "actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3"
```
jonathan343, line 23: Why this specific hash? Same question for the ones below?

nateprewitt: I'm going to add another commit to move everything to pins. I have the end result in the S3transfer PR.

The main goal is to ensure that we're always using a known state when running actions. That means if a repository is compromised and a new commit with malicious code is made under a known tag, we don't run the malicious code. The ability to compromise a repository _and_ rewrite the history to create the same commit hash is almost impossible.

Full information can be found here: https://github.com/ossf/scorecard/blob/e1eede2d3fa09d08bbc00ca2331c2bffebdc602d/docs/checks.md#pinned-dependencies

jonathan343 (approved):
Looks like the new CodeQL job passed👍

PR #2990, opened 2023-07-28
Move 100-continue behavior to use high-level request interface

nateprewitt:
## Overview

Starting in urllib3 2.0, the HTTPConnection behavior was reorganized to incorporate request chunking under the same top-level `request` method. This removed the underlying call to `_send_request` that's utilized by CPython's HTTPConnection which is the parent class of the urllib3 implementation. This change effectively short circuited botocore's provisional HTTP 100-continue support.

To provide a path forward for supporting urllib3 2.0, we're moving to us urllib3's `request` interface. Fortunately, this also exists in CPython as a [thin wrapper around `_send_request`](https://github.com/python/cpython/blob/main/Lib/http/client.py#L1316-L1319). This will reenable _most_ of our internal patches to the class to support 100-continue.

## Changes in urllib3's behavior
The only significant pain point this will introduce is urllib3 has moved from using `endheaders` to send the body in CPython to bundling the send logic in `request` itself. That circumvents our ability to drop sending the body since it's not dispatched to another method that can be overwritten. In order to work around this, we need to stub out all `send` calls to be no-op once a non-100 response is observed. We're able to do this relatively easily because urllib3's current behavior matches what was in the standard library prior to Python 2.7.

The only other issue we needed to work around was a ~50% regression in response time for non-100 responses. This is due to the number of times urllib3 invokes send with its forced chunking. We previously only saw one `send` call but now for large bodies it's potentially thousands of invocations. When logging is enabled, this not only pollutes the logs but takes a non-trivial amount of time. We'll move our logging to only fire once on the first `send` call and let the others pass silently.

## Release plan

This PR will add a new GHA test runner that will always use "urllib3<2" to ensure we're still operational with urllib3 1.26.x. It will also move to the new provisional interface that should be compatible across both urllib3 major versions _without_ moving the urllib3 pin.

This let's us ensure we don't have any unexpected gaps and unblocks distro maintainers from migrating their code/test suites to use urllib3 2.0 if needed. Once we have confidence in the change, we'll look at moving the pin. There isn't a strict timeline on this as there are still OpenSSL issues that may affect some Lambda/Codebuild environments. We'll provide an update to the tracking ticket #2926 once we have more specifics.

### DIFF
--- a/.changes/next-release/enhancement-HTTP-88756.json
+++ b/.changes/next-release/enhancement-HTTP-88756.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "HTTP",
+  "description": "Move 100-continue behavior to use `HTTPConnections` request interface."
+}
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -33,3 +33,24 @@ jobs:
         uses: codecov/codecov-action@v3
         with:
           directory: tests
+
+
+  urllib3:
+    name: 'urllib3 1.x'
+    runs-on: 'ubuntu-latest'
+    strategy:
+      fail-fast: true
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: 'Set up Python 3.10'
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: Install dependencies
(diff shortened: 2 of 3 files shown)


kyleknap (approved):
:ship:

PR #2996, opened 2023-08-09
User-Agent component customization

jonathan343:


botocore/useragent.py
```diff
@@ -116,6 +116,14 @@ def to_string(self):
         return self._value
 
 
+try:
+    from botocore.customizations.useragent import modify_components
```
nateprewitt, line 120: Can we add a disclaimer here to effect of:
```
# This is not a public interface and subject to abrupt breaking changes.
# Any usage is not advised or supported in external code bases.

File: tests/unit/test_useragent.py [hunk 2]
```diff
@@ -65,20 +65,22 @@ def test_basic_user_agent_string():
...
+        'md/arch#x86_64',
+        'lang/python#3.8.20',
+        'md/pyimpl#Dpython',
+        'exec-env/AWS_Lambda_python3.8',
+        'cfg/retry-mode#legacy',
+        'app/fooapp',
+    ]
+
+    indices = [actual.index(el) for el in expected_in_exact_order]
+    assert indices == list(sorted(indices)), 'Elements were found out of order'
```
nateprewitt, line 83: Can we potentially do something like this on the tests rather than rewriting them to try to work around injected ordering?

```python
@mock.patch.object(botocore.useragent, 'modify_components', noop_modify)
```

nateprewitt (changes requested):
Two minor notes, otherwise looks good!

botocore/useragent.py
jonathan343: Added in latest changes.

tests/unit/test_useragent.py
jonathan343: I reverted these changes and instead added `@mock.patch.object` to pass a function that doesn't modify components.

nateprewitt (approved):
Looks good, thanks @jonathan343!

PR #3000, opened 2023-08-11
Add base dependabot config

nateprewitt:
This PR will add a base dependabot config to check our github actions for updates once a week. This will ignore patch updates to minimize noise for the time being.


jonathan343 (approved):
:shipit:

jonathan343:
Not a blocker but we might want to apply a label to these for easier filtering. Something like:
```
    labels:
      - "some-github-actions-label"
```

PR #3020, opened 2023-09-22
Update endpoint-rules to latest format

nateprewitt:
This PR will update our endpoint rules to the latest version, adding in performance optimizations and shrinking the overall size.


jonathan343:
I think we should add an `enhancement` changelog entry for this PR.

dlm6693 (approved):
🚢

PR #3027, opened 2023-09-28
Add `account_id` to `Credentials`

dlm6693:
Adds `account_id` as an optional input to `Credentials`, `RefreshableCredentials` and `ReadOnlyCredentials`. It is also added to `DeferredRefreshableCredentials` as an internal variable.

It is resolved to a non-null value when possible in a number of supported credential providers:
* `AssumeRoleProvider`/`AssumeRoleWithWebIdentityProvider`: parsed from `AssumedRoleUser.Arn` in assume role response.
* `SSOProvider`: sourced from `sso_account_id` profile parameter in AWS config file
* `EnvProvider`: sourced from `AWS_ACCOUNT_ID` environment variable
* `ConfigProvider`: sourced from `aws_account_id` profile parameter in AWS config file
* `SharedCredentialsProvider`: sourced from `aws_account_id` profile parameter in AWS shared credentials file
* `ProcessProvider`: sourced from `AccountId` parameter returned in credential process response or `aws_account_id` profile parameter if the former isn't available.

It can also be configured statically to a botocore client via the `aws_account_id` arg in `Session.create_client` or the `account_id` arg in `Session.set_credentials`

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -44,9 +44,11 @@ from botocore.exceptions import (
 )
 from botocore.tokens import SSOTokenProvider
 from botocore.utils import (
+    ArnParser,
     ContainerMetadataFetcher,
     FileWebIdentityTokenLoader,
     InstanceMetadataFetcher,
+    InvalidArnException,
     JSONFileCache,
     SSOTokenLoader,
     parse_key_val_file,
@@ -55,7 +57,9 @@ from botocore.utils import (
 
 logger = logging.getLogger(__name__)
 ReadOnlyCredentials = namedtuple(
-    'ReadOnlyCredentials', ['access_key', 'secret_key', 'token']
+    'ReadOnlyCredentials',
+    ['access_key', 'secret_key', 'token', 'account_id'],
+    defaults=(None,),
 )
 
 _DEFAULT_MANDATORY_REFRESH_TIMEOUT = 10 * 60  # 10 min
@@ -307,9 +311,17 @@ class Credentials:
     :param str token: The security token, valid only for session credentials.
     :param str method: A string which identifies where the credentials
         were found.
+    :param str account_id: The account ID associated with the credentials.
     """
 
-    def __init__(self, access_key, secret_key, token=None, method=None):
+    def __init__(
+        self,
+        access_key,
+        secret_key,
+        token=None,
+        method=None,
+        account_id=None,
+    ):
         self.access_key = access_key
         self.secret_key = secret_key
         self.token = token
@@ -317,6 +329,7 @@ class Credentials:
         if method is None:
             method = 'explicit'
         self.method = method
+        self.account_id = account_id
 
         self._normalize()
 
@@ -332,7 +345,7 @@ class Credentials:
 
     def get_frozen_credentials(self):
         return ReadOnlyCredentials(
-            self.access_key, self.secret_key, self.token
+            self.access_key, self.secret_key, self.token, self.account_id
         )
 
 
@@ -349,6 +362,7 @@ class RefreshableCredentials(Credentials):
     :param str method: A string which identifies where the credentials
         were found.
     :param function time_fetcher: Callback function to retrieve current time.
+    :param str account_id: The account ID associated with the credentials.
     """
 
     # The time at which we'll attempt to refresh, but not
@@ -367,6 +381,7 @@ class RefreshableCredentials(Credentials):
         refresh_using,
         method,
         time_fetcher=_local_now,
+        account_id=None,
     ):
         self._refresh_using = refresh_using
         self._access_key = access_key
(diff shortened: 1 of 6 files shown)


tests/integration/test_credentials.py
```diff
@@ -72,7 +72,10 @@ def test_access_secret_vs_profile_code(self, credentials_cls):
         )
 
         credentials_cls.assert_called_with(
-            access_key='code', secret_key='code-secret', token=mock.ANY
+            access_key='code',
+            secret_key='code-secret',
+            token=mock.ANY,
+            account_id=mock.ANY,
```
dlm6693, line 78: We can avoid making this change by only setting account ID in `ConfigProvider.load` if the value isn't `None`. However, we haven't conditionally provided values to a credentials object before so I elected not to for the time being.

tests/unit/test_credentials.py
```diff
@@ -720,6 +772,7 @@ def get_expected_creds_from_response(self, response):
             'secret_key': response['Credentials']['SecretAccessKey'],
             'token': response['Credentials']['SessionToken'],
             'expiry_time': expiration,
+            'account_id': None,
```
dlm6693, line 775: Similar to the changed integration tests, this can be avoided if we only conditionally set `account_id` in the credential dictionary in the cached credential fetcher.  However, we haven't conditionally set values in the dict either even if the value is None so I elected not to.

tests/functional/test_credentials.py
```diff
@@ -198,7 +198,7 @@ def create_assume_role_response(self, credentials, expiration=None):
             },
             'AssumedRoleUser': {
                 'AssumedRoleId': 'myroleid',
-                'Arn': 'arn:aws:iam::1234567890:user/myuser',
+                'Arn': f'arn:aws:iam::{credentials.account_id}:user/myuser',
```
dlm6693, line 201: Since the `AssumedRoleUser` object with a valid ARN was already a part of this test helper function, the `account_id` in the resulting credentials object will always resolve to a non-None value. So this needed to be updated. Another option would be to delete `AssumedRoleUser`, which would let us remove the below account ID line.

tests/unit/test_credentials.py
```diff
@@ -3587,6 +3836,7 @@ def test_load_sso_credentials_without_cache(self):
             self.assertEqual(credentials.access_key, 'foo')
             self.assertEqual(credentials.secret_key, 'bar')
             self.assertEqual(credentials.token, 'baz')
+            self.assertEqual(credentials.account_id, '1234567890')
```
dlm6693, line 3839: Since account ID for SSO is attached directly to the provider, these existing tests will always have the value set. Because of this I think it makes more sense to just update the existing tests.

tests/unit/test_session.py
```diff
@@ -549,6 +549,23 @@ def test_cred_provider_called_when_partial_creds_provided(self):
                 aws_secret_access_key='foo',
             )
 
+    @mock.patch('botocore.client.ClientCreator')
+    def test_credential_params(self, client_creator):
```
dlm6693, line 553: I couldn't find a test that explicitly checked if each of the credential values that can be statically provided to `create_client` were properly set on the credentials, so I added one. We would need something like this for account ID anyway.

botocore/credentials.py
```diff
@@ -55,7 +57,9 @@
 
 logger = logging.getLogger(__name__)
 ReadOnlyCredentials = namedtuple(
-    'ReadOnlyCredentials', ['access_key', 'secret_key', 'token']
+    'ReadOnlyCredentials',
+    ['access_key', 'secret_key', 'token', 'account_id'],
+    defaults=(None,),
```
dlm6693, line 62: The syntax is a little weird, but `defaults` are applied to the rightmost argument first, so this will only let `account_id` have a default of `None`.

botocore/credentials.py
```diff
@@ -756,6 +790,18 @@ def _create_cache_key(self):
         argument_hash = sha1(args.encode('utf-8')).hexdigest()
         return self._make_file_safe(argument_hash)
 
+    def _resolve_account_id(self, response):
+        user_arn = response.get('AssumedRoleUser', {}).get('Arn')
+        if user_arn is not None:
+            try:
+                account_id = self._arn_parser.parse_arn(user_arn)['account']
+            except InvalidArnException:
+                logger.debug(
```
dlm6693, line 799: Not sure if the debug log is necessary, but figured it couldn't hurt.

botocore/credentials.py
```diff
@@ -1184,6 +1251,11 @@ def fetch_credentials(require_expiry=True):
                     provider=method, cred_var=mapping['expiry_time']
                 )
 
+            credentials['account_id'] = None
+            account_id = environ.get(mapping['account_id'], '')
+            if account_id:
+                credentials['account_id'] = account_id
+
```
dlm6693, line 1258: This can be written in one line, but I chose to mimic the format used by the other variables above.
`credentials['account_id'] = environ.get(mapping['account_id'])`

dlm6693:
Closed in favor of https://github.com/boto/botocore/pull/3031

Issue #2361, opened 2021-04-22
Typo in BOTOCORE_DEFAUT_SESSION_VARIABLES variable name


adamantoise:
**Describe the bug**
In `botocore/configprovider.py`, the variable name `BOTOCORE_DEFAUT_SESSION_VARIABLES` contains a typo: the word `DEFAUT` should be spelled `DEFAULT`.

stobrien89:
Hi @adamantoise,

Thanks for the heads-up!

nateprewitt:
Hi @adamantoise,

Thanks for reaching out! We're aware of the typo in the variable, but unfortunately, fixing it isn't straight forward. Due to the age of the variable, it's difficult to know what downstream systems may be relying on this particular spelling. Any update would be a breaking change. We could rename the variable and create an alias for the old spelling, but this means users will still get incorrect suggestions in their IDE prompts etc.

For that reason, we haven't really touched it because the options available don't provide a lot of value. That's the conclusion we reached last time this was discussed at least, but please let us know if you have further thoughts.

adamantoise:
No problem if you want to close this as WONTFIX, it's obviously very minor and inconsequential.  It's certainly not worth putting in a lot of effort to fix.  (You could optionally add a comment explaining the unfortunate name if you'd like.)

nateprewitt:
Yeah, I think a brief comment callout is a reasonable suggestion. We'll leave this open to track that as a quick follow up item. Thanks again for reviving the conversation.

adamantoise:
This issue is still present (but, again, of little consequence).  I'd prefer it though if a _human_ closed this rather than a bot, if the decision is not to fix it.

nateprewitt:
This was resolved in #3029, we should be set to close this now.

PR #3029, opened 2023-10-02
Fix typo in configprovider.py

mgenereu:


nateprewitt:
Hi @mgenereu, thanks for the PR. Unfortunately, this is a breaking change. We've discussed this in #2361 previously. We could add a short note about this for future readers, but otherwise won't be changing the variable name.

mgenereu:
@nateprewitt Like that?

nateprewitt (approved):

nateprewitt:
Yep, thanks @mgenereu!

PR #3031, opened 2023-10-04
Account ID endpoint routing

dlm6693:
Adds the ability for AWS services to route customers to account ID-specific endpoints. This PR can be reviewed in two parts.

## Part 1 (commits 1-2)
Adds `account_id` as an optional input to `Credentials`, `RefreshableCredentials` and `ReadOnlyCredentials`. It is also added to `DeferredRefreshableCredentials` as an internal variable.

It is resolved to a non-null value when possible in a number of supported credential providers:
* `AssumeRoleProvider`/`AssumeRoleWithWebIdentityProvider`: parsed from `AssumedRoleUser.Arn` in assume role response.
* `SSOProvider`: sourced from `sso_account_id` profile parameter in AWS config file
* `EnvProvider`: sourced from `AWS_ACCOUNT_ID` environment variable
* `ConfigProvider`: sourced from `aws_account_id` profile parameter in AWS config file
* `SharedCredentialsProvider`: sourced from `aws_account_id` profile parameter in AWS shared credentials file
* `ProcessProvider`: sourced from `AccountId` parameter returned in credential process response or `aws_account_id` profile parameter if the former isn't available.

It can also be configured statically to a botocore client via the `aws_account_id` arg in `Session.create_client` or the `account_id` arg in `Session.set_credentials`

## Part 2 (commits 3-4)
Adds the config setting `account_id_endpoint_mode` and used during endpoint resolution. When enabled and configured in a service's ruleset, the endpoint resolver will attempt to resolve `AccountId` as an input parameter to the endpoint provider.
* Account ID is treated as an endpoint builtin parameter `Aws::Auth::AccountId` defaulting to `None`. If the config setting is enabled, the parameter is configured in the ruleset and the builtin hasn't been set during the `before-endpoint-resolution` event, account ID will be resolved from credentials.
* valid config settings for `account_id_endpoint_mode` are `preferred`, `required` and `disabled` (case sensitive). Defaults to `preferred` if not supplied by a customer.
* Account ID based routing is always disabled for `UNSIGNED` requests and presigned URLs

## Updates since opening the PR
* Account ID based routing is not disabled for presigned URLs
* endpoint builtin resolution is layered through a couple of abstractions. `CredentialBuiltinResolver` is responsible for actually resolving the account ID value (and future credential builtin values). If/when other types of builtin resolvers are required they will all be passed to `EndpointBuiltinResolver`. This construct has one public API `resolve` which is called by the ruleset resolver.

### DIFF
--- a/.changes/next-release/feature-endpoints-57979.json
+++ b/.changes/next-release/feature-endpoints-57979.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "endpoints",
+  "description": "Enables account ID endpoint routing"
+}
--- a/.changes/next-release/feature-endpoints-57979.json
+++ b/.changes/next-release/feature-endpoints-57979.json
@@ -1,5 +1,5 @@
 {
   "type": "feature",
   "category": "endpoints",
-  "description": "Enables account ID endpoint routing"
+  "description": "Adds support for account ID endpoint routing"
 }
--- a/botocore/config.py
+++ b/botocore/config.py
@@ -230,8 +230,18 @@ class Config:
     :type account_id_endpoint_mode: str
     :param account_id_endpoint_mode: Enables or disables account ID based
         endpoint routing for supported operations.
+        Valid options are:
 
-        Defaults to None.
+        * ``preferred`` -- Attempt to resolve account ID during endpoint
+            resolution if supported by the service. If account ID cannot be
+            resolved, fallback to a different endpoint.
+        * ``required`` -- Require account ID to be resolved during endpoint
+            resolution. If account ID cannot be resolved, raises
+            ``AccountIDNotFound`` exception.
+        * ``disabled`` -- Disable account ID based endpoint routing. The SDK
+            will not attempt to resolve account ID during endpoint resolution.
+
+        If not specified, the default behavior is ``preferred``.
     """
 
     OPTION_DEFAULTS = OrderedDict(
--- a/botocore/config.py
+++ b/botocore/config.py
@@ -228,18 +228,14 @@ class Config:
         documentation. Invalid parameters or ones that are not used by the
         specified service will be ignored.
     :type account_id_endpoint_mode: str
-    :param account_id_endpoint_mode: Enables or disables account ID based
-        endpoint routing for supported operations.
+    :param account_id_endpoint_mode: Set account ID based endpoint behavior
+        for supported operations.
         Valid options are:
 
-        * ``preferred`` -- Attempt to resolve account ID during endpoint
-            resolution if supported by the service. If account ID cannot be
-            resolved, fallback to a different endpoint.
-        * ``required`` -- Require account ID to be resolved during endpoint
-            resolution. If account ID cannot be resolved, raises
-            ``AccountIDNotFound`` exception.
-        * ``disabled`` -- Disable account ID based endpoint routing. The SDK
-            will not attempt to resolve account ID during endpoint resolution.
+        * ``preferred`` -- Use account ID based endpoint routing if available.
+        * ``required`` -- Raise ``AccountIDNotFound`` exception if account ID
+            based endpoint routing is unavailable.
+        * ``disabled`` -- Disable account ID based endpoint routing.
 
         If not specified, the default behavior is ``preferred``.
     """
--- a/botocore/config.py
+++ b/botocore/config.py
@@ -227,6 +227,9 @@ class Config:
         the ``Client Context Parameters`` section of the service client's
         documentation. Invalid parameters or ones that are not used by the
         specified service will be ignored.
+    :type account_id_endpoint_mode: str
+    :param account_id_endpoint_mode: Enables or disables account ID based
+        endpoint routing for supported operations.
 
         Defaults to None.
     """
@@ -257,6 +260,7 @@ class Config:
             ('request_min_compression_size_bytes', None),
             ('disable_request_compression', None),
             ('client_context_params', None),
+            ('account_id_endpoint_mode', None),
         ]
     )
 
--- a/botocore/config.py
+++ b/botocore/config.py
@@ -233,7 +233,7 @@ class Config:
         Valid options are:
 
         * ``preferred`` -- Use account ID based endpoint routing if available.
-        * ``required`` -- Raise ``AccountIDNotFound`` exception if account ID
+        * ``required`` -- Raise ``AccountIdNotFound`` exception if account ID
             based endpoint routing is unavailable.
         * ``disabled`` -- Disable account ID based endpoint routing.
 
--- a/botocore/config.py
+++ b/botocore/config.py
@@ -227,6 +227,7 @@ class Config:
         the ``Client Context Parameters`` section of the service client's
         documentation. Invalid parameters or ones that are not used by the
         specified service will be ignored.
+
     :type account_id_endpoint_mode: str
     :param account_id_endpoint_mode: Set account ID based endpoint behavior
         for supported operations.
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1084,9 +1084,8 @@ class ProcessProvider(CredentialProvider):
         return creds_dict
 
     def _resolve_account_id(self, parsed_response):
-        return parsed_response.get('AccountId') or self.profile_config.get(
-            'aws_account_id'
-        )
+        account_id = parsed_response.get('AccountId')
+        return account_id or self.profile_config.get('aws_account_id')
 
     @property
     def _credential_process(self):
@@ -1096,9 +1095,8 @@ class ProcessProvider(CredentialProvider):
     def profile_config(self):
         if self._loaded_config is None:
             self._loaded_config = self._load_config()
-        return self._loaded_config.get('profiles', {}).get(
-            self._profile_name, {}
-        )
+        profiles = self._loaded_config.get('profiles', {})
+        return profiles.get(self._profile_name, {})
 
 
 class InstanceMetadataProvider(CredentialProvider):
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1251,10 +1251,7 @@ class EnvProvider(CredentialProvider):
                     provider=method, cred_var=mapping['expiry_time']
                 )
 
-            credentials['account_id'] = None
-            account_id = environ.get(mapping['account_id'], '')
-            if account_id:
-                credentials['account_id'] = account_id
+            credentials['account_id'] = environ.get(mapping['account_id'])
 
             return credentials
 
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -708,13 +708,16 @@ class CachedCredentialFetcher:
 
         creds = response['Credentials']
         expiration = _serialize_if_needed(creds['Expiration'], iso=True)
-        return {
+        creds_dict = {
             'access_key': creds['AccessKeyId'],
             'secret_key': creds['SecretAccessKey'],
             'token': creds['SessionToken'],
             'expiry_time': expiration,
-            'account_id': creds.get('AccountId'),
         }
+        account_id = creds.get('AccountId')
+        if account_id is not None:
+            creds_dict['account_id'] = account_id
+        return creds_dict
 
     def _load_from_cache(self):
         if self._cache_key in self._cache:
@@ -1037,7 +1040,7 @@ class ProcessProvider(CredentialProvider):
             secret_key=creds_dict['secret_key'],
             token=creds_dict.get('token'),
             method=self.METHOD,
-            account_id=creds_dict['account_id'],
+            account_id=creds_dict.get('account_id'),
         )
 
     def _retrieve_credentials_using(self, credential_process):
@@ -1063,18 +1066,22 @@ class ProcessProvider(CredentialProvider):
                 ),
             )
         try:
-            return {
+            creds_dict = {
                 'access_key': parsed['AccessKeyId'],
                 'secret_key': parsed['SecretAccessKey'],
                 'token': parsed.get('SessionToken'),
                 'expiry_time': parsed.get('Expiration'),
-                'account_id': self._resolve_account_id(parsed),
             }
         except KeyError as e:
             raise CredentialRetrievalError(
                 provider=self.METHOD,
                 error_msg=f"Missing required key in response: {e}",
             )
+
+        account_id = self._resolve_account_id(parsed)
+        if account_id is not None:
+            creds_dict['account_id'] = account_id
+        return creds_dict
 
     def _resolve_account_id(self, parsed_response):
         return parsed_response.get('AccountId') or self.profile_config.get(
@@ -1200,7 +1207,7 @@ class EnvProvider(CredentialProvider):
                     expiry_time,
                     refresh_using=fetcher,
                     method=self.METHOD,
-                    account_id=credentials['account_id'],
+                    account_id=credentials.get('account_id'),
                 )
 
             return Credentials(
@@ -1208,7 +1215,7 @@ class EnvProvider(CredentialProvider):
                 credentials['secret_key'],
                 credentials['token'],
                 method=self.METHOD,
-                account_id=credentials['account_id'],
+                account_id=credentials.get('account_id'),
             )
         else:
             return None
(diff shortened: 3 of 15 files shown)


tests/unit/test_credentials.py
```diff
@@ -257,6 +265,7 @@ def get_expected_creds_from_response(self, response):
             'secret_key': response['Credentials']['SecretAccessKey'],
             'token': response['Credentials']['SessionToken'],
             'expiry_time': expiration,
+            'account_id': None,
```
dlm6693, line 268: This change is required since account ID is not conditionally added to the creds dict. We haven't needed to conditionally set params to it in the past and I chose not to here. If it's preferred it's a simple change.

tests/integration/test_credentials.py
```diff
@@ -72,7 +72,10 @@ def test_access_secret_vs_profile_code(self, credentials_cls):
         )
 
         credentials_cls.assert_called_with(
-            access_key='code', secret_key='code-secret', token=mock.ANY
+            access_key='code',
+            secret_key='code-secret',
+            token=mock.ANY,
+            account_id=mock.ANY,
```
dlm6693, line 78: Similar to the assume role unit tests, this was needed since we don't conditionally send the token either even if it is none.

tests/functional/test_credentials.py
```diff
@@ -198,7 +198,7 @@ def create_assume_role_response(self, credentials, expiration=None):
             },
             'AssumedRoleUser': {
                 'AssumedRoleId': 'myroleid',
-                'Arn': 'arn:aws:iam::1234567890:user/myuser',
+                'Arn': f'arn:aws:iam::{credentials.account_id}:user/myuser',
```
dlm6693, line 201: Since the `AssumedRoleUser` object with a valid arn was already in these tests, account ID will never resolve to a non-None value.

tests/functional/test_credentials.py
```diff
@@ -789,7 +790,8 @@ def assert_session_credentials(self, expected_params, **kwargs):
         expected_creds = self.create_random_credentials()
         response = self.create_assume_role_response(expected_creds)
         session = StubbedSession(**kwargs)
-        stubber = session.stub('sts')
+        config = Config(signature_version=UNSIGNED)
+        stubber = session.stub('sts', config=config)
```
dlm6693, line 794: Without this change `test_assume_role` would always get stuck. I believe it's because the stubber doesn't take full effect until the `before-call` event is emitted, which isn't until after endpoint resolution. So the `assume_role_with_web_identity` call would now only partially be mocked.

Adding this config setting resolved the issue and I believe is an acceptable change because the provider uses `UNSIGNED` in actuality https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L904

botocore/config.py
```diff
@@ -219,6 +219,12 @@ class Config:
     :param disable_request_compression: Disables request body compression if
         set to True.
 
+        Defaults to None.
+
+    :type account_id_endpoint_mode: str
+    :param account_id_endpoint_mode: Enables or disables account ID based
+        endpoint routing for supported operations.
+
```
jonathan343, line 227: We should be providing more information about the possible config options. For example:
```
...
Valid options are:

* ``preferred`` -- ...
* ``disabled`` -- ...
* ``required`` -- ...
```

botocore/config.py
```diff
@@ -219,6 +219,12 @@ class Config:
...
         Defaults to None.
```
jonathan343, line 228: We should explain the default behavior in more detail.

jonathan343 (commented):
Dropping a few comments after a first pass. I can take a deeper look sometime tomorrow.

botocore/exceptions.py
```diff
@@ -814,3 +814,7 @@ class EndpointResolutionError(EndpointProviderError):
 
 class UnknownEndpointResolutionBuiltInName(EndpointProviderError):
     fmt = 'Unknown builtin variable name: {name}'
+
+
+class AccountIDNotFound(EndpointResolutionError):
+    fmt = '`account_id_endpoint_mode is set to `required` but no account ID was found.'
```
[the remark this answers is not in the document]
dlm6693: This was just a miss on my part. I'll update to use double quotes instead.

botocore/regions.py
```diff
@@ -562,6 +575,71 @@ def _get_provider_params(
...
+    def _do_resolve_account_id_builtin(
+        self, account_id_endpoint_mode, builtins
+    ):
+        # This will make a call to resolve credentials if they are not already
+        # or need to be refreshed.
+        frozen_creds = self._credentials.get_frozen_credentials()
+        account_id = frozen_creds.account_id
+        if account_id is None:
+            if account_id_endpoint_mode == 'preferred':
+                LOG.debug(
```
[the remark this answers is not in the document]
dlm6693: There was a concern about blowing up logs. I'm not opposed to changing this though

botocore/config.py
dlm6693: Yeah I'll add more info for this as well as the available options

botocore/config.py
```diff
@@ -220,6 +220,22 @@ class Config:
...
+        * ``preferred`` -- Attempt to resolve account ID during endpoint
+            resolution if supported by the service. If account ID cannot be
+            resolved, fallback to a different endpoint.
+        * ``required`` -- Require account ID to be resolved during endpoint
+            resolution. If account ID cannot be resolved, raises
+            ``AccountIDNotFound`` exception.
+        * ``disabled`` -- Disable account ID based endpoint routing. The SDK
+            will not attempt to resolve account ID during endpoint resolution.
+
+        If not specified, the default behavior is ``preferred``.
```
nateprewitt, line 238:
```suggestion
    :param account_id_endpoint_mode: Set account ID based
        endpoint behavior for supported operations.
        Valid options are:

        * ``preferred`` -- Use Account ID based routing if available.
        * ``required`` -- Raise ``AccountIDNotFound`` exception if routing is unavailable.
        * ``disabled`` -- Disable account ID based endpoint routing.

        If not specified, the default behavior is ``preferred``.
```

botocore/regions.py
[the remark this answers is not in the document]
nateprewitt: This will be _very_ noisy for the average user who likely doesn't want/need this feature. The default is preferred which means every call will log a warning.

tests/unit/test_credentials.py
nateprewitt: This seems like a risk. Why do the responses have to include this value now? Can anything subclassing credentials or response structure that may not return this format break?

tests/unit/test_endpoint_provider.py
```diff
@@ -111,6 +115,11 @@ def ruleset_dict():
         return json.load(f)
 
 
+@pytest.fixture(scope="module")
+def ruleset_dict():
+    return _ruleset_dict()
```
nateprewitt, line 120: What prompted this?

nateprewitt (changes requested):

botocore/credentials.py
```diff
@@ -1184,6 +1251,11 @@ def fetch_credentials(require_expiry=True):
                     provider=method, cred_var=mapping['expiry_time']
                 )
 
+            credentials['account_id'] = None
+            account_id = environ.get(mapping['account_id'], '')
+            if account_id:
+                credentials['account_id'] = account_id
+
```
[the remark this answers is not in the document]
dlm6693: Left a comment on the old PR and forgot to move it here. I'll simplify
https://github.com/boto/botocore/pull/3027#discussion_r1340681852

botocore/exceptions.py
```diff
@@ -814,3 +814,7 @@ class EndpointResolutionError(EndpointProviderError):
 
 class UnknownEndpointResolutionBuiltInName(EndpointProviderError):
     fmt = 'Unknown builtin variable name: {name}'
+
+
+class AccountIDNotFound(EndpointResolutionError):
```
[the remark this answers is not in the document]
dlm6693: Nope. Will update

botocore/regions.py
```diff
@@ -562,6 +575,71 @@ def _get_provider_params(
...
+        config setting.
+        """
+        not_presign = not request_context.get('is_presign_request', False)
+        should_sign = self._requested_auth_scheme != UNSIGNED
+        creds_available = self._credentials is not None
+        if all((not_presign, should_sign, creds_available)):
+            config = request_context['client_config']
+            act_id_ep_mode = config.account_id_endpoint_mode
+            return self._validate_account_id_endpoint_mode(act_id_ep_mode)
+        return 'disabled'
```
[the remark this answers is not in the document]
dlm6693: This was a decision that was made with the author of the spec. I'll touch base about this with them today.

botocore/regions.py
```diff
@@ -562,6 +575,71 @@ def _get_provider_params(
...
+            if act_id_ep_mode == 'disabled':
+                # if account ID has been set with a custom handler, but the mode
+                # is disabled, we must unset it so it won't be passed to the
+                # endpoint provider.
+                builtins[act_id_builtin_key] = None
+                return
+
+            act_id_builtin = builtins.get(act_id_builtin_key)
+            if act_id_builtin is None:
+                self._do_resolve_account_id_builtin(act_id_ep_mode, builtins)
```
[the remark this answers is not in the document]
dlm6693: I'll clean this up, but I think it should be this instead
```python
acct_id_builtin_key = EndpointResolverBuiltins.AWS_ACCOUNT_ID
if acct_id_ep_mode == 'disabled':
    # Unset the account ID if endpoint mode is disabled.
    builtins[acct_id_builtin_key] = None
elif builtins.get(acct_id_builtin_key) is None:
    self._do_resolve_account_id_builtin(acct_id_ep_mode, builtins)

File: tests/unit/test_endpoint_provider.py (hunk 10 above)
Comment (dlm6693) [reply tests/unit/test_endpoint_provider.py:120]: Needed a ruleset without account ID configured for a test case here
https://github.com/boto/botocore/blob/b819d815398162ea6a9603678239d3159e8ffcbf/tests/unit/test_endpoint_provider.py#L622

I could separate this out into it's own test, which would let me use it as a fixture and remove the need for this change.

File: tests/unit/test_args.py [hunk 15]
```diff
@@ -614,6 +614,28 @@ def test_bad_value_disable_request_compression(self):
...
+
+    def test_acct_id_ep_mode_client_cfg_overrides_cfg_store(self):
+        self.config_store.set_config_variable(
+            'account_id_endpoint_mode', 'preferred'
+        )
+        config = Config(account_id_endpoint_mode='required')
+        config = self.call_get_client_args(client_config=config)
+        client_config = config['client_config']
+        self.assertEqual(client_config.account_id_endpoint_mode, 'required')
+
```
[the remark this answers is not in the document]
dlm6693: Agreed. I'll remove these

tests/unit/test_credentials.py
dlm6693: Yeah I see what you're getting at here. I'll shoot up an update that only sets `account_id` if the value has been successfully resolved to a value that isn't `None`.

tests/functional/test_credentials.py
dlm6693: After coming back to this, I'm not seeing the test hanging anymore. I think it must've been something wrong with my local environment. I'll revert the change and keep an eye on it in CI

tests/functional/test_credentials.py
```diff
@@ -209,6 +209,7 @@ def create_random_credentials(self):
             'fake-%s' % random_chars(15),
             'fake-%s' % random_chars(35),
             'fake-%s' % random_chars(45),
+            account_id='fake-%s' % random_chars(12),
```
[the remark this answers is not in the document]
dlm6693: https://github.com/boto/botocore/pull/3031/files#r1346289525. I could do something similar to the unit tests and write tests that have `AssumedRoleUser` in the response instead.

botocore/regions.py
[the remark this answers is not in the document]
dlm6693: This was a decision made with the author of the spec. The reasoning is that presigned URLs are primarily for S3, which will likely never support account ID routing because there are bucket-specific endpoints in plain S3 and account-specific endpoints in S3 control. If you disagree or would like to discuss further, let me know.

botocore/regions.py
```diff
@@ -46,6 +48,12 @@
 LOG = logging.getLogger(__name__)
 DEFAULT_URI_TEMPLATE = '{service}.{region}.{dnsSuffix}'  # noqa
 DEFAULT_SERVICE_DATA = {'endpoints': {}}
+# Allowed values for the ``account_id_endpoint_mode`` config field.
+VALID_ACCOUNT_ID_ENDPOINT_MODES = (
+    'preferred',
+    'disabled',
+    'required',
+)
```
nateprewitt, line 56: Do these actually need to be a public global constant? It looks like we have these relegated to the `EndpointRulesetResolver`.

botocore/session.py
```diff
@@ -910,6 +916,10 @@ def create_client(
         :rtype: botocore.client.BaseClient
         :return: A botocore client instance
 
+        :type aws_account_id: string
+        :param aws_account_id: The AWS account ID to use when creating the client.
+            Same semantics as aws_access_key_id above.
+
```
nateprewitt, line 922:
```suggestion
```

.changes/next-release/feature-endpoints-57979.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "endpoints",
+  "description": "Enables account ID endpoint routing"
```
nateprewitt, line 4:
```suggestion
  "description": "Adds support for account ID endpoint routing."
```

tests/unit/test_credentials.py
```diff
@@ -190,6 +190,14 @@ def test_refresh_returns_partial_credentials(self):
         with self.assertRaises(botocore.exceptions.CredentialRetrievalError):
             self.creds.access_key
 
+    def test_account_id_refresh(self):
+        metadata = self.metadata.copy()
+        metadata['account_id'] = '123456789012'
+        self.refresher.return_value = metadata
+        self.mock_time.return_value = datetime.now(tzlocal())
+        self.assertTrue(self.creds.refresh_needed())
+        self.assertEqual(self.creds.account_id, '123456789012')
```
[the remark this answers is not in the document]
dlm6693: Accessing `self.creds.account_id` triggers the refresh. I've updated to change the expiry time so we can refresh again and set account ID the second time.

botocore/regions.py
```diff
@@ -450,6 +452,65 @@ class EndpointResolverBuiltins(str, Enum):
...
+    def _do_resolve_account_id_builtin(self, builtins):
+        frozen_creds = self._credentials.get_frozen_credentials()
+        account_id = frozen_creds.account_id
+        if account_id is None:
+            acct_id_ep_mode = self._account_id_endpoint_mode
+            msg = f'"account_id_endpoint_mode" is set to {acct_id_ep_mode}'
+            if acct_id_ep_mode == 'preferred':
+                LOG.debug('%s, but account ID was not found.', msg)
+            elif acct_id_ep_mode == 'required':
+                raise AccountIdNotFound(msg=msg)
```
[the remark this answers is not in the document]
dlm6693: I've provided more granular messaging based on `method` attribute available on the credentials object. I didn't write any tests for them though. Let me know what you think.

botocore/regions.py
```diff
@@ -450,6 +452,100 @@ class EndpointResolverBuiltins(str, Enum):
...
+                f'{base_msg} Please ensure that your credential '
+                'source is configured to return an account ID.'
+            )
+        else:
+            supported = ", ".join(self.SUPPORTED_CREDENTIAL_METHODS)
+            msg = (
+                f'{base_msg} Please change your credential source to one of: '
+                f'{supported} or set "account_id_endpoint_mode" to "disabled".'
+            )
+        return msg
```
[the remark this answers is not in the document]
dlm6693: I think this was how I interpreted this comment https://github.com/boto/botocore/pull/3031#discussion_r1378230399. I'll add a backlog item to update our developer guide.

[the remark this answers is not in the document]
nateprewitt: Ah ok, yeah what I was getting at there was attaching either a snapshot or metadata to the error to help with debugging. We don't need to include all the info in the message itself.

tests/unit/test_credentials.py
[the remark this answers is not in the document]
dlm6693: Since there is extra work required to trigger a refresh twice in the same test, I refactored to make the unset check a separate test.

nateprewitt (changes requested):

nateprewitt:

nateprewitt:

PR #3032, opened 2023-10-04
Remove duplicate line

tim-finnigan:
https://github.com/boto/botocore/blob/49008e50529ef19076eb644aac32ae238b6e1a38/tests/unit/retries/test_standard.py#L28C1-L29C29

### DIFF
--- a/tests/unit/retries/test_standard.py
+++ b/tests/unit/retries/test_standard.py
@@ -21,6 +21,7 @@ RETRYABLE_THROTTLED_RESPONSES = [
     (400, 'RequestThrottledException', True),
     (400, 'TooManyRequestsException', True),
     (400, 'ProvisionedThroughputExceededException', True),
+    (400, 'TransactionInProgressException', True),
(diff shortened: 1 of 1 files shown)


tests/unit/retries/test_standard.py
```diff
@@ -26,7 +26,6 @@
     (400, 'LimitExceededException', True),
     (403, 'RequestThrottled', True),
     (503, 'SlowDown', True),
-    (503, 'SlowDown', True),
```
nateprewitt, line 29: I think we're actually missing an item in the list. It looks like we should have this line in its place:

```
(400, 'TransactionInProgressException', True),
```

Can we add that on line 24 while we're here?

nateprewitt (changes requested):

tests/unit/retries/test_standard.py
tim-finnigan: Sure just added that to line 24.

nateprewitt (approved):

PR #3034, opened 2023-10-05
Add support for urllib3 2.0 in Python 3.10+

nateprewitt:
Starting in Python 3.10, the introduction of [PEP 644](https://peps.python.org/pep-0644/) requires that new versions of Python require OpenSSL 1.1.1+. This will let us confidently move the pin for Python 3.10+ to support the growing adoption of urllib3 2.0.

Details on the current blockers for allowing this for all version can be found in the [main tracking thread](https://github.com/boto/botocore/issues/2926#issuecomment-1673963526). Long term, we'll continue looking at options to expand support safely for users as we find other opportunities. For now, we'd encourage anyone looking to use urllib3 2.0 with Botocore to migrate to Python 3.10 or later.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,7 +5,8 @@ universal = 0
 requires_dist =
     jmespath>=0.7.1,<2.0.0
     python-dateutil>=2.1,<3.0.0
-    urllib3>=1.25.4,<1.27
+    urllib3>=1.25.4,<1.27; python_version<"3.10"
+    urllib3>=1.25.4,<2.1; python_version>="3.10"
 
 [options.extras_require]
 crt = awscrt==0.16.26
--- a/setup.py
+++ b/setup.py
@@ -25,8 +25,13 @@ def find_version(*file_paths):
 requires = [
     'jmespath>=0.7.1,<2.0.0',
     'python-dateutil>=2.1,<3.0.0',
-    'urllib3>=1.25.4,<1.27',
+    'urllib3>=1.25.4,<1.27 ; python_version < "3.10"',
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -25,9 +25,14 @@ def find_version(*file_paths):
...
     'python-dateutil>=2.1,<3.0.0',
-    'urllib3>=1.25.4,<1.27',
+    'urllib3>=1.25.4,<1.27 ; python_version < "3.10"',
+    'urllib3>=1.25.4,<2.1 ; python_version >= "3.10"',
 ]
 
+extras_require = {
+    'crt': ['awscrt==0.16.26'],
+}
+
```
nateprewitt, line 35: Adding the environment markers to setup.cfg doesn't work with extras for older versions of setuptools. I've replicated the extras info here (similar to how we have the requires info in setup.cfg) to make sure we retain compatibility in those older setuptools versions.

setu4993 (commented):
Thanks for the update + fix here!

dhhoang:
Thanks for this. Can I ask when we can expect this to be released? We are currently blocked by urllib3 conflict with Google API.

pquentin:
@dhhoang I believe google-auth relaxed its urllib3 pin so there's no conflict right now. Unless you're talking about something else?

setup.cfg
```diff
@@ -5,7 +5,8 @@ universal = 0
 requires_dist =
```
kyleknap, line 5: It's probably worth adding a changelog entry to this.

kyleknap (commented):
Looks good. I just had one comment.

kyleknap (approved):
:ship:

PR #3035, opened 2023-10-06
Sanitize __type metadata out of response parsing for tagged unions

SamRemis:
Some AWS services will return the __type metadata in a response alongside a tagged union.  Since this isn't modeled as a response member, it's never shown to SDK users, but needs to be ignored in the response parser to allow prevent an error from being thrown.


nateprewitt (approved):
Looks good, thanks @SamRemis!

PR #3036, opened 2023-10-09
Remove strict kwarg that was deprecated in Python 3.4

nateprewitt:
This PR will stop passing the no-longer required `strict` argument to the urllib3 ConnectionPool class. This argument was originally needed in `Python<3.5` (specifically Python 2) to specify support for older HTTP status lines before http/1.0. The functionality went away in HTTPConnection in Python 3.4 ([ref](https://docs.python.org/3/library/http.client.html#http.client.HTTPConnection)).

Starting in urllib3 2.1, this parameter will be removed from the APIs we're using. Given it has no effect on version of Python we support, this should be a no-op for users on _all_ versions of urllib3 we support. You can find a [similar PR](https://github.com/psf/requests/pull/6434) where we've done this in Requests which has been in place for ~6 months without issue.

### DIFF
--- a/botocore/httpsession.py
+++ b/botocore/httpsession.py
@@ -335,7 +335,6 @@ class URLLib3Session:
 
     def _get_pool_manager_kwargs(self, **extra_kwargs):
         pool_manager_kwargs = {
-            'strict': True,
             'timeout': self._timeout,
             'maxsize': self._max_pool_connections,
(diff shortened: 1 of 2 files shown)


dlm6693 (approved):

GavinZZ:
Hi there, any idea when this would be released to PyPI?

PR #3037, opened 2023-10-12
Add generic support for client context params

dlm6693:
Adds client context parameters generically across all AWS services and adds them to docs. This PR can be reviewed in two parts. Should be merged in tandem with https://github.com/boto/boto3/pull/3894.

## Commit 1
Adds `client_context_params` config option. If the service is s3 or s3control, still use the s3 config param over client context params if it is set. Params are modeled as capital case but are expected in snake case by the endpoint resolver. This commit also includes some code cleanup for the existing client context params test.

## Commit 2
Adds client context params docs to each service client's homepage if any are modeled. Existing s3/s3control params are excluded because overlapping behavior already exists in the s3 config. There are no other modeled client context params today, but when one is added to a service model, the new docs section will look something like this:

Screenshot 2023-10-11 at 2 58 03 PM

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -266,6 +266,7 @@ class ClientArgsCreator:
                 disable_request_compression=(
                     client_config.disable_request_compression
                 ),
+                client_context_params=client_config.client_context_params,
             )
         self._compute_retry_config(config_kwargs)
         self._compute_connect_timeout(config_kwargs)
@@ -641,14 +642,16 @@ class ClientArgsCreator:
             client_endpoint_url=endpoint_url,
             legacy_endpoint_url=endpoint.host,
         )
-        # botocore does not support client context parameters generically
-        # for every service. Instead, the s3 config section entries are
-        # available as client context parameters. In the future, endpoint
-        # rulesets of services other than s3/s3control may require client
-        # context parameters.
-        client_context = (
-            s3_config_raw if self._is_s3_service(service_name_raw) else {}
-        )
+        # Client context params for s3 conflict with the available settings
+        # in the `s3` parameter on the `Config` object. The s3 config will
+        # always take precedence over the client context params for s3 and
+        # s3control if set.
+        if self._is_s3_service(service_name_raw) and s3_config_raw:
+            client_context = s3_config_raw
+        elif client_config is not None and client_config.client_context_params:
+            client_context = client_config.client_context_params
+        else:
+            client_context = {}
         sig_version = (
             client_config.signature_version
             if client_config is not None
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -643,15 +643,15 @@ class ClientArgsCreator:
             legacy_endpoint_url=endpoint.host,
         )
         # Client context params for s3 conflict with the available settings
-        # in the `s3` parameter on the `Config` object. The s3 config will
-        # always take precedence over the client context params for s3 and
-        # s3control if set.
-        if self._is_s3_service(service_name_raw) and s3_config_raw:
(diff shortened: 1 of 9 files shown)


botocore/config.py
```diff
@@ -220,6 +220,16 @@ class Config:
...
+
+    :type client_context_params: dict
+    :param client_context_params: A dictionary of parameters specific to
+        individual services. If available, valid parameters can be found in the
+        ``Client Context Parameters`` section of the service's client homepage.
+        Invalid parameters or ones that are not used by the specified service
+        will be ignored.
+
+        Defaults to None.
+
```
nateprewitt, line 232:
```suggestion
```

botocore/args.py
```diff
@@ -641,14 +642,16 @@ def _build_endpoint_resolver(
...
-        # context parameters.
-        client_context = (
-            s3_config_raw if self._is_s3_service(service_name_raw) else {}
-        )
+        # Client context params for s3 conflict with the available settings
+        # in the `s3` parameter on the `Config` object. The s3 config will
+        # always take precedence over the client context params for s3 and
+        # s3control if set.
+        if self._is_s3_service(service_name_raw) and s3_config_raw:
+            client_context = s3_config_raw
```
nateprewitt, line 650: You've set up the documentation to exclude old params, but are documenting any new ones. If I supply my old config but want to opt into a new client context param, what does that look like? It's not clear how this handles that resolution.

dlm6693: Yeah you're right. I'll take another look at this.

botocore/docs/service.py
```diff
@@ -11,11 +11,25 @@
...
+OMITTED_CONTEXT_PARAMS = {
+    's3': (
+        'Accelerate',
+        'DisableMultiRegionAccessPoints',
+        'ForcePathStyle',
+        'UseArnRegion',
+    ),
+    's3control': ('UseArnRegion',),
+}
+
```
nateprewitt, line 32: Why is this being pulled up to the high level documentation classes instead of staying localized to the `ClientContextParamsDocumenter`?

nateprewitt (commented):
A couple minor comments but otherwise this looks good. Can you add an enhancement changelog to this?

botocore/docs/client.py
```diff
@@ -399,3 +400,56 @@ def _add_response_params(self, section, shape):
             shape,
             include=[self._GENERIC_ERROR_SHAPE],
         )
+
+
+class ClientContextParamsDocumenter:
+    _CONFIG_GUIDE_LINK = (
+        'https://boto3.amazonaws.com/'
+        'v1/documentation/api/latest/guide/configuration.html'
+    )
```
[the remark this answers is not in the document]
dlm6693: Per offline discussion we can't internally link to boto3 docs. There is also precedence for hardcoding the whole URL.

botocore/config.py
```diff
@@ -219,6 +219,15 @@ class Config:
     :param disable_request_compression: Disables request body compression if
         set to True.
 
+        Defaults to None.
+
+    :type client_context_params: dict
+    :param client_context_params: A dictionary of parameters specific to
+        individual services. If available, valid parameters can be found in
+        the ``Client Context Parameters`` section of the service's client's
+        documentation. Invalid parameters or ones that are not used by the
```
nateprewitt, line 228:
```suggestion
        the ``Client Context Parameters`` section of the service client's
        documentation. Invalid parameters or ones that are not used by the
```

nateprewitt (approved):

PR #3039, opened 2023-10-17
EKS provider updates

SamRemis:
This PR updates the existing container credential fetcher to support retrieving credentials from EKS

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1927,12 +1927,12 @@ class ContainerProvider(CredentialProvider):
                 auth_token = token_file.read()
         elif self.ENV_VAR_AUTH_TOKEN in self._environ:
             auth_token = self._environ[self.ENV_VAR_AUTH_TOKEN]
-        if auth_token:
+        if auth_token is not None:
             self._validate_auth_token(auth_token)
             return {'Authorization': auth_token}
 
     def _validate_auth_token(self, auth_token):
-        if "\r\n" in auth_token:
+        if "\r" in auth_token or "\n" in auth_token:
             raise ValueError("Auth token value is not a legal header value")
 
     def _create_fetcher(self, full_uri, headers):
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1886,6 +1886,7 @@ class ContainerProvider(CredentialProvider):
     ENV_VAR = 'AWS_CONTAINER_CREDENTIALS_RELATIVE_URI'
     ENV_VAR_FULL = 'AWS_CONTAINER_CREDENTIALS_FULL_URI'
     ENV_VAR_AUTH_TOKEN = 'AWS_CONTAINER_AUTHORIZATION_TOKEN'
+    ENV_VAR_AUTH_TOKEN_FILE = 'AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE'
 
     def __init__(self, environ=None, fetcher=None):
         if environ is None:
@@ -1919,9 +1920,20 @@ class ContainerProvider(CredentialProvider):
         )
 
     def _build_headers(self):
-        auth_token = self._environ.get(self.ENV_VAR_AUTH_TOKEN)
-        if auth_token is not None:
+        auth_token = None
+        if self.ENV_VAR_AUTH_TOKEN_FILE in self._environ:
+            auth_token_file_path = self._environ[self.ENV_VAR_AUTH_TOKEN_FILE]
+            with open(auth_token_file_path, 'r') as token_file:
+                auth_token = token_file.read()
+        elif self.ENV_VAR_AUTH_TOKEN in self._environ:
+            auth_token = self._environ[self.ENV_VAR_AUTH_TOKEN]
+        if auth_token:
+            self._validate_auth_token(auth_token)
             return {'Authorization': auth_token}
+
+    def _validate_auth_token(self, auth_token):
+        if "\r\n" in auth_token:
+            raise ValueError("Auth token value is not a legal header value")
 
     def _create_fetcher(self, full_uri, headers):
         def fetch_creds():
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1923,7 +1923,7 @@ class ContainerProvider(CredentialProvider):
         auth_token = None
         if self.ENV_VAR_AUTH_TOKEN_FILE in self._environ:
             auth_token_file_path = self._environ[self.ENV_VAR_AUTH_TOKEN_FILE]
-            with open(auth_token_file_path, 'r') as token_file:
+            with open(auth_token_file_path) as token_file:
                 auth_token = token_file.read()
         elif self.ENV_VAR_AUTH_TOKEN in self._environ:
             auth_token = self._environ[self.ENV_VAR_AUTH_TOKEN]
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -2938,8 +2938,8 @@ class ContainerMetadataFetcher:
         is_whitelisted_host = self._check_if_whitelisted_host(parsed.hostname)
         if not is_whitelisted_host:
             raise ValueError(
-                "Unsupported host '%s'.  Can only "
-                "retrieve metadata from these hosts: %s"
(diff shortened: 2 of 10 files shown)


tests/unit/test_credentials.py
```diff
@@ -3128,6 +3134,61 @@ def test_can_pass_basic_auth_token(self):
...
+        token_file_path = '/some/path/token.jwt'
+        environ = {
+            'AWS_CONTAINER_CREDENTIALS_FULL_URI': 'http://localhost/foo',
+            'AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE': token_file_path,
+        }
+        fetcher = self.create_fetcher()
+        provider = credentials.ContainerProvider(environ, fetcher)
+
+        with self.assertRaises(FileNotFoundError):
+            provider.load()
```
dlm6693, line 3177: Is this test necessary?

tests/unit/test_credentials.py
```diff
@@ -3128,6 +3134,61 @@ def test_can_pass_basic_auth_token(self):
...
+        environ = {
+            'AWS_CONTAINER_CREDENTIALS_FULL_URI': 'http://localhost/foo',
+            'AWS_CONTAINER_AUTHORIZATION_TOKEN': "invalid\r\ntoken",
+        }
+        fetcher = self.create_fetcher()
+        provider = credentials.ContainerProvider(environ, fetcher)
+
+        with self.assertRaisesRegex(
+            ValueError, "Auth token value is not a legal header value"
+        ):
```
dlm6693, line 3189: I'm not sure we need to test the exact error message here. Just asserting the right error gets raised should be sufficient.

tests/unit/test_utils.py
```diff
@@ -2598,6 +2598,24 @@ def test_can_use_127_ip_addr(self):
...
+
+    def test_can_use_eks_ipv6_addr(self):
+        self.assert_can_retrieve_metadata_from(
+            'http://[fd00:ec2::23]/credentials'
+        )
+
+    def test_can_use_complex_uri(self):
+        self.assert_can_retrieve_metadata_from(
+            'http://127.0.0.1:8080/credentials?foo=bar%20baz'
+        )
```
dlm6693, line 2614: nit: for formatting we can alias the url so we only use two lines
```suggestion
        url = 'http://127.0.0.1:8080/credentials?foo=bar%20baz'
        self.assert_can_retrieve_metadata_from(url)
```

dlm6693 (changes requested):

botocore/utils.py
```diff
@@ -2903,7 +2904,12 @@ class ContainerMetadataFetcher:
     RETRY_ATTEMPTS = 3
     SLEEP_TIME = 1
     IP_ADDRESS = '169.254.170.2'
-    _ALLOWED_HOSTS = [IP_ADDRESS, 'localhost', '127.0.0.1']
```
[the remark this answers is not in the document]
SamRemis: It's a loopback address, so now that's covered by the is_loopback function

tests/unit/test_credentials.py
SamRemis: Debatable; it's part of the specification that an error needs to be thrown here, but it should always be thrown by the open() method, and maybe this test should be added if we ever use something other than open, but for now this test is not necessary?  

I'll leave it in for now, but will likely delete it before this gets merged

botocore/credentials.py
```diff
@@ -1919,10 +1920,21 @@ def _retrieve_or_fail(self):
...
-        auth_token = self._environ.get(self.ENV_VAR_AUTH_TOKEN)
-        if auth_token is not None:
+        auth_token = None
+        if self.ENV_VAR_AUTH_TOKEN_FILE in self._environ:
+            auth_token_file_path = self._environ[self.ENV_VAR_AUTH_TOKEN_FILE]
+            with open(auth_token_file_path) as token_file:
+                auth_token = token_file.read()
+        elif self.ENV_VAR_AUTH_TOKEN in self._environ:
+            auth_token = self._environ[self.ENV_VAR_AUTH_TOKEN]
+        if auth_token:
```
nateprewitt, line 1930: This got more specific, what was the reason for changing it from `is not None`?

botocore/credentials.py
```diff
@@ -1919,10 +1920,21 @@ def _retrieve_or_fail(self):
...
+            self._validate_auth_token(auth_token)
             return {'Authorization': auth_token}
 
+    def _validate_auth_token(self, auth_token):
+        if "\r\n" in auth_token:
+            raise ValueError("Auth token value is not a legal header value")
```
nateprewitt, line 1936: It can be either character, we don't want any newlines/carriage returns in the header. Either of these options is probably sufficiently correct:

```python
_TOKEN_VALIDATOR = re.compile(r"^[^\r\n]*$")
[...]
    def _validate_auth_token(self, auth_token):
        if not _TOKEN_VALIDATOR.match(auth_token):
            raise ValueError("Auth token value is not a legal header value")
```

or just:

```python
if "\r" in auth_token or "\n" in auth_token:

**Review by nateprewitt — COMMENTED** [review]
A couple of minor comments, otherwise I think it looks good. Thanks Sam!

File: botocore/credentials.py (hunk 5 above)
Comment (SamRemis) [reply botocore/credentials.py:1930]: If the string winds up being empty - say due to an empty file being read.  This should probably be changed back to is not none and then we should validate it is a non empty string in the validate function instead.  Should we also be adding debug statements to let users know if the token came from a file or environment variable?

File: tests/unit/test_utils.py [hunk 7]
```diff
@@ -2218,7 +2218,7 @@ def get_s3_accesspoint_request(
         self,
         accesspoint_name=None,
         accesspoint_context=None,
-        **s3_request_kwargs
+        **s3_request_kwargs,
```
nateprewitt, line 2221: Did a formatter prompt this change? It's kind of odd to have a comma here since this is a catchall for everything.

tests/unit/test_utils.py
```diff
@@ -2598,6 +2597,17 @@ def test_can_use_127_ip_addr(self):
...
+        uri = 'http://169.254.170.23/credentials'
+        self.assert_can_retrieve_metadata_from(uri)
+
+    def test_can_use_eks_ipv6_addr(self):
+        uri = 'http://[fd00:ec2::23]/credentials'
+        self.assert_can_retrieve_metadata_from(uri)
+
+    def test_can_use_loopback_v6_uri(self):
+        self.assert_can_retrieve_metadata_from('http://[::1]/credentials')
+
```
nateprewitt, line 2610: One thing we're not testing is an IPv6 addr with a port.

nateprewitt (changes requested):
Last set of comments and I think this looks good. Would you mind also adding a changelog entry? You can use `scripts/new-change` to generate it. This would be a `feature`/`ContainerProvider`/`Added support for EKS container credentials` entry.

tests/unit/test_utils.py
nateprewitt: I see this came from the automated linter checks, I'm curious why this wasn't caught before.

nateprewitt (approved):

botocore/utils.py
```diff
@@ -2938,8 +2938,8 @@ def _validate_allowed_url(self, full_url):
         is_whitelisted_host = self._check_if_whitelisted_host(parsed.hostname)
         if not is_whitelisted_host:
             raise ValueError(
-                "Unsupported host '%s'.  Can only "
-                "retrieve metadata from these hosts: %s"
+                "Unsupported host '%s'.  Can only retrieve metadata "
+                "from a loopback address or one of these hosts: %s"
                 % (parsed.hostname, ', '.join(self._ALLOWED_HOSTS))
```
nateprewitt, line 2943: nit; when we update strings, we should try to move them towards either the `format()` or `f-string` style. There is however a single exception to this for log statements which are still optimized for this format.
```suggestion
                f"Unsupported host '{parsed.hostname}'.  Can only retrieve metadata from"
                f"a loopback address or one of these hosts: {', '.join(self._ALLOWED_HOSTS)}"
```

PR #3044, opened 2023-10-26
Feature eks credentials

SamRemis:
This PR updates the existing container credential fetcher to support retrieving credentials from EKS


botocore/utils.py
```diff
@@ -2983,22 +2997,20 @@ def _get_response(self, full_url, headers, timeout):
...
                 )
             try:
                 return json.loads(response_text)
             except ValueError:
-                error_msg = (
-                    "Unable to parse JSON returned from ECS metadata services"
-                )
-                logger.debug('%s:%s', error_msg, response_text)
+                error_msg = "Unable to parse JSON returned from container metadata services"
+                logger.debug(f"{error_msg}: {response_text}")
```
nateprewitt, line 3008: This is what I was noting in the previous comment from https://github.com/boto/botocore/pull/3039#discussion_r1372386119. We can leave log statements alone because they're actually more performant with the older syntax. This is mostly for error messages we're returning in exceptions.

nateprewitt (changes requested):
One minor comment, but otherwise looks good. Thanks @SamRemis!

nateprewitt (approved):

PR #3045, opened 2023-11-02
Add ec2_metadata_v1_disabled option

SamRemis:


botocore/utils.py
```diff
@@ -393,8 +393,10 @@ def __init__(
 
         if env is None:
             env = os.environ.copy()
-        self._disabled = env.get('AWS_EC2_METADATA_DISABLED', 'false').lower()
-        self._disabled = self._disabled == 'true'
+        self._disabled = (
+            env.get('AWS_EC2_METADATA_DISABLED', 'false').lower() == 'true'
+        )
```
dlm6693, line 398: I agree that setting an instance var only to immediately change it is probably bad practice. I'm guessing it was done to reduce the number of lines. We can still do that, improve the quality and utilize something more reusable if we di this instead.
```suggestion
        raw_disabled = env.get('AWS_EC2_METADATA_DISABLED', 'false')
        self._disabled = ensure_boolean(raw_disabled)
```
At the very least we should use `ensure_boolean` which lowercases string inputs already.

botocore/utils.py
```diff
@@ -554,6 +558,12 @@ def _log_imds_response(self, response, reason_to_log, log_body=False):
             logger_args.append(response.content)
         logger.debug(statement, *logger_args)
 
+    def _check_imdsv1_disabled(self):
+        if self._imds_v1_disabled:
```
dlm6693, line 562: I think we should be a little more explicit here. What if someone called this class directly and set  `ec2_metadata_v1_disabled` to `foo`?

.changes/next-release/enhancement-MetadataFetcher-40786.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "MetadataFetcher",
+  "description": "Adds a config option to opt out of IMDSv1 fallback if the fetcher is unable to acquire a token"
```
dlm6693, line 4:
```suggestion
  "description": "Adds a config option to opt out of IMDSv1 fallback"
```

botocore/utils.py
nateprewitt: Ideally this should have never been a boolean, and enum for the mode makes more sense long term. I left feedback on the specification but that ship has probably already sailed.

botocore/utils.py
```diff
@@ -457,10 +459,12 @@ def _fetch_metadata_token(self):
                 if response.status_code == 200:
                     return response.text
                 elif response.status_code in (404, 403, 405):
+                    self._check_imdsv1_disabled()
```
nateprewitt, line 462: Instead of adding checks in all the failure cases, could we do this once to the beginning of `_get_request` similar to how we did `_assert_enabled`? That way we're sure we don't emit a request without a token and we avoid accidentally regressing this if a new failure path is added.

botocore/utils.py
nateprewitt: Both are private variables, so I think we can say that's outside of the scope of expected behavior. We're already ensuring the variable we assign is a boolean, if they decide to mutate it, then here be dragons.

nateprewitt (changes requested):

botocore/utils.py
dlm6693: It looks like we only coerce to a boolean for `self._disabled` not `self._imds_v1_disabled`

botocore/utils.py
SamRemis: This was the original design, but I misunderstood some feedback.  Just reverted it, thank you :)

botocore/utils.py
nateprewitt: We should be using the same logic for both. Deviating where/how we check is going to create more pain in the future.

botocore/utils.py
```diff
@@ -475,7 +477,6 @@ def _fetch_metadata_token(self):
                     raise InvalidIMDSEndpointError(endpoint=url, error=e)
                 else:
                     raise
-        return None
```
nateprewitt, line 478: Was there a reason we removed this?

SamRemis: A misunderstanding of best practice.  Since this is an implicit return, I thought it was more idiomatic to remove it; I see now that since there's another branch that does return something, it's better to have it here.  Thank you

nateprewitt (approved):

PR #3048, opened 2023-11-08
Credential Scope

dlm6693:
This PR adds a `scope` to credentials that is used for endpoint resolution through the `AWS::Auth::CredentialScope` builtin parameter. Similar to account ID, it is treated as completely optional and only set on credential fetchers if the value is present. A couple of properties are added to the endpoint builtin resolver and piped through to the ruleset resolver in order to raise a configuration warning for customers that still rely on legacy behavior.

The base branch is https://github.com/boto/botocore/tree/account-id-endpoints since this PR is largely built on top of that branch and both will be merged at the same time. Once this is approved, account ID will be merged, this branch will be rebased with those changes and the base branch will be updated to https://github.com/boto/botocore/tree/develop/


botocore/client.py
```diff
@@ -1091,13 +1092,34 @@ def _resolve_endpoint_ruleset(
...
 
+    def _maybe_warn_signing_region_mismatch(self, signing_context):
+        legacy_signing_region = self._request_signer.region_name
+        signing_region = signing_context.get('region')
+        if (
+            signing_region is not None
+            and legacy_signing_region != signing_region
+            and not self._ruleset_resolver.uses_builtin_data_path
+            and self._ruleset_resolver.credential_scope_set
+        ):
```
dlm6693, line 1111: There's a variety of ways and places we can do this, but I think at the client level makes the most sense because we need state from the endpoint resolver and request signer. This also avoids adding a new attribute to the signer or adding additional data to the request context for a very niche case.

tests/functional/test_endpoint_rulesets.py
```diff
@@ -305,3 +306,28 @@ def builtin_overwriter_handler(builtins, **kwargs):
...
+    'botocore.regions.EndpointRulesetResolver.auth_schemes_to_signing_ctx',
+    return_value=('v4', {'region': 'us-west-2'}),
+)
+@mock.patch(
+    'botocore.regions.EndpointRulesetResolver.uses_builtin_data_path', False
+)
+@mock.patch(
+    'botocore.regions.EndpointRulesetResolver.credential_scope_set', True
+)
+def test_signing_region_mismatch_warns(auth_schemes_mock, patched_session):
```
dlm6693, line 322: It's unfortunate we have to mock so much out, but this is an edge case and would require a ton of scaffolding to be able to handle a warning. We could do this with a mock ruleset instead, but would need to mock out the function that returns the ruleset data to the client. The uses builtin check is already mocked for other behavior [here](https://github.com/boto/botocore/blob/7c1e00cd2bb38a3d41f1ca6849decfbe30605d56/tests/unit/test_args.py#L84) and would be painful to set up. It also seems like too much effort to make a custom endpoints file that would also need to be mocked out.

botocore/regions.py
```diff
@@ -510,19 +514,43 @@ def _validate_account_id_endpoint_mode(self):
...
+        return frozen_creds.scope
+
 
 class EndpointBuiltinResolver:
     """Resolves endpoint builtins during endpoint construction."""
 
-    def __init__(self, resolver_map):
+    def __init__(self, resolver_map, uses_builtin_data_path=True):
         self._resolver_map = resolver_map
+        self._uses_builtin_data_path = uses_builtin_data_path
```
dlm6693, line 536: We could consider renaming this to prevent confusion between and endpoint builtin parameter and builtin data. Maybe we change the name to `uses_custom_data` and flip the bool around.

botocore/regions.py
```diff
@@ -454,6 +455,9 @@ class EndpointResolverBuiltins(str, Enum):
     SDK_ENDPOINT = "SDK::Endpoint"
     # An account ID sourced from the credential resolution process.
     AWS_ACCOUNT_ID = "AWS::Auth::AccountId"
+    # The AWS credential scope configured for the SDK client. Clients may not
+    # make requests to regions outside their credential scope.
```
nateprewitt, line 459: I don't think this is quite right. Credentials are not valid outside of their credential scope region.

botocore/args.py
```diff
@@ -620,12 +621,15 @@ def _ensure_boolean(self, val):
...
+        self, credentials, client_config, uses_builtin_data
+    ):
         credential_builtin_resolver = CredentialBuiltinResolver(
-            credentials, client_config.account_id_endpoint_mode
+            credentials,
+            client_config.account_id_endpoint_mode,
         )
         resolver_map = {'credentials': credential_builtin_resolver}
-        return EndpointBuiltinResolver(resolver_map)
+        return EndpointBuiltinResolver(resolver_map, uses_builtin_data)
```
nateprewitt, line 632: If this is only for the warning, we shouldn't be changing the function signature for this. Let's use a simpler patch, if we resolve a custom credential scope. We'll warn the first time that it may not be honored for that endpoint if one is defined for the supplied credentials.

We should validate this _only_ warns when there's a custom `endpoints.json` in `~/.aws/models`.

botocore/client.py
nateprewitt: I think we should probably pull this and do something closer to the example above. I'm also not clear on why we're associating the region name on the signer with `credentialScope` in the `endpoint.json` file. Won't this fire for any S3 bucket that's in a different region?

nateprewitt (changes requested):

botocore/credentials.py
```diff
@@ -58,8 +58,8 @@
 logger = logging.getLogger(__name__)
 ReadOnlyCredentials = namedtuple(
     'ReadOnlyCredentials',
-    ['access_key', 'secret_key', 'token', 'account_id'],
-    defaults=(None,),
+    ['access_key', 'secret_key', 'token', 'account_id', 'scope'],
+    defaults=(None, None),
```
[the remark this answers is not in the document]
dlm6693: black splits this into multiple lines when including the trailing comma. Let me know if you'd like to see it reverted, leave as-is or do something else

botocore/client.py
dlm6693: No because S3 will never have credential scope configured on its ruleset. I'll still work on simplifying this though per your suggestion in the other comment.

botocore/regions.py
```diff
@@ -510,19 +513,43 @@ def _validate_account_id_endpoint_mode(self):
...
+
 
 class EndpointBuiltinResolver:
     """Resolves endpoint builtins during endpoint construction."""
 
-    def __init__(self, resolver_map):
+    def __init__(self, resolver_map, uses_builtin_data_path=True):
         self._resolver_map = resolver_map
+        self._uses_builtin_data_path = uses_builtin_data_path
+        self._credential_scope_set = False
```
nateprewitt, line 536: This can't exist. We're adding in a race condition where this isn't going to be accurate in a multi-threaded context.

nateprewitt (changes requested):

botocore/args.py
```diff
@@ -219,7 +219,6 @@ def compute_client_args(
             client_config=client_config,
             endpoint_url=endpoint_url,
         )
-
```
nateprewitt: Was this intentional?

botocore/client.py
```diff
@@ -788,8 +789,17 @@ def _pick_region_values(self, resolved, region_name, endpoint_url):
...
+                self._maybe_warn_custom_signing_region(signing_region)
         return region_name, signing_region
 
+    def _maybe_warn_custom_signing_region(self, signing_region):
+        if not self.resolver_uses_builtin_data():
+            warnings.warn(
+                "Detected a signing region resolved from a custom "
+                f"endpoints.json file: {signing_region}. This value "
+                "may not be honored during request signing."
+            )
```
nateprewitt, line 801:
```suggestion
    def _warn_custom_signing_region(self, signing_region):
        if not self.resolver_uses_builtin_data():
            warnings.warn(
                "Detected a signing region resolved from a custom "
                f"endpoints.json file: {signing_region}. This value "
                "may not be honored during request signing."
            )
```

botocore/regions.py
```diff
@@ -510,6 +513,19 @@ def _validate_account_id_endpoint_mode(self):
...
+        self, builtin_configured, builtin_value
+    ):
+        """Resolve the ``AWS::Auth::CredentialScope`` builtin."""
+        if not builtin_configured or self._credentials is None:
+            return None
+
+        if builtin_value is not None:
+            return builtin_value
+
+        frozen_creds = self._credentials.get_frozen_credentials()
```
nateprewitt, line 526: Is this introducing a risk for mismatches in values? We call `get_frozen_credentials` separately in `resolve_account_id_builtin` and `resolve_credential_scope_builtin` which may both [invoke `_refresh`](https://github.com/boto/botocore/blob/4181e681982e60cf49285e17d21c35737299d78c/botocore/credentials.py#L611) at different times.

I think this got called out in the other PR at one point, but how are you ensuring the values don't change between the time each of these is resolved? Shouldn't they all be using the same set of frozen credentials? What testing did we do to ensure this was safe?

botocore/session.py
```diff
@@ -919,6 +923,11 @@ def create_client(
         :type aws_account_id: string
         :param aws_account_id: The AWS account ID to use when creating the client.
             Same semantics as aws_access_key_id above.
+
+        :type aws_credential_scope: string
+        :param aws_credential_scope: The AWS credential scope to use when creating
+            creating the client. Same semantics as aws_access_key_id above. Must be
+            equal to region_name.
```
nateprewitt, line 930: > Must be equal to region_name.

I think we discussed this not necessarily being true. We should notify through the exceptions rather than hardcoding this in the description.

tests/functional/test_regions.py
```diff
@@ -545,3 +545,31 @@ def test_endpoint_resolver_knows_its_datasource(patched_session, is_builtin):
...
+    num_expected_warnings,
+    mock_resolved_endpoint,
+    patched_session,
+    recwarn,
+):
+    construct_endpoint_mock.return_value = mock_resolved_endpoint
+    loader = patched_session.get_component('data_loader')
+    with mock.patch.object(loader, 'is_builtin_path', return_value=is_builtin):
+        patched_session.create_client('iam')
+        assert len(recwarn) == num_expected_warnings
```
nateprewitt, line 575: We may want to be more specific with this test because it has potential to either break if new warnings are added or misrepresent coverage in those cases.

Can we do this at a higher level so we're not having to get as minute with the mocks? Could we stub out AWS_DATA_PATH to a temporary file and then just ensure a call with the client raises the warning?

tests/unit/data/endpoints/valid-rules/aws-credential-scope.json
```diff
@@ -0,0 +1,95 @@
+{
```
nateprewitt, line 1: Has there been any discussion about making this a more comprehensive ruleset so we don't have the same boiler plate each time?

tests/unit/test_endpoint_provider.py
```diff
@@ -542,29 +555,53 @@ def operation_model_empty_context_params():
...
+    token="token",
+    account_id="1234567890",
+    scope=US_WEST_2,
+)
 REQUIRED = "required"
 PREFERRED = "preferred"
 DISABLED = "disabled"
 URL_NO_ACCOUNT_ID = "https://amazonaws.com"
 URL_WITH_ACCOUNT_ID = "https://1234567890.amazonaws.com"
+URL_NO_SCOPE = URL_NO_ACCOUNT_ID
```
nateprewitt, line 595: Is this needed? Should we have a clearer name and define it once? Otherwise, we may accidentally regress the other based on the assumption we're modifying something specific to account id.

nateprewitt (changes requested):

botocore/regions.py
dlm6693: I called this out in the design doc. Neither account ID nor credential scope are expected to change throughout the lifetime of a client so I opted against pulling frozen creds out to a single call. We can add a frozen credentials property here though.

If there were a risk however, using the same frozen credentials instance wouldn't solve the underlying race condition problem because there is still a separate call at the outset of request signing.

tests/unit/data/endpoints/valid-rules/aws-credential-scope.json
dlm6693: I don't think so. I did write the last few by hand whereas the existing files were all shipped with the endpoints refactor last year. Since they're all pretty specific I elected to keep them separate. If we want to combine them I think it'd be better to do that as a followup PR.

nateprewitt (changes requested):

botocore/regions.py
```diff
@@ -522,20 +547,42 @@ def resolve(self, param_definitions, builtins):
...
 
     def _resolve_credential_builtins(self, param_definitions, builtins):
-        self._resolve_account_id_builtin(param_definitions, builtins)
+        frozen_creds_cache = {}
+        self._resolve_account_id_builtin(
+            param_definitions, builtins, frozen_creds_cache
+        )
+        self._resolve_credential_scope_builtin(
+            param_definitions, builtins, frozen_creds_cache
+        )
```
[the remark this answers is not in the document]
dlm6693: The point is to cache it only through a single endpoint resolution call. What you suggest is naturally a much simpler implementation, but it resolves credentials regardless of state.  My goal was to maintain the behavior of only resolving credentials given certain conditions, but also ensuring that the `get_frozen_credentials` is only called once per resolved endpoint.

I can make this change, but doing so will mean that credentials are always resolved earlier in the call stack, moving from request signing to here.

tests/unit/test_credentials.py
```diff
@@ -3765,6 +3968,47 @@ def test_raises_helpful_message_on_unauthorized_exception(self):
...
+                'accessKeyId': 'foo',
+                'secretAccessKey': 'bar',
+                'sessionToken': 'baz',
+                'expiration': self.now_timestamp + 1000000,
+                'credentialScope': 'us-west-2',
+            }
+        }
+        client = mock.Mock()
+        client.get_role_credentials.return_value = expected_response
+        client_creator = mock.Mock(return_value=client)
```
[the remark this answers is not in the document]
dlm6693: Yeah this was a miss on my part for not mentioning earlier. `credentialScope` isn't modeled for SSO `get_role_credentials` yet so we're always going to get a `ParamValidationError` if we use the stubbed client. However, we really should test this behavior still so I decided to use a similar approach that we have in `TestAssumeRoleCredentialFetcher`:
https://github.com/boto/botocore/blob/develop/tests/unit/test_credentials.py#L241-L249
This is also why we don't experience the same issue in those suites of tests.

tests/functional/test_regions.py
```diff
@@ -10,10 +10,15 @@
...
 # language governing permissions and limitations under the License.
+import json
+import os
+import tempfile
+
 import pytest
 
 from botocore.client import ClientEndpointBridge
 from botocore.exceptions import NoRegionError
+from botocore.session import Session
```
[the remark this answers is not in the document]
dlm6693: We need to use something similar. Unfortunately both `patched_session` and `BaseSessionTest` rely on a cached `Loader` instance which cannot be used for this specific case

tests/functional/test_regions.py
```diff
@@ -545,3 +551,58 @@ def test_endpoint_resolver_knows_its_datasource(patched_session, is_builtin):
...
+class TestCustomEndpointsFile(BaseSessionTest):
+    def setUp(self):
+        super().setUp()
+        self.tempdir = tempfile.TemporaryDirectory()
+        self.environ['AWS_DATA_PATH'] = self.tempdir.name
+        self.session = Session()
+        self.session.set_config_variable(
+            'credentials_file', 'noexist/foo/botocore'
+        )
+        self.session.config_filename = 'no-exist-foo'
```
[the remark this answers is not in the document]
dlm6693: Discussed offline. The point of this was to mimic everything we're doing in `BaseSessionTest` and `create_session`. We needed to create our own session because the method uses a cached `Loader` instance which does not have the custom data path attached to it. However, since there are other tests that check that the env vars work as expected, we're going to modify `create_session` instead to pass a custom loader

tests/unit/test_credentials.py
```diff
@@ -3658,6 +3783,84 @@ def test_account_id_from_process_takes_precedence(self):
...
+        provider = self.create_process_provider()
+        creds = provider.load()
+        self.assertIsNotNone(creds)
+        self.assertEqual(creds.access_key, 'foo')
+        self.assertEqual(creds.secret_key, 'bar')
+        self.assertEqual(creds.token, 'baz')
+        self.assertEqual(creds.method, 'custom-process')
+        self.assertEqual(creds.scope, 'us-west-2')
+
+    def test_credential_scope_from_process_takes_precedence(self):
```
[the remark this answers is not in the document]
dlm6693: The latter. However, I think this is maybe the only case where credential values can be retrieved from more than one source. I don't know if it's a guarantee that all future credential values will share this behavior though so we may want to keep the individual checks. Here is the wording in the spec for account ID. Behavior for scope is expected to be the same. Note that it does not say to fall back to all credentials in the config/credentials file, just the individual values.
```
If the process credentials provider is provisioned through the config file, and successful credential response from the process as described above does not contain a value for account ID, SDKs SHOULD attempt to resolve the value of the profile property aws_account_id and store it.
```

botocore/regions.py
```diff
@@ -466,27 +469,78 @@ class CredentialBuiltinResolver:
...
+    def _should_resolve_scope(self, param_definitions):
+        return self._builtin_configured(param_definitions, 'CredentialScope')
+
+    def _builtin_configured(self, param_definitions, param_name):
+        param_def = param_definitions.get(param_name)
+        return param_def is not None and param_def.builtin is not None
+
+    def _validate_builtin(self, builtin_name, value):
+        def noop_validator(value):
+            return value
```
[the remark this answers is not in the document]
dlm6693: I thought namespacing it with a specific purpose was more readable than the lambda, but I don't feel strongly about it.

botocore/regions.py
```diff
@@ -466,27 +469,78 @@ class CredentialBuiltinResolver:
...
+        has_builtin = self._builtin_configured(param_definitions, 'AccountId')
+        mode_enabled = self._account_id_endpoint_mode != 'disabled'
+        return has_builtin and mode_enabled
+
+    def _should_resolve_scope(self, param_definitions):
+        return self._builtin_configured(param_definitions, 'CredentialScope')
+
+    def _builtin_configured(self, param_definitions, param_name):
+        param_def = param_definitions.get(param_name)
+        return param_def is not None and param_def.builtin is not None
```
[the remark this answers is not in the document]
dlm6693: We can. I think I opted to keep it because it is similar syntactically to the `_validate_builtin` logic and avoids the use of conditionals. Additionally we will need to update the mapping to use the pascal case `AccountId` and `CredentialScope`. I'll have something up soon.

nateprewitt:

PR #3062, opened 2023-11-14
Remove Python 3.6 compatibility code from datetime2timestamp()

akx:
The library has targeted 3.7+ since #2684; no need to retain compatibility code that is never called.

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -1049,11 +1049,7 @@ def datetime2timestamp(dt, default_timezone=None):
             default_timezone = tzutc()
         dt = dt.replace(tzinfo=default_timezone)
     d = dt.replace(tzinfo=None) - dt.utcoffset() - epoch
-    if hasattr(d, "total_seconds"):
-        return d.total_seconds()  # Works in Python 3.6+
-    return (
-        d.microseconds + (d.seconds + d.days * 24 * 3600) * 10**6
-    ) / 10**6
+    return d.total_seconds()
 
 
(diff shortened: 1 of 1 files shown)


nateprewitt:
Thanks for the PR @akx. While we don't officially support Python 3.6, we still have some downstreams that are consuming current versions of botocore using Python 3.6. We've intentionally left a few of these changes and our [syntax upgrade](https://github.com/boto/botocore/blob/develop/.pre-commit-config.yaml#L30-L35) setup to avoid unnecessarily breaking this.

That said, we have changes staged for our upcoming Python 3.7 deprecation that will address some of this. I can leave this open to be integrated into that branch or we can resolve this as a duplicate.

akx:
@nateprewitt Thanks – I'm curious: how are those downstreams using new `botocore`s when a [`python_requires='>=3.7'`](https://github.com/boto/botocore/blame/4181e681982e60cf49285e17d21c35737299d78c/setup.py#L53) is in place? Just `pip install --ignore-requires-python` and hope for the best? 😁

nateprewitt:
Unfortunately, there are still distant reaches that don't rely on `pip` or any of the PyPA infrastructure to manage their Python packages. We'll hopefully be leaving that behind shortly 😉

nateprewitt (approved):
Ok, we're ready to ship 3.7 deprecation. This should be good to go.

I saw the other PR, we still have some concerns around the original windows issue that lead to that PR. It's likely safe to clean up as you've suggested but we'll need to do some manual testing first.

PR #3068, opened 2023-11-21
Add private shim to fetch credentials

nateprewitt:
This is for internal use for retrieving resolved credentials off a client when a Session credential resolver isn't readily available. It is subject to unannounced breakages and shouldn't be accessed publicly.


kyleknap (approved):
:ship:

PR #3074, opened 2023-11-26
Remove deprecated IOError

jxdv:
Since python version 3.3 IOError has been deprecated and merged (with various different errors) into OSError.


nateprewitt:
Thanks for the PR, @jxdv. This is a vendored package we don't intend to modify at this time.

Issue #3088, opened 2023-12-18
botocore using deprecated datetime.datetime.utcnow() call


paulschreiber:
### Describe the bug

botocore is using `datetime.datetime.utcnow()`, which is deprecated. It generates this warning:
```
 DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    datetime_now = datetime.datetime.utcnow()
```

### Expected Behavior

No warnings

### Current Behavior

DeprecationWarning

### Reproduction Steps

n/a

### Possible Solution

Use datetime.datetime.now(datetime.UTC)

### SDK version used

1.34.2

### Environment details (OS name and version, etc.)

Python 3.12.0

nateprewitt:
Hi @paulschreiber,

Thanks for reaching out. We're currently tracking this in boto/boto3#3889. Unfortunately, this isn't as straight forward as replacing a function call. We already attempted this in https://github.com/boto/botocore/pull/3003/commits/96ca907d74d27ebb5ad715db4a75228fdc029a20 when 3.12 support was released and needed to roll it back due to breaking date serialization changes.

We're currently looking at removing the warnings long term in a backwards compatible manner, but don't have a timeline to provide. The `utcnow` deprecation is currently slated sometime after Python 3.16, so it's at least few years out. We'll work on getting it addressed more timely.

iam-decoder:
After upgrading from python 3.11.11 to python 3.13, I get errors because of this:

```
    <class 'TypeError'> - account.py#453: issubclass() arg 2 must be a class, a tuple of classes, or a union
    Traceback (most recent call last):
      File "/path/to/my/code/models/account.py", line 453, in __verify_account_creds
        ....client('sts').get_caller_identity()
          ~~~~~~~~~~~~~~~~~~~^^
      File "/path/to/lib/python3.13/site-packages/botocore/client.py", line 570, in _api_call
        return self._make_api_call(operation_name, kwargs)
               ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
      File "/path/to/lib/python3.13/site-packages/botocore/context.py", line 124, in wrapper
        return func(*args, **kwargs)
      File "/path/to/lib/python3.13/site-packages/botocore/client.py", line 1013, in _make_api_call
        http, parsed_response = self._make_request(
[...]

PR #3089, opened 2023-12-18
Replace deprecated datetime.datetime.utcnow() call with datetime.datetime.now(datetime.UTC)

paulschreiber:
### Description
Replace deprecated datetime.datetime.utcnow() call with datetime.datetime.now(datetime.UTC).

[From the Python 3.12 readme](https://docs.python.org/3/whatsnew/3.12.html):

>[datetime](https://docs.python.org/3/library/datetime.html#module-datetime): [datetime.datetime](https://docs.python.org/3/library/datetime.html#datetime.datetime)’s [utcnow()](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) and [utcfromtimestamp()](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp) are deprecated and will be removed in a future version. Instead, use timezone-aware objects to represent datetimes in UTC: respectively, call [now()](https://docs.python.org/3/library/datetime.html#datetime.datetime.now) and [fromtimestamp()](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp) with the tz parameter set to [datetime.UTC](https://docs.python.org/3/library/datetime.html#datetime.UTC). (Contributed by Paul Ganssle in [gh-103857](https://github.com/python/cpython/issues/103857).)

Fixes https://github.com/boto/botocore/issues/3088.

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -416,7 +416,7 @@ class SigV4Auth(BaseSigner):
     def add_auth(self, request):
         if self.credentials is None:
             raise NoCredentialsError()
-        datetime_now = datetime.datetime.utcnow()
+        datetime_now = datetime.datetime.now(datetime.UTC)
         request.context['timestamp'] = datetime_now.strftime(SIGV4_TIMESTAMP)
         # This could be a retry.  Make sure the previous
         # authorization header is removed first.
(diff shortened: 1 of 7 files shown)


dlm6693:
Thanks for the PR @paulschreiber! I actually have been working on this myself and just opened https://github.com/boto/botocore/pull/3093 so I'm going to close this one.  Also please note that the `datetime.UTC` syntax was only introduced in python 3.11. Since we still support 3.8 - 3.10, we have to use the old syntax `datetime.timezone.utc`.

PR #3090, opened 2023-12-18
Remove legacy pseudo FIPS region tests

kyleknap:
The tests were based around legacy endpoint resolution that is no longer used by default in botocore. Any new S3 FIPS entries in the `endpoints.json` does not affect endpoint resolution. The only affect that adding the new pseudo FIPS region to the `endpoints.json` file is the pseudo region will be included in enumeration of regions when calling the `get_available_regions` method with `allow_non_regional` set to `True`, which historically has include pseudo FIPS regions.


nateprewitt (approved):
⛵

PR #3093, opened 2023-12-26
remove `utcnow()` calls to fix deprecation warnings encountered in python 3.12

dlm6693:
Replace `utcnow()` with `now(datetime.timezone.utc)`. Integ tests were run successfully on my local machine. One deprecation warning from `python-dateutil` is still expected in GitHub CI. The unknown timezone warning is a separate issue that should be addressed in another PR.


nateprewitt:

Issue #3098, opened 2024-01-13
botocore does not ignore credential_process when that configuration value is blank (empty string)


rdctmeconomou:
### Describe the bug

[The AWS CLI documentation on changing configuration settings using commands says](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-methods):

> To remove a setting, use an empty string as the value, or manually delete the setting in your config and credentials files in a text editor.

However, if one sets `credential_process` to the empty string, botocore neither removes the configuration variable from the configuration file nor ignores its value.  Instead, subsequent invocations of the AWS CLI or botocore library result in an `IndexError` when `botocore.credentials.ProcessProvider._retrieve_credentials_using()` passes an empty list derived from the empty string value to `subprocess.Popen()` at [botocore/credentials.py line 1019](https://github.com/boto/botocore/blob/cd683293c485ac9f965ab398e0a6437ad711f55b/botocore/credentials.py#L1019C25-L1019C25).

### Expected Behavior

botocore (and by extension the AWS CLI) should act as if the `credential_process` configuration variable is unset when it has an empty string value as described in the AWS CLI documentation linked above.

```
$ rm ~/.aws/*
$ aws configure set credential_process "" --profile test
$ cat ~/.aws/config
[profile test]
credential_process = 
$ aws s3api list-buckets --profile test

Unable to locate credentials. You can configure credentials by running "aws configure".
```

### Current Behavior

One cannot remove the `credential_process` setting without editing the configuration file.

```
$ rm ~/.aws/*
$ aws configure set credential_process "" --profile test
$ cat ~/.aws/config
[profile test]
credential_process = 
$ aws s3api list-buckets --profile test

list index out of range
```

### Reproduction Steps

I've trimmed the output to focus on the expected behavior (`NoCredentialsError` in `botocore.auth.add_auth()` when no credentials are set) versus the current behavior (`IndexError` in `subprocess._execute_child()` when `credentials_process` is set to the empty string).

```
$ rm -rf ~/.aws
$ python3.11 -m venv .venv
$ source .venv/bin/activate
$ pip install awscli
$ aws configure set credential_process "" --profile test
$ python
>>> import botocore.session
>>> session = botocore.session.get_session()
>>> client = session.create_client('ec2', region_name='us-west-2')
>>> for reservation in client.describe_instances()['Reservations']:
...     for instance in reservation['Instances']:
...         print(instance['InstanceId'])
... 
[...]

PR #3099, opened 2024-01-13
Ignore credential_process when set to the empty string

rdctmeconomou:
This implements the fix to `botocore.credentials.ProcessProvider.load()` described in #3098.  I added a test that reproduced the bug.  The test passes with my fix applied.

Please note that linting fails in test code released as part of botocore 1.34.18.  I didn't think it wise to modify anything outside the scope of my bug report but am happy to commit a style fix if necessary.


rdctmeconomou:
I noticed #3100 and will rebase my PR as soon as that's merged.

PR #3107, opened 2024-01-26
add support for memoryview objects

jakkdl:
Fixes https://github.com/boto/boto3/issues/3423

I know it's been labeled needs-discussion, but I wanted to check if it really was as easy as just modifying a couple `isinstance` checks, and after trying it and seeing it really was just that I thought I might as well write up a PR for it.

I've been working on https://github.com/piskvorky/smart_open/issues/380 and it would save significant amounts of memory if it was possible to directly pass [`memoryview`](https://docs.python.org/3/library/stdtypes.html#memoryview) objects to botocore.

I opted to modify all cases where I found `isinstance.*bytearray` in the code, even though https://github.com/boto/boto3/issues/3423 only directly needs the modifications in `validate.py` and `utils.py` afaik. But I see no reason not to accept it in all those places.

I also noticed there was no test for `bytearray` + httpchecksum, and the error message for type validation didn't include `str` despite the `isinstance` line above accepting it, so quickly fixed those as well.

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -528,7 +528,7 @@ class AWSPreparedRequest:
         # the entire body contents again if we need to).
         # Same case if the body is a string/bytes/bytearray type.
 
-        non_seekable_types = (bytes, str, bytearray)
+        non_seekable_types = (bytes, str, bytearray, memoryview)
         if self.body is None or isinstance(self.body, non_seekable_types):
             return
         try:
--- a/botocore/compress.py
+++ b/botocore/compress.py
@@ -71,7 +71,7 @@ def _is_compressible_type(request_dict):
     if isinstance(body, dict):
         body = urlencode(body, doseq=True, encoding='utf-8').encode('utf-8')
         request_dict['body'] = body
-    is_supported_type = isinstance(body, (str, bytes, bytearray))
+    is_supported_type = isinstance(body, (str, bytes, bytearray, memoryview))
     return is_supported_type or hasattr(body, 'read')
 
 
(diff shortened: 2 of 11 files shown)


jakirkham:
Thanks for working on this John! 🙏

This looks like a great improvement. We ran into a similar issue

Wonder if it makes sense to use Python 3.12's [`Buffer`]( https://docs.python.org/3.12/library/collections.abc.html#collections.abc.Buffer ) to catch all of these types. In the end most C/C++ code working with the memory itself (like in IO logic) would check for [the Python Buffer Protocol]( https://docs.python.org/3/c-api/buffer.html ), which this type matches. For older Python versions [`typing_extensions`]( https://github.com/python/typing_extensions ) includes it starting in `4.6.0` ( https://github.com/python/typing_extensions/commit/37909ec67adf56070fa7218efa475e87812834e5 ).  This would help simplify the existing logic improving maintainability

Think we may also want to ensure `memoryview`'s are normalized by [casting them]( https://docs.python.org/3/library/stdtypes.html#memoryview.cast ) to a byte-like type. For example

```python
mv = mv.cast("B")
```

This will ensure code that checks for the number of bytes in a message works with `len`. Otherwise `memoryview` will otherwise return the number of items which could differ if the items are larger than 1 byte like `double`. This would help in the code below

https://github.com/boto/botocore/blob/590de686e6afd22bdfd74fd134fd37da364dc986/botocore/utils.py#L3183-L3187

Not sure the exact right place to do this, but this code path looks like a good option

https://github.com/boto/botocore/blob/590de686e6afd22bdfd74fd134fd37da364dc986/botocore/serialize.py#L499-L505

jakkdl:
using `Buffer` sounds like a good idea, though I'm not as familiar with memoryview casting - but on a glance looks good as well.
But given the complete lack of interest from maintainers for over a year I'm vary of putting in extra efforts for marginal improvements to this PR. But would be open to doing so if I was given an indication that this might be merged at some point.

PR #3109, opened 2024-01-31
Fix pytest-8.0.0 test failures due to incorrect `pytest.warns()`

mgorny:
Remove two incorrect `pytest.warns()` invocations
for `.get_component('internal')`, in order to fix test failures with pytest 8.0.0.  This version of pytest fixes `pytest.warns()` to work inside `pytest.raises()`, and therefore causes the tests to fail because the warning is not emitted when `get_component()` raises `ValueError`.

### DIFF
--- a/tests/unit/test_session.py
+++ b/tests/unit/test_session.py
@@ -781,9 +781,7 @@ class TestSessionComponent(BaseSessionTest):
             self.session._get_internal_component('internal'), component
         )
         with self.assertRaises(ValueError):
-            # get_component has been deprecated to the public
-            with pytest.warns(DeprecationWarning):
-                self.session.get_component('internal')
+            self.session.get_component('internal')
 
     def test_internal_endpoint_resolver_is_same_as_deprecated_public(self):
         endpoint_resolver = self.session._get_internal_component(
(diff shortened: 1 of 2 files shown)


tests/unit/test_session.py
```diff
@@ -781,9 +781,7 @@ def test_internal_component(self):
             self.session._get_internal_component('internal'), component
         )
         with self.assertRaises(ValueError):
-            # get_component has been deprecated to the public
-            with pytest.warns(DeprecationWarning):
-                self.session.get_component('internal')
+            self.session.get_component('internal')
```
nateprewitt, line 784: For reference, we're handing the correct case [here](https://github.com/boto/botocore/pull/3109/files#diff-7824b4fcea790555e99c27ab732bde7187f87896ab367892c6ec1b8949c064f4R801-R806).

tests/unit/test_session_legacy.py
```diff
@@ -795,9 +795,7 @@ def test_internal_component(self):
             self.session._get_internal_component('internal'), component
         )
         with self.assertRaises(ValueError):
-            # get_component has been deprecated to the public
-            with pytest.warns(DeprecationWarning):
-                self.session.get_component('internal')
```
nateprewitt, line 800: For reference, we're handling the correct case [here](https://github.com/boto/botocore/pull/3109/files#diff-2d1a03ec5b80ddbfd9b9c273d7cace9e6c14b3b6facc646ee00c6d2587fc466bR815-R820).

nateprewitt (approved):
Thanks, @mgorny! This does looks like an oversight in the test.

mgorny:
Thanks!

PR #3114, opened 2024-02-06
Fix EKS credentials refresh bug

SamRemis:
This PR fixes a bug with EKS credentials not properly fetching the token from the token file if the file contents have been updated

### DIFF
--- a/tests/unit/test_credentials.py
+++ b/tests/unit/test_credentials.py
@@ -3199,7 +3199,6 @@ class TestContainerProvider(BaseEnvVar):
         time.sleep(3)
         with open(token_file_path, 'w') as token_file:
             token_file.write('Second auth token')
-        provider = credentials.ContainerProvider(environ, fetcher)
         provider.load()
         fetcher.retrieve_full_uri.assert_called_with(
             'http://localhost/foo',
--- a/tests/unit/test_credentials.py
+++ b/tests/unit/test_credentials.py
@@ -3191,7 +3191,7 @@ class TestContainerProvider(BaseEnvVar):
             },
         ]
         provider = credentials.ContainerProvider(environ, fetcher)
-        provider.load()
+        creds = provider.load()
         fetcher.retrieve_full_uri.assert_called_with(
             'http://localhost/foo',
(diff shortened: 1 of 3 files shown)


tests/unit/test_credentials.py
```diff
@@ -3165,6 +3166,45 @@ def test_can_pass_auth_token_from_file(self):
         self.assertEqual(creds.token, 'token')
         self.assertEqual(creds.method, 'container-role')
 
+    def test_reloads_auth_token_from_file(self):
+        token_file_path = os.path.join(self.tempdir, 'token.jwt')
```
nateprewitt, line 3170: For this, we're going to have collisions/race conditions if the tests happen to get run more than once on the same system. It may be a good idea to use [temporary_file](https://github.com/boto/botocore/blob/9bacd6c033cc42b2f8fdde68be3a98e5447454db/tests/__init__.py#L111-L133) to ensure we're both isolated and cleaning up immediately.

tests/unit/test_credentials.py
```diff
@@ -3165,6 +3166,45 @@ def test_can_pass_auth_token_from_file(self):
...
+            'AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE': token_file_path,
+        }
+        fetcher = self.create_fetcher()
+        timeobj = datetime.now(tzlocal())
+        fetcher.retrieve_full_uri.side_effect = [
+            {
+                "AccessKeyId": "first_key",
+                "SecretAccessKey": "first_secret",
+                "Token": "first_token",
+                "Expiration": (timeobj + timedelta(seconds=2)).isoformat(),
```
nateprewitt, line 3184: nit; this could probably we 1s with a 1.5-2s sleep but that may end up creating more noise in the tests with false failures. Something to keep an eye on after release.

nateprewitt (commented):
Cool, looks good overall! A couple minor comments, after that I think we're good to ship it.

nateprewitt (approved):

arielbeckjit:
Hey @SamRemis was this fix? we're experiencing something similar -
https://github.com/boto/botocore/issues/3480

PR #3115, opened 2024-02-08
Support pagination of App Runner ListServices

dgholz:
For aws/aws-cli#8124


SamRemis:
Hi @dgholz,
Thank you for the contribution.  These model files are controlled by upstream services, and if we were to alter them could be overwritten and broken with any given release.  Ideally the app runner service team can make this feature upstream and will vend it to be used by all of our SDKs rather than only boto3 and the CLI. I have opened [a feature request](https://github.com/aws/aws-sdk/issues/704) in the repo where we place issues that impact all SDKs and have made an internal ticket to the team as well.

PR #3123, opened 2024-02-16
Add experimental environment variable to opt out of 100 continue with empty body

SamRemis:
Adds the header `BOTO_EXPERIMENTAL__NO_EMPTY_CONTINUE`; when enabled, we will make a best effort to inspect the size of the body and only add the 100 continue header if the body is non-empty.  

Note: this feature is marked as experimental and the functionality and may be altered or removed in a minor patch without warning.  Users relying on this feature should strictly pin their botocore version.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -346,6 +346,14 @@ def add_expect_header(model, params, **kwargs):
     if 'body' in params:
         body = params['body']
         if hasattr(body, 'read'):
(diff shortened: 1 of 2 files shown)


nateprewitt (approved):
We should note in the PR overview that this is not a GA feature and is subject to abrupt breaking changes.

Otherwise, this looks good, thanks @SamRemis!

Issue #3124, opened 2024-02-22
100-continue handled incorrectly


dbrumley:
### Describe the bug

 From RFC: https://tools.ietf.org/html/rfc7231#section-5.1.1
Requirement for clients:
- A client MUST NOT generate a 100-continue expectation in a request that does not include a message body.

The current code always returns true, even with an empty message body.

### Expected Behavior

Change to self._expect_header_set = headers.get('Content-Length') != '0'

### Current Behavior

botocore will incorrectly generate a 100-continuation.

### Reproduction Steps

Had this bug with minio. I don't have a repro local, but did cite the RFC.

### Possible Solution

I'll submit a PR. It's a simple change.

### SDK version used

develop

### Environment details (OS name and version, etc.)

macos, python 3.9

SamRemis:
Thank you for bringing this to our attention

We have had this issue reported a number of times; one of the boto maintainers gave an in depth explanation of the complexity of the issue [here](https://github.com/boto/boto3/issues/3757#issuecomment-1603242542).  We currently have [an open PR](https://github.com/boto/botocore/pull/3123) for an experimental opt in header to solve this issue.

PR #3125, opened 2024-02-22
resolves issue 3124

dbrumley:
Fix for https://github.com/boto/botocore/issues/3124

We've been maintaining this patch locally in our company for awhile, and just noticed. Figured it's best to get upstream.

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -89,7 +89,11 @@ class AWSConnection:
             headers = {}
         self._response_received = False
         if headers.get('Expect', b'') == b'100-continue':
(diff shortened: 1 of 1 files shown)


SamRemis:
Hi @dbrumley,

Thank you for the contribution.

I just opened up a [PR](https://github.com/boto/botocore/pull/3123) for an experimental opt in header for exactly this; we are hesitant to make this change the default since customers may be relying on the existing behavior which is fully functional for S3.  In the future, this may become the default once we are confident that this is not a breaking change.

The content-length header is also not guaranteed to be present, so making a best-effort check on the length of the body will work for more use cases

PR #3131, opened 2024-03-01
Update setup.py to fix library dateutil issue

m4vazquez:
fix issue with new install of aws-cli using library dateutil 2.9.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ def find_version(*file_paths):
 
 requires = [
     'jmespath>=0.7.1,<2.0.0',
-    'python-dateutil>=2.1,<3.0.0',
+    'python-dateutil>=2.1,<=2.8.2',
     'urllib3>=1.25.4,<1.27 ; python_version < "3.10"',
     'urllib3>=1.25.4,<2.1 ; python_version >= "3.10"',
 ]


nateprewitt:
Hi @m4vazquez 

Thanks the for the PR. As already stated in https://github.com/dateutil/dateutil/issues/1344 and the issue you opened, this is a problem with dependencies for dateutils 2.9.0 which we do not support. Had this pin been moved, even more people would have potentially been broken. Please note we do not support Python versions older than Python 3.8 so this change wouldn't help the specific use case you noted.

Paul has already released a post release with the correct fix from their side. I don't think we intend to update immediately until any other issues have time to surface. I'll close this out since it's not actionable at this time.

m4vazquez:
Hi @nateprewitt, yes I'm using python 3.5.7, but are you tell in me that, we can only user python 3.8 as minimun?
because the dateutil library has a new release?

thanks for the answer

nateprewitt:
Boto3 and the AWS CLI ended support for Python 3.5 in February 2021 as announced [here](https://aws.amazon.com/blogs/developer/announcing-the-end-of-support-for-python-3-4-and-3-5-in-the-aws-sdk-for-python-and-aws-cli-v1/). While older releases should continue to work, all current releases only support Python 3.8+ as noted in our [README](https://github.com/boto/botocore/blob/develop/README.rst) and [PyPI](https://pypi.org/project/botocore/). No updates merged in this repo will be backported to older releases of the library.

The change in dateutil was a mistake which has since been corrected upstream but is unrelated to our codebase. Note future breakages may not be easily fixed, so it may be worth investing in a version of Python still receiving updates. You can find the full list of support on [python.org](https://devguide.python.org/versions/).

Issue #2916, opened 2023-04-25
Add the option to set socket.TCP_KEEPIDLE and socket.TCP_KEEPINTVL


mtoma:
### Describe the feature

With the botocore.config.Config option tcp_keepalive=True the TCP socket is opened with the keep alive socket option (socket.SO_KEEPALIVE) but the interval used for TCP keepalive probes is taken from the system default values in the proc filesystem /proc/sys/net/ipv4/tcp_keepalive_time on Linux.

Changing this value requires the root access which is often not available or is just read-only as in the AWS Fargate containers.

The linux default value for this parameter is 7200 seconds which exceeds by far the AWS VPC timeout of 350 seconds which makes boto3 invoke() call to start a Lambda function in synchronous mode (RequestResponse) loose contact with the Lambda backend ent times out without getting back any response if the Lambda execution time exceeds 350 seconds and is started from a VPC.

### Use Case

It is currently impossible to get an answer from a Lambda function:
- which is started in synchronous mode (InvocationType='RequestResponse')
- inside a VPC (which cuts inactive connections after 350 seconds)
- and with an execution time greater than 350 seconds

The invoke() call times out after read_timeout=XXX seconds because the TCP connection is lost.

I want this call to successfully finish maintaining the TCP connection alive with TCP probes sent with a user defined interval.

### Proposed Solution

The only workaround today is to override the `_compute_socket_options` private method of the  `botocore.args.ClientArgsCreator` class.

The `botocore.config.Config`class should accept an additional parameter (for example tcp_keepalive_time) to allow the user to set the value for (on linux, but is different on OSX and possibly windows)
`(socket.SOL_TCP, socket.TCP_KEEPIDLE, 60), (socket.SOL_TCP, socket.TCP_KEEPINTVL, 60)` parameters.

### Other Information

The current workaround is quite ugly:

# Create a nev version of the botocore private method _compute_socket_options:

```
original_compute_socket_options = botocore.args.ClientArgsCreator._compute_socket_options
def add_aditionnal_tcp_keepalive_options(self, scoped_config, client_config=None):
    print('add_aditionnal_tcp_keepalive_options')
    return original_compute_socket_options(self, scoped_config, client_config) + [(socket.SOL_TCP, socket.TCP_KEEPIDLE, 60), (socket.SOL_TCP, socket.TCP_KEEPINTVL, 60)]
```

# Override the class method before the invoke call:

```
[...]

PR #3140, opened 2024-03-12
fix: tcp_keepalive socket

ShaneNolan:
When using the botocore.config.Config option tcp_keepalive=True, the TCP socket is configured with the keep alive socket option (`socket.SO_KEEPALIVE`). By default, Linux sets the TCP keepalive time parameter to 7200 seconds, which exceeds the AWS NAT Gateway default timeout of 350 seconds [[source](https://docs.aws.amazon.com/vpc/latest/userguide/nat-gateway-troubleshooting.html#nat-gateway-troubleshooting-timeout)].

This limitation leads to an inability to receive a response from a Lambda function under the following conditions:

- The Lambda function is invoked in synchronous mode (InvocationType='RequestResponse').
- The invocation occurs within VPC where a NAT gateway is required to access the internet from a private subnet.
- The execution time of the Lambda function exceeds 350 seconds.

Therefore, by configuring `socket.TCP_KEEPIDLE`, `socket.TCP_KEEPINTVL` and `socket.TCP_KEEPCNT` when `tcp_keepalive` during the `_compute_socket_options` function call we can overcome this limitation. 

`socket.IPPROTO_TCP` is used to support cross platform compatibility. 

The code submitted automatically calculates these values based on the read timeout. Another option would be to have supplied in the scope/client object.

Fixes issues: https://github.com/boto/boto3/issues/2424, https://github.com/boto/boto3/issues/2510 and https://github.com/boto/botocore/issues/2916.

Fargate recently had a similar solution implemented to support this use case: https://aws.amazon.com/blogs/containers/announcing-additional-linux-controls-for-amazon-ecs-tasks-on-aws-fargate/.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -548,37 +548,54 @@ class ClientArgsCreator:
             scoped_config.get("tcp_keepalive", False)
         )
         # Enables TCP Keepalive if specified in client config object or shared config file.
-        if not client_keepalive and not scoped_keepalive:
-            return socket_options
+        if client_keepalive or scoped_keepalive:
+            socket_options.append((socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1))
+        return socket_options
 
-        seconds_in_a_minute = 60
+    def _compute_socket_options(self, scoped_config, client_config=None):
+        # This disables Nagle's algorithm and is the default socket options in urllib3.
+        socket_options = [(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]
 
-        client_read_timeout = client_config and client_config.read_timeout
-        scoped_read_timeout = scoped_config.get("read_timeout", None)
-
-        read_timeout = (
-            scoped_read_timeout if scoped_read_timeout else client_read_timeout
-        ) or seconds_in_a_minute
-
-        maximum_keepalive_probes = int(read_timeout / seconds_in_a_minute) or 1
-        keep_idle = (
-            seconds_in_a_minute
-            if read_timeout > seconds_in_a_minute
-            else read_timeout
+        # Enables TCP Keepalive if specified in client config object or shared config file.
+        client_keepalive = client_config and client_config.tcp_keepalive
+        scoped_keepalive = scoped_config and self._ensure_boolean(
+            scoped_config.get("tcp_keepalive", False)
         )
+        if client_keepalive or scoped_keepalive:
+            socket_options.append((socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1))
 
-        socket_options.extend(
-            [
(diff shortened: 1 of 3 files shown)


adammcdonagh:
This is also impacting me. Unfortunately we are invoking Lambda from ECS via AWS Batch, which doesn't support adding these new options in the task definition yet.

smasa1112:
This issue is same for me.
In my case, lambda connection is read_timeout when ec2 by codebuild try to connect lambda.
It is OK, when lambda sleep 300sec but read_timeout is occured when lambda sleep 450 sec.

pankajastro:
Hi Maintainers, I’ve tested these changes in my scenario, and they seem promising. I’m curious if there’s anything else needed for the review and merging of this PR. I would greatly appreciate any feedback and updates on its progress. Thank you!

spawn-guy:
bump. any movement on this PR? my 200s lambda sync invocations are constantly failing with `botocore.ReadTimeoutError` on `Amazon Linux 2023`

rodrigofp-possiblefinance:
Bump. Any new?
I'm facing the same issue with a Lambda function invoked at Airflow through LambdaInvokeFunctionOperator

rawwar:
just tagging active contributor's on the repo to get some attention:
@alexgromero , @nateprewitt , @ubaskota

ShaneNolan:
one year anniversary for this pr; I still need to use this work around.

also tagging active contributors:
@ubaskota @nateprewitt @arandito @SamRemis

MartinBlanchard3012:
Bumping this issue too, as I'm having the same problems with Lambdas that take more than 350 seconds.

paperlinguist:
We are also facing this issue and it's impacting our production pipelines.

SamRemis:
While this is definitely worth addressing, these new defaults are significantly different from the old behavior and this would apply to every customer who has opted in to TCP keepalive.  Merging this could break existing customer workflows for users who are relying on the current default configurations, and it wouldn't give them any ability to opt out back into the old behavior.

To preserve backwards compatibility, I'd be more in favor of making this an opt in client level configuration - the other solution proposed in the description of this PR.  I will bring this up to the botocore team to get some more thoughts and see where others stand.

xaerto:
Also facing same issue, when  waiting reply from lambda, triggered by MWAA based DAG.    Applying proposed fix has resolved issue

ShaneNolan:
Hey @SamRemis, I've refactored the code to preserve backwards compatibility and make it configuration based. ☕

kwikadi:
Bumping since this is affecting me too. @SamRemis thoughts on merging this now? Seems like the fix is backwards compatible now

paulrinaldi-juno:
Can we get this merged?

PR #3144, opened 2024-03-19
Make bucket name injection for S3Express more generic

SamRemis:
Abstracts the s3 Express bucket name from the context to be generically usable as well as adding additional keys to be stored


nateprewitt (approved):
Minor note, but otherwise this seems reasonable. Thanks @SamRemis!

PR #3145, opened 2024-03-20
Remove datetime.datetime.utcnow

azliu0:
resolves #3889


adarshraj007:
Is this issue resolved?   We are still getting when we upgrade python to 3.12
DeprecationWarning: datetime.datetime.utcnow() is dep
recated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    datetime_now = datetime.datetime.utcnow()

botocore/auth.py
```diff
@@ -416,7 +416,7 @@ def signature(self, string_to_sign, request):
     def add_auth(self, request):
         if self.credentials is None:
             raise NoCredentialsError()
-        datetime_now = datetime.datetime.utcnow()
+        datetime_now = datetime.datetime.now(datetime.timezone.utc)
```
kurtmckee, line 419: The minimal change here is:

```suggestion
        datetime_now = datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)
```

The code as it currently stands makes `datetime_now` timezone-aware. The change I'm suggesting above resolves the `DeprecationWarning` without fundamentally altering `datetime_now` in the process.

You can see how `datetime_now` was affected by looking in the test suite, where all of the existing tests had to be modified to suddenly be timezone-aware.

botocore/crt/auth.py
```diff
@@ -54,9 +54,7 @@ def add_auth(self, request):
         if self.credentials is None:
             raise NoCredentialsError()
 
-        # Use utcnow() because that's what gets mocked by tests, but set
-        # timezone because CRT assumes naive datetime is local time.
-        datetime_now = datetime.datetime.utcnow().replace(
+        datetime_now = datetime.datetime.now(datetime.timezone.utc).replace(
             tzinfo=datetime.timezone.utc
         )
```
kurtmckee, line 59: No need to call `.replace()` here anymore; it's already set.

```suggestion
        datetime_now = datetime.datetime.now(datetime.timezone.utc)
```

botocore/crt/auth.py
```diff
@@ -251,9 +249,7 @@ def add_auth(self, request):
...
```
kurtmckee, line 254: Same as above.

```suggestion
        datetime_now = datetime.datetime.now(datetime.timezone.utc)
```

botocore/endpoint.py
```diff
@@ -152,10 +152,10 @@ def prepare_request(self, request):
     def _calculate_ttl(
         self, response_received_timestamp, date_header, read_timeout
     ):
-        local_timestamp = datetime.datetime.utcnow()
+        local_timestamp = datetime.datetime.now(datetime.timezone.utc)
```
kurtmckee, line 155: As noted above, this can be simplified to avoid fundamentally changing `local_timestamp`:

```suggestion
        local_timestamp = datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)
```

kurtmckee (commented):
_I am not a botocore contributor but I want to see this get merged!_

This PR appears to resolve the `DeprecationWarning`s that are giving me heartburn, but it fundamentally alters the `datetime` instances, causing ripples throughout the code and test suite.

The correct, minimal swap is:

```diff
- datetime.datetime.utcnow()
+ datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)
```

I've left comments, but not at every location to be updated.

kurtmckee:
I've submitted an alternative, #3239, which makes significantly fewer changes to accomplish the same thing, and passes CI.

nateprewitt:

PR #3146, opened 2024-03-22
Per-request credential override

SamRemis:
Add ability to use credentials from the context to override credentials used for a given request

### DIFF
--- a/tests/functional/test_credentials.py
+++ b/tests/functional/test_credentials.py
@@ -1143,3 +1143,28 @@ class SSOSessionTest(BaseEnvVar):
             }
         }
         stubber.add_response(body=json.dumps(response).encode('utf-8'))
+
+
+class TestContextCredentials(unittest.TestCase):
+    ACCESS_KEY = "access-key"
+    SECRET_KEY = "secret-key"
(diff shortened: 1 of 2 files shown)


tests/functional/test_credentials.py
```diff
@@ -1143,3 +1143,28 @@ def add_credential_response(self, stubber):
...
+        assert self.ACCESS_KEY in request.headers.get('Authorization')
+
+    def test_credential_context_override(self):
+        session = StubbedSession()
+        with SessionHTTPStubber(session) as stubber:
+            s3 = session.create_client('s3')
+            s3.meta.events.register('before-sign', self._add_fake_creds)
+            s3.meta.events.register(
+                'request-created', self._verify_bucket_and_key_in_context
+            )
```
nateprewitt, line 1168: Instead of adding our test as an event, can we make an assertion on the end request instead? That should be possible to check with the stubber while avoiding ambiguity and regressions in the event system.

nateprewitt (changes requested):

nateprewitt (approved):

PR #3161, opened 2024-04-18
Add flags to CI test script

SamRemis:
Adds a marker to tests which allows us to run only tests that validate models


nateprewitt (approved):
:shipit:

PR #3177, opened 2024-05-14
Fix s3 `put_object` example

tim-finnigan:
Fixes https://github.com/boto/boto3/issues/4126


SamRemis:
Yes, I'd like to see these examples be reworked if we have the bandwidth- would you be okay making those changes and updating this PR @tim-finnigan ?

PR #3178, opened 2024-05-15
Fix calculation of truncated binary exponential backoff

jamesls:
The intent of the jitter in the exponential backoff calculation is to create a uniform random distribution.  However, the current calculation applies the truncation after the randomization.  This results in clamped/truncated distribution which is not what we want.

This only affects cases where the calculated backoff exceeds the `MAX_BACKOFF`, and is more pronounced the more you exceed `MAX_BACKOFF`. The current default of 3 max attempts is well below this value so this change has no effect on the default behavior.  A user would have to have a max attempts of at least 6 before this started to affect backoff times.

As an example, this is a distribution based on 10,000 events with `attempt=10` (x-axis is backoff time in seconds):

current

And here's the distribution with 10,000 events with `attempt=10` with the changes in this PR:

fixed

### DIFF
--- a/.changes/next-release/bugfix-retries-19921.json
+++ b/.changes/next-release/bugfix-retries-19921.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "retries",
+  "description": "Fix backoff calculation for truncated binary exponential backoff (`#3178 <https://github.com/boto/botocore/issues/3178>`__)"
+}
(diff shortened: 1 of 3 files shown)


PPenglis:
Is there a chance that the back off is very close to 0 each time? Depending on how random the random number is and how many retries. 
Surely it would be better to create jitter with non-zero values for random such as rand(0.1,1)

I assume the initial intent would have been for rand(0,1) + 2 ** attempt however this was likely changed to deal with small values of MAX_BACKOFF.

expoential backoff would likely to fix up flow control, not sure a very short back off makes sense and cause errors that are hard to debug.

PR #3206, opened 2024-06-14
Ruff update

nateprewitt:
This is a follow up to boto/boto3#4161 to move our repos to use [`ruff`](https://github.com/astral-sh/ruff) in favor of our existing `pre-commit` cocktail. For botocore, we have some unused imports that have been left in place due to third-party packages importing from modules importing from other modules. `ruff` does not currently supply a way to comprehensively skip specific lines, only disabling sorting for entire import blocks. Because of this we've shut off sorting in `ruff` and will keep `isort` in place for now.

This should be largely inline with our existing configuration with some minor changes to fix f-string upgrades missed with our previous setup and some newline fixes. This PR keeps our existing quote preservation behavior until we get all projects moved to `ruff`. At that point, we can look at turning it to "double".


jonathan343 (commented):
Looks good, just had two questions:
* We should add a `.git-blame-ignore-revs` entry for the related commits like we did for boto3: https://github.com/boto/boto3/pull/4168
*  I think we're making logging slightly inefficient by migrating onto fstrings. I'm not strongly against this since performance hit should be insignificant, but wanted to make sure this is a conscious change.

nateprewitt:
> We should add a .git-blame-ignore-revs entry for the related commits like we did for boto3

I usually save these for follow up PRs since we have to change the commit every time we force push to this branch and if this gets squash merged we have to redo the hash as well. Those have been messed up enough that just keeping it as a post-release step has been most effective.

> I think we're making logging slightly inefficient by migrating onto fstrings. I'm not strongly against this since performance hit should be insignificant, but wanted to make sure this is a conscious change.

Hmm, good shout. So this is interesting, Jonas pointed out a minor perf improvement for using `logger.debug("message %s", value)` over the f-strings which is why we've gone that direction in net-new code. Almost everything this is modifying is turning code in the format `logger.debug("message %s" % value)` into f-strings which _is_ more performant (note `%` vs `,`). There are two outliers though:

* https://github.com/boto/botocore/pull/3206/files#diff-93392e4565eff0b6a2f3cb5a9f0e7fbe1e8855fc9d189ae6ed954fd012d65811R993
* https://github.com/boto/botocore/pull/3206/files#diff-93392e4565eff0b6a2f3cb5a9f0e7fbe1e8855fc9d189ae6ed954fd012d65811R1001

I don't actually know why these got reformatted, but lines like this didn't:
* https://github.com/boto/botocore/pull/3206/files#diff-c91f38d9d6f1217097937507bc104cce7b60db41414e823b06d0a812cfcb2cd3R311-R315

I can take a look at trying to manually move the first two lines back to the older format. Rewriting everything to the comma format by hand is likely not worth the minor performance difference imo.

> Is there a reason for keeping isort?

I added a blip about that in the overview of the PR. The main issue is ruff does not have a way to ignore specific import lines only blocks. We have to keep isort for backwards compatibility at the moment.

jonathan343 (approved):
Cool, thanks for the insight. I don't think that issue is worth blocking this PR. Approving

SamRemis:
Can we add updates to botocore/CONTRIBUTING.rst?

nateprewitt:
Yep, there are still a number of cleanup actions needed. The PR is still in draft to be added to as we go. I'll get everything finalized before putting it up for review.

jonathan343 (approved):
🚀

PR #3208, opened 2024-06-19
Add a lock file for requirements-docs.txt

jonathan343:
This PR uses the `pip-compile` command from `pip-tools` to generate a lock file for `requirments-docs.txt`. This was generated in `Python 3.8` virtual environment.

**NOTE:**
The following urllib3 requirements were added to `requirements-docs.txt` to force the same version constraints as botocore when generating the `requirements-docs-lock.txt`:
```
urllib3>=1.25.4,<1.27 ; python_version < "3.10"
urllib3>=1.25.4,!=2.2.0,<3 ; python_version >= "3.10"
```
This specifically becomes an issue when using Python 3.8 since `Sphinx` has a transitive dependency on the latest version of urllib3 and botocore will require `urllib3>=1.25.4,<1.27`.


docs/README.md
```diff
@@ -8,7 +8,7 @@ Sphinx is used for documentation. You can generate HTML locally with the
 following:
 
 ```
-$ pip install -r requirements-docs.txt
+$ pip install -r requirements-docs-lock.txt
```
nateprewitt, line 11: We can probably leave this change out to not overly constrain users when building the docs.

nateprewitt (changes requested):

nateprewitt (approved):

requirements-docs.txt
```diff
@@ -2,3 +2,6 @@ sphinx==5.3.0
 furo==2022.12.7
 sphinx_copybutton==0.5.1
 sphinx-remove-toctrees==0.0.3
+# Matches the urllib3 requirements for botocore.
+urllib3>=1.25.4,<1.27 ; python_version < "3.10"
+urllib3>=1.25.4,!=2.2.0,<3 ; python_version >= "3.10"
```
nateprewitt, line 7: If we're going to add these, I think we may want to simplify it to the strict requirements. Otherwise, we may end up with odd behaviors as these change and I'd rather we not have to worry about keeping this in-sync in more places.

```suggestion
# Avoid urllib3 2.x below Python 3.10
urllib3<2.0 ; python_version < "3.10"
```

nateprewitt (approved):

PR #3227, opened 2024-07-26
Update Protocol Tests

jonathan343:
## Context

This PR replaces the existing protocol tests in `botocore/tests/unit/protocols/...` with new tests generated from [Smithy protocol test models](https://github.com/smithy-lang/smithy/tree/main/smithy-aws-protocol-tests/model). We use a version of these Smithy tests that are converted to the format currently supported by our existing test runner (`test_protocols.py`).

## Issues

Many of the new protocol tests are failing due to test runner, serialization, and parsing issues. I've highlighted the notable issues below to provide additional context for reviwers:

### Test Runner (`test_protocols.py`)

* **Response Body Normalization** - The new protocol tests define expected body values with JSON and XML that includes extra white space and newlines. This prohibits us from continuing to do direct assertions between expected and actual values. Instead, we normalize the expected body value by attempting to parse the content as a JSON or XML object based on the protocol.
* **Handle Special Float Types** - In the protocol test suite, certain special float types are represented as strings: `"Infinity"`, `"-Infinity"`, and `"NaN"`. However, we parse these values as actual floats types, so we need to convert them back to their string representation before comparing with the expected values.

### Input Serialization (`serialize.py`)

...

### Response Parsing (`parsers.py`)
* **Infer Root XML Node** - We should be trying to infer the following root nodes when parsing responses.
  * **ec2** - Serializes XML responses within an XML root node with the name of the operation's output suffixed with "Response" such as `<OperationNameResponse>`. ([see more](https://smithy.io/2.0/aws/protocols/aws-ec2-query-protocol.html#response-serialization))
  * **query** - Serializes a nested element with the name of the operation's output suffixed with "Result" such as `<OperationNameResult>`. ([see more](https://smithy.io/2.0/aws/protocols/aws-query-protocol.html#response-serialization))
 * **JSON Error Parsing** - Parse errors as described in [Operation error serialization](https://smithy.io/2.0/aws/protocols/aws-restjson1-protocol.html#operation-error-serialization) for ALL json-based protocols. This behavior currently only exists for the rest-json protocol and doesn't handle `:` characters. 
 * **JSON Parse Header Values** - This PR includes updates to the rest-json parser to handle boolean, float, and double values when represented as strings in a header.
...

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -124,6 +124,7 @@ import re
 from botocore.compat import ETree, XMLParseError
 from botocore.eventstream import EventStream, NoInitialResponseError
 from botocore.utils import (
+    ensure_boolean,
     is_json_value_header,
     lowercase_dict,
     merge_dicts,
@@ -204,7 +205,7 @@ class ResponseParser:
     # This is a list of known values for the "location" key in the
     # serialization dict. The location key tells us where in the response
     # to parse the value.
-    KNOWN_LOCATIONS = ['statusCode', 'header', 'headers']
+    KNOWN_LOCATIONS = ('statusCode', 'header', 'headers')
 
     def __init__(self, timestamp_parser=None, blob_parser=None):
         if timestamp_parser is None:
@@ -437,11 +438,10 @@ class BaseXMLResponseParser(ResponseParser):
             member_shape = members[member_name]
             location = member_shape.serialization.get('location')
             if (
-                location
-                and location in self.KNOWN_LOCATIONS
+                location in self.KNOWN_LOCATIONS
                 or member_shape.serialization.get('eventheader')
             ):
-                # All members with know locations have already been handled,
+                # All members with known locations have already been handled,
                 # so we don't need to parse these members.
                 continue
             xml_name = self._member_key_name(member_shape, member_name)
@@ -551,6 +551,8 @@ class BaseXMLResponseParser(ResponseParser):
 
 
 class QueryParser(BaseXMLResponseParser):
+    ROOT_NODE_SUFFIX = 'Result'
+
     def _do_error_parse(self, response, shape):
         xml_contents = response['body']
         root = self._parse_xml_string_to_dom(xml_contents)
@@ -589,9 +591,8 @@ class QueryParser(BaseXMLResponseParser):
                 operation_name = response.get("context", {}).get(
                     "operation_name", ""
                 )
-                inferred_wrapper_name = operation_name + "Result"
                 inferred_wrapper = self._find_result_wrapped_shape(
-                    inferred_wrapper_name, root
+                    f"{operation_name}{self.ROOT_NODE_SUFFIX}", root
                 )
                 if inferred_wrapper is not None:
                     start = inferred_wrapper
@@ -615,6 +616,8 @@ class QueryParser(BaseXMLResponseParser):
 
 
 class EC2QueryParser(QueryParser):
+    ROOT_NODE_SUFFIX = 'Response'
+
     def _inject_response_metadata(self, node, inject_into):
         mapping = self._build_name_to_xml_node(node)
         child_node = mapping.get('requestId')
@@ -979,7 +982,7 @@ class BaseRestParser(ResponseParser):
         for name in member_shapes:
             member_shape = member_shapes[name]
             location = member_shape.serialization.get('location')
-            if location is None or location not in self.KNOWN_LOCATIONS:
+            if location is None:
                 continue
             elif location == 'statusCode':
                 final_parsed[name] = self._parse_shape(
@@ -1052,13 +1055,7 @@ class RestJSONParser(BaseRestParser, BaseJSONParser):
         return self._parse_body_as_json(body_contents)
 
     def _handle_boolean(self, shape, value):
-        # It's possible to receive a boolean as a string
-        if isinstance(value, str):
-            if value == 'true':
-                return True
-            else:
-                return False
-        return value
+        return ensure_boolean(value)
 
     def _handle_integer(self, shape, value):
         return int(value)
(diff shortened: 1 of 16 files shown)


botocore/parsers.py
```diff
@@ -1011,28 +1051,23 @@ class RestJSONParser(BaseRestParser, BaseJSONParser):
...
-            error['Error']['Code'] = code
-        elif 'code' in body or 'Code' in body:
-            error['Error']['Code'] = body.get('code', body.get('Code', ''))
+    def _handle_boolean(self, shape, value):
+        # It's possible to receive a boolean as a string
+        if isinstance(value, str):
+            if value == 'true':
+                return True
+            else:
+                return False
```
nateprewitt, line 1060: This seems potentially breaking. Why are we only accepting the exact value `true` from a string? Is `True` in a response indicative of False? We have a function for normalization [here](https://github.com/boto/botocore/blob/547b20801770c8ea4255ee9c3b809fea6b9f6bc4/botocore/utils.py#L193-L203), which we might consider. I'm a little concerned about this being impactful for both the Stubber and any third-party service emulators.

tests/unit/test_protocols.py
```diff
@@ -98,7 +99,12 @@
     'rest-json': RestJSONParser,
     'rest-xml': RestXMLParser,
 }
-PROTOCOL_TEST_BLACKLIST = ['Idempotency token auto fill']
+PROTOCOL_TEST_BLACKLIST = [
+    # These cases test functionality outside the serializers and parsers.
+    "Test cases for QueryIdempotencyTokenAutoFill operation",
+    "Test cases for PutWithContentEncoding operation",
+    "Test cases for HttpChecksumRequired operation",
+]
```
nateprewitt, line 107: We'll want to update usage elsewhere when this is used.

```suggestion
PROTOCOL_TEST_BLOCKLIST = (
    # These cases test functionality outside the serializers and parsers.
    "Test cases for QueryIdempotencyTokenAutoFill operation",
    "Test cases for PutWithContentEncoding operation",
    "Test cases for HttpChecksumRequired operation",
)
```

tests/unit/test_protocols.py
```diff
@@ -394,11 +430,32 @@ def _serialize_request_description(request_dict):
...
 
-def _assert_requests_equal(actual, expected):
-    assert_equal(
-        actual['body'], expected.get('body', '').encode('utf-8'), 'Body value'
-    )
+def _assert_requests_equal(actual, expected, protocol_type):
+    expected_body = expected.get('body', '').encode('utf-8')
+    actual_body = actual['body']
+    # The expected bodies in our consumed protocol tests have extra
+    # whitespace and newlines that need to handled. We need to normalize
```
nateprewitt, line 437:
```suggestion
    # whitespace and newlines that need to be handled. We need to normalize
```

tests/unit/test_protocols.py
```diff
@@ -394,11 +430,32 @@ def _serialize_request_description(request_dict):
...
+            assert_equal(tree1, tree2, 'Body value')
+        else:
+            assert_equal(actual_body, expected_body, 'Body value')
+    except (json.JSONDecodeError, ET.ParseError):
+        assert_equal(actual_body, expected_body, 'Body value')
+
     actual_headers = HeadersDict(actual['headers'])
+    if protocol_type in ['query', 'ec2']:
+        if expected.get('headers', {}).get('Content-Type'):
+            expected['headers']['Content-Type'] += '; charset=utf-8'
```
nateprewitt, line 458: Why are we only getting charset information with `query`/`ec2`?

tests/unit/test_protocols.py
```diff
@@ -394,11 +430,32 @@ def _serialize_request_description(request_dict):
...
+                json.loads(actual_body),
+                json.loads(expected_body),
+                'Body value',
+            )
+        elif protocol_type in ['rest-xml']:
+            tree1 = ET.canonicalize(actual_body, strip_text=True)
+            tree2 = ET.canonicalize(expected_body, strip_text=True)
+            assert_equal(tree1, tree2, 'Body value')
+        else:
+            assert_equal(actual_body, expected_body, 'Body value')
```
nateprewitt, line 451: We should consider making these individual functions instead of overloading this one with a bunch of conditional logic.

tests/unit/test_protocols.py
```diff
@@ -394,11 +430,32 @@ def _serialize_request_description(request_dict):
...
+    except (json.JSONDecodeError, ET.ParseError):
+        assert_equal(actual_body, expected_body, 'Body value')
```
nateprewitt, line 453: Do we have test cases that are giving us back bad bodies that we expect to work? This seems like an anti-pattern because we're potentially letting breakages through the test suite with this.

tests/unit/test_protocols.py
```diff
@@ -279,32 +292,55 @@ def _fixup_parsed_result(parsed):
...
+
+def _convert_bytes_to_str(parsed):
+    return _convert(parsed, [(bytes, _bytes_to_str)])
+
+
+def _special_floats_to_str(value):
+    if isinstance(value, float):
+        if value in [float('Infinity'), float('-Infinity')] or math.isnan(
+            value
+        ):
```
nateprewitt, line 329: Is there something unique about `NaN` that we don't use `float('NaN')` for the check like we do inf?

nateprewitt (changes requested):
Looks like a good start! I left some initial feedback to make sure we're not regressing things with these changes. Ideally we should be able to tie code blocks to individual test(s) changes.

botocore/parsers.py
jonathan343: Updating this to use the `ensure_boolean`. Was the concern with only accepting `true` and not `True`? Or with the change as a whole?

jonathan343 (commented):
Replied/resolved some of the comments and suggestions. Still looking deeper into a few of them but keeping changes local for now because I think GitHub will hide the unresolved comments and make it harder to keep track of them if I push new changes.

jonathan343:
Closing this PR in favor of the following:
* https://github.com/boto/botocore/pull/3247
* https://github.com/boto/botocore/pull/3378

PR #3232, opened 2024-08-06
s3 expires implementation

alexgromero:
This PR customizes the S3 `Expires` shape in preparation for model changes. Instead of throwing an error, operations will now complete with fallback behavior and a warning.
* Ensures the `Expires` header remains a timestamp for backwards compatibility.
* Omits the `Expires` key in the response if parsing fails and logs a warning.

Operations with `Expires` in their output shapes will:
* Add a synthetic member called `ExpiresString` to retain the raw value of `Expires`.
* Deprecate `Expires` and encourage users to use the new field `ExpiresString` through documentation.


botocore/handlers.py
```diff
@@ -596,6 +596,38 @@ def parse_get_bucket_location(parsed, http_response, **kwargs):
...
+        doc_section.write(
+            '*This member has been deprecated*. Please use ``ExpiresString`` instead.'
+        )
+        param_section.add_new_section('ExpiresString')
+        new_param = param_section.get_section('ExpiresString')
+        new_param.style.start_li()
+        new_param.write('**ExpiresString** (*string*) --')
+        new_param.style.end_li()
+        new_param.style.new_line()
+        new_param.write('\tThe raw unparsed value of the ``Expires`` field.')
```
SamRemis, line 628: Nit: raw and unparsed usually has a comma between it in my experience.

tests/functional/test_s3.py
```diff
@@ -1340,6 +1340,79 @@ def test_500_error_with_non_xml_body(self):
             self.assertEqual(len(http_stubber.requests), 2)
 
 
+class TestS3Parser(BaseS3OperationTest):
+    def get_operations_with_expires(self):
+        s3 = self.session.create_client("s3")
+        service_model = s3.meta.service_model
+        operations_with_expires = []
+        operations = service_model.operation_names
+        for operation in operations:
```
SamRemis, line 1349: I'd rather just pick one where we know it is and use that than iterate through and test each operation.

If something goes wrong here, then we may wind up having this return an empty list, and these couple of tests will not make any assertions and just pass.  Also, what do we gain from testing this multiple times on multiple operations?

tests/functional/test_s3.py
```diff
@@ -1340,6 +1340,79 @@ def test_500_error_with_non_xml_body(self):
...
+
+    def test_valid_expires_value_in_response(self):
+        operations_with_expires = self.get_operations_with_expires()
+        expires_value = "Thu, 01 Jan 1970 00:00:00 GMT"
+        for operation in operations_with_expires:
+            mock_headers = {'expires': expires_value}
+            s3 = self.session.create_client("s3")
+            with ClientHTTPStubber(s3) as http_stubber:
+                http_stubber.add_response(headers=mock_headers)
+                try:
```
SamRemis, line 1364: What is the purpose of this try/except?  Since we are stubbing this, shouldn't it always be passing?  And if it does fail with an error, it will fail the test anyway.

tests/functional/test_s3.py
```diff
@@ -1340,6 +1340,79 @@ def test_500_error_with_non_xml_body(self):
...
+                    response = getattr(s3, xform_name(operation))(
+                        Bucket="mybucket", Key="mykey"
+                    )
+                except ClientError as e:
+                    self.fail(f"Operation {operation} failed with error: {e}")
+                headers = response['ResponseMetadata']['HTTPHeaders']
+                self.assertIn(
```
SamRemis, line 1371: What is this check doing?  Won't it always be there since you set it?  Isn't this effectively tested below with the other assertions?

tests/unit/test_parsers.py
```diff
@@ -1622,6 +1622,97 @@ def test_can_parse_route53_with_missing_message(self):
...
+            'ResponseMetadata': {
+                'HTTPStatusCode': 200,
+                'HTTPHeaders': {'expires': expires_value},
+            },
+            'Expires': datetime.datetime(2015, 1, 1, tzinfo=tzutc()),
+            'ExpiresString': expires_value,
+        }
+        self.assertEqual(parsed, expected)
+
+    def test_invalid_expires_response_parsed(self):
```
jonathan343, line 1663: We should be testing all these known invalid timestamp values:
```
2034-01-01T00:00:00Z
2023-11-02 08:43:04 -0400
Invalid Date
access plus 1 month
2023-10-12 23:00:00 UTC
Thu, 01 Jan 1970 00:00:00 UTC
Tue, 15 Apr 2014 07:17:09 UTC
10/21/2018
Expires: Thu, 9 Sep 2013 14:19:41 GMT
01 dec 2100
Sun, 31 Aug 2031 20:56:53 GMT+00:00
Sun, 22 Oct 23 00:45:02 UTC
access plus 1 month
{ts '2023-10-10 09:27:14'}
Tue, 01 Jan 2014 12:00:00 GMT
Sat, 10 Jan 2049 00:00:00 GMT
```

botocore/handlers.py
```diff
@@ -596,6 +596,38 @@ def parse_get_bucket_location(parsed, http_response, **kwargs):
     parsed['LocationConstraint'] = region
 
 
+def document_s3_expires_shape(section, event_name, **kwargs):
```
jonathan343, line 599: Can you include an example for what each of these gets rendered as?

botocore/parsers.py
```diff
@@ -1113,8 +1113,39 @@ def _handle_string(self, shape, text):
...
+        if shape.name == 'Expires':
+            handler = getattr(self, '_handle_expires_timestamp')
+            return handler(shape, node)
+        return super()._parse_shape(shape, node)
+
+    def _handle_expires_timestamp(self, shape, timestamp):
+        try:
+            return self._timestamp_parser(timestamp)
+        except ValueError as e:
+            LOG.warning(
```
jonathan343, line 1139: nit - I think we can make this a bit more clear. Something like:
```
f'Failed to parse the "Expires" member as a timestamp: {e}. '
f'The unparsed value is available in the response under "ExpiresString"'
```

botocore/data/s3/2006-03-01/service-2.sdk-extras.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": 1.0,
+  "merge": {
+    "metadata":{
+      "protocol":"s3"
```
jonathan343, line 5: There currently also exists a `"protocols":["rest-xml"],` value in the s3 service model. We should probably update this too. I don't think it's used anywhere yet though.

tests/unit/test_parsers.py
alexgromero: The thing is not all of these are considered invalid in the SDK. As long as it's in a timestamp format, our parsers are able to parse them as such. For example, `Tue, 01 Jan 2014 12:00:00 GMT` is "invalid" in this list because 01/01/14 was on a Wednesday instead of a Tuesday.
However, I can add the ones which are not parsable as a timestamp to the test.

botocore/handlers.py
alexgromero: For the 'response-example' section it would look like this:
syntax
For the 'response-params' section it would look like this:
params

tests/functional/test_s3.py
```diff
@@ -1340,6 +1340,46 @@ def test_500_error_with_non_xml_body(self):
...
+        mock_headers = {'expires': expires_value}
+        s3 = self.session.create_client("s3")
+        with ClientHTTPStubber(s3) as http_stubber:
+            http_stubber.add_response(headers=mock_headers)
+            response = s3.get_object(Bucket='mybucket', Key='mykey')
+            self.assertIn('Expires', response)
+            self.assertIsInstance(response['Expires'], datetime.datetime)
+            self.assertIn('ExpiresString', response)
+            self.assertIsInstance(response['ExpiresString'], str)
+            self.assertEqual(response['ExpiresString'], expires_value)
```
SamRemis, line 1355: An extra assertion isn't blocking, but the `isInstance` calls are redundant with the `assertEqual` calls.  - this goes for many of these tests

alexgromero: You're right. I removed those.

botocore/data/s3/2006-03-01/service-2.sdk-extras.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": 1.0,
+  "merge": {
+    "metadata":{
+      "protocol":"s3",
+      "protocols":["s3"]
```
nateprewitt, line 6: This piece seems a little concerning. We can probably assume this will never expand for a service like s3, but how do we know if it does? We may want some form of validation on the base model via a test.

tests/functional/test_s3.py
```diff
@@ -1340,6 +1341,41 @@ def test_500_error_with_non_xml_body(self):
...
+            with ClientHTTPStubber(s3) as http_stubber:
+                http_stubber.add_response(headers=mock_headers)
+                response = s3.get_object(Bucket='mybucket', Key='mykey')
+                self.assertNotIn('Expires', response)
+                self.assertIn('ExpiresString', response)
+                self.assertEqual(response['ExpiresString'], expires_value)
+                self.assertIn(
+                    'Failed to parse the "Expires" member as a timestamp',
+                    log.output[0],
+                )
```
nateprewitt, line 1375: Can we make this a bit more robust in the event another Warning starts getting raised in this flow? Is it sufficient to check that at least one output in logs contains this warning?

tests/unit/test_parsers.py
```diff
@@ -1622,6 +1622,103 @@ def test_can_parse_route53_with_missing_message(self):
         self.assertEqual(error['Message'], '')
 
 
+class TestS3ExpiresResponses(unittest.TestCase):
```
nateprewitt, line 1625: Ideally we don't want to be adding net-new tests as unittest TestCases. We can use [pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) for what you're doing in the setup section and [`@pytest.mark.parametrize`](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html) to break the timestamps into individual tests.

The reason this is valuable is it allows us to better parallelize tests and it gives us more granular failures. When we test a bunch of use cases in a single test, it becomes harder to determine what is actually failing.

botocore/parsers.py
```diff
@@ -1113,8 +1113,38 @@ def _handle_string(self, shape, text):
...
 
+class S3Parser(RestXMLParser):
+    def parse(self, response, shape):
+        parsed = super().parse(response, shape)
+        if 'Expires' in parsed:
+            parsed['ExpiresString'] = (
+                parsed.get('ResponseMetadata', {})
+                .get('HTTPHeaders', {})
+                .get('expires')
+            )
```
jonathan343, line 1124: I'm concerned about relying on the `expires` header here as opposed to use the original unparsed `Expires` value

jonathan343: To elaborate, S3 currently models "Expires" like:
```
"Expires":{
  "shape":"Expires",
  "documentation":"<p>The date and time at which the object is no longer cacheable.</p>",
  "location":"header",
  "locationName":"Expires"
},
```
If they start sending this value under a new header name ( and update `locationName` from `Expires` to something different), this implementation breaks. I'd be in favor of storing the originally pre-parsed `Expires` value and using that for `ExpiresString` as opposed hard coding an `expires` header lookup.

botocore/parsers.py
```diff
@@ -964,11 +964,14 @@ def _parse_non_payload_attrs(
...
                 )
             elif location == 'header':
-                header_name = member_shape.serialization.get('name', name)
-                if header_name in headers:
-                    final_parsed[name] = self._parse_shape(
-                        member_shape, headers[header_name]
-                    )
+                self._parse_headers(final_parsed, headers, member_shape, name)
+
+    def _parse_headers(self, final_parsed, headers, member_shape, name):
```
jonathan343, line 969: nit: I would use the singular version of this name (i.e. `_parse_header`) since this function only parses a single header.

botocore/parsers.py
```diff
@@ -1113,8 +1116,36 @@ def _handle_string(self, shape, text):
...
 
+class S3Parser(RestXMLParser):
+    def _parse_headers(self, final_parsed, headers, member_shape, name):
+        header_name = member_shape.serialization.get('name', name)
+        if header_name in headers:
+            if header_name == 'Expires':
+                final_parsed[name] = self._handle_expires_timestamp(
+                    member_shape, headers[header_name]
+                )
+                final_parsed['ExpiresString'] = headers[header_name]
```
jonathan343, line 1127: super nit: I'd make this the first line in the conditional for readability. It doesn't rely on the previous statement and is in-between two statements that are related to each other.

tests/functional/test_s3.py
```diff
@@ -2079,6 +2118,22 @@ def test_presign_uses_v2_for_us_east_1_with_us_east_1_regional(self):
         self.assert_is_v2_presigned_url(url)
 
 
+def test_s3_protocols_unchanged():
+    # Test to ensure that the protocols list in the base S3 model remains unchanged.
+    # This is important to verify because updates to the protocol occur in the
```
SamRemis, line 2123: If you're going to be making edits anyway, I may change what this says.  This tells us about what, but not the why.  

We need to verify this so that if S3 ever adds support for a new protocol, it's not getting silently overwritten and we are aware of the change being ignored and can make a decision accordingly.

tests/unit/test_parsers.py
```diff
@@ -1646,3 +1646,92 @@ def test_can_handle_generic_error_message(parser, body):
...
+    assert parsed == expected
+
+
+@pytest.mark.parametrize(
+    "expires",
+    [
+        "Invalid Date",
+        "access plus 1 month",
+        "Expires: Thu, 9 Sep 2013 14:19:41 GMT",
+        "{ts '2023-10-10 09:27:14'}",
```
nateprewitt, line 1714: Let's add another test case here for `-3370280040400`. This offset is going to trigger [the issue](https://github.com/boto/botocore/pull/3232/files#r1729384969) I pointed out above where it's falls within our Overflow error handling but doesn't trigger anything else. That falls through to the `RuntimeError` which will mean we can't create an `ExpiresString` entry.

tests/functional/test_s3.py
```diff
@@ -1340,6 +1343,42 @@ def test_500_error_with_non_xml_body(self):
...
+        with ClientHTTPStubber(s3) as http_stubber:
+            http_stubber.add_response(headers=mock_headers)
+            response = s3.get_object(Bucket='mybucket', Key='mykey')
+            self.assertIn('Expires', response)
+            self.assertEqual(
+                response['Expires'],
+                datetime.datetime(1970, 1, 1, tzinfo=tzutc()),
+            )
+            self.assertIn('ExpiresString', response)
+            self.assertEqual(response['ExpiresString'], expires_value)
```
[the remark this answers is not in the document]
alexgromero: I was originally checking that the request wouldn't be retried, but I realize that is not relevant to this test so I will remove it.

botocore/handlers.py
```diff
@@ -596,6 +596,38 @@ def parse_get_bucket_location(parsed, http_response, **kwargs):
...
+        doc_section.write(
+            '*This member has been deprecated*. Please use ``ExpiresString`` instead.'
+        )
+        param_section.add_new_section('ExpiresString')
+        new_param = param_section.get_section('ExpiresString')
+        new_param.style.start_li()
+        new_param.write('**ExpiresString** (*string*) --')
+        new_param.style.end_li()
+        new_param.style.new_line()
+        new_param.write('\tThe raw, unparsed value of the ``Expires`` field.')
```
nateprewitt, line 628: Did the `\t` here come from somewhere else? I don't think that's actually rendered in the browser, it's going to have the same indentation as the above list won't it?

Actually, I'm also a little confused about the `end_li` before writing the last line. Looking at the live HTML, it looks like both values are within paragraph tags inside a single list element.

```html
<li>
<p><strong>Expires</strong> <em>(datetime) –</em></p>
<p>The date and time at which the object is no longer cacheable.</p>
</li>
```

Am I incorrectly conflating concepts? Do we have a sample of the actual HTML being generated?

SamRemis: https://github.com/boto/botocore/pull/3232#discussion_r1725190777

nateprewitt: Sure, I've read the comment thread previously. What I'm asking about is the specific mechanics of the code. What's written doesn't make sense conceptually. Leading whitespace is not actually used in HTML, so there must be some other mechanism that is enforcing that tab before we render the page.

SamRemis: Sure, I pulled this down and generated it real quick.  
With the `\t`: 
```
<ul>
<li><p><strong>ExpiresString</strong> (<em>string</em>) –</p>
<p>The raw, unparsed value of the <code class="docutils literal notranslate"><span class="pre">Expires</span></code> field.</p>
</li>
</ul>
```
without:
```
<ul class="simple">
<li><p><strong>ExpiresString</strong> (<em>string</em>) –</p></li>
</ul>
<p>The raw, unparsed value of the <code class="docutils literal notranslate"><span class="pre">Expires</span></code> field.</p>
```

That winds up being rendered out of alignment like this: 
image

Looking at the full rendered HTML of the version with the `\t`, the list of parameters is kind of weirdly split into three unordered lists that don't have a separator between them (everything before `Expires`, `Expires`/`ExpiresString`, and everything after).  Not sure if that may render strangely in some contexts or be confusing for screen readers.

alexgromero: The docs handler was fixed in a new commit. The HTML and RST now render how a modeled parameter would.

tests/functional/test_service_headers.py
```diff
@@ -0,0 +1,50 @@
...
+    services_that_target_expires_header = set()
+
+    for service in services:
+        service_model = session.get_service_model(service)
+        for operation in service_model.operation_names:
+            operation_model = service_model.operation_model(operation)
+            if output_shape := operation_model.output_shape:
+                if _shape_targets_expires_header(output_shape):
+                    services_that_target_expires_header.add(service)
+    assert services_that_target_expires_header == {'s3'}
```
jonathan343, line 36: Can we make this a more descriptive failure so we can more easily detect the issue?

SamRemis (approved):

botocore/parsers.py
```diff
@@ -966,9 +977,17 @@ def _parse_non_payload_attrs(
...
+                        final_parsed[name] = self._handle_expires_timestamp(
+                            member_shape, headers[header_name]
+                        )
+                        if final_parsed[name] is None:
+                            del final_parsed[name]
+                        final_parsed['ExpiresString'] = headers[header_name]
+                    else:
+                        final_parsed[name] = self._parse_shape(
+                            member_shape, headers[header_name]
+                        )
```
nateprewitt, line 990: How are we scoping this behavior to S3?

botocore/parsers.py
```diff
@@ -536,6 +536,17 @@ def _handle_float(self, shape, text):
     def _handle_timestamp(self, shape, text):
         return self._timestamp_parser(text)
 
+    @_text_content
+    def _handle_expires_timestamp(self, shape, text):
```
nateprewitt, line 540: Why are we adding this to the `BaseXMLResponseParser` but invoking it in the `BaseRestParser`?

tests/functional/test_service_headers.py
```diff
@@ -0,0 +1,54 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import pytest
+
+from botocore.session import get_session
+
+# This test verifies that S3 is the only service that uses the 'expires' header
+# in their response. If any other service is found to target this header in any
+# of their output shapes, the test will fail and alert us of the change.
```
nateprewitt, line 19: This kind of information should be a docstring for the actual test rather than a comment for the whole module. This also appears `s3` specific, is there a reason we're creating a new test module instead of using `test_s3.py`?

jonathan343 (approved):
LGTM!

tests/unit/test_handlers.py
```diff
@@ -1780,3 +1780,109 @@ def test_remove_bucket_from_url_paths_from_model(
...
+
+@pytest.fixture()
+def document_s3_expires_mocks():
+    section = mock.Mock()
+    parent = mock.Mock()
+    param_line = mock.Mock()
+    param_section = mock.Mock()
+    doc_section = mock.Mock()
+    new_param_line = mock.Mock()
+    new_param_section = mock.Mock()
```
nateprewitt, line 1793: Do we need to declare all of these variables or could we set each value in our mock dictionary to `mock.Mock()`?

nateprewitt (changes requested):

botocore/parsers.py
alexgromero: We are using `test_service_headers.py` to verify that S3 is the only service that targets the expires header. (which I can move to `test_s3.py`)

botocore/parsers.py
alexgromero: I will move it to `BaseRestParser`.

tests/functional/test_service_headers.py
alexgromero: I thought since we were testing all services that it would be a new test module, but I will move it to `test_s3.py`.

tests/unit/test_endpoint.py
```diff
@@ -239,16 +242,16 @@ def test_retry_events_can_alter_behavior(self):
...
                 'needs-retry.ec2.DescribeInstances',
             ]
             * 2,
         )
 
     def test_retry_on_socket_errors(self):
-        self.event_emitter.emit.side_effect = self.get_emitter_responses(
-            num_retries=1
-        )
+        val = self.get_emitter_responses(num_retries=1, num_events=3)
```
SamRemis, line 253: This is a leftover from some testing; we should revert this change, except for the num_events part

botocore/endpoint.py
```diff
@@ -302,10 +302,19 @@ def _do_get_response(self, request, operation_model, context):
...
+            **{
+                'response_dict': response_dict,
+                'additional_response_keys': additional_response_keys,
+            },
+        )
         parser = self._response_parser_factory.create_parser(protocol)
         parsed_response = parser.parse(
             response_dict, operation_model.output_shape
         )
+        parsed_response.update(additional_response_keys)
```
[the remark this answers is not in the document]
alexgromero: yes, `response_dict` doesn't persist because we use this dict as a temporary variable to help us populate `parsed_response` with all the modeled attributes. after that, we discard the `response_dict` variable.

botocore/handlers.py
```diff
@@ -596,6 +596,56 @@ def parse_get_bucket_location(parsed, http_response, **kwargs):
     parsed['LocationConstraint'] = region
 
 
+def handle_expires_header(response_dict, customized_response_dict, **kwargs):
```
jonathan343, line 599: Is there a reason why we're also not checking the operation model output shape to make sure if has the `Expires` member modeled? I may have missed that conversation

alexgromero: fixing that in new commit

SamRemis (approved):

PR #3233, opened 2024-08-06
Multi-auth trait support

SamRemis:
Adds support for the new 'auth' trait on our models that will allow a priority list of auth types for a service or operation


botocore/config.py
```diff
@@ -256,6 +262,7 @@ class Config:
             ('tcp_keepalive', None),
             ('request_min_compression_size_bytes', None),
             ('disable_request_compression', None),
+            ('sigv4a_signing_region_set', None),
             ('client_context_params', None),
```
nateprewitt, line 266: Was there a reason we're changing order for existing entries? If not, we should append to the end.
```suggestion
            ('client_context_params', None),
            ('sigv4a_signing_region_set', None),
```

tests/functional/test_auth_config.py
```diff
@@ -0,0 +1,76 @@
...
+@pytest.mark.parametrize("auth_service, operation_model", AUTH_OPERATIONS)
+def test_all_requirements_match_for_operation(auth_service, operation_model):
+    # Validates that all operation-level signature types have the same requirements
+    message = f'Found mixed signer requirements for operation: {auth_service}.{operation_model.name}'
+    auth_config = operation_model.auth
+    assert_all_requirements_match(auth_config, message)
+
+
+def assert_all_requirements_match(auth_config, message):
+    if len(auth_config) > 1:
```
nateprewitt, line 70: It's usually best to avoid conditionals in tests if possible. Branching like this leads to subtle failures where we may no longer meet the conditional and now the tests "pass" but do nothing.

tests/unit/auth/test_auth_trait.py
```diff
@@ -0,0 +1,46 @@
...
+    def test_auth_resolves_first_available(self):
+        auth = ['aws.auth#foo', 'aws.auth#bar']
+        bar_signer = mock.Mock(spec=BaseSigner)
+
+        auth_types = AUTH_TYPE_MAPS.copy()
+        auth_types['bar'] = bar_signer
+
+        auth_type_conversions = AUTH_TYPE_TO_SIGNATURE_VERSION.copy()
+        auth_type_conversions['aws.auth#foo'] = "foo"
+        auth_type_conversions['aws.auth#bar'] = "bar"
```
nateprewitt, line 34: Just curious on why we're adding the extra mocking scaffolding here instead of something like:
```suggestion
        # Don't declare a signer for "foo"
        auth_types = {'bar': mock.Mock(spec=BaseSigner)}
        auth_type_conversions = {'aws.auth#foo': 'foo', 'aws.auth#bar': 'bar'}
```

If we're unit testing the ability to skip an entry, we can keep the inputs minimal.

tests/unit/test_client.py
```diff
@@ -1713,6 +1713,17 @@ def test_client_internal_credential_shim(self):
...
+    def test_unsigned_payload_disables_signing(self):
+        creator = self.create_client_creator()
+        service_client = creator.create_client('myservice', 'us-west-2')
+        test_operation = self.service_description['operations'][
+            'TestOperation'
+        ]
+        test_operation['unsignedPayload'] = True
+        service_client.test_operation(Foo='one')
+        context = self.endpoint.make_request.call_args[0][1]['context']
+        self.assertFalse(context['payload_signing_enabled'])
```
nateprewitt, line 1725: Do we have any positive tests enforcing the other side of this?

tests/unit/test_handlers.py
```diff
@@ -1042,7 +1042,7 @@ def test_set_operation_specific_signer_v4a_existing_signing_context(self):
             context=context, signing_name=signing_name
         )
         # region has been updated
-        self.assertEqual(context['signing']['region'], '*')
+        self.assertEqual(context['signing']['region'], 'abc')
```
nateprewitt, line 1045: I think this change is wrong. It's no longer asserting what we're intending and doesn't match the comment. If we're moving this functionality out of `set_operation_specific_signer`, we need another unit or functional test to confirm this behavior remains in place for MRAP.

This also points to a breakage we're making. I'm not sure if there will be any tangible impact from that, but this is a shift in how sigv4a requests would work in our event system.

tests/unit/test_handlers.py
```diff
@@ -1073,6 +1073,29 @@ def test_set_operation_specific_signer_s3v4_unsigned_payload(self):
...
+            'auth_type': 'v4a',
+        }
+        handlers.set_operation_specific_signer(
+            context=context, signing_name=signing_name
+        )
+        self.assertEqual(context['signing']['region'], '*')
+
+    def test_set_operation_specific_signer_prefers_client_config(self):
+        signing_name = 'myservice'
+        context = {
```
nateprewitt, line 1088: We should also have a region declared in a signing context here to demonstrate it won't be used if this config is present.

botocore/client.py
```diff
@@ -955,6 +960,10 @@ def _make_api_call(self, operation_name, api_params):
             'has_streaming_input': operation_model.has_streaming_input,
             'auth_type': operation_model.auth_type,
         }
+
+        if operation_model.unsigned_payload:
+            request_context['payload_signing_enabled'] = False
+
```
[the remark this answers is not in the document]
SamRemis: It was between adding it to the context and putting it here. I have added it back to context and it is now being used in set_operation_specific_signer

tests/unit/test_client.py
SamRemis: Since this logic moved out of client.py, the test moved to a similar one in handler.py.  I have added an inverse of this test as well

tests/functional/test_auth_config.py
SamRemis: Leaving this comment unresolved as there is still an if statement in this test, but this one is gone.  Now there is just the if statement above: 
```
           if operation_model.auth:
                auth_operations.append([service, operation_model])
```

Any alternatives that I can come up with involve loosening the checks of this test for no benefit and checking many empty lists to confirm that they are indeed empty.

tests/functional/test_auth_config.py
```diff
@@ -0,0 +1,77 @@
...
+    # Validates that all operation-level signature types have the same requirements
+    message = f'Found mixed signer requirements for operation: {auth_service}.{operation_model.name}'
+    auth_config = operation_model.auth
+    assert_all_requirements_match(auth_config, message)
+
+
+def assert_all_requirements_match(auth_config, message):
+    auth_requirements = set()
+    for auth_type in auth_config:
+        auth_requirements.add(AUTH_TYPE_REQUIREMENTS[auth_type])
```
nateprewitt, line 76: nit; a comprehension is more idiomatic and performant in most of these cases. If you don't intend to add complex logic in a loop, it's better to create the set with all values at once than add them one at a time.

```suggestion
    auth_requirements = set(
        AUTH_TYPE_REQUIREMENTS[auth_type] for auth_type in auth_config
    )
```

nateprewitt (commented):
Couple more small comments but otherwise I think this is looking good.

tests/unit/test_handlers.py
SamRemis: I hope that customers are not currently relying on something they have set manually being overwritten, especially when it's in the signing context which is considered a private interface.  At the moment, all sigv4a endpoint rule-sets will evaluate to `*`.  

There are already [existing tests](https://github.com/boto/botocore/blob/08877232170abe910198e3c1f758ae6d0f2df6a0/tests/functional/test_s3.py#L1234) that confirm this behavior for MRAPs.

.changes/next-release/enhancement-signing-83434.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
```
nateprewitt, line 2:
```suggestion
  "type": "feature",
```

PR #3243, opened 2024-09-03
only add the content-sha256 header when it's already exists for CrtSigV4QueryAuth and CrtSigV4AsymQueryAuth

TingDaoK:
- For CRT signer, override `_should_add_content_sha256_header` for `CrtSigV4QueryAuth` and `CrtSigV4AsymQueryAuth` to add the `content_sha256` header if the request to sign already have the header, instead of relying on the mix of the configurations and the headers in request to sign.
- So that it matches the SDK signer behavior to not add the extra `x-amz-content-sha256` header when `_should_sha256_sign_payload` is disabled.

Extra notes:
- Currently the CRT signer `CrtSigV4QueryAuth` and `CrtSigV4AsymQueryAuth` will never add the `x-amz-content-sha256` header, with the change from CRT, we consider to add `x-amz-content-sha256` header to the signed request if your original request to sign has the header set to "UNSIGNED-PAYLOAD". From https://github.com/awslabs/aws-c-auth/pull/248
- This change ensures that `request.context["payload_signing_enabled"] = False` will not add the extra header to the signed request.

Will update the CRT version and add tests.

### DIFF
--- a/botocore/crt/auth.py
+++ b/botocore/crt/auth.py
@@ -81,7 +81,7 @@ class CrtSigV4Auth(BaseSigner):
         else:
             explicit_payload = UNSIGNED_PAYLOAD
 
-        if self._should_add_content_sha256_header(explicit_payload):
+        if self._should_add_content_sha256_header(existing_sha256, explicit_payload):
             body_header = (
                 awscrt.auth.AwsSignedBodyHeaderType.X_AMZ_CONTENT_SHA_256
             )
@@ -169,7 +169,7 @@ class CrtSigV4Auth(BaseSigner):
         # bit of metadata through the request context.
         return request.context.get('payload_signing_enabled', True)
 
-    def _should_add_content_sha256_header(self, explicit_payload):
+    def _should_add_content_sha256_header(self, existing_sha256, explicit_payload):
         # only add X-Amz-Content-SHA256 header if payload is explicitly set
         return explicit_payload is not None
 
@@ -278,7 +278,7 @@ class CrtSigV4AsymAuth(BaseSigner):
         else:
             explicit_payload = UNSIGNED_PAYLOAD
 
-        if self._should_add_content_sha256_header(explicit_payload):
+        if self._should_add_content_sha256_header(existing_sha256, explicit_payload):
             body_header = (
                 awscrt.auth.AwsSignedBodyHeaderType.X_AMZ_CONTENT_SHA_256
             )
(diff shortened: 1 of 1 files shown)


botocore/crt/auth.py
```diff
@@ -81,7 +81,7 @@ def add_auth(self, request):
         else:
             explicit_payload = UNSIGNED_PAYLOAD
 
-        if self._should_add_content_sha256_header(explicit_payload):
+        if self._should_add_content_sha256_header(existing_sha256, explicit_payload):
```
nateprewitt, line 84: We should always add new positional arguments at the end of the function signature. I would also recommend making this an optional parameter that defaults to `None`.

botocore/crt/auth.py
```diff
@@ -494,6 +494,10 @@ def _apply_signing_changes(self, aws_request, signed_crt_request):
         # fragment - 4
         aws_request.url = urlunsplit((p[0], p[1], p[2], signed_query, p[4]))
 
+    def _should_add_content_sha256_header(self, existing_sha256, explicit_payload):
+        # only add X-Amz-Content-SHA256 header if header already set to UNSIGNED_PAYLOAD
+        return existing_sha256 == UNSIGNED_PAYLOAD
```
nateprewitt, line 499: Why are we only adding this if it's unsigned? What if we have an actual payload we're signing for?

nateprewitt (changes requested):

botocore/crt/auth.py
nateprewitt: It looks like we also missed an update somewhere? I think this will be resolved if you follow the advice above, but we should figure out where we're missing.
````
TypeError: _should_add_content_sha256_header() takes 2 positional arguments but 3 were given
```

botocore/crt/auth.py
TingDaoK: It's a temporary workaround we want to add for vpc-lattice and not want to introduce the change for others (Even though that it might be more reasonable to add the header to sign for other service as well, and it will match the behavior for the botocore signer).

We decided to have an other workaround as the short term patch within CRT now. https://github.com/awslabs/aws-c-auth/pull/249

We will have further discussion about what's the proper way to handle their use case in the near future, and it should be across SDKs not only for CRT.

So, no need for this PR now.

botocore/crt/auth.py
TingDaoK: I saw the error, but I don't understand why... Looks like we are not declaring `_should_add_content_sha256_header` anywhere else, and I updated the param for `_should_add_content_sha256_header` in `CrtSigV4QueryAuth`. Interesting.

TingDaoK:
I'll close this PR as we don't need this change now.

We already kicked off the release for the alternative change in CRT, as long as we pick up the latest version v0.21.4 once it's release, we should be good

PR #3244, opened 2024-09-04
Add S3 tests to ensure dot segments are preserved

jonathan343:
### Summary

This PR adds s3 specific tests to ensures the following:
* Leading dot segments in the key value are preserved in the URL path.
  * Example: `../key.txt`
* Embedded dot segments in the key value are preserved in the URL path.
  * * Example: `foo/../key.txt`


nateprewitt (approved):
:shipit:

Issue #3245, opened 2024-09-05
Feature: Paginator For OpenSearch Client


digglife:
### Describe the feature

Add paginator support for OpenSearch Client.

### Use Case

With this feature, we can use `paginator` for the OpenSearch APIs with Max Results limit, instead of looping with `NextToken` manually. Something like

```python
instances = []
opensearch = boto3.client('opensearch')
paginator = opensearch.get_paginator('describe_reserved_instances')
for page in paginator.paginate():
    instances.extend(page['ReservedInstances'])
```

### Proposed Solution

Update `paginators-1.json` of opensearch

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

Python SDK Botocore 1.x

### Environment details (OS name and version, etc.)

macos

tim-finnigan:
Thanks for the feature request and for creating that PR. Paginator models are owned and maintained upstream by service teams, because they are used across AWS SDKs. Therefore the OpenSearch team would need to release these paginator models directly in order for SDKs like Boto3 to use them. 

I have created an internal feature request for that service team to add the paginator models from your PR. I'm going to close this since the feature request won't be tracked here, but please refer to the [CHANGELOG](https://github.com/boto/botocore/blob/develop/CHANGELOG.rst?plain=1) for future updates.

PR #3246, opened 2024-09-05
feat: add paginator for opensearch

digglife:
With this change, we can use paginator for the OpenSearch APIs with Max Results limit, instead of looping with NextToken manually.

```python
# a trivial example
instances = []
opensearch = boto3.client('opensearch')
paginator = opensearch.get_paginator('describe_reserved_instances')
for page in paginator.paginate():
    instances.extend(page['ReservedInstances'])
```

fixes: #3245


tim-finnigan:
Thanks for creating this PR, but we can't accept these changes for reasons described in https://github.com/boto/botocore/issues/3245#issuecomment-2332598532

PR #3247, opened 2024-09-06
Update Response Parsing Protocol Tests

jonathan343:
> **Note:** Request serialization tests will be added in a follow up PR in effort to reduce the scope of this one.
> * Request Serialization PR: https://github.com/boto/botocore/pull/3378

## Summary

This PR replaces the existing response parsing protocol tests in `botocore/tests/unit/protocols/output` with new tests generated from [Smithy protocol test models](https://github.com/smithy-lang/smithy/tree/main/smithy-aws-protocol-tests/model). We use a version of these Smithy tests that are converted to the format currently supported by our existing test runner (`test_protocols.py`). This PR makes updates to the botocore Parsers and protocol test runner to comply with a majority of the new test cases.

## Granular Protocol Ignore List
This PR introduces a `protocol-tests-ignore-list.json` file that can be used to ignore specific protocol test suites or cases.
The structure for this list is defined below:
```
{
  "general": {
    "<test_type>": {
      "suites": [
        "Example test suite description",
      ],
      "cases": [
        "example-test-case-id"
      ]
    }
  },
  "protocols": {
    "<protocol_name>": {
      # Same as "general"  
    }
  }
}
```
* `<test_type>` - There are two types of protocol tests. 
  * `input` - Request serialization protocol test. 
  * `output` - Response parsing protocol test.
  * **Note:** Only `output` will work until the request serialization tests are updated.
* `suites` - A list of test suite descriptions to ignore. When specified, all test cases for the related suite will be ignored.
* `cases` - A list of test case ids to ignore.
* `<protocol_name>` - The protocol name as represented by its corresponding protocol test file name (without the `.json`) extension.
  * Supported values are `ec2`, `json`, `json_1_0`, `query`, `rest-json`, `rest-xml`. This may grow as we add more protocols.


botocore/parsers.py
```diff
@@ -577,7 +590,7 @@ def _do_parse(self, response, shape):
         return self._parse_body_as_xml(response, shape, inject_metadata=True)
 
     def _parse_body_as_xml(self, response, shape, inject_metadata=True):
-        xml_contents = response['body']
+        xml_contents = response['body'] or b'<xml/>'
```
nateprewitt, line 593: Is there something unique about this case or should this be handled farther down in `_parse_xml_string_to_dom`? I'm curious what valid cases we're expecting the service to send us a body and we're getting nothing.

botocore/parsers.py
```diff
@@ -707,8 +720,10 @@ def _do_error_parse(self, response, shape):
             # code has a couple forms as well:
             # * "com.aws.dynamodb.vAPI#ProvisionedThroughputExceededException"
             # * "ResourceNotFoundException"
+            if ':' in code:
+                code = code.split(':', 1)[0]
             if '#' in code:
-                code = code.rsplit('#', 1)[1]
+                code = code.split('#', 1)[1]
```
nateprewitt, line 726: This is semantically different. Are we enforcing that the code is everything following the first `#`?

botocore/parsers.py
```diff
@@ -707,8 +720,10 @@ def _do_error_parse(self, response, shape):
             # code has a couple forms as well:
             # * "com.aws.dynamodb.vAPI#ProvisionedThroughputExceededException"
             # * "ResourceNotFoundException"
+            if ':' in code:
+                code = code.split(':', 1)[0]
```
nateprewitt, line 724: There are now 3 distinct `if` statements here that aren't using `elif/else`. Is the intent to continually mutate the `code` value as it goes through each check?

e.g.

`com.test.mypackage#this:that#1` -- `split(':', 1)` --> `com.test.mypackage#this`
`com.test.mypackage#` -- `split('#', 1)` --> `this`

Then `this` is evaluated for the `x-amzn-query-error` code?

botocore/parsers.py
```diff
@@ -1020,19 +1035,29 @@ def _inject_error_code(self, error, response):
...
-            error['Error']['Code'] = code
         elif 'code' in body or 'Code' in body:
-            error['Error']['Code'] = body.get('code', body.get('Code', ''))
+            code = body.get('code', body.get('Code', ''))
+        if code is not None:
+            if ':' in code:
+                code = code.split(':', 1)[0]
+            if '#' in code:
+                code = code.split('#', 1)[1]
+            error['Error']['Code'] = code
```
nateprewitt, line 1048: It looks like we're changing behavior again here. Is there additional scope, or why are we post-processing the body for the `code` case now?

tests/unit/test_protocols.py
```diff
@@ -99,6 +98,10 @@
     'rest-xml': RestXMLParser,
 }
 PROTOCOL_TEST_BLACKLIST = ['Idempotency token auto fill']
+IGNORE_LIST_FILENAME = "protocol-tests-ignore-list.json"
+PROTOCOL_TEST_IGNORE_LIST_PATH = os.path.join(TEST_DIR, IGNORE_LIST_FILENAME)
+with open(PROTOCOL_TEST_IGNORE_LIST_PATH) as f:
+    PROTOCOL_TEST_IGNORE_LIST = json.load(f)
```
nateprewitt, line 104: Do we use `PROTOCOL_TEST_IGNORE_LIST_PATH ` somewhere else? Should this whole thing be a self-contained function or `pytest` fixture instead?

tests/unit/test_protocols.py
```diff
@@ -181,39 +192,51 @@ def stream(self):
...
         # We load the json as utf-8, but the response parser is at the
         # botocore boundary, so it expects to work with bytes.
-        body_bytes = case['response']['body'].encode('utf-8')
-        case['response']['body'] = body_bytes
+        # If a test case doesn't define a response body, set it to `None`.
+        if 'body' in case['response']:
+            body_bytes = case['response']['body'].encode('utf-8')
+            case['response']['body'] = body_bytes
+        else:
+            case['response']['body'] = b''
```
nateprewitt, line 214: Where are we setting the body to `None`? It looks like our fallback is empty bytes. Is this what's causing issues with the xml test above?

tests/unit/test_protocols.py
```diff
@@ -462,7 +503,7 @@ def _get_suite_test_id():
         if len(split) == 2:
             suite_id, test_id = int(split[0]), int(split[1])
         else:
-            suite_id = int(split([0]))
+            suite_id = int(split[0])
```
nateprewitt, line 506: Is this code even reachable? How did this work before?

tests/unit/protocols/output/ec2.json
```diff
@@ -1,612 +1,1847 @@
 [
```
nateprewitt, line 1: high-level question; do we know if the new files cover all the existing cases we're deleting?

nateprewitt (commented):
Handful of questions about the current implementation.

botocore/parsers.py
```diff
@@ -427,11 +439,12 @@ def _handle_structure(self, shape, node):
             return self._handle_unknown_tagged_union_member(tag)
         for member_name in members:
             member_shape = members[member_name]
+            location = member_shape.serialization.get('location')
             if (
-                'location' in member_shape.serialization
+                location in self.KNOWN_LOCATIONS
```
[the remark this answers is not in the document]
jonathan343: Request object members can be modeled with `location` values that aren't applied to response parsing such as  `uri` and `querystring`. In the case where the same structure is modeled for the input and output of an operation and one of location values mentioned above is modeled, we would incorrectly assume the member has already been handled and fail to parse the related member.

**Solution:** Add a list (`KNOWN_LOCATIONS = ('statusCode', 'header', 'headers')` of know locations used to determine if a member is parsed in a location that is not the body. We'll only skip the member if one of these locations is specified, preventing the case where we unintentionally skip when `"location": "uri"` is specified.

Related Protocol Test Id: `IgnoreQueryParamsInResponse`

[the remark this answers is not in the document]
SamRemis: Sorry, I think I'm just reading this wrong, but what you wrote doesn't make sense to me.  

> Request object members can be modeled with location values that aren't applied to response parsing such as uri and querystring.

This seems to contradict what you say here:
> We'll only skip the member if one of these locations is specified, preventing the case where we unintentionally skip when "location": "uri" is specified.

So we do want to parse members that are in the uri after all?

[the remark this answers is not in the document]
SamRemis: Also, is there a reason we aren't doing this [here](https://github.com/boto/botocore/blob/08877232170abe910198e3c1f758ae6d0f2df6a0/botocore/parsers.py#L953-L971) where we handle the rest of the locations?  Is this somehow specific to XML parsing?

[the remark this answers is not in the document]
jonathan343: The `uri` and `querystring` locations are specific to input serialization (handled in `serialize.py`) and shouldn't be handled by our parsers. We shouldn't care about these locations when parsing responses (in `parsers.py`), but we do today. This changes is defining a list of the location values we do care about: `KNOWN_LOCATIONS = ('statusCode', 'header', 'headers')`.

Input serialization currently has a similar list which looks like: `KNOWN_LOCATIONS = ['uri', 'querystring', 'header', 'headers']`.

Lmk if you need further clarification!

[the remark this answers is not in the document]
SamRemis: Okay, this makes more sense now.  To give an example for anyone else confused by this, here's my current understanding:
1) These traits are specific to XML parsing, which is why they were put here.  This still needs to be confirmed.
2) This solves an _extremely_ specific case.  Best way to describe it is an example:

Let's say a service team models a `FooShape` and uses that shape as a top level input and a top level output on an operation.  If `FooShape` has a member called `BarString` that targets the `querystring`, we want to serialize that on the input by adding `?barstring=<input value>` to the query string; this all makes sense and is in line with how many of our APIs behave.

Since the service has modeled `FooShape` as _both_ the input and output, we need to have some way to parse the result for `BarString` that the service returns.  But since the query string is controlled entirely by the requester, the service can't reply with a value there, so instead they put it in the body.  This code ensures that `BarString` won't get skipped during parsing and instead will be returned to the customer.

tests/unit/protocols/protocol-tests-ignore-list.json
```diff
@@ -0,0 +1,72 @@
+{
+  "general": {},
```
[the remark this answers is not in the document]
jonathan343: Discussing this internally.

botocore/parsers.py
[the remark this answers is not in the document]
SamRemis: [Non-Blocking] approving anyhow, but I'd like to know where this investigation led.  If nothing else, I'd like to see this renamed to something that clarifies it's the list of known _response, non-body_ locations.  

Better yet would be to remove this logic entirely if we don't have a strong reason to add it.  Simple code is far more maintainable, and this seems to be adding a layer of complexity without real customer benefit.

tests/unit/protocols/output/ec2.json
jonathan343: We'll be leaving in legacy tests to ensure full coverage on the initial release. We can follow up with removal once we upstream coverage gaps.

botocore/parsers.py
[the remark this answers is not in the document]
jonathan343: I reverted the changes related to this test case and added it to our ignore list to unblock this PR. We will follow up with an deeper investigation on whether or not supporting this test is valuable.

SamRemis (approved):

SamRemis (approved):

PR #3254, opened 2024-09-12
Update event stream tests

JordonPhillips:
This fixes a pair of event stream tests. Implementations of `application/vnd.amazon.eventstream` are expected to check the CRC before attempting to semantically parse the prelude. The tests meant to enforce that were present, but the wrong error was being asserted.

This change fixes the ordering of checks, fixes the tests, and adds two new test cases to make assertions about the error types that were previously being applied to the old tests.


nateprewitt (approved):
:shipit:

Issue #2742, opened 2022-08-19
Allow Source Address and Port to be specified when creating a client


jmulcock-rbbn:
### Describe the feature

We should be able to set the source IP address and source port when creating the client using an optional parameter, in the same manner as urllib3. 

E.g. 

- session.create_client('ec2', source_address=None)) # Default value for parameter, allow random port and default IP address
- session.create_client('ec2', source_address=('10.10.10.10', 0)) # Use address 10.10.10.10, allow randomly selected port
- session.create_client('ec2', source_address=('', 1234)) # Use default address, set source port as 1234
- session.create_client('ec2', source_address=('10.10.10.10', 1234)) # Use address 10.10.10.10, set source port as 1234

### Use Case

As the endpoint IP address for requests is not static, local firewall rules on the machine need to allow traffic from port 443 to make requests. For some business use cases this unacceptable. Allowing to set source address and port allows greater granularity to the firewall rules.

### Proposed Solution

The source_address parameter should be set in pool_manager_kwargs in URLLib3Session, for it to work with urllib3

Add the optional parameter into to Session.create_client() with default set as _None_ (which will continue current behavior).  Pass that value all the way through until create the URLLib3Session and it gets added to pool_manager_kwargs, 

e.g.
   
    def _get_pool_manager_kwargs(self, **extra_kwargs):
        pool_manager_kwargs = {
            'strict': True,
            'timeout': self._timeout,
            'maxsize': self._max_pool_connections,
            'ssl_context': self._get_ssl_context(),
            'socket_options': self._socket_options,
            'cert_file': self._cert_file,
            'key_file': self._key_file,
            **'source_address': self._source_address,**
        }
        pool_manager_kwargs.update(**extra_kwargs)
        return pool_manager_kwargs

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

1.20.0

### Environment details (OS name and version, etc.)

Debian 10

tim-finnigan:
Hi @jmulcock-rbbn thanks for the feature request. I brought this up for discussion with the team and it’s something that may be considered at a future date. We suggest others who are also interested in this feature to 👍 the issue or leave additional feedback in the comments. And if there’s anything else you want to expand on or clarify regarding your use case please let us know.

nveloso:
Hi @tim-finnigan @jmulcock-rbbn, I've opened up a PR that adds support to specify the source IP address and source port. Please check #3256. Feel free to add any comments.

PR #3256, opened 2024-09-17
Add option to set the source IP address

nveloso:
I was looking into how I could set the source IP address and port when using boto3 but could not find a way. It turns out that I was not the only one looking into that. See #2742.

I looked into the botocore code and tried to add this functionality. Please let me know if something is not right.

I tested my code by creating a simple s3 listing where I specified my config with the new `source_address` parameter.

```python3
boto_config = Config(source_address=('...', 0))

s3 = boto3.client('s3', config=boto_config)
response = s3.list_objects_v2(Bucket='...', Prefix='...')

for content in response['Contents']:
    print(content['Key'])
```

Then I captured all packets using wireshark and ran my script. Confirmed on wireshark that the traffic to AWS was using the source address I specified in the `Config` object.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -147,6 +147,7 @@ class ClientArgsCreator:
             socket_options=socket_options,
             client_cert=new_config.client_cert,
             proxies_config=new_config.proxies_config,
+            source_address=new_config.source_address,
         )
 
         serializer = botocore.serialize.create_serializer(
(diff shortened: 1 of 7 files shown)


nveloso:
Hey @adev-code @RyanFitzSimmonsAK!
I've open up this PR more than a year ago and got no feedback. Would you help me get the right people involved in this PR?
Thank you!

PR #3259, opened 2024-09-19
feat: clamp PageSize to metadata-defined min/max

rustyconover:
Clamps the user-supplied PageSize in PaginationConfig to the min/max values specified in the limit key's shape metadata. This removes the need for users to manually account for operation-specific PageSize limits (e.g., 50 for list_regions, 1000 for list_images).


tim-finnigan:
Closing for reasons mentioned in https://github.com/boto/botocore/issues/3260#issuecomment-2371950077.

PR #3264, opened 2024-09-24
Remove worklink client following the service deprecation

alexgromero:
This PR removes the Amazon WorkLink client from Boto3 following the service deprecation.


jonathan343 (approved):
🚀

PR #3271, opened 2024-10-02
Add Flexible Checksum v2 Config Options

jonathan343:
## Overview
This PR adds the ``request_checksum_calculation`` and ``response_checksum_validation`` config options. 

> [!NOTE]  
> This PR only implements the functionality to configure the new options and doesn't actually use them for anything yet.

## Request Checksum Calculation
A new configuration option will be introduced that allows users to configure request checksum calculation behavior. The following configuration methods will be supported:

* Environment Variable - ``AWS_REQUEST_CHECKSUM_CALCULATION``
* Profile Config - ``request_checksum_calculation``
* Client Config - ``request_checksum_calculation``

Supported values:

* ``WHEN_SUPPORTED`` (Default) - If operations are modeled with  ``"requestChecksumRequired":true`` OR a ``requestAlgorithmMember``, then a checksum will be calculated for all request payloads.
* ``WHEN_REQUIRED`` - If operations are modeled with  ``"requestChecksumRequired":true`` or ``requestAlgorithmMember`` is modeled and set by the user, then a checksum will be calculated for request payloads.

## Response Checksum Validation
A new configuration option will be introduced that allows users to configure response checksum validation behavior. The following configuration methods will be supported:

* Environment Variable - ``AWS_RESPONSE_CHECKSUM_VALIDATION``
* Profile Config - ``response_checksum_validation``
* Client Config - ``response_checksum_validation``

Supported Values:

* ``WHEN_SUPPORTED`` (Default) - Checksum validation MUST be performed on all response payloads for operations modeled with the ``responseAlgorithms`` trait, unless no modeled checksum is supported.
* ``WHEN_REQUIRED`` - Checksum validation MUST NOT be performed on response payloads of operations UNLESS botocore supports the modeled checksum algorithms AND the user has set the ``requestValidationMemeber`` to ``ENABLED``.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -26,6 +26,10 @@ import botocore.parsers
 import botocore.serialize
 from botocore.config import Config
 from botocore.endpoint import EndpointCreator
+from botocore.httpchecksum import (
+    RequestChecksumCalculation,
+    ResponseChecksumValidation,
+)
 from botocore.regions import EndpointResolverBuiltins as EPRBuiltins
 from botocore.regions import EndpointRulesetResolver
 from botocore.signers import RequestSigner
@@ -271,11 +275,18 @@ class ClientArgsCreator:
                 sigv4a_signing_region_set=(
                     client_config.sigv4a_signing_region_set
                 ),
+                request_checksum_calculation=(
+                    client_config.request_checksum_calculation
+                ),
+                response_checksum_validation=(
+                    client_config.response_checksum_validation
+                ),
             )
         self._compute_retry_config(config_kwargs)
         self._compute_connect_timeout(config_kwargs)
         self._compute_user_agent_appid_config(config_kwargs)
         self._compute_request_compression_config(config_kwargs)
+        self._compute_checksum_config(config_kwargs)
         s3_config = self.compute_s3_config(client_config)
 
         is_s3_service = self._is_s3_service(service_name)
@@ -771,3 +782,40 @@ class ClientArgsCreator:
                 f'maximum length of {USERAGENT_APPID_MAXLEN} characters.'
             )
         config_kwargs['user_agent_appid'] = user_agent_appid
+
+    def _compute_checksum_config(self, config_kwargs):
+        self._handle_checksum_config(
+            config_kwargs,
+            config_key="request_checksum_calculation",
+            default_value="when_supported",
+            valid_options_enum=RequestChecksumCalculation,
+        )
+        self._handle_checksum_config(
+            config_kwargs,
+            config_key="response_checksum_validation",
+            default_value="when_supported",
+            valid_options_enum=ResponseChecksumValidation,
+        )
+
+    def _handle_checksum_config(
+        self,
+        config_kwargs,
+        config_key,
+        default_value,
+        valid_options_enum,
+    ):
+        value = config_kwargs.get(config_key)
+        if value is None:
+            value = (
+                self._config_store.get_config_variable(config_key)
+                or default_value
+            )
+        value = value.lower()
+        valid_options = valid_options_enum.values()
+        if value not in valid_options:
+            raise botocore.exceptions.InvalidChecksumConfigError(
+                config_key=config_key,
+                config_value=value,
+                valid_options=valid_options,
+            )
+        config_kwargs[config_key] = value
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -26,10 +26,6 @@ import botocore.parsers
 import botocore.serialize
 from botocore.config import Config
 from botocore.endpoint import EndpointCreator
-from botocore.httpchecksum import (
-    RequestChecksumCalculation,
-    ResponseChecksumValidation,
-)
 from botocore.regions import EndpointResolverBuiltins as EPRBuiltins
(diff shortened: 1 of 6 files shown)


botocore/args.py
```diff
@@ -771,3 +782,40 @@ def _compute_user_agent_appid_config(self, config_kwargs):
...
+        valid_options_enum,
+    ):
+        value = config_kwargs.get(config_key)
+        if value is None:
+            value = (
+                self._config_store.get_config_variable(config_key)
+                or default_value
+            )
+        value = value.lower()
+        valid_options = valid_options_enum.values()
```
nateprewitt, line 814: I know we talked about this a little previously. I'm curious what value we're getting out of using these enums vs a validation list like we do elsewhere in the codebase. Given we expect users to be providing strings for configuration, is the extra scaffolding we're adding here to create enums a net-benefit or is this additional overhead without a clear use case?

botocore/config.py
```diff
@@ -235,6 +235,37 @@ class Config:
         specified service will be ignored.
 
         Defaults to None.
+
+    :type request_checksum_calculation: ``RequestChecksumCalculation``
```
nateprewitt, line 239: This seems wrong. The type is a string, we're using the enum to manage/represent that string. We shouldn't be adding additional configuration complexity to this if it can be avoided.

botocore/config.py
```diff
@@ -235,6 +235,37 @@ class Config:
...
+            ``requestAlgorithmMember`` is modeled.
+
+        * ``WHEN_REQUIRED`` -- When set, a checksum will only be calculated
+            for request payloads of operations modeled with the ``httpChecksum``
+            trait where ``requestChecksumRequired`` is ``true`` or where a
+            ``requestAlgorithmMember`` is modeled and supplied.
+
+        Defaults to None.
+
+    :type response_checksum_validation: ``ResponseChecksumValidation``
```
nateprewitt, line 255: Same note here. We can consolidate the conversation to one place, so don't feel like you need to reply to each of these. Just flagging for adjustments depending on outcome.

nateprewitt (changes requested):
A couple high level comments about our `enum` usage. Otherwise looks good.

botocore/args.py
jonathan343: I liked the idea of using an Enum since this it would consolidate the valid options for each config. It would prevent some issues like typos when using these options on both the user and maintainer side. However, looking at this now, I don't think this provides enough value and introduces a new way to specify config options which can be more confusing to users. I can update the PR to move away from using the Enums.

botocore/config.py
jonathan343: Updated to `str`.

botocore/config.py
```diff
@@ -235,6 +235,37 @@ class Config:
         specified service will be ignored.
 
         Defaults to None.
+
+    :type request_checksum_calculation: str
+    :param request_checksum_calculation: Determines when a checksum will be
+        calculated for request payloads. Valid values are:
+
+        * ``when_supported`` -- When set, a checksum will be calculated for
```
jonathan343, line 243: Note: I changed these to lowercase to be consistent with our other config values. This gets normalized anyway before validation.

botocore/config.py
```diff
@@ -235,6 +235,37 @@ class Config:
...
+            all request payloads of operations modeled with the ``httpChecksum``
+            trait where ``requestChecksumRequired`` is ``true`` and/or a
+            ``requestAlgorithmMember`` is modeled.
+
+        * ``when_required`` -- When set, a checksum will only be calculated
+            for request payloads of operations modeled with the ``httpChecksum``
+            trait where ``requestChecksumRequired`` is ``true`` or where a
+            ``requestAlgorithmMember`` is modeled and supplied.
+
+        Defaults to None.
```
jonathan343, line 253: We say `Defaults to None.` for a few options in this file which is correct, however, during evaluation there may be an actual default value used. For both checksum config options, we default to `when_supported`. The reason for defaulting to None here and not `when_supported` is to prevent overwriting any values a user might configure at the session level, config file, or environment variable.

I wanted to point this out because it seems confusing from a user perspective when they read `Defaults to None.` and experience different behavior. How do we feel about updating our user documentation to mention this nuance.

botocore/args.py
```diff
@@ -61,6 +61,15 @@
...
 
+VALID_REQUEST_CHECKSUM_CALCULATION_CONFIG = (
+    "when_supported",
+    "when_required",
+)
+VALID_RESPONSE_CHECKSUM_VALIDATION_CONFIG = (
+    "when_supported",
+    "when_required",
+)
+
```
nateprewitt, line 72: We might consider if these should live here/if they're public. Once we expose them, it'll be harder to move/reuse these later. Importing from `args` into other modules is likely to create import cycles, so we just want to make sure we're accounting for that.

nateprewitt (approved):
I noted a minor concern we should look at, but I wouldn't consider it blocking for moving this to the dev branch.

botocore/args.py
jonathan343: I'll merge the PR as is for now and make a note to reevaluate the constants so we don't forget about this.

PR #3276, opened 2024-10-08
S3 200 errors implementation

alexgromero:
The S3 service may occasionally respond with an HTTP 200 OK status code while including errors in the response body. This PR addresses this handling to ensure accurate error reporting.


botocore/handlers.py
```diff
@@ -180,6 +190,16 @@ def _looks_like_special_case_error(http_response):
     return False
 
 
+def _has_modeled_body(operation_model):
```
jonathan343, line 193: To me it seems slightly odd for this function to live in `handlers.py`. I would consider moving this to `utils.py` or added as a property of the `OperationModel` class.

tests/functional/test_s3.py
```diff
@@ -462,19 +461,35 @@ def test_s3_copy_object_with_empty_response(self):
         self.assertEqual(response["ResponseMetadata"]["HTTPStatusCode"], 200)
         self.assertTrue("CopyObjectResult" in response)
 
-    def test_s3_copy_object_with_incomplete_response(self):
```
jonathan343, line 465: Why did we remove the `test_s3_copy_object_with_incomplete_response` test?

tests/functional/test_s3.py
```diff
@@ -462,19 +461,35 @@ def test_s3_copy_object_with_empty_response(self):
...
+            b"<Error>"
+            b"<Code>SlowDown</Code>"
+            b"<Message>Please reduce your request rate.</Message>"
+            b"</Error>"
+        )
+        self.http_stubber.add_response(status=200, body=error_body)
+        self.http_stubber.add_response(status=200, body=error_body)
+        self.http_stubber.add_response(status=200, body=error_body)
+        self.http_stubber.add_response(status=200, body=error_body)
+        self.http_stubber.add_response(status=200, body=error_body)
```
jonathan343, line 478: why do we have to add the response 5 times?

tests/unit/test_handlers.py
```diff
@@ -1175,49 +1148,6 @@ def test_non_ascii_characters(self):
         )
 
 
-class TestRetryHandlerOrder(BaseSessionTest):
```
jonathan343, line 1178: Just to make sure we're on the same page: Was this test removed because we now register the handler to the `before-parse` event and guarantees it gets called before the retry logic?

tests/functional/test_s3.py
alexgromero: This was to ensure we are reaching the 4 max retries to assert we are raising the `ClientError`.

tests/functional/test_s3.py
alexgromero: We no longer go through the parser first which will raise a `ResponseParserError` for an incomplete response. This operation now runs through the `check_for_200_error` handler first which will lead to retry a partial response instead of raising the error. 
Alternatively, we can check whether we retried and received a complete body, but the test above this one ([test_s3_copy_object_with_empty_response](https://github.com/boto/botocore/blob/develop/tests/functional/test_s3.py#L438C9-L463)) is already doing that with an empty body.

tests/unit/test_handlers.py
alexgromero: Yes

tests/functional/test_credentials.py
```diff
@@ -1060,7 +1060,7 @@ def test_token_chosen_from_provider(self):
         session = Session(profile='sso-test')
         with SessionHTTPStubber(session) as stubber:
             self.add_credential_response(stubber)
-            stubber.add_response()
+            stubber.add_response(body=b'<Test></Test>')
```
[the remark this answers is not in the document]
alexgromero: Yes, this change was introduced because the current tests fail as they go through the `check_for_200_error` handler, which expects certain operations to include a body.

tests/functional/test_s3.py
jonathan343: Ah, my bad. I missed that on the first read. Thank you!

botocore/handlers.py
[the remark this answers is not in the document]
alexgromero: Great suggestion! I agree it would make more sense to move it to the `OperationModel` class.

botocore/endpoint.py
```diff
@@ -314,6 +314,11 @@ def _do_get_response(self, request, operation_model, context):
             response_dict, operation_model.output_shape
         )
         parsed_response.update(customized_response_dict)
+        if updated_status_code := customized_response_dict.get(
+            'updated_status_code'
+        ):
+            http_response.status_code = updated_status_code
+            del parsed_response['updated_status_code']
```
SamRemis, line 321: This feels kind of awkard; if we have to do this, we can move this added section one line up and delete it from the customized response dict instead.  

Better yet would be updating the ResponseMetadata.HTTPStatusCode in the customized response dictionary, then checking the parsed response in the future handler.  Not sure if this is possible without trying it out.

botocore/model.py
```diff
@@ -707,6 +707,26 @@ def _get_streaming_body(self, shape):
                 return payload_shape
         return None
 
+    @CachedProperty
+    def has_modeled_body_input(self):
```
SamRemis, line 711: Is this being used anywhere?  If not, we should hold off on adding it until it's needed

nateprewitt (changes requested):
Some high level comments of what may make this a bit simpler. We should be able to drop the empty body requirement after our discussion.

tests/functional/test_credentials.py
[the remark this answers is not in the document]
alexgromero: Adding bodies to our test suite is breaking changes. This will be reverted in the next revision.

tests/unit/test_handlers.py
```diff
@@ -501,33 +501,6 @@ def test_presigned_url_casing_changed_for_rds(self):
         )
         self.assertIn('X-Amz-Signature', params['PreSignedUrl'])
 
-    def test_500_status_code_set_for_200_response(self):
```
SamRemis, line 504: If we are leaving the method there for backward compatibility, why are we deleting the tests?  (I see the names were reused below, but we should be making new tests for those)

alexgromero: Got it, I will add them back.

tests/unit/test_handlers.py
```diff
@@ -1944,3 +1917,80 @@ def test_document_response_params_without_expires(document_expires_mocks):
...
+            b"<Message>Please reduce your request rate.</Message>"
+            b"</Error>"
+        ),
+    }
+
+
+def test_500_status_code_set_for_200_response(
+    operation_model_for_200_error, response_dict_for_200_error
+):
+    handlers._handle_200_error(
```
nateprewitt, line 1945: Ideally we aren't invoking private functions outside of their respective modules. Do we have specific cases we're testing in these unit tests that aren't already verified with the functional tests?

tests/unit/test_handlers.py
```diff
@@ -1944,3 +1917,80 @@ def test_document_response_params_without_expires(document_expires_mocks):
...
+    parsed = {}
+    parsed.setdefault('ResponseMetadata', {})['HTTPStatusCode'] = 500
+    handlers._retry_200_error((http_response, parsed))
+    assert http_response.status_code == 500
+
+
+def test_500_response_can_be_none():
+    # A 500 response can raise an exception, which means the response
+    # object is None.  We need to handle this case.
+    handlers._retry_200_error(None)
```
nateprewitt, line 1996: For future reference, we should always be asserting _something_ in our tests to ensure accidental no-op regressions don't sneak in. If we wanted to validate that `None` is a valid, we should be testing what the behavior of the function is when that's passed. In this case, we know it will return `None` if it's working as expected.

Read the above comment about testing private functions though before making any changes.

nateprewitt (changes requested):
Looking good! Have a few minor notes on comments and unit tests. We want to try to be succinct with comments/logging as long messages are often ignored.

.changes/next-release/enhancement-s3-47846.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "``s3``",
+  "description": "Adds logic to gracefully handle HTTP 200 OK responses that contain error information in the response body."
```
nateprewitt, line 4:
```suggestion
  "description": "Handle HTTP 200 responses with error information for all supported s3 operations."
```

botocore/handlers.py
```diff
@@ -1239,6 +1240,35 @@ def document_expires_shape(section, event_name, **kwargs):
         )
 
 
+def _handle_200_error(operation_model, response_dict, **kwargs):
+    # S3 can return a 200 response with an error embedded in the body.
+    # Convert the 200 to a 500 for retry resolution in ``_retry_200_error``.
```
nateprewitt, line 1245:
```suggestion
    # Convert the 200 to a 500 for retry resolution in ``_update_status_code``.
```

botocore/handlers.py
```diff
@@ -1239,6 +1242,35 @@ def document_expires_shape(section, event_name, **kwargs):
...
+
+def _update_status_code(response, **kwargs):
+    # Update the http_response status code when the parsed response has been
+    # modified in a handler. This enables retries for cases like ``_handle_200_error``.
+    if response is None:
+        return
+    http_response, parsed = response
+    parsed_status_code = parsed.get('ResponseMetadata', {}).get(
+        'HTTPStatusCode'
+    )
```
nateprewitt, line 1269: Ok, hopefully last note. We don't want to set the status_code to `None`. If we end up with this not being available with the `get`, we should be falling back to what `http_response.status_code` is already set to so the rest of this function is a no-op.

```suggestion
    parsed_status_code = parsed.get('ResponseMetadata', {}).get(
        'HTTPStatusCode', http_response.status_code
    )
```

nateprewitt (changes requested):
Looking good! One last thing to fix and then I think we're set.

botocore/handlers.py
alexgromero: Good catch, thanks!

nateprewitt (approved):
Things look good, assuming tests pass I think this is ready to ship. Great job, @alexgromero!

nateprewitt:
Closing/reopening to rerun tests against fix in #3282.

PR #3277, opened 2024-10-08
Request/Response Checksum Behavior Updates

jonathan343:
This PR makes the following updates to the botocore flexible checksum behavior:

## Request Checksum Calculation
* Remove MD5 checksums and make CRC32 the default checksum algorithm.
  * CRC32 is a suitable default algorithm because it is available with and without the optional `awscrt` dependency.
* When the `request_checksum_calculation` config is set to `when_supported` (default value), a checksum will be generated whenever one of the following conditions is true:
  * An operation applies the `httpchecksum` trait and defines a `requestAlgorithmMember`.
  * An operation applies the `httpchecksum` trait and defines `requestChecksumRequired: true`.
* When the `request_checksum_calculation` config is set to `when_required`, a checksum will be generated only when:
  *  An operation applies the `httpchecksum` trait and defines `requestChecksumRequired: true`.

### S3 Customizations:
* Deprecates existing customizations for MD5 checksums.

---

## Response Checksum Validation
* A value set for `requestValidationModeMember` by the user will be used by the SDK.
* If the `requestValidationModeMember` value is not set by the user, the following behavior is experienced:
  * When the `response_checksum_validation` config is set to `when_supported` (default value), the `requestValidationModeMember` will be set to `ENABLED`.
  * When the `response_checksum_validation` config is set to `when_required`, the `requestValidationModeMember` will not be set and checksum validation is skipped.


botocore/utils.py
```diff
@@ -3223,96 +3220,18 @@ def get_encoding_from_headers(headers, default='ISO-8859-1'):
...
-    # assume a checksum has already been provided and an md5 is not needed
+    # assume a checksum has already been provided by the user.
     for header in headers:
         if CHECKSUM_HEADER_PATTERN.match(header):
             return True
 
     return False
 
 
-def conditionally_calculate_checksum(params, **kwargs):
```
nateprewitt, line 3271: Removing these may be breaking, I don't know if we've generally deleted functions out of the `utils.py` package.

tests/functional/test_httpchecksum.py
```diff
@@ -0,0 +1,326 @@
...
+    checksum_algorithm,
+    request_payload,
+    expected_headers,
+):
+    patch_load_service_model(
+        patched_session,
+        monkeypatch,
+        TEST_CHECKSUM_SERVICE_MODEL,
+        TEST_CHECKSUM_RULESET,
+    )
```
nateprewitt, line 217: Can we do this as a decorator or context manager?

tests/functional/test_s3.py
```diff
@@ -28,6 +27,7 @@
     UnsupportedS3AccesspointConfigurationError,
     UnsupportedS3ConfigurationError,
 )
+from botocore.httpchecksum import _CHECKSUM_CLS
```
nateprewitt, line 30: Same note here, we should try to avoid importing anything marked as private.

tests/functional/test_s3.py
```diff
@@ -1454,30 +1454,61 @@ def setUp(self):
     def get_sent_headers(self):
         return self.http_stubber.requests[0].headers
 
-    def test_content_md5_set(self):
+    def test_trailing_checksum_set(self):
```
nateprewitt, line 1457: It doesn't look like we're actually validating the "trailer" here?

tests/functional/test_s3.py
```diff
@@ -1454,30 +1454,61 @@ def setUp(self):
...
+            b"STREAMING-UNSIGNED-PAYLOAD-TRAILER",
+        )
 
-    def test_content_sha256_set_if_config_value_is_true(self):
-        # By default, put_object() does not include an x-amz-content-sha256
-        # header because it also includes a `Content-MD5` header. The
-        # `payload_signing_enabled` config overrides this logic and forces the
-        # header.
+    def test_content_sha256_not_set_if_config_value_is_true(self):
+        # By default, put_object() provides a trailing checksum and includes the
```
nateprewitt, line 1505: Is this right? I wouldn't expect us to be using trailing checksums by default on `put_object`.

tests/functional/test_s3.py
```diff
@@ -1488,11 +1519,17 @@ def test_content_sha256_set_if_config_value_is_true(self):
             self.client.put_object(Bucket="foo", Key="bar", Body="baz")
         sent_headers = self.get_sent_headers()
         sha_header = sent_headers.get("x-amz-content-sha256")
-        self.assertNotEqual(sha_header, b"UNSIGNED-PAYLOAD")
+        self.assertIsNotNone(sha_header)
```
nateprewitt, line 1522: Do we need this check if we're asserting it's equal to a non-None value below?

tests/functional/test_s3.py
```diff
@@ -1488,11 +1519,17 @@ def test_content_sha256_set_if_config_value_is_true(self):
...
+        self.assertEqual(sha_header, b"STREAMING-UNSIGNED-PAYLOAD-TRAILER")
 
     def test_content_sha256_not_set_if_config_value_is_false(self):
+        # By default, put_object() provides a trailing checksum and includes the
```
nateprewitt, line 1526: Same note here, is this right?

nateprewitt (changes requested):

tests/functional/test_s3.py
jonathan343: Because CRC32 is the new default, `put_object` will use trailing checksums by default. This is because `put_object` has streaming input.

You can see this behavior in the current version of the SDK by running:
```
import boto3
boto3.set_stream_logger('')

s3 = boto3.client("s3")

response = s3.put_object(
    Body=b"Example data.",
    Bucket="<bucket_name>",
    Key="example_key.txt",
    ChecksumAlgorithm="CRC32",
)
```

tests/functional/test_s3.py
jonathan343: See response in previous thread.

botocore/httpchecksum.py
```diff
@@ -32,10 +32,7 @@
     MissingDependencyException,
 )
 from botocore.response import StreamingBody
-from botocore.utils import (
-    conditionally_calculate_md5,
-    determine_content_length,
-)
+from botocore.utils import _has_checksum_header, determine_content_length
```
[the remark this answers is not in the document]
nateprewitt: If we're going to start importing it, then we'll need to make it public since it's used cross module. We can leave a note that it's not intended for external usage.

botocore/httpchecksum.py
```diff
@@ -280,35 +290,30 @@ def resolve_request_checksum_algorithm(
...
-            "name": f"x-amz-checksum-{algorithm_name}",
-        }
-
-        if algorithm["name"] in request["headers"]:
-            # If the header is already set by the customer, skip calculation
-            return
+    location_type = "header"
+    if operation_model.has_streaming_input:
+        # Operations with streaming input must support trailers.
+        if request["url"].startswith("https:"):
```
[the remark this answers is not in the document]
nateprewitt: Right, I was asking if this can be collapsed into:

```python
    if (
        operation_model.has_streaming_input
        and urlparse(request["url"]).scheme == "https"
    ): 
        # Operations with streaming input must support trailers.
```

or is there a reason we've made them distinct conditionals?

botocore/handlers.py
```diff
@@ -1271,6 +1269,17 @@ def _update_status_code(response, **kwargs):
...
 
+def handle_request_validation_mode_member(params, model, **kwargs):
+    client_config = kwargs.get("context", {}).get("client_config")
+    if client_config is None:
+        return
+    response_checksum_validation = client_config.response_checksum_validation
+    http_checksum = model.http_checksum
+    mode_member = http_checksum.get("requestValidationModeMember")
+    if mode_member and response_checksum_validation == "when_supported":
+        params.setdefault(mode_member, "ENABLED")
```
nateprewitt, line 1280: Why do we need to do this as a handler instead of part of the actual code base?

tests/functional/test_s3.py
```diff
@@ -3702,10 +3737,12 @@ def _verify_presigned_url_addressing(
 
 
 class TestS3XMLPayloadEscape(BaseS3OperationTest):
-    def assert_correct_content_md5(self, request):
-        content_md5_bytes = get_md5(request.body).digest()
-        content_md5 = base64.b64encode(content_md5_bytes)
-        self.assertEqual(content_md5, request.headers["Content-MD5"])
+    def assert_correct_crc32_checksum(self, request):
+        checksum = CrtCrc32Checksum() if HAS_CRT else Crc32Checksum()
```
nateprewitt, line 3741: Should we be surfacing `_CHECKSUM_CLASS` in some way to do this automatically. Otherwise, we're desyncing the code from the test.

tests/unit/test_httpchecksum.py
```diff
@@ -106,20 +111,26 @@ def test_request_checksum_algorithm_model_opt_in(self):
             http_checksum={"requestAlgorithmMember": "Algorithm"}
         )
 
-        # Param is not present, no checksum will be set
+        # Param is not present, crc32 checksum will be set by default
```
nateprewitt, line 114: The test name isn't really correct anymore. This is specifically that we won't turn on checksums by default because it's breaking. Now that S3 has decided they're no longer upholding that, we're not checking that things are opt-in. This may be better as `test_request_checksum_algorithm_model_default`.

jonathan343: Updated in 75db0fe6792083943d1af499275fbcfea77b3666. Thanks for calling this out.

botocore/httpchecksum.py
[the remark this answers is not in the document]
jonathan343: Updated in 75db0fe6792083943d1af499275fbcfea77b3666

botocore/utils.py
```diff
@@ -3255,7 +3255,9 @@ def _is_s3express_request(params):
     return endpoint_properties.get('backend') == 'S3Express'
 
 
-def _has_checksum_header(params):
+# This is not a public interface and is subject to abrupt breaking changes.
+# Any usage is not advised or supported in external code bases.
```
nateprewitt, line 3259: Let's do this comment inside the function rather than before it.

botocore/httpchecksum.py
```diff
@@ -259,7 +258,18 @@ def resolve_request_checksum_algorithm(
...
+    # If the header is already set by the customer, skip calculation
+    if _has_checksum_header(request):
+        return
+
+    request_checksum_calculation = request["context"][
+        "client_config"
+    ].request_checksum_calculation
     http_checksum = operation_model.http_checksum
+    request_checksum_required = (
+        operation_model.http_checksum_required
```
SamRemis, line 270: Since httpChecksumRequired is being deprecated in favor of requestChecksumRequired, do we want to add something to make sure we will delete this once all operations migrate to requestChecksumRequired and remove the old logic?

botocore/httpchecksum.py
```diff
@@ -259,7 +258,18 @@ def resolve_request_checksum_algorithm(
     params,
     supported_algorithms=None,
 ):
+    # If the header is already set by the customer, skip calculation
+    if _has_checksum_header(request):
+        return
+
+    request_checksum_calculation = request["context"][
```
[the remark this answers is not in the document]
nateprewitt: We know it'll be there: https://github.com/jonathan343/botocore/blob/66d047b5cdb033e4406e306afc5ab1c3e4785f16/botocore/client.py#L958

botocore/httpchecksum.py
nateprewitt: We will likely never be migrated in this case. If that does happen, that'll be a larger initiative.

botocore/httpchecksum.py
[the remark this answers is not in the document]
jonathan343: Missed this comment yesterday. That makes much more sense, I misinterpreted your question. I updated this in 009ea3d3b3c07886c07863673d3dd471c8241647

botocore/utils.py
jonathan343: Updated in 009ea3d3b3c07886c07863673d3dd471c8241647. Let me know if the wording looks good.

tests/functional/test_s3.py
```diff
@@ -28,6 +27,7 @@
     UnsupportedS3AccesspointConfigurationError,
     UnsupportedS3ConfigurationError,
 )
+from botocore.httpchecksum import HAS_CRT, Crc32Checksum, CrtCrc32Checksum
```
nateprewitt, line 30: We shouldn't be doing transitive imports. `HAS_CRT` comes from botocore.compat, so we should add that on line 21.
```suggestion
from botocore.httpchecksum import Crc32Checksum, CrtCrc32Checksum
```

tests/functional/test_s3.py
nateprewitt: Yep, I don't think we should be importing the private variable directly in code. For this test harness since it's not library code, we may establish a codified reliance on this even those it's private. That way the test fails on removal and it's clear what we're doing.

```python
import botocore.httpchecksum
[...]
def get_default_checksum_cls():
    # This pass through is grabbing our internally supported list of checksums
    # to ensure we stay in sync, while not exposing them publicly.
    return botocore.httpchecksum._CHECKSUM_CLS["crc32"]
```

tests/functional/test_s3.py
```diff
@@ -1519,30 +1519,67 @@ def setUp(self):
...
         with self.http_stubber:
             self.client.put_object(Bucket="foo", Key="bar", Body="baz")
-        self.assertIn("content-md5", self.get_sent_headers())
+        sent_headers = self.get_sent_headers()
+        self.assertEqual(sent_headers["Content-Encoding"], b"aws-chunked")
+        self.assertEqual(sent_headers["Transfer-Encoding"], b"chunked")
+        self.assertEqual(
+            sent_headers["X-Amz-Trailer"], b"x-amz-checksum-crc32"
+        )
+        self.assertEqual(sent_headers["X-Amz-Decoded-Content-Length"], b"3")
```
nateprewitt, line 1531: Should we be validating the checksum algo header we're adding as well?

botocore/httpchecksum.py
```diff
@@ -259,7 +259,19 @@ def resolve_request_checksum_algorithm(
     params,
     supported_algorithms=None,
 ):
+    # If the header is already set by the customer, skip calculation
+    if has_checksum_header(request):
+        return
+
+    extra_headers = {}
```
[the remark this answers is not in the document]
jonathan343: I updated this to instead store the following in our request checksum context:
```
"request_algorithm_header": {
    "name": "foo",
    "value": "bar",
},
```
We can later check `if "request_algorithm_header" in request["context"]["checksum"]` and apply the header if it exists. Let me know if that's better.

tests/functional/test_s3.py
jonathan343: Updated in the latest commit.

jonathan343:
This work is being tracked internally.

PR #3279, opened 2024-10-15
Sigv4a signing region set config bugfix

SamRemis:
Fixes a bug where the sigv4a_signing_region_set configuration in the config file and the corresponding environment variable would be ignored

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -276,6 +276,7 @@ class ClientArgsCreator:
         self._compute_connect_timeout(config_kwargs)
         self._compute_user_agent_appid_config(config_kwargs)
         self._compute_request_compression_config(config_kwargs)
+        self._compute_sigv4a_signing_region_set(config_kwargs)
         s3_config = self.compute_s3_config(client_config)
 
         is_s3_service = self._is_s3_service(service_name)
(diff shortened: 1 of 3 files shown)


botocore/args.py
```diff
@@ -771,3 +772,17 @@ def _compute_user_agent_appid_config(self, config_kwargs):
...
+            'sigv4a_signing_region_set'
+        )
+        if sigv4a_signing_region_set is None:
+            sigv4a_signing_region_set = self._config_store.get_config_variable(
+                'sigv4a_signing_region_set'
+            )
+        if not sigv4a_signing_region_set:
+            raise botocore.exceptions.InvalidConfigError(
+                error_msg="sigv4a_signing_region_set configuration must be non-empty"
+            )
```
nateprewitt, line 787: This doesn't seem like something we should be evaluating. The default is `None` which means we're going to fail here with the default in every case. We only fallback to `*` in the handler which I believe was a specific design decision made in your last PR.

nateprewitt (changes requested):
Can we add a functional test that exercises the ConfigProvider to ensure we don't regress this again in the future?

tests/functional/test_auth_config.py
```diff
@@ -75,3 +75,17 @@ def assert_all_requirements_match(auth_config, message):
...
+    environ = {
+        'AWS_CONFIG_FILE': 'no-exist-foo',
+        'AWS_SIGV4A_SIGNING_REGION_SET': 'foo',
+    }
+    with mock.patch('os.environ', environ):
+        session = create_session()
+        session.config_filename = 'no-exist-foo'
+        session.create_client('s3')
+        assert (
+            session.get_config_variable('sigv4a_signing_region_set') == 'foo'
```
hssyoo, line 90: I think this assertion passes without the changes in this commit

hssyoo (approved):
🏆

aemous (commented):
Nit: Looks like the `message` arg in `test_auth_config#assert_all_requirements_match` is unused.

tests/functional/test_auth_config.py
```diff
@@ -75,3 +75,15 @@ def assert_all_requirements_match(auth_config, message):
         AUTH_TYPE_REQUIREMENTS[auth_type] for auth_type in auth_config
     )
     assert len(auth_requirements) == 1
+
+
+def test_sigv4a_signing_region_set_config_from_environment():
```
nateprewitt, line 80: It may be worth parameterizing this so we're testing environment, set Config, and a default Config to exercise all your code paths in `_compute_sigv4a_signing_region_set_config`.

nateprewitt (changes requested):

nateprewitt (approved):

Issue #3284, opened 2024-10-22
Botocore `1.35.45` breaks S3:SelectObjectContent


bpandola:
### Describe the bug

The S3 action `SelectObjectContent` fails with the latest version of botocore.  An exception is raised in the recently-added `_handle_200_error` handler (see PR #3276).

### Regression Issue

- [X] Select this option if this issue appears to be a regression.

### Expected Behavior

Confirmed working in previous version(s) of botocore.

### Current Behavior

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../env/python3.11/lib/python3.11/site-packages/botocore/client.py:569: in _api_call
    return self._make_api_call(operation_name, kwargs)
../env/python3.11/lib/python3.11/site-packages/botocore/client.py:1005: in _make_api_call
    http, parsed_response = self._make_request(
../env/python3.11/lib/python3.11/site-packages/botocore/client.py:1029: in _make_request
    return self._endpoint.make_request(operation_model, request_dict)
../env/python3.11/lib/python3.11/site-packages/botocore/endpoint.py:119: in make_request
    return self._send_request(request_dict, operation_model)
../env/python3.11/lib/python3.11/site-packages/botocore/endpoint.py:197: in _send_request
    success_response, exception = self._get_response(
../env/python3.11/lib/python3.11/site-packages/botocore/endpoint.py:239: in _get_response
    success_response, exception = self._do_get_response(
../env/python3.11/lib/python3.11/site-packages/botocore/endpoint.py:306: in _do_get_response
    self._event_emitter.emit(
../env/python3.11/lib/python3.11/site-packages/botocore/hooks.py:412: in emit
    return self._emitter.emit(aliased_event_name, **kwargs)
../env/python3.11/lib/python3.11/site-packages/botocore/hooks.py:256: in emit
    return self._emit(event_name, kwargs)
../env/python3.11/lib/python3.11/site-packages/botocore/hooks.py:239: in _emit
    response = handler(**kwargs)
../env/python3.11/lib/python3.11/site-packages/botocore/handlers.py:1252: in _handle_200_error
    if _looks_like_special_case_error(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[...]

PR #3285, opened 2024-10-22
Fix event stream regression

alexgromero:
This PR addresses a regression in the `SelectObjectContent` functionality that was introduced in Botocore `1.35.45`. This issue was identified in issue #3284.


nateprewitt (approved):
:shipit:

PR #3288, opened 2024-10-23
Fix S3 200 Error handling

alexgromero:
This PR updates the S3 200 Errors implementation (https://github.com/boto/botocore/pull/3276) to address bugs introduced in payload operations.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -1245,11 +1245,7 @@ def document_expires_shape(section, event_name, **kwargs):
 def _handle_200_error(operation_model, response_dict, **kwargs):
     # S3 can return a 200 response with an error embedded in the body.
     # Convert the 200 to a 500 for retry resolution in ``_update_status_code``.
-    if (
-        not response_dict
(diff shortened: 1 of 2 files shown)


botocore/handlers.py
```diff
@@ -1262,6 +1258,25 @@ def _handle_200_error(operation_model, response_dict, **kwargs):
...
+        or not output_shape
+    ):
+        return False
+    payload = output_shape.serialization.get('payload')
+    if payload is not None:
+        payload_shape = output_shape.members[payload]
+        if (
+            payload_shape.type_name == 'blob'
+            or payload_shape.type_name == 'string'
+        ):
```
nateprewitt, line 1275:
```suggestion
        if payload_shape.type_name in ('blob', 'string'):
```

nateprewitt (changes requested):
nit; otherwise looks good. Thanks @alexgromero!

nateprewitt (approved):

PR #3292, opened 2024-10-30
Bucket name doc updates

alexgromero:
This PR updates the bucket names to meet the policy which requires that special reserved bucket names be used for all examples and documentation content.


nateprewitt (approved):
🚢

jonathan343 (approved):

PR #3293, opened 2024-10-31
Inherit from standard ConnectionError in boto3.exceptions.ConnectionError

daniel-leicht:
Please consider pulling this commit where boto3.exceptions.ConnectionError has multiple inheritance from the built-in ConnectionError and BotoCoreError.

It improves compatibility with error handling in Python by allowing `ConnectionError` from `boto3.exceptions` to be caught alongside other `ConnectionError` exceptions.

### DIFF
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -107,7 +107,7 @@ class HTTPClientError(BotoCoreError):
         )
 
 
-class ConnectionError(BotoCoreError):
+class ConnectionError(BotoCoreError, ConnectionError):
     fmt = 'An HTTP Client failed to establish a connection: {error}'
 
 


nateprewitt:
Hi @daniel-leicht, this would be a breaking change for the library. Adding `ConnectionError` has generally been considered a mistake in other Python networking libraries like Requests and Urllib3 as it inaccurately conflates the exception origin for the excepter. If you're seeing actual `ConnectionError` exceptions leaked to you as an end user, that's another discussion.

Can you give some specific examples of how you arrived at this PR being needed?

daniel-leicht:
Hi @nateprewitt , I’m using a library that relies on botocore, and it currently raises the botocore.exceptions.ConnectionError directly. I’d like to handle this exception and avoid importing it from botocore, as my project itself does not use it, only the library I rely on.

If incorporating Python's built-in ConnectionError into this exception could cause issues, I completely understand, and feel free to close the PR if that’s the case. Thank you!

nateprewitt:
Thanks for the info, @daniel-leicht! From a principle of how we've developed most of Botocore, we try to keep exceptions raised within our namespace. I definitely understand how that can be cumbersome if there's another library unexpectedly surfacing our exceptions. Ideally, they're handling those (or wrapping them) if you're not intended to be aware of botocore's usage.

I'm not sure we can use multiple inheritance to use any of the std lib exceptions since that may change control flow in existing applications. I think the best option for now may be to reach out to the library author to see if they have any considerations for handling these, or you can use `botocore.exceptions.BotocoreError` as a catch all for our exceptions to avoid having to micromanage imports.

I'll close this out for now since I don't think we can immediately address the issue in a backwards compatible manner.

PR #3239, opened 2024-08-16
Resolve Python 3.12 `.utcnow()` `DeprecationWarning`

kurtmckee:
This PR resolves the Python 3.12 `DeprecationWarning` thrown when botocore calls `datetime.datetime.utcnow()`.

@nateprewitt, this PR differs from other attempts posted to this repo by _ensuring `datetime` instances are still timezone-unaware_:

```python
import datetime

# Current form: timezone-unaware, raises DeprecationWarnings in >=py3.12
datetime.datetime.utcnow()

# Common transformation -- makes the instance timezone-aware!
datetime.datetime.now(datetime.timezone.utc)

# This PR's approach -- ensures instance is still timezone-unaware
datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)
```

I ran this through CI in my own repo and [it passes cleanly](https://github.com/kurtmckee/pr-botocore/actions).

Fixes boto/boto3#3889.

PR #3294, opened 2024-11-03
Replaced deprecated datetime.utcnow() with datetime.now(datetime.UTC)

gautamgiri-dev:
Fixes: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: `datetime.datetime.now(datetime.UTC)`.
  `datetime_now = datetime.datetime.utcnow()`
warning that shows due to change in the behavior of datetime module in latest python

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -421,7 +421,7 @@ class SigV4Auth(BaseSigner):
     def add_auth(self, request):
         if self.credentials is None:
             raise NoCredentialsError()
-        datetime_now = datetime.datetime.utcnow()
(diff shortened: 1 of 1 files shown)


kurtmckee (commented):
These are not the only places that reference `.utcnow()`, and this change makes the datetime instances timezone-aware.

There have been multiple PRs to address this, but I think that #3239 is the only one that doesn't make the datetime instances timezone-aware.

gautamgiri-dev:
Yes. You are right @kurtmckee. Sorry for hurrying in opening the PR without understanding the complete picture.

PR #3295, opened 2024-11-04
Query compatibility mode header

alexgromero:
This PR adds a custom header to inform a service that the SDK is running in query compatibility mode. This change enables support for services migrating away from the AWS Query protocol.

### DIFF
--- a/.changes/next-release/enhancement-QueryCompatibilityModeHeader-5257.json
+++ b/.changes/next-release/enhancement-QueryCompatibilityModeHeader-5257.json
@@ -1,5 +1,5 @@
 {
   "type": "enhancement",
   "category": "Query Compatibility Mode Header",
-  "description": "Adds logic for custom header to support services migrating away from the AWS Query protocol."
+  "description": "Added support for header enabling service migration off the AWS Query protocol."
 }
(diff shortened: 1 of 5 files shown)


.changes/next-release/enhancement-QueryCompatibilityModeHeader-5257.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Query Compatibility Mode Header",
+  "description": "Adds logic for custom header to support services migrating away from the AWS Query protocol."
```
nateprewitt, line 4: super nit; We may consider avoiding wording like "Adding logic" or "custom header" in the release notes. Everything we write will probably fall under those categories so they're more filler words than informative. Especially for features like this that don't have a lot of external visibility to the end-user, it's good to keep the brief short and leave any additional details for the customer in the PR if needed.

```suggestion
  "description": "Added support for header enabling service migration off the AWS Query protocol."
```

or

```suggestion
  "description": "New header added to enable service migration off the AWS Query protocol."
```

nateprewitt (changes requested):
Minor notes on the changelog but the feature otherwise looks good. Thanks @alexgromero!

nateprewitt (approved):

PR #3296, opened 2024-11-05
Fixes the issue in EndpointExceptionError by changing the format of the error message.

ubaskota:
The current base class for EndpointResolutionError takes variable-length arguments; kwarg, and passing a positional argument to it returns a TypeError. This fix solves it by passing an argument in a proper format.

### DIFF
--- a/botocore/endpoint_provider.py
+++ b/botocore/endpoint_provider.py
@@ -638,7 +638,7 @@ class ParameterDefinition:
                 return self.default
             if self.required:
                 raise EndpointResolutionError(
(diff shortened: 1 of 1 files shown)


nateprewitt (approved):
Thanks for the PR, @ubaskota! It's probably worth adding a quick note to the description to the effect of:

> This PR addresses a miss in the initial implementation of error handling in our EndpointProvider. The missing `kwarg` for passing the exception message results in an error being raised while trying to raise our `EndpointResolutionError`.

jonathan343 (approved):

PR #3301, opened 2024-11-14
Add string array context params

alexgromero:


botocore/endpoint_provider.py
```diff
@@ -577,6 +577,7 @@ class ParameterType(Enum):
 
     string = str
     boolean = bool
+    stringarray = tuple
```
jonathan343, line 580: Why not use `string_array` to be more idiomatic?

botocore/endpoint_provider.py
```diff
@@ -600,7 +601,7 @@ def __init__(
         except AttributeError:
             raise EndpointResolutionError(
                 msg=f"Unknown parameter type: {parameter_type}. "
-                "A parameter must be of type string or boolean."
+                "A parameter must be of type string, boolean, or stringarray."
```
jonathan343, line 604: `stringarray` isn't a type in python. We should probably instead say something like `list[str]` which lets the user know they should use a list of strings.

botocore/regions.py
```diff
@@ -600,6 +609,14 @@ def _resolve_param_as_client_context_param(self, param_name):
             client_ctx_varname = client_ctx_params[param_name]
             return self._client_context.get(client_ctx_varname)
 
+    def _resolve_param_as_operation_context_param(
```
jonathan343, line 612: We don't see to be handling default values when the following parameter is modeled:
```
"defaultStringParam": { 
    "type": "StringArray",
    "required": true,
    "default": ["A", "B", "C"] 
}
```
This seems like a miss.

botocore/utils.py
```diff
@@ -1491,6 +1491,9 @@ def func_with_weakref(weakref_to_self, *args, **kwargs):
 
         @functools.wraps(func)
         def inner(self, *args, **kwargs):
+            for kwarg_key, kwarg_value in kwargs.items():
+                if isinstance(kwarg_value, list):
+                    kwargs[kwarg_key] = tuple(kwarg_value)
```
jonathan343, line 1496: Have we manually tested that we get a cache hit when using operationContextParams?

tests/unit/data/endpoints/valid-rules/array-index.json
```diff
@@ -0,0 +1,45 @@
+{
+  "version": "1.3",
+  "parameters": {
+    "ResourceList": {
+      "type": "stringArray"
+    }
```
jonathan343, line 6: Related to my pervious comment: We should included tests for string array parameters that define default values.

jonathan343 (changes requested):
Looks like a great start! I left some feedback and questions that I think should be addressed. Let me know if you have any questions.

botocore/endpoint_provider.py
alexgromero: `stringarray` isn't a type, but we have added it as an enum value in the class `ParameterType` which is a "translation from `type` attribute to native Python type."

botocore/utils.py
alexgromero: Yes, we manually tested this and we got a cache hit.

botocore/regions.py
alexgromero: Good catch! Will add in the next commit.

botocore/utils.py
SamRemis: Yes, we ran a test with some hypothetical models and ensured that the endpoint was not resolved twice.  This code is necessary to run before we check the cache, so these few lines do get called multiple times.

botocore/regions.py
SamRemis: Defaults are handled elsewhere in the code.

When we are making a representation of the ruleset to attach to the client, we make a [parameter definition](https://github.com/boto/botocore/blob/b3cb3452225877f7fa823ab488d86bacd3211be9/botocore/endpoint_provider.py#L583) where we define a default if it is existed on the client.

This code is run later during endpoint resolution; we will try to evaluate that parameter.  If a default exists and we fail to resolve a value any other way, we will fall back to that default.  

While this PR does not require any changes to make the default functional, the testing bit below is a good callout.  I'll resolve this comment for now since it won't require a code change, but keep the comment about the testing until that is addressed.

botocore/endpoint_provider.py
SamRemis: This is the definition of a parameter, where there are custom, language-agnostic types.  This is defined in the ruleset and not by the end-users; here we are checking the name of the parameter type.

We are only checking the name of the parameter that's defined in the ruleset, for instance as it is shown in the test below:

```
  "parameters": {
    "ResourceList": {
      "type": "stringArray"
    }
  }
```

The valid types are `string`, `boolean`, and `stringArray`.  All of these get lowercased on line 599, so I've kept it lowercase in the error message, but I'm happy to add back in the uppercase or modify the message as you see fit.  

Based on this, what do you think is the right formatting for the message?

botocore/endpoint_provider.py
SamRemis: This comes straight from the model and needs to match what is in there.  We can do a transformation on it if that makes sense; right now we are only calling `lower()` on it.  Making it `string_array` would require us to do do a more complex transformation, which I prefer to keep simple unless we have a good reason to make it more idiomatic.

I'm not closed off to the idea though

jonathan343: Got it, I think it makes sense to leave it as `stringarray` then. Thanks

botocore/endpoint_provider.py
jonathan343: Thank you both for the explanation. Let's leave this as `stringarray`.

jonathan343 (approved):
LGTM 🚀

PR #3302, opened 2024-11-15
Add support for tuples in get_attr

SamRemis:
The get_attr function is intended to be able to support an input with a path such as `[0]` and a value of `['foo']`and parse out the fist index of the value to get `foo` as the result.  

Without the change in this PR, the such an input will not match the indexing regex, and instead it will reach the else statement and reach the code intended to access dictionary indexes.  This will fail with a type error, so this PR updates the regex to properly reach the part of the code that handles indexes without a word string before them.  

That code still fails for the same input; it assumes that it is getting the index of a list from a dictionary, so it attempts to call `.get()` on the base input.  This will not work if the input value is not a dictionary, so we need to add the conditional to only call `.get()` if there is a string before the first open bracket in the part of the path that we are parsing.  

Finally, this code can't be reached currently with a list due to how we memoize input parameters to endpoints; all parameter types have to be hash-able. While this function could theoretically get called with a list as input from another context, this is a private interface and it is only internally called from the endpoint provider.  Because of this, I've updated the docblock to specify `tuple` as the input type instead of `list`.

### DIFF
--- a/botocore/endpoint_provider.py
+++ b/botocore/endpoint_provider.py
@@ -173,9 +173,9 @@ class RuleSetStandardLibrary:
         :type path: str
         :rtype: Any
         """
-        if value is None:
-            return None
         for part in path.split("."):
+            if value is None:
+                return None
             match = GET_ATTR_RE.search(part)
             if match is not None:
                 name, index = match.groups()
--- a/botocore/endpoint_provider.py
+++ b/botocore/endpoint_provider.py
@@ -173,12 +173,14 @@ class RuleSetStandardLibrary:
         :type path: str
         :rtype: Any
         """
+        if value is None:
+            return None
         for part in path.split("."):
             match = GET_ATTR_RE.search(part)
             if match is not None:
                 name, index = match.groups()
                 index = int(index)
-                if name and value is not None:
+                if name:
                     value = value.get(name)
                 if index >= len(value):
                     return None
--- a/botocore/endpoint_provider.py
+++ b/botocore/endpoint_provider.py
@@ -42,7 +42,7 @@ from botocore.utils import (
 logger = logging.getLogger(__name__)
 
 TEMPLATE_STRING_RE = re.compile(r"\{[a-zA-Z#]+\}")
-GET_ATTR_RE = re.compile(r"(\w+)\[(\d+)\]")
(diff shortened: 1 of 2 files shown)


alexgromero (approved):
LGTM

botocore/endpoint_provider.py
```diff
@@ -169,17 +169,20 @@ def get_attr(self, value, path):
...
             if match is not None:
                 name, index = match.groups()
                 index = int(index)
-                value = value.get(name)
-                if value is None or index >= len(value):
+                if name:
+                    value = value.get(name)
+                if index >= len(value):
                     return None
                 return value[index]
```
nateprewitt, line 187: This seems broken compared to our old behavior. We're setting `value = value.get(name)` which can be None. We then assume it's non-None in the index check and again index into `value` on line 187. Is there something else mitigating that issue?

Do we have a more comprehensive set of tests to make sure this function does what we want?

tests/unit/test_endpoint_provider.py
```diff
@@ -510,3 +510,23 @@ def test_aws_is_virtual_hostable_s3_bucket_allow_subdomains(
...
+
+
+def test_get_attr_returns_none_on_too_high_index(rule_lib):
+    result = rule_lib.get_attr({"foo": ['bar']}, 'foo[1]')
+    assert result is None
+
+
+def test_get_attr_can_get_list_index(rule_lib):
+    result = rule_lib.get_attr(("foo",), '[0]')
+    assert result == "foo"
```
nateprewitt, line 532: If there's not a strong reason for breaking the tests up, we should try to parametrize similar tests.
```suggestion
@pytest.mark.parametrize(
    "value, path, expected_value",
    [
        ({"foo": ['bar']}, 'baz[0]', None),  # Missing index
        ({"foo": ['bar']}, 'foo[1]', None),  # Out of range index
        ({"foo": ['bar']}, 'foo[0]', "bar"),  # Named index
        (("foo",), '[0]', "foo"),  # Bare index
    ],
)
def test_get_attr(rule_lib, value, path, expected_value):
    result = rule_lib.get_attr(value, path)
    assert result == expected_value
```

nateprewitt (approved):

Issue #3305, opened 2024-11-18
ECS: Waiter: Increase ServicesStable Waiter Time to "x" Time (30 or 60 minutes?)


colinbjohnson:
### Describe the feature

I would like to propose that AWS increase the "ServicesStable" waiter time (link as of 2024-11-18: https://github.com/boto/botocore/blob/develop/botocore/data/ecs/2014-11-13/waiters-2.json#L42-L72) to _some_ amount of time greater than 10 minutes.

### Use Case

We (and a number of AWS customers) that use the `aws ecs wait services-stable` command (or similar commands within SDKs) may have deployments that take > 10 minutes to become stable.

### Proposed Solution

* Near Term Solution: change this to 30 or 60 minutes
* Long Term Solution: expose the waiter's `delay` and `maxAttempts` values to the SDK and/or AWS CLI (you can see this ticket here: https://github.com/aws/aws-cli/issues/1295 where there are three comments about this)

### Other Information

# Alternative Solutions

* writing our own waiter (you can see these online)
  * https://github.com/misterjoshua/ecs-deployment-wait
* wrapping the AWS waiter with code so the waiter _itself_ is called repeatedly
* using a different method for determining if/when a deployment is complete (EventBridge?)
* not utilizing a waiter (and assuming the deployment succeeded)
* using a different method for deploying ECS Service updates
* something else?

# Other Information

* This _feels_ like something (near-term, changing the values for `delay` and `maxAttempts`) that would be relatively easy to do (I submitted a PR: https://github.com/boto/botocore/issues/3305)
* There is some AWS documentation that would need to be updated, most notably:
  * https://docs.aws.amazon.com/cli/latest/reference/ecs/wait/services-stable.html
  * https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ecs/wait/services-stable.html

### Acknowledgements

- [X] I may be able to implement this feature request
- [X] This feature might incur a breaking change

### SDK version used

aws-cli/2.15.0 Python/3.11.6

### Environment details (OS name and version, etc.)

Darwin/24.1.0

colinbjohnson:
Am closing this issue - I understand that the "waiter configuration" is owned by the service teams. I've opened this issue instead: [[ECS] Waiter: Increase ServicesStable Waiter Time to "x" Time (30 or 60 minutes?)](https://github.com/aws/containers-roadmap/issues/2474)

PR #3306, opened 2024-11-19
move to 30 minute timeout for aws ecs wait services-stable

colinbjohnson:
This moves the `aws ecs wait services-stable` to having a 30 minute timeout. I don't have any data on if 30 minutes is ideal but anecdotally based on requests and internet posts 10 minutes may _not_ be enough.

This is in response to issue [[ECS] Waiter: Increase ServicesStable Waiter Time to "x" Time (30 or 60 minutes?)](https://github.com/aws/containers-roadmap/issues/2474) and, previously, https://github.com/boto/botocore/issues/3305 .

I do believe the appropriate long term solution is to expose this variable - a ticket for the AWS CLI has already been opened, originally on April 17, 2015 (https://github.com/aws/aws-cli/issues/1295).


adev-code:
I have submitted a request to the Service team in this regard. We do not have a timeline yet on when/if they will release the change. 

For Internal Ref: P368550249

PR #3332, opened 2025-01-08
Add account ID to the environment variable credential provider

alexgromero:
### Overview
This PR introduces `account_id` to the environment variable credential provider. With this update, users can now provide an `AWS_ACCOUNT_ID` environment variable, which will be associated with the credentials when they are loaded.

### Testing
Functional testing to ensure the `account_id` gets propagated and resolved to the correct endpoint for each credential provider will be added in an upcoming PR.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -312,7 +312,7 @@ class Credentials:
     :param str token: The security token, valid only for session credentials.
     :param str method: A string which identifies where the credentials
         were found.
-    :param str account_id: The account ID associated with the credentials.
+    :param str account_id: (optional) An account ID associated with the credentials.
     """
 
     def __init__(
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -312,6 +312,7 @@ class Credentials:
     :param str token: The security token, valid only for session credentials.
     :param str method: A string which identifies where the credentials
         were found.
+    :param str account_id: The account ID associated with the credentials.
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


botocore/credentials.py
```diff
@@ -312,6 +312,7 @@ class Credentials:
     :param str token: The security token, valid only for session credentials.
     :param str method: A string which identifies where the credentials
         were found.
+    :param str account_id: The account ID associated with the credentials.
```
alexgromero, line 315: Forgot to add param comment for `account_id` in previous PR #3328.

botocore/credentials.py
```diff
@@ -1127,8 +1129,12 @@ def __init__(self, environ=None, mapping=None):
         :param mapping: An optional mapping of variable names to
             environment variable names.  Use this if you want to
             change the mapping of access_key->AWS_ACCESS_KEY_ID, etc.
-            The dict can have up to 3 keys: ``access_key``, ``secret_key``,
-            ``session_token``.
+            The dict can have up to 5 keys:
```
alexgromero, line 1132: The current mapping is:
```python
{
    'access_key': 'AWS_ACCESS_KEY_ID',
    'secret_key': 'AWS_SECRET_ACCESS_KEY',
    'token': ['AWS_SECURITY_TOKEN', 'AWS_SESSION_TOKEN'],
    'expiry_time': 'AWS_CREDENTIAL_EXPIRATION',
    'account_id': 'AWS_ACCOUNT_ID',
}

File: botocore/credentials.py (hunk 1 above)
Comment (nateprewitt) [reply botocore/credentials.py:315]: We should probably briefly note the that this is an optional field. Either in the string or with something like:
```suggestion
    :param str account_id: (optional) An account ID associated with the credentials.
```

jonathan343 (approved):

PR #3336, opened 2025-01-14
Removes inline scripts and inline styles in botocore documentation site to make it compatible with the newly added security headers

ubaskota:
Due to the recently added Content Security Policy(CSP), all inline scripts and styles are automatically blocked. In order to fix this, we had to move the inline scripts and styles, and keep them in separate files. This change performs all necessary edits to ensure that the documentation site for Botocore works as expected along with the newly added security headers.


jonathan343 (approved):
LGTM 🚀

jonathan343 (commented):
LGTM! 🚢

PR #3345, opened 2025-01-21
Add account_id_endpoint_mode and consume account_id in ruleset

alexgromero:
## Account ID Endpoint Mode
This PR introduces the `account_id_endpoint_mode` configuration setting, which determines the client’s behavior for account ID based endpoint routing. When enabled and configured in a service’s ruleset, the endpoint resolver will attempt to resolve the account ID as an input parameter to the endpoint provider. The valid values for this setting are:
* `preferred`: The endpoint should include the account ID if available.
* `disabled`: The endpoint will not include the account ID, even if available.
* `required`: The endpoint must include the account ID. An exception will be raised if the account ID is unavailable.

If no value is provided, the default behavior will be `preferred`.

## Consuming Account ID in the Ruleset
This PR also adds support for consuming the account ID in the ruleset during endpoint resolution. Specifically, the following built-in parameters will now be resolved:
* `AWS::Auth::AccountId`: The AWS account ID that can be optionally configured for the SDK client.
* `AWS::Auth::AccountIdEndpointMode`: The setting that determines whether the endpoint should include the account ID.

Account ID will be treated as a deferred property, using the `get_deferred_property` method in the `Credentials` class. This ensures that the account ID is resolved only when needed, deferring the resource-intensive resolution of account ID tied to `DeferredRefreshableCredentials` until endpoint resolution time. This approach avoids unnecessary credential fetches during client initialization and aligns with the behavior of deferred credentials.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -880,19 +880,28 @@ class ClientArgsCreator:
         config_kwargs[config_key] = value
 
     def _compute_account_id_endpoint_mode_config(self, config_kwargs):
-        account_id_endpoint_mode = config_kwargs.get(
-            'account_id_endpoint_mode'
-        )
+        config_key = 'account_id_endpoint_mode'
+        account_id_endpoint_mode = config_kwargs.get(config_key)
+
         if account_id_endpoint_mode is None:
             account_id_endpoint_mode = self._config_store.get_config_variable(
-                'account_id_endpoint_mode'
+                config_key
             )
+
+        if isinstance(account_id_endpoint_mode, str):
+            account_id_endpoint_mode = account_id_endpoint_mode.lower()
+
         if (
             account_id_endpoint_mode
             not in VALID_ACCOUNT_ID_ENDPOINT_MODE_CONFIG
         ):
             raise botocore.exceptions.InvalidConfigError(
-                error_msg=f"The configured value '{account_id_endpoint_mode}' for 'account_id_endpoint_mode' is "
+                error_msg=f"The configured value '{account_id_endpoint_mode}' for '{config_key}' is "
                 f"invalid. Valid values are: {VALID_ACCOUNT_ID_ENDPOINT_MODE_CONFIG}."
             )
-        config_kwargs['account_id_endpoint_mode'] = account_id_endpoint_mode
+
+        signature_version = config_kwargs.get('signature_version')
+        if signature_version is botocore.UNSIGNED:
+            account_id_endpoint_mode = 'disabled'
+
+        config_kwargs[config_key] = account_id_endpoint_mode
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -78,6 +78,12 @@ PRIORITY_ORDERED_SUPPORTED_PROTOCOLS = (
     'ec2',
 )
 
+VALID_ACCOUNT_ID_ENDPOINT_MODE_CONFIG = (
+    'preferred',
+    'disabled',
+    'required',
+)
+
 
 class ClientArgsCreator:
     def __init__(
@@ -136,6 +142,7 @@ class ClientArgsCreator:
         configured_endpoint_url = final_args['configured_endpoint_url']
         signing_region = endpoint_config['signing_region']
         endpoint_region_name = endpoint_config['region_name']
+        endpoint_mode = config_kwargs['account_id_endpoint_mode']
 
         event_emitter = copy.copy(self._event_emitter)
         signer = RequestSigner(
@@ -183,6 +190,8 @@ class ClientArgsCreator:
             is_secure,
             endpoint_bridge,
             event_emitter,
+            credentials,
+            endpoint_mode,
         )
 
         # Copy the session's user agent factory and adds client configuration.
@@ -294,6 +303,7 @@ class ClientArgsCreator:
                 response_checksum_validation=(
                     client_config.response_checksum_validation
                 ),
+                account_id_endpoint_mode=client_config.account_id_endpoint_mode,
             )
         self._compute_retry_config(config_kwargs)
         self._compute_connect_timeout(config_kwargs)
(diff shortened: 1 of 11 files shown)


botocore/args.py
```diff
@@ -781,3 +804,21 @@ def _compute_sigv4a_signing_region_set_config(self, config_kwargs):
                 'sigv4a_signing_region_set'
             )
         config_kwargs['sigv4a_signing_region_set'] = sigv4a_signing_region_set
+
+    def _compute_account_id_endpoint_mode_config(self, config_kwargs):
+        account_id_endpoint_mode = config_kwargs.get(
+            'account_id_endpoint_mode'
```
jonathan343, line 810: nit - we use `'account_id_endpoint_mode'` 5 times in this function. We can probably create a variable and use that instead:
```py
config_key = 'account_id_endpoint_mode'
...
```

botocore/args.py
```diff
@@ -119,6 +125,7 @@ def get_client_args(
         configured_endpoint_url = final_args['configured_endpoint_url']
         signing_region = endpoint_config['signing_region']
         endpoint_region_name = endpoint_config['region_name']
+        endpoint_mode = config_kwargs['account_id_endpoint_mode']
```
jonathan343, line 128: Does this ever get used?

botocore/args.py
```diff
@@ -781,3 +804,21 @@ def _compute_sigv4a_signing_region_set_config(self, config_kwargs):
                 'sigv4a_signing_region_set'
             )
         config_kwargs['sigv4a_signing_region_set'] = sigv4a_signing_region_set
+
+    def _compute_account_id_endpoint_mode_config(self, config_kwargs):
```
jonathan343, line 808: Should we make the config value case insensitive? As in should we accept `PREFERRED` as a valid value. This is what we do with other values.

botocore/configprovider.py
```diff
@@ -168,6 +168,12 @@
         None,
         None,
     ),
+    'account_id_endpoint_mode': (
+        'account_id_endpoint_mode',
+        'AWS_ACCOUNT_ID_ENDPOINT_MODE',
+        'preferred',
```
jonathan343, line 174: I believe we should be setting this value to `disabled` when request signing has been disabled. I don't see that in this logic in this PR.

jonathan343 (commented):
Great start! I just had a few suggestions and questions

botocore/args.py
alexgromero: Yes, we pass this [below](https://github.com/boto/botocore/pull/3345/files#diff-019afcd91cca6d46dd1046132b61e78198063592925d01e9a4db1fd9f1bf8c3fR177) to `_build_endpoint_resolver()`

botocore/configprovider.py
alexgromero: Good catch. Will update in next revision.

botocore/config.py
```diff
@@ -234,6 +234,17 @@ class Config:
...
+        Defaults to None.
+
+    :type account_id_endpoint_mode: str
+    :param account_id_endpoint_mode: The value used to determine the client's
+        behavior for account ID based endpoint routing. Valid values are:
+        * ``preferred`` - The endpoint should include account ID if available.
+        * ``disabled`` - A resolved endpoint does not include account ID.
+        * ``required`` - The endpoint must include account ID. If the account ID
+          isn't available, an exception will be raised.
+        If a value is not provided, the client will default to ``preferred``.
```
jonathan343, line 246: This documentation won't render properly. I've included a screenshot of how this renders right now and a suggestion with how to fix it.

---

Screenshot 2025-02-07 at 11 00 06 AM

---

```suggestion
        behavior for account ID based endpoint routing. Valid values are:
        
        * ``preferred`` - The endpoint should include account ID if available.
        * ``disabled`` - A resolved endpoint does not include account ID.
        * ``required`` - The endpoint must include account ID. If the account ID
          isn't available, an exception will be raised.

        If a value is not provided, the client will default to ``preferred``.
```

jonathan343 (changes requested):

botocore/config.py
alexgromero: Thank you for catching that!

botocore/args.py
```diff
@@ -781,3 +804,30 @@ def _compute_sigv4a_signing_region_set_config(self, config_kwargs):
...
+            account_id_endpoint_mode
+            not in VALID_ACCOUNT_ID_ENDPOINT_MODE_CONFIG
+        ):
+            raise botocore.exceptions.InvalidConfigError(
+                error_msg=f"The configured value '{account_id_endpoint_mode}' for '{config_key}' is "
+                f"invalid. Valid values are: {VALID_ACCOUNT_ID_ENDPOINT_MODE_CONFIG}."
+            )
+
+        signature_version = config_kwargs.get('signature_version')
+        if signature_version is botocore.UNSIGNED:
```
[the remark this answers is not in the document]
alexgromero: Hmm, you're right. Did a quick investigation and it seems like the CLI uses the [`disable_signing`](https://github.com/boto/botocore/blob/develop/botocore/handlers.py#L346-L351) handler which returns `botocore.UNSIGNED` for a [no_sign_request](https://github.com/aws/aws-cli/blob/a403f9c9503c45672b1260d0a6ccac9704273806/awscli/customizations/globalargs.py#L85-L93). However, this is a triggered by an event so the logic here would not handle this case. I will look into it further.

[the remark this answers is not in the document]
SamRemis: Nit: if we wind up going with this logic, it should be the first thing in this method.  It will override the rest of the work we have done in this function, so it makes sense to do this early in the method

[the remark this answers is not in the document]
alexgromero: I agree. Will update this in next revision.

jonathan343 (approved):

PR #3348, opened 2025-01-22
Add account ID to static credentials providers

alexgromero:
This PR introduces account ID to `SharedCredentialProvider` and `ConfigProvider`. The update allows users to specify an `aws_account_id` variable either in the shared credentials file, config file, or during client creation. The account ID will then be associated with the credentials when they are loaded.

### DIFF
--- a/botocore/session.py
+++ b/botocore/session.py
@@ -915,10 +915,7 @@ class Session:
 
         :type aws_account_id: string
         :param aws_account_id: The account id to use when creating
-            the client.  This is entirely optional, and if not provided,
-            the credentials configured for the session will automatically
-            be used.  You only need to provide this argument if you want
-            to override the credentials used for this specific client.
+            the client.  Same semantics as aws_access_key_id above.
 
         :rtype: botocore.client.BaseClient
         :return: A botocore client instance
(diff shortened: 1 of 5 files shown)


jonathan343 (changes requested):

botocore/session.py
```diff
@@ -840,6 +845,7 @@ def create_client(
         aws_access_key_id=None,
         aws_secret_access_key=None,
         aws_session_token=None,
+        aws_account_id=None,
```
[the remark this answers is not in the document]
alexgromero: Ah, thanks for catching that! Will update.

jonathan343:
I think these changes look good. The only thing I'd suggest is adding a test that verifies the correct precedence for `account_id` if specified in multiple places such as the shared config file, environment variable, client creation parameter, ...

botocore/session.py
```diff
@@ -945,6 +958,7 @@ def create_client(
                 access_key=aws_access_key_id,
                 secret_key=aws_secret_access_key,
                 token=aws_session_token,
```
SamRemis, line 960: Right now, this looks like we are just silently dropping `token` and `account_id` if they are set without `aws_access_key_id` and `aws_secret_access_key`, is that correct?

That's probably fine- we haven't had any customer complaints yet that I know of, but it makes sense to me to throw a debug statement here that we are discarding the `token` or `account_id`(unless they can get consumed some other way?)

alexgromero:
@jonathan343 We don't source `account_id` from multiple places. Instead, we source credentials from multiple providers in the chain and `account_id` is resolved based on the credentials that are used.

botocore/session.py
```diff
@@ -954,6 +965,13 @@ def create_client(
                 ),
             )
         else:
+            if ignored_credentials := self._get_ignored_credentials(
+                aws_session_token, aws_account_id
+            ):
+                logger.debug(
+                    f"Ignoring the following credential-related values which were set without "
+                    f"an access key id and secret key on the client: {ignored_credentials}"
```
SamRemis, line 973:
```suggestion
                    f"an access key id and secret key on the session or client: {ignored_credentials}"
```

PR #3351, opened 2025-01-23
Remove Transfer-Encoding from signed headers

nateprewitt:
This PR will exclude `Transfer-Encoding` from the headers considered for signing with SigV4. This prevents unintentional SignatureMismatch errors in the event a network intermediary drops the `Transfer-Encoding` header to replace it with a computed `Content-Length`.

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -65,6 +65,7 @@ ISO8601 = '%Y-%m-%dT%H:%M:%SZ'
 SIGV4_TIMESTAMP = '%Y%m%dT%H%M%SZ'
 SIGNED_HEADERS_BLACKLIST = [
     'expect',
+    'transfer-encoding',
     'user-agent',
     'x-amzn-trace-id',
 ]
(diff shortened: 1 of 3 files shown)


jonathan343 (approved):

botocore/auth.py
```diff
@@ -65,6 +65,7 @@
 SIGV4_TIMESTAMP = '%Y%m%dT%H%M%SZ'
 SIGNED_HEADERS_BLACKLIST = [
```
alexgromero, line 66: nit: should this list also be renamed to blocklist like the tests below?

jonathan343: We generally avoid renaming existing constant's since it's possible customers might be importing them directly using `from botocore.auth import SIGNED_HEADERS_BLACKLIST`. If we rename this constant, this code would break when they pull in the new version of botocore.

However, any net new names should use the term "blocklist".

alexgromero: Ah got it, thank you!

PR #3360, opened 2025-01-30
Ignore Content-Length header when setting 'chunked' for Transfer-Encoding

jonathan343:


botocore/httpchecksum.py
```diff
@@ -409,6 +409,12 @@ def _apply_request_trailer_checksum(request):
         # services such as S3 may require the decoded content length
         headers["X-Amz-Decoded-Content-Length"] = str(content_length)
 
+        if "Content-Length" in headers:
+            del headers["Content-Length"]
+            logger.warning(
+                "Ignoring the Content-Length header since 'chunked' is specified for Transfer-Encoding."
```
nateprewitt, line 415: This only occurs when the user manually supplies this right? I'm not sure if we want this as a warning since we know this is out of the customers control with some tools. We may want to downgrade this to debug.

SamRemis: Kind of a nit - maybe we should replace  "Ignoring" with "Removing" to show what's actually happening here.

nateprewitt (approved):
:shipit:

nateprewitt (approved):

PR #3370, opened 2025-02-03
Restrict CRC64NVME support based on CRT version

nateprewitt:
Some platforms automatically include a copy of the CRT in the environment, unbeknownst to the end user. In these cases, when the default checksum CRC64NVME is added to object uploaded through the console, boto3 will fail trying to validate the returned checksum.

This patch moves CRC64NVME out of the hot path for validation and will only attempt to process the received checksum when it's supported by the AWS CRT version on the machine.

### DIFF
--- a/botocore/httpchecksum.py
+++ b/botocore/httpchecksum.py
@@ -25,7 +25,7 @@ import logging
 from binascii import crc32
 from hashlib import sha1, sha256
 
-from botocore.compat import HAS_CRT, urlparse
+from botocore.compat import HAS_CRT, has_minimum_crt_version, urlparse
 from botocore.exceptions import (
     AwsChunkedWrapperError,
     FlexibleChecksumError,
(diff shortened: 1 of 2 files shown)


botocore/httpchecksum.py
```diff
@@ -522,8 +522,12 @@ def _handle_bytes_response(http_response, response, algorithm):
     _CRT_CHECKSUM_CLS = {
         "crc32": CrtCrc32Checksum,
         "crc32c": CrtCrc32cChecksum,
-        "crc64nvme": CrtCrc64NvmeChecksum,
     }
+
+    if has_minimum_crt_version((0, 23, 0)):
+        # CRC64NVME support wasn't officially added until 0.23.x
```
jonathan343, line 528: Is this right? I installed `awscrt==0.22.4` and was able to import `crc64nvme` from the `awscrt.checksums` module.
```py
>>> import awscrt
>>> awscrt.__version__
'0.22.4'
>>> from awscrt.checksums import crc64nvme
>>> 
```

The version we used when updating flexible checksum defaults was `0.23.4`, but it seems like `crc64nvme` has existed for a while now.

nateprewitt: That's why I used the wording "officially". There were some testing versions earlier but I think it's best to put the floor where we'd stated we now had CRC64NVME support. That prevents us using older copies that may have bugs.

That said, it looks like there were further changes after 0.23.0 and the actual official release is what we use (0.23.4). Since we're not following SemVer here, I'll update the version number to be exact.

nateprewitt: Should be updated.

jonathan343 (approved):
If we really wanted to we could add a functional tests to verify the following behavior:
* Mock the awscrt version to something < 0.23.4 and confirm we don't support CRC64NVME
* Mock the awscrt version to something >= 0.23.4 and confirm we do support CRC64nvme

I wouldn't consider this blocking and think it's more important to get a fix out for customers. Approving

PR #3371, opened 2025-02-03
Remove payload length cap from Event Stream messages

nateprewitt:
In initial implementations of Event Streams, there was a 16MB payload limit for Event Stream payloads. Anything exceeding that was considered malformed. In more modern implementations, that limit has been raised with some services using 24MB. We'll be removing this client-side check going forward as it doesn't serve a practical purpose. The checksum will be used to validate the payload and size limitations will be left to the service to determine going forward.


nateprewitt:
Windows builds seem to be struggling with the larger payload. I'll investigate what's currently failing, it's not clear in the logs.

jonathan343 (approved):
:shipit:

PR #3375, opened 2025-02-04
Optimizes iter_lines handling long lines

### LINKED ISSUE #2774: Very slow performance when using iter_lines on an s3 object with long lines
### Describe the bug

`StreamingBody.iter_lines` seems to be extremely slow when dealing with a file with very long lines. We were using this in ECS to parse a ~500mb json lines file containing fairly large objects and it was taking upwards of 10 minutes to run.

### Expected Behavior

Performance should be comparable to `read` or `iter_chunks`, regardless of how long the lines are.

### Current Behavior

It is way slower than reading the file in other ways. 

### Reproduction Steps

Minimal repro (requires `moto`) that reads a 5MB file with no line breaks:
```python
from moto import mock_s3
import boto3
import time

BUCKET = "test_bucket"
KEY = "long_lines.txt"
KILOBYTES = 1024
MEGABYTES = 1024 * KILOBYTES

@mock_s3
def slow_iter_lines():
    s3 = boto3.resource("s3")
    s3.create_bucket(Bucket=BUCKET)
    obj = s3.Object(BUCKET, KEY)
    obj.put(Body=b"a" * (10*MEGABYTES))

    start1 = time.perf_counter()
    obj.get()["Body"].read()
    end1 = time.perf_counter()
    print(f"Normal read took {end1-start1}s")

    start2 = time.perf_counter()
    list(obj.get()["Body"].iter_lines())
    end2 = time.perf_counter()
    print(f"Line iterator took {end2-start2}s")

slow_iter_lines()
```

Output on `botocore==1.27.87`:
```
Normal read took 0.004170743000031507s
Line iterator took 24.49556041300002s
```

### Possible Solution

The implementation of `iter_lines` looks to be quadratic in the length of the lines:
```python
    def iter_lines(self, chunk_size=_DEFAULT_CHUNK_SIZE, keepends=False):
        pending = b''
        for chunk in self.iter_chunks(chunk_size):
            lines = (pending + chunk).splitlines(True)
            for line in lines[:-1]:
                yield line.splitlines(keepends)[0]
            pending = lines[-1]
        if pending:
            yield pending.splitlines(keepends)[0]
```
If there are no line breaks in any of the chunks, then every time it goes round in this loop it is doing:
* `(pending + chunk)`, which requires allocating a new buffer
* `splitlines`, which requires iterating through the whole buffer again looking for line breaks
So pending keeps growing longer, and every time we have to copy it and iterate through it, so it gets slower quadratically until there is a line break.

A better implementation would probably be to maintain a list of pending chunks and concatenate them only when a line break is reached.

Increasing `chunk_size` to 1MB fixed the immediate problem we were having.

### SDK version used

1.27.87

Ubuntu 22.04, python 3.10.4

mfulgo:
In the previous implementation of `iter_lines`, lines that spanned chunks could cause a quadratic iteration of the line data. This could result in a much slower read time for files with long lines or lines significantly longer than the chunk size used when reading them.

To avoid the issue, the StreamingBody now collects the line data and only constructs the line (copying the bytes into a new buffer) when it encounters a line ending.

Closes #2774

### DIFF
--- a/botocore/response.py
+++ b/botocore/response.py
@@ -54,6 +54,7 @@ class StreamingBody(IOBase):
         self._raw_stream = raw_stream
         self._content_length = content_length
         self._amount_read = 0
+        self._pending_line_data = []
 
     def __del__(self):
         # Extending destructor in order to preserve the underlying raw_stream.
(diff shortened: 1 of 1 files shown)


mfulgo:
I wasn't quite sure how you all might want a performance test written for this, so I just added and ran the following unit test before and after the change:

```python
    def test_streaming_line_iterator_long_lines(self):
        line = b'1234567890'*1000000
        body = itertools.repeat(line, 1)
        stream = response.StreamingBody(BytesIO(b'\n'.join(body)), content_length=(1*10*1000000)+1-1)
        start = perf_counter()
        for l in stream.iter_lines():
            self.assertEqual(l, line)
        raise RuntimeError(f"It took {perf_counter()-start:.3f} seconds")
```

Prior to this commit, it took around `23` seconds to run. After the change, it consistently ran in about `0.009` seconds.

PR #3376, opened 2025-02-04
Protocols trait support

SamRemis:
This PR adds support for the new `protocols` trait, which allows a service to list a number of protocols it supports; botocore will automatically choose the highest priority supported protocol and use that to make requests to the service.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -820,8 +820,8 @@ class ClientArgsCreator:
 
     def _resolve_protocol(self, service_model):
         # We need to ensure `protocols` exists in the metadata before attempting to
-        #  access it directly since referencing service_model.protocols directly will
-        #  raise an UndefinedModelAttributeError if protocols is not defined
+        # access it directly since referencing service_model.protocols directly will
+        # raise an UndefinedModelAttributeError if protocols is not defined
         if service_model.metadata.get('protocols'):
             for protocol in PRIORITY_ORDERED_SUPPORTED_PROTOCOLS:
                 if protocol in service_model.protocols:
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -819,7 +819,8 @@ class ClientArgsCreator:
         )
 
     def _compute_protocol(self, service_model):
-        # If we don't have a protocols trait, fall back to the legacy protocol trait
+        # If a service does not have a `protocols trait`, fall back to the legacy
+        # `protocol` trait
         if 'protocols' not in service_model.metadata:
             return service_model.metadata['protocol']
         for protocol in PRIORITY_ORDERED_SUPPORTED_PROTOCOLS:
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -822,7 +822,6 @@ class ClientArgsCreator:
         # If we don't have a protocols trait, fall back to the legacy protocol trait
         if 'protocols' not in service_model.metadata:
             return service_model.metadata['protocol']
-
         for protocol in PRIORITY_ORDERED_SUPPORTED_PROTOCOLS:
             if protocol in service_model.protocols:
                 return protocol
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -70,14 +70,12 @@ VALID_RESPONSE_CHECKSUM_VALIDATION_CONFIG = (
     "when_required",
 )
 
-PRIORITY_ORDERED_SUPPORTED_PROTOCOLS = list(
-    (
-        'json',
-        'rest-json',
-        'rest-xml',
-        'query',
-        'ec2',
-    )
+PRIORITY_ORDERED_SUPPORTED_PROTOCOLS = (
+    'json',
+    'rest-json',
+    'rest-xml',
+    'query',
+    'ec2',
 )
 
 
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -70,6 +70,16 @@ VALID_RESPONSE_CHECKSUM_VALIDATION_CONFIG = (
     "when_required",
 )
 
+PRIORITY_ORDERED_SUPPORTED_PROTOCOLS = list(
+    (
+        'json',
+        'rest-json',
+        'rest-xml',
+        'query',
+        'ec2',
+    )
+)
+
 
 class ClientArgsCreator:
     def __init__(
@@ -210,7 +220,7 @@ class ClientArgsCreator:
         scoped_config,
     ):
         service_name = service_model.endpoint_prefix
-        protocol = service_model.metadata['protocol']
+        protocol = self._compute_protocol(service_model)
         parameter_validation = True
         if client_config and not client_config.parameter_validation:
             parameter_validation = False
@@ -808,6 +818,20 @@ class ClientArgsCreator:
             config_kwargs,
             config_key="response_checksum_validation",
             valid_options=VALID_RESPONSE_CHECKSUM_VALIDATION_CONFIG,
+        )
+
+    def _compute_protocol(self, service_model):
+        # If we don't have a protocols trait, fall back to the legacy protocol trait
+        if 'protocols' not in service_model.metadata:
+            return service_model.metadata['protocol']
+
+        for protocol in PRIORITY_ORDERED_SUPPORTED_PROTOCOLS:
+            if protocol in service_model.protocols:
+                return protocol
+        raise botocore.exceptions.NoSupportedProtocolError(
+            botocore_supported_protocols=PRIORITY_ORDERED_SUPPORTED_PROTOCOLS,
+            service_supported_protocols=service_model.protocols,
+            service=service_model.service_name,
         )
 
     def _handle_checksum_config(
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -218,7 +218,7 @@ class ClientArgsCreator:
         scoped_config,
     ):
         service_name = service_model.endpoint_prefix
-        protocol = self._compute_protocol(service_model)
+        protocol = self._resolve_protocol(service_model)
         parameter_validation = True
         if client_config and not client_config.parameter_validation:
             parameter_validation = False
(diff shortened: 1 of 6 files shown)


botocore/exceptions.py
```diff
@@ -823,3 +823,12 @@ class InvalidChecksumConfigError(BotoCoreError):
         'Unsupported configuration value for {config_key}. '
         'Expected one of {valid_options} but got {config_value}.'
     )
+
+
+class NoSupportedProtocolError(BotoCoreError):
```
jonathan343, line 828: What do you think of a name like `UnsupportedServiceProtocolError` instead?

jonathan343 (commented):
Looks great so far! I left a few suggestions, but nothing major.

botocore/exceptions.py
SamRemis: `UnsupportedServiceProtocolsError` is another alternative, I just want to make sure that protocols is plural

botocore/args.py
```diff
@@ -810,6 +818,20 @@ def _compute_checksum_config(self, config_kwargs):
             valid_options=VALID_RESPONSE_CHECKSUM_VALIDATION_CONFIG,
         )
 
+    def _resolve_protocol(self, service_model):
+        # If a service does not have a `protocols` trait, fall back to the legacy
+        # `protocol` trait
+        if 'protocols' not in service_model.metadata:
```
nateprewitt, line 824: This one feels a little odd to me. We may want to be validating the truthiness of `protocols` rather than the key existing. It should be equivalent or a supeset of protocols, but if it's `[]` and protocol exists, it seems better to be using the protocol instead of hard failing?

botocore/args.py
```diff
@@ -810,6 +818,20 @@ def _compute_checksum_config(self, config_kwargs):
             valid_options=VALID_RESPONSE_CHECKSUM_VALIDATION_CONFIG,
         )
 
+    def _compute_protocol(self, service_model):
+        # If a service does not have a `protocols trait`, fall back to the legacy
+        # `protocol` trait
+        if 'protocols' not in service_model.metadata:
+            return service_model.metadata['protocol']
```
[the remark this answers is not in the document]
jonathan343: I'm kind of confused now lol.

If the metadata was being mocked, wouldn't the property return what they expect since it calls metadata under the hood?

```py
    @CachedProperty
    def protocol(self):
        return self._get_metadata_property('protocol')

    ...

    def _get_metadata_property(self, name):
        try:
            return self.metadata[name]
        except KeyError:
            raise UndefinedModelAttributeError(
                f'"{name}" not defined in the metadata of the model: {self}'
            )
```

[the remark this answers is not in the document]
nateprewitt: That's what I pointed out to Sam. He said there's a case this wouldn't work like that (I'm not sure how), but even if that's the case, I don't think it's reasonable to expect that as a reasonable mock.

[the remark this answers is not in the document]
SamRemis: I just tested making this change and it still happens today, but maybe I'm missing something.  If you go to the tip of the develop and only change `protocol = service_model.metadata['protocol']` to `protocol = service_model.protocol`, then 45 tests will fail in `test_args.py`.  

These tests go through [this code path](https://github.com/boto/botocore/blob/5c409244cfdf3b26486c8c6a4c3b65fcf9e45c5c/tests/unit/test_args.py#L64-L78), where we mock out the service_model using `service_model = mock.Mock(ServiceModel)`.  This means that when we call `service_model.protocol`, a `Mock` object is returned, not the actual metadata key's value.  For example, calling `service_model.protocol` will return `<Mock name='mock.protocol' id='01234567'>`, even though calling `service_model.metadata.get('protocol')` will return `query`.  

When the mock object is returned as the protocol, these tests fail on [this line](https://github.com/boto/botocore/blob/08877232170abe910198e3c1f758ae6d0f2df6a0/botocore/serialize.py#L68) with a the following error when we try to create a `Serializer` object from the resolved protocol:
```
KeyError: <Mock name='mock.protocol' id='01234567'>
```

This can be fixed by adding `service_model.protocol = 'query'` to the method which creates the mock, then `query` gets returned instead.

botocore/args.py
SamRemis: While Nate likely knows this, I'm still going to put this here and in a new code comment for posterity.  I don't think we can access any of the new `CachedProperty` properties directly unless we are somehow guaranteed they will be there in all models.  This guarantee is impossible because some users have custom models, so we will always need some way to fall back to the existing properties.   The reason is that when we access any of these properties- `service_model.protocols` for instance, it calls `_get_metadata_property` under the hood, which will raise an error if the key doesn't exist:
```
    def _get_metadata_property(self, name):
        try:
            return self.metadata[name]
        except KeyError:
            raise UndefinedModelAttributeError(
                f'"{name}" not defined in the metadata of the model: {self}'
            )
```

While I'm not sure how we could get into the codepath where properties is an empty array given the tests below that prevent that, I think that the readability of this code could be improved by Nate's suggestion, so I'll go make that change and push up a revision shortly :) 

I'm not sure that there's a way to safely call `service_model.protocols` directly though, so I'll have to do it through `service_model.metadata.get('protocol')`

botocore/args.py
```diff
@@ -810,6 +818,23 @@ def _compute_checksum_config(self, config_kwargs):
             valid_options=VALID_RESPONSE_CHECKSUM_VALIDATION_CONFIG,
         )
 
+    def _resolve_protocol(self, service_model):
+        # We need to ensure `protocols` exists in the metadata before attempting to
+        #  access it directly since referencing service_model.protocols directly will
+        #  raise an UndefinedModelAttributeError if protocols is not defined
```
jonathan343, line 824:
```suggestion
        # access it directly since referencing service_model.protocols directly will
        # raise an UndefinedModelAttributeError if protocols is not defined
```

tests/functional/test_supported_protocols.py
```diff
@@ -0,0 +1,56 @@
...
+    services_models_with_protocols = []
+    services_models_without_protocols = []
+
+    for service in services:
+        service_model = session.get_service_model(service)
+        if 'protocols' in service_model.metadata:
+            services_models_with_protocols.append(service_model)
+        else:
+            services_models_without_protocols.append(service_model)
+    return (services_models_with_protocols, services_models_without_protocols)
```
jonathan343, line 33: [Not Blocking]This seems pretty memory inefficient as opposed to using `yield`. I don't feel strongly though given this pattern is used in other tests and this isn't actually SDK code.

tests/unit/test_args.py
```diff
@@ -907,3 +945,15 @@ def test_sdk_endpoint_legacy_set_without_builtin_data(self):
             legacy_endpoint_url='https://my.legacy.endpoint.com',
         )
         self.assertEqual(bins['SDK::Endpoint'], None)
+
+
+class TestProtocolPriorityList:
```
jonathan343, line 950: We should provide an error message here to make it more obvious when debugging this error.

jonathan343 (commented):
I have some pretty minor feedback. Happy to approve once they've been addressed or we talk about why they shouldn't.

botocore/args.py
```diff
@@ -810,6 +818,23 @@ def _compute_checksum_config(self, config_kwargs):
...
+        if service_model.metadata.get('protocols'):
```
nateprewitt, line 825: I'm assuming this has been done but before we release, do we know these are all identical (`protocol` v `protocols`) for the currently released models?

tests/functional/test_supported_protocols.py
```diff
@@ -0,0 +1,56 @@
...
+
+    services = loader.list_available_services('service-2')
+    services_models_with_protocols = []
+    services_models_without_protocols = []
+
+    for service in services:
+        service_model = session.get_service_model(service)
+        if 'protocols' in service_model.metadata:
+            services_models_with_protocols.append(service_model)
+        else:
```
nateprewitt, line 31: Do we have models that don't have `protocols` entries currently? If not, should this just be an error case?

tests/unit/test_args.py
```diff
@@ -63,9 +67,11 @@ def _get_service_model(self, service_name=None):
         service_model = mock.Mock(ServiceModel)
         service_model.service_name = service_name
         service_model.endpoint_prefix = service_name
+        service_model.protocol = 'query'
+        service_model.protocols = ['query']
         service_model.metadata = {
             'serviceFullName': 'MyService',
-            'protocol': 'query',
```
nateprewitt, line 68: Why did we remove this line?  They both exist, right?

tests/functional/test_supported_protocols.py
SamRemis: It would break custom models, no?

nateprewitt: I don't know if we're intending to test custom models? Since this isn't in the runtime path, if it does break, I think the fix at that point is simple and ensures the quality of the model improves to match what we expect in production.

nateprewitt: I think this also means we never run the second test?

SamRemis: My bad, I misread which line the comment was referring to from GH mobile.  

Yes, we have models that don't have `protocols`.  148 at the time of writing this comment.

tests/unit/test_args.py
SamRemis: They do :) it's just no longer necessary for the test, so it's one less line we need in the codebase.  It doesn't make a real difference to me, so I'm happy to add it back

nateprewitt: The reason I'm calling it out is you added `service_model.protocol` on line 70. If that's required, we should be keeping the metadata in sync. Otherwise we're creating surface area for future bugs unnecessarily.

botocore/args.py
SamRemis: That's true- for all models that currently have the `protocols` trait.  Every service with the trait also only has a length of one for the moment, so no customers should see any change in behavior from this change today.

tests/functional/test_supported_protocols.py
SamRemis: I've updated it to something "better"?  It doesn't load them all into memory, but it feels a bit awkward.  I'd love some feedback on it when you get the chance.

PR #3380, opened 2025-02-07
Upgrade ruff, fix issues, use ruff for import sorting

akx:
This PR:

* upgrades the version of Ruff used from 0.4.8 to 0.9.4
* copies the exclude rules from `pre-commit` configuration to the Ruff configuration, to make it possible to run `ruff` outside of `pre-commit`
* reapplies `ruff format` (mechanical)
* fixes other new complaints surfaced by the new version of Ruff
* gets rid of `isort` for import sorting in favor of just ruff's `I` rule.
  * This had some manual work go into it, to keep the context of the `noqa` comments intact.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,10 +19,6 @@ repos:
       - id: check-yaml
       - id: end-of-file-fixer
       - id: trailing-whitespace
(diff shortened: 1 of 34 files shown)


akx:
Well, looks like @nateprewitt did this work some 4 months later in #3495 😄

PR #3381, opened 2025-02-07
Add account id to STS and SSO providers

alexgromero:
This PR adds support to source an account ID from `AssumeRoleProvider`, `AssumeRoleWithWebIdentityProvider`, and `SSOProvider`. The account ID will be resolved after receiving  a successful response from a given provider and then be associated with the credentials object.

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -735,9 +735,8 @@ class CachedCredentialFetcher:
             'secret_key': creds['SecretAccessKey'],
             'token': creds['SessionToken'],
             'expiry_time': expiration,
+            'account_id': creds.get('AccountId'),
         }
-        if account_id := creds.get('AccountId'):
-            credentials['account_id'] = account_id
 
         return credentials
 
@@ -814,7 +813,7 @@ class BaseAssumeRoleCredentialFetcher(CachedCredentialFetcher):
         argument_hash = sha1(args.encode('utf-8')).hexdigest()
         return self._make_file_safe(argument_hash)
 
-    def _get_account_id(self, response):
+    def _add_account_id_to_response(self, response):
         role_arn = response.get('AssumedRoleUser', {}).get('Arn')
         if ArnParser.is_arn(role_arn):
             arn_parser = ArnParser()
@@ -884,7 +883,7 @@ class AssumeRoleCredentialFetcher(BaseAssumeRoleCredentialFetcher):
         kwargs = self._assume_role_kwargs()
         client = self._create_client()
         response = client.assume_role(**kwargs)
-        self._get_account_id(response)
+        self._add_account_id_to_response(response)
         return response
 
     def _assume_role_kwargs(self):
@@ -973,7 +972,7 @@ class AssumeRoleWithWebIdentityCredentialFetcher(
         config = Config(signature_version=UNSIGNED)
         client = self._client_creator('sts', config=config)
         response = client.assume_role_with_web_identity(**kwargs)
-        self._get_account_id(response)
+        self._add_account_id_to_response(response)
         return response
 
     def _assume_role_kwargs(self):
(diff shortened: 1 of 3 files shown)


botocore/credentials.py
```diff
@@ -863,7 +877,9 @@ def _get_credentials(self):
         """Get credentials by calling assume role."""
         kwargs = self._assume_role_kwargs()
         client = self._create_client()
-        return client.assume_role(**kwargs)
+        response = client.assume_role(**kwargs)
+        self._get_account_id(response)
```
SamRemis, line 881: Nit:  `_get_account_id()` is a function that I would assume would retrieve and return the account id without any side effects.

This retrieves it from the response, extracts it from the arn, then changes the response from the operation to include this.  Considering at `_get_credentials` previously returned the raw, unaltered response from the service, it seems odd to be changing that.  

Does it make more sense to put this logic closer to where we create the Credentials object?  Or at a minimum, name this method to be clearer that it's altering the response.

botocore/credentials.py
```diff
@@ -723,12 +724,16 @@ def _get_cached_credentials(self):
...
         creds = response['Credentials']
         expiration = _serialize_if_needed(creds['Expiration'], iso=True)
-        return {
+        credentials = {
             'access_key': creds['AccessKeyId'],
             'secret_key': creds['SecretAccessKey'],
             'token': creds['SessionToken'],
             'expiry_time': expiration,
         }
+        if account_id := creds.get('AccountId'):
```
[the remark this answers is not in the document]
alexgromero: You're right. We chatted about this offline and explained that I made this change due to some failing unit tests which we will update in the next revision. 
The issue was that these tests were relying on `get_expected_creds_from_response()` which returns a set of credentials from response. Since we will now be sending `account_id` with the credentials dict, we will be updating this function to also return `account_id`.

botocore/credentials.py
alexgromero: Also chatted about this offline and agreed to change the method name to be more clearer. We will rename this to `_add_account_id_to_response`.  
Moving this logic closer to where we create the Credentials object would require a bit of refactoring since the `_get_cached_credentials` method in the `CachedCredentialFetcher` class is used by different providers that send back different responses.

botocore/credentials.py
```diff
@@ -1207,6 +1259,11 @@ def fetch_credentials(require_expiry=True):
                     provider=method, cred_var=mapping['expiry_time']
                 )
 
+            credentials['account_id'] = None
```
SamRemis, line 1262: [Non-blocking] Nit: is this any different than doing it all in one line with the following?:
```credentials['account_id'] = environ.get(mapping['account_id'])```

I see it is similar to the expiry_time logic above, but I don't quite get why that would be necessary.

File: botocore/credentials.py [hunk 4]
```diff
@@ -2169,6 +2226,7 @@ def _get_credentials(self):
                 'SecretAccessKey': credentials['secretAccessKey'],
                 'SessionToken': credentials['sessionToken'],
                 'Expiration': self._parse_timestamp(credentials['expiration']),
+                'AccountId': self._account_id,
```
SamRemis, line 2229: To confirm an assumption here cause I don't see it in this PR:

This will always exist because you added it on `CachedCredentialFetcher` in another PR, correct?

SamRemis (approved):

botocore/credentials.py
alexgromero: You're right. I just followed the pattern with the other variables. This change is part of another PR that has already been approved (#3332). The commit ended up in this PR accidentally.

botocore/credentials.py
alexgromero: No, I added this because the `GetRoleCredentials` call must be called with an account id, per the SSO SEP, so this is a value that we already have and are sending back as part of the Credentials.

botocore/credentials.py
alexgromero: Just force pushed the branch to rebase and fix outdated commits.

SamRemis (approved):

PR #3383, opened 2025-02-09
trim now unneeded usage of botocore.compat shim

a-detiste:
There are 37 more source files to review, but I'd first like to see how this PR goes before finishing this tedious work.

My ultimate goal is to continue to removed `six` from Debian. Having a more auditable usage of `botocore.compat` is one step forward.

https://wiki.debian.org/Python3-six-removal

Greetings


a-detiste:
(PS: of course we should keep `tests/unit/test_compat.py` as-is)

toabctl (commented):
LGTM

PR #3389, opened 2025-02-12
User agent 2.1: Track client features

hssyoo:
#### Description
This PR implements scaffolding for tracking client feature usage in the user agent header. Features that we track are defined in `botocore.useragent._USERAGENT_FEATURE_MAPPINGS`. Whenever a trackable feature is used, its associated metric id is stored. When the user agent string is built, it retrieves all registered features and creates a comma-separated string with the `m/` prefix. For example, if a request is made using SigV4A signing, the user agent string may look like the following:
```
Boto3/1.36.13 md/Botocore#1.36.13 md/awscrt#0.23.8 ua/2.0 os/macos#24.2.0 md/arch#arm64 lang/python#3.11.10 md/pyimpl#CPython m/S cfg/retry-mode#legacy Botocore/1.36.13
```
Note the `m/S` component.

#### Design
I prioritized an approach that tries to minimize invasiveness to the current codebase. Specifically, I wanted to avoid having to plumb some `features` variable to every method that registers a function, mainly so we wouldn't need to update the function signature of potentially all methods. This would also require higher level packages (Boto3, S3Transfer, AWS CLI) to plumb features they register.

#### Initial features to track
There are 5 features tracked to start:
- `WAITER`: An operation called using a waiter.
- `PAGINATOR`: An operation called using a paginator.
- `S3_TRANSFER`: An operation called using the S3 Transfer Manager.
- `ENDPOINT_OVERRIDE`: An operation called using a user provided endpoint URL.
- `SIGV4A_SIGNING`: An operation called using sigv4a signing.

S3Transfer usage will need to be registered in `boto3` and will require a small update to `s3transfer`(to pass context from parent to child threads).
- S3Transfer: https://github.com/boto/s3transfer/pull/335
- Boto3: https://github.com/boto/boto3/pull/4441

#### Caveats
- Features are tracked in a set. Since sets don't preserve order in which elements were added, if the feature metadata string exceeds 1KB and it's truncated, the truncation strategy won't be true LIFO. I thought this was an acceptable tradeoff since we're nowhere near the limit and we can explore storage solutions or even metric collection strategy in the future if this becomes a real issue.
- Some features are "cacheable". For example, if a Session object is creating a new Client and there are cached credentials, it will skip invoking the method that registers the relevant credentials feature. This PR doesn't prescribe a single way to deal with cacheable features since the way feature values are cached and gotten can vary. For cached credentials, I imagine we would inspect the `METHOD` attributes to determine the correct feature to register.

### DIFF
--- a/botocore/context.py
+++ b/botocore/context.py
@@ -58,6 +58,15 @@ def set_context(ctx):
     return token
 
 
+def reset_token(token):
+    """Reset the current ``_context`` context variable.
+
+    :type token: contextvars.Token
+    :param token: Token object to reset the context variable.
+    """
+    _context.reset(token)
+
+
 @contextmanager
 def start_as_current_context(ctx=None):
     """
@@ -86,7 +95,7 @@ def start_as_current_context(ctx=None):
     try:
         yield
     finally:
-        _context.reset(token)
+        reset_token(token)
 
 
 def with_current_context(hook=None):
--- a/botocore/context.py
+++ b/botocore/context.py
@@ -26,7 +26,7 @@ from typing import Set
 @dataclass
 class ClientContext:
     """
-    Encapsulation of objects tracked within the ``__context`` context variable.
+    Encapsulation of objects tracked within the ``_context`` context variable.
 
     ``features`` is a set responsible for storing features used during
     preparation of an AWS request. ``botocore.useragent.register_feature_id``
--- a/botocore/context.py
+++ b/botocore/context.py
@@ -58,7 +58,7 @@ def set_context(ctx):
     return token
 
 
-def reset_token(token):
+def reset_context(token):
     """Reset the current ``_context`` context variable.
 
     :type token: contextvars.Token
@@ -95,7 +95,7 @@ def start_as_current_context(ctx=None):
     try:
         yield
     finally:
-        reset_token(token)
+        reset_context(token)
 
 
 def with_current_context(hook=None):
--- a/botocore/context.py
+++ b/botocore/context.py
@@ -1,0 +1,119 @@
+# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""
+NOTE: All classes and functions in this module are considered private and are
+subject to abrupt breaking changes. Please do not use them directly.
+"""
+
+from contextlib import contextmanager
+from contextvars import ContextVar
+from copy import deepcopy
+from dataclasses import dataclass, field
+from functools import wraps
+from typing import Set
+
+
+@dataclass
+class ClientContext:
+    """
+    Encapsulation of objects tracked within the ``__context`` context variable.
+
+    ``features`` is a set responsible for storing features used during
+    preparation of an AWS request. ``botocore.useragent.register_feature_id``
+    is used to add to this set.
+    """
+
+    features: Set[str] = field(default_factory=set)
+
+
+_context = ContextVar("_context")
+
+
+def get_context():
+    """Get the current ``_context`` context variable if set, else None."""
+    return _context.get(None)
+
+
+def set_context(ctx):
+    """Set the current ``_context`` context variable.
+
+    :type ctx: ClientContext
+    :param ctx: Client context object to set as the current context variable.
+
+    :rtype: contextvars.Token
+    :returns: Token object used to revert the context variable to what it was
+        before the corresponding set.
+    """
+    token = _context.set(ctx)
+    return token
+
+
+@contextmanager
+def start_as_current_context(ctx=None):
+    """
+    Context manager that copies the passed or current context object and sets
+    it as the current context variable. If no context is found, a new
+    ``ClientContext`` object is created. It mainly ensures the context variable
+    is reset to the previous value once the executed code returns.
+
+    Example usage:
+
+        def my_feature():
+            with start_as_current_context():
+                register_feature_id('MY_FEATURE')
+                pass
+
+    :type ctx: ClientContext
+    :param ctx: The client context object to set as the new context variable.
+        If not provided, the current or a new context variable is used.
+    """
+    current = ctx or get_context()
+    if current is None:
+        new = ClientContext()
+    else:
+        new = deepcopy(current)
+    token = set_context(new)
+    try:
+        yield
+    finally:
+        _context.reset(token)
+
+
+def with_current_context(hook=None):
+    """
+    Decorator that wraps ``start_as_current_context`` and optionally invokes a
+    hook within the newly-set context. This is just syntactic sugar to avoid
+    indenting existing code under the context manager.
+
+    Example usage:
+
+        @with_current_context(partial(register_feature_id, 'MY_FEATURE'))
+        def my_feature():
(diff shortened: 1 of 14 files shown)


botocore/context.py
```diff
@@ -0,0 +1,128 @@
...
+    current = ctx or get_context()
+    if current is None:
+        new = ClientContext()
+    else:
+        new = deepcopy(current)
+    token = set_context(new)
+    try:
+        yield
+    finally:
+        reset_token(token)
```
jonathan343, line 98: Could you help me understand why the context is reset after the nested code finishes executing?. I'm imagining a scenario where we have sequential functions that register features. The features are only added to a copy of the parent context available in the current scope. I could be misusing `with_current_context`.

Example:
```python
from botocore.useragent import register_feature_id
from botocore.context import with_current_context


@with_current_context(partial(register_feature_id, 'WAITER'))
def _foo_waiter():
    print(f"features: {get_context().features}")
    ...

@with_current_context(partial(register_feature_id, 'PAGINATOR'))
def _foo_paginator():
    print(f"features: {get_context().features}")
    ...

@with_current_context()
def _foo():
    _foo_waiter()
    _foo_paginator()
    print(f"features: {get_context().features}")

_foo()
```

**Output:**
```
features: {'B'}
features: {'C'}
features: set() # I imagine this being {'B', 'C'} instead
```

hssyoo: Ah good question. Your own answer is right and here's why:

In your example using waiters and paginators, we actually _would_ want the final print statement to contain 0 features registered because waiters and paginators actually make AWS requests, and feature tracking is done on a per-request basis.

If we changed up your example a bit and say we were calling `_foo_new_signer()` and `_foo_new_feature()` under the `_foo()` function, then we would be using `register_feature_id` in the 2 feature functions.

> Could you help me understand why the context is reset after the nested code finishes executing?

The context manager is used when we want to:
1. Ensure that we have some already-instantiated context to work with.
2. Ensure that when the decorated method returns, the context is reset. This prevents features registered during preparation of request A from bleeding into the feature set of request B.

This means that, at a minimum, we want to use the context manager on a method that serves as an "entrypoint" to creating an encapsulated set of features. "Entrypoint" here refers to how a user may be interacting with the client. You can see in the PR that the decorated methods are:
1. Client instantiation. We want to register, set, and reset client-specific features here.
2. Modeled API call. This sends an AWS request, so we want to reset any features registered after the request is made.
3. Paginator call. This also sends AWS request(s), so we want to reset after returning.
4. Waiter call. This also sends AWS request(s), so we want to reset after returning.

And finally, `register_feature_id` can be directly used to register features if we want the feature to persist beyond the scope of the method it was called in.

botocore/useragent.py
```diff
@@ -94,11 +141,40 @@ def to_string(self):
...
+            clean_string = f'{clean_prefix}/{clean_name}#{clean_value}'
+        if self.size_config is not None:
+            clean_string = self._truncate_string(
+                clean_string,
+                self.size_config.max_size_in_bytes,
+                self.size_config.delimiter,
+            )
+        return clean_string
+
+    def _truncate_string(self, string, max_size, delimiter):
```
jonathan343, line 158: nit - Should `max_size` be inclusive?

botocore/useragent.py
```diff
@@ -94,11 +141,40 @@ def to_string(self):
...
+        """
+        Pop ``delimiter``-separated values until encoded string is less than or
+        equal to ``max_size``.
+        """
+        orig = string
+        encoded_string = string.encode('utf-8')
+        while len(encoded_string) > max_size:
```
jonathan343, line 165: nit - Passing in a negative value for `max_size` would result in an infinite loop. Kind of a silly thing to guard against since we control what gets configured provided. I don't feel strongly about it, but maybe something to consider.

jonathan343, line 165: nit - if we change this to `while len(string.encode('utf-8')) > max_size:`, the need for `encoded_string` goes away.

botocore/useragent.py
```diff
@@ -71,19 +103,34 @@ def sanitize_user_agent_string_component(raw_str, allow_hash):
     )
 
 
+class UserAgentSizeConfig(NamedTuple):
```
jonathan343, line 106: Should this be `UserAgentComponentSizeConfig`?

jonathan343 (changes requested):
This PR looks great! Left a few minor comments, but overall this seems very polished.

botocore/useragent.py
jonathan343: Resolving. After taking a second look, the `while len(encoded_string) > max_size:` condition is correct.

botocore/useragent.py
```diff
@@ -94,11 +141,40 @@ def to_string(self):
...
+        orig = string
+        encoded_string = string.encode('utf-8')
+        while len(encoded_string) > max_size:
+            parts = string.split(delimiter)
+            parts.pop()
+            string = delimiter.join(parts)
+            encoded_string = string.encode('utf-8')
+
+        if string == '':
+            raise ValueError(
```
jonathan343, line 172: Is raising an error (and failing the request) actually the behavior we want in this case? That seems wrong

botocore/useragent.py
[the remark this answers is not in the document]
hssyoo: No this is a miss

botocore/useragent.py
jonathan343: To provide more context, I'm suggesting this incase we ever decide to add a size config for the entire user agent string and not just a component. We can probably leave the name as is if we think this same config could be reused for that case. If not, renaming might make sense.

botocore/useragent.py
jonathan343: I'm thinking the behavior we want to experience in this _hopefully_ rare case is the following:
- Don't add the empty component to the user agent string
- Add a debug log statement explaining the issue/reason for omitting the UA component.

We should double check if this violates any requirements of the UA spec(s).

botocore/useragent.py
hssyoo: Addressed in https://github.com/boto/botocore/pull/3389/commits/a743c33783692038adf7f573449eab721e06fab2

hssyoo: Addressed in https://github.com/boto/botocore/pull/3389/commits/a743c33783692038adf7f573449eab721e06fab2

botocore/useragent.py
hssyoo: Addressed in https://github.com/boto/botocore/pull/3389/commits/a743c33783692038adf7f573449eab721e06fab2

botocore/useragent.py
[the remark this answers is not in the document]
hssyoo: Addressed in https://github.com/boto/botocore/pull/3389/commits/f2aacbbc2d4ad3aa9950b5cad84b91292e7d176e

jonathan343 (approved):
LGTM!

tests/unit/test_useragent.py
```diff
@@ -193,3 +198,57 @@ def test_from_environment_unknown_platform(monkeypatch):
...
+def test_register_feature_id(client_context):
+    register_feature_id('WAITER')
+    ctx = get_context()
+    assert ctx.features == {'B'}
+
+
+def test_register_unknown_feature_id_raises(client_context):
+    with pytest.raises(ValueError) as excinfo:
+        register_feature_id('MY_FEATURE')
+    assert "Unknown feature id" in str(excinfo.value)
```
ashovlin, line 223: Is there a world where someone is on a newer downstream package, say `s3transfer`, and it registers a feature that isn't in botocore's mapping? 

Wondering if it's better to fail more gracefully? Thinking about the mismatch issues we had with the S3 checksums rollout.

Otherwise we have to update the minimum pin each time we add a new feature, right?

hssyoo: Addressed in https://github.com/boto/botocore/pull/3389/commits/6eec8771bf168ed82df2ac9845edca2004a1100b

ashovlin (commented):
🚢 Looks good to me, I can't resolve comments or approve on this repo.

PR #3392, opened 2025-02-13
Update setup and install commands

nateprewitt:
Modernize the commands to use the same python interpreter. This will avoid mismatched links that may lead to confusion about what's being used/installed.


jonathan343 (approved):
Looks good to me.
We may want to consider adding `.venv` to our `.gitignore`, but I know many people will include this in their global git ignore

PR #3396, opened 2025-02-18
RPCv2 Serializer Support

SamRemis:
This PR adds support for serializing input to the RPCv2/CBOR protocol

### DIFF
--- a/botocore/serialize.py
+++ b/botocore/serialize.py
@@ -1025,16 +1025,16 @@ class RpcV2CBORSerializer(BaseRpcV2Serializer, CBORSerializer):
         serialized['headers']['smithy-protocol'] = 'rpc-v2-cbor'
 
         if operation_model.has_event_stream_output:
-            header_val = 'application/vnd.amazon.eventstream'
+            content_type = 'application/vnd.amazon.eventstream'
         else:
-            header_val = 'application/cbor'
+            content_type = 'application/cbor'
 
         has_body = serialized['body'] != b''
         has_content_type = has_header('Content-Type', serialized['headers'])
 
-        serialized['headers']['Accept'] = header_val
+        serialized['headers']['Accept'] = content_type 
         if not has_content_type and has_body:
-            serialized['headers']['Content-Type'] = header_val
+            serialized['headers']['Content-Type'] = content_type
 
 
 SERIALIZERS = {
--- a/botocore/serialize.py
+++ b/botocore/serialize.py
@@ -43,6 +43,7 @@ import calendar
 import datetime
 import json
 import re
+import struct
 from xml.etree import ElementTree
 
 from botocore import validate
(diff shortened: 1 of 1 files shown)


botocore/serialize.py
```diff
@@ -802,10 +1013,35 @@ def _default_serialize(self, xmlnode, params, shape, name):
...
+        self._serialize_data_item(body, parameters, input_shape)
+        return bytes(body)
+
+    def _serialize_headers(self, serialized, operation_model):
+        serialized['headers']['smithy-protocol'] = 'rpc-v2-cbor'
+
+        if operation_model.has_event_stream_output:
+            header_val = 'application/vnd.amazon.eventstream'
+        else:
+            header_val = 'application/cbor'
```
nateprewitt, line 1030:
```suggestion
        if operation_model.has_event_stream_output:
            content_type = 'application/vnd.amazon.eventstream'
        else:
            content_type = 'application/cbor'
```

botocore/serialize.py
```diff
@@ -802,10 +1013,35 @@ def _default_serialize(self, xmlnode, params, shape, name):
...
+
+        has_body = serialized['body'] != b''
+        has_content_type = has_header('Content-Type', serialized['headers'])
+
+        serialized['headers']['Accept'] = header_val
```
nateprewitt, line 1035: Unrelated to the naming change, is this right? Are we stating we only accept a single response representation?
```suggestion
        serialized['headers']['Accept'] = content_type 
```

botocore/serialize.py
```diff
@@ -802,10 +1013,35 @@ def _default_serialize(self, xmlnode, params, shape, name):
...
+        if not has_content_type and has_body:
+            serialized['headers']['Content-Type'] = header_val
```
nateprewitt, line 1037:
```suggestion
            serialized['headers']['Content-Type'] = content_type
```

botocore/serialize.py
```diff
@@ -669,6 +845,41 @@ def _convert_header_value(self, shape, value):
...
+            f'/service/{service_name}/operation/{operation_name}'
+        )
+
+        input_shape = operation_model.input_shape
+        if input_shape is not None:
+            self._serialize_payload(
+                parameters, serialized, input_shape
+            )
+
+        self._serialize_headers(serialized, operation_model)
```
nateprewitt, line 873: Is `X-Amz-Target` no longer relevant for RPCv2?

nateprewitt: We also don't appear to be implementing stubbed versions of these methods.

botocore/serialize.py
SamRemis: [Spec](https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html):
> Requests for the rpcv2Cbor protocol MUST use the following behavior for setting an Accept header:

>    For requests with event streaming responses: the value of the Accept
        header MUST be application/vnd.amazon.eventstream.

>    For requests with all other response types: the value of the Accept
        header MUST be application/cbor.

That's correct

botocore/serialize.py
SamRemis: From the [spec](https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html):
 
>Requests for the rpcv2Cbor protocol MUST NOT contain an X-Amz-Target or X-Amzn-Target header.  An rpcv2Cbor request is malformed if it contains either of these headers. Server-side implementations MUST reject such requests for security reasons.

botocore/serialize.py
SamRemis: [From the spec](https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html):

    Requests for the rpcv2Cbor protocol MUST use the following behavior for setting an Accept header:

    For requests with event streaming responses: the value of the Accept
    header MUST be application/vnd.amazon.eventstream.

    For requests with all other response types: the value of the Accept
    header MUST be application/cbor.

That's correct

botocore/serialize.py
```diff
@@ -431,6 +432,181 @@ def _serialize_type_blob(self, serialized, value, shape, key):
...
+        if timestamp < 24:
+            serialized.extend(bytes([timestamp]))
+        elif timestamp < 256:
+            serialized.extend(b'\x18' + timestamp.to_bytes(1, "big"))
+        elif timestamp < 65536:
+            serialized.extend(b'\x19' + timestamp.to_bytes(2, "big"))
+        elif timestamp < 4294967296:
+            serialized.extend(b'\x1a' + timestamp.to_bytes(4, "big"))
+        else:
+            serialized.extend(b'\x1b' + timestamp.to_bytes(8, "big"))
```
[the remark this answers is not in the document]
SamRemis: This is what I came up with:

```
    def _serialize_type_integer(self, serialized, value, shape, key):
        # Major type is 0 for unsigned integers, and 1 for negative integers
        if value >= 0:
            major_type_bytes = 0 << 5
        else:
            major_type_bytes = 1 << 5
            #The only differences in serializing negative and positive integers is 
            # that for negative, we set the major type to 1 and set the value to -1 
            # minus the value
            value = -1 - value
        additional_info, num_bytes = self._get_additional_info_and_num_bytes(value)
        if num_bytes == 0:
            serialized.extend((major_type_bytes | value).to_bytes(1, "big"))
        else: 
            initial_byte = (major_type_bytes | additional_info).to_bytes(1, "big")
            serialized.extend(initial_byte + value.to_bytes(num_bytes, "big"))

    def _get_additional_info_and_num_bytes(self, value):
        # Values under 24 can be stored in the initial byte and don't need further
        # encoding
        if value < 24:
            return value, 0
        # Values between 24 and 255 (inclusive) can be stored in 1 byte and
        # correspond to additional info 24
        elif value < 256:
            return 24, 1
        elif value < 65536:
            return 25, 2
        elif value < 4294967296:
            return 26, 4
        else:
            return 27, 8
```

I'm not sure if I like it or not, it feels less readable to me, but maybe it would be more readable to someone else?  The helper method is at least reusable.

botocore/serialize.py
```diff
@@ -431,6 +433,218 @@ def _serialize_type_blob(self, serialized, value, shape, key):
...
+        else:
+            return 27, 8
+
+    def _get_initial_byte(self, major_type, additional_info):
+        # The first three bits are the major type, so we need to bitshift the major
+        # type by 5
+        major_type_bytes = major_type << 5
+        return (major_type_bytes | additional_info).to_bytes(1, "big")
+
+    def _get_bytes_for_special_numbers(self, value):
```
jonathan343, line 635: So special numbers are handled the same for both floats and doubles?

botocore/serialize.py
```diff
@@ -431,6 +433,218 @@ def _serialize_type_blob(self, serialized, value, shape, key):
         serialized[key] = self._get_base64(value)
 
 
+class CBORSerializer(Serializer):
```
jonathan343, line 436: The CBOR RFC defines the major types pretty well here: https://www.rfc-editor.org/rfc/rfc8949.html#name-major-types

In most methods were starting them with:

```python
major_type = # some number 1-7

...

... _get_initial_byte(major_type, ...)
```

Can we make these descriptive class-level constants for the `CBORSerializer` instead?

botocore/serialize.py
```diff
@@ -431,6 +433,218 @@ def _serialize_type_blob(self, serialized, value, shape, key):
...
+
+    def _get_bytes_for_special_numbers(self, value):
+        major_type = 7
+        additional_info = 25
+        initial_byte = self._get_initial_byte(major_type, additional_info)
+        if value == float('inf'):
+            return initial_byte + struct.pack(">H", 0x7C00)
+        elif value == float('-inf'):
+            return initial_byte + struct.pack(">H", 0xFC00)
+        elif math.isnan(value):
```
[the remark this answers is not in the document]
SamRemis: From my REPL:
```
>>> import numpy
>>> import math
>>> math.isnan(numpy.nan)
True
```

botocore/serialize.py
```diff
@@ -431,6 +433,218 @@ def _serialize_type_blob(self, serialized, value, shape, key):
...
+            major_type = 7
+            initial_byte = self._get_initial_byte(major_type, 27)
+            serialized.extend(initial_byte + struct.pack(">d", value))
+
+    def _serialize_type_boolean(self, serialized, value, shape, key):
+        major_type = 7
+        additional_info = 21 if value else 20
+        serialized.extend(self._get_initial_byte(major_type, additional_info))
+
+    def _get_additional_info_and_num_bytes(self, value):
```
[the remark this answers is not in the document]
SamRemis: That's mostly right.  It's effectively checking how many bytes we are going to need to store that particular value.  
> Based on the value, we're determining the number of additional bytes we need to represent the value. Is that correct? Can it ever by anything else?

For values under 24, that value just goes straight in the initial byte.   That's the only caveat here.

botocore/serialize.py
SamRemis: Yep :) they're the same in CBOR.  Both are major type 7, just stored in different number of bytes.  

The reason we have different methods is that the serializer is for going from C2J types to CBOR data, and they are different types in C2J.  We serialize to the proper major type based on the C2J type, so both need to be handled.  

We could theoretically consolidate some of this logic into a common method, but I can't think of a more readable way to do that than what I've written here.  It would have to either branch into an if/else based on the C2J type or do unnecessary checks.  I'm open to suggestions here.

botocore/serialize.py
```diff
@@ -431,6 +432,181 @@ def _serialize_type_blob(self, serialized, value, shape, key):
         serialized[key] = self._get_base64(value)
 
 
+class CBORSerializer(Serializer):
+
+
+    def _serialize_data_item(self, serialized, value, shape, key=None):
+        method = getattr(self, f'_serialize_type_{shape.type_name}')
```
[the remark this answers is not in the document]
jonathan343: Just to follow up here, have we collected a list of all shape types that exist in the SDK? I though at one point we supported `character` which I don't see here. `character` may have never been used or removed entirely, I actually don't know, but our `parsers.py` does have support for it. That might be dead code now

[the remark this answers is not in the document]
SamRemis: Please ignore the _terrible_ code here, but here's the script I ran to confirm this using a file I modified from 2 years ago:

```
import os
import json
shapes = set()
data_path = '/Users/<user>/<projects-directory>/botocore/botocore/data'

for root, dirs, files in os.walk(data_path):
    level = root.replace(data_path, '').count(os.sep)
    if level == 2:
        if os.path.isfile(os.path.join(root, "service-2.json")):
            with open(os.path.join(root, "service-2.json")) as f:
                data = json.load(f)
                for _, shape in data['shapes'].items():
                    shapes.add(shape['type'])
print(shapes)
```

Output: 
`{'timestamp', 'float', 'boolean', 'long', 'integer', 'string', 'structure', 'blob', 'map', 'double', 'list'}`

botocore/serialize.py
```diff
@@ -431,6 +433,246 @@ def _serialize_type_blob(self, serialized, value, shape, key):
...
+                initial_byte + map_length.to_bytes(num_bytes, "big")
+            )
+
+        members = shape.members
+        for member_key, member_value in value.items():
+            member_shape = members[member_key]
+            if 'name' in member_shape.serialization:
+                member_key = member_shape.serialization['name']
+            if member_value is not None:
+                self._serialize_type_string(serialized, member_key, None, None)
```
jonathan343, line 569: In `_serialize_type_map` we used `_serialize_data_item` instead of `_serialize_type_string` for the key. What is the need for this difference? The [Smithy docs](https://smithy.io/2.0/spec/aggregate-types.html#map) on maps say the key MUST be a string.

> 3.2. Map
The map type represents a map data structure that maps string keys to homogeneous values. A map requires a member named key that MUST target a string shape and a member named value.

SamRemis: That is the key we are serializing on this line :)

SamRemis: Ah, I see what you're saying.  Yeah, if you make that code suggestion, I'll commit it.  Or I can push up a revision on the other method

jonathan343 (approved):
PR looks good to merge into the feature branch. @SamRemis will address the failing test when opening the PR to merge from the feature branch into `develop`. This is necessary since the test can't be fixed until https://github.com/boto/botocore/pull/3391/files is merged.

PR #3399, opened 2025-02-20
Refactored usages of deprecated method utcnow()

TehWeifu:
Refactored usages of utcnow() method for method now(datetime.UTC). The first method is being deprecated and displaying deprecation warnings

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -422,7 +422,7 @@ class SigV4Auth(BaseSigner):
     def add_auth(self, request):
         if self.credentials is None:
             raise NoCredentialsError()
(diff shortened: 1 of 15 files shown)


ArzelaAscoIi:
+1 🙏

kurtmckee:
I've already opened a PR to address this: #3239.

The method used in this PR [has been previously attempted](https://github.com/boto/botocore/issues/3088#issuecomment-1861405074), but had to be discarded because it made the datetime objects timezone-aware.

In addition, the `datetime.UTC` alias used in this PR is only available in Python 3.11 and higher, and would cause `AttributeError`s in lower Python versions.

nateprewitt:

PR #3413, opened 2025-03-03
Add concurrency control to run-tests and run-crt-test workflows

arandito:
This PR adds concurrency control to the `run-tests` and `run-crt-test` GitHub Actions workflows.

Problem Description

Currently, when multiple commits are pushed to the same branch in succession, the `run-tests` and `run-crt-test` workflows are triggered for each push. This results in excessive use of GitHub Actions minutes as it's only necessary to run the workflows for the most recent push within a pull request.

Solution in This PR

This PR adds concurrency control to the `run-tests` and `run-crt-test` runners to cancel in-progress jobs when new commits are pushed to the same branch. Only the latest push will now trigger the workflows. However, this behavior does not apply to the `develop` and `master` branches, where all pushed commits should continue triggering `run-tests` and `run-crt-test`.

### DIFF
--- a/.github/workflows/run-crt-test.yml
+++ b/.github/workflows/run-crt-test.yml
@@ -8,6 +8,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ (github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/master') && github.run_id || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: '${{ matrix.os }}'
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -8,6 +8,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ (github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/master') && github.run_id || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: '${{ matrix.os }}'


.github/workflows/run-crt-test.yml
```diff
@@ -8,6 +8,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
```
jonathan343, line 11: Just curious, why not use something like in this example from GitHub?
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-only-cancel-in-progress-jobs-on-specific-branches

jonathan343: I imagine something like `cancel-in-progress: ${{ !contains(github.ref, 'develop') && !contains(github.ref, 'master') }}`

arandito: I initially implemented it that way, but during testing, I found that `cancel-in-progress` does not fully prevent all job cancellations. By default, the `concurrency` feature allows for at most one running job and one pending job per concurrency group (see [GitHub docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#using-concurrency-in-different-scenarios)).  

For example, if two pushes occur in quick succession, one job runs while the other remains pending. A third push cancels the pending job and replaces it. Enabling `cancel-in-progress` changes this behavior by allowing only one running job at a time, with no pending jobs. If set to `false`, `concurrency` behaves as expected by default.  

To fully prevent job cancellations in `develop` and `master`, each run needs a unique concurrency group based on `github.run_id`. This ensures that every run is isolated and does not cancel in-progress jobs regardless of the value of `cancel-in-progress`. For other branches, the group remains branch-specific, limiting concurrency to one running job per branch at a time.

jonathan343 (approved):
🚢

PR #3421, opened 2025-03-07
Add SigV4 note to generate_presigned_post documentation

adev-code:
## Issue
https://github.com/boto/boto3/issues/4351 
Title: (short issue description)generate_presigned_post documentation should provide a link to S3 doc explaining regional default variation in response #4351

## Summary
This PR adds a documentation note to `generate_presigned_post` in `signers.py`, explaining the difference 
between Signature Version 2 (SigV2) and Signature Version 4 (SigV4). As suggested on Issue 4351, it would be feasible if the information from the S3 API Documentation https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html would be included to the Boto3 Doc: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/generate_presigned_post.html#generate-presigned-post 

## Changes
- Added a `.. note::` section in `signers.py` docstring.
- Included a link to AWS's official SigV4 documentation and its note.

## Testing
- Ran `make html` in `docs/` to verify the doc renders correctly.

### DIFF
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -848,7 +848,7 @@ def generate_presigned_post(
         For more details, see:
         `AWS Signature Version 4 <https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html>`_.
 
-        
+
     :type Bucket: string
     :param Bucket: The name of the bucket to presign the post to. Note that
         bucket related conditions should not be included in the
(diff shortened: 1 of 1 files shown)


SamRemis:
I'd rather see us make a more useful comment that clarifies which signature version will be used by default with botocore, and why customers may want to change it, then link to the docs page where this was copied from.

Just copy/pasting the note from another AWS docs site could create conflicting information if the other page ever updates that note.

PR #3422, opened 2025-03-07
Track RPC v2 CBOR feature

arandito:
Description

This PR adds support for tracking usage of the `PROTOCOL_RPC_V2_CBOR` feature.

User agent strings can now include an 'M' in the `m/` metrics field when an operation is called using the Smithy RPC v2 CBOR protocol. 

User agent string example:
```
Boto3/1.37.9 md/Botocore#1.37.9 ua/2.1 os/macos#24.3.0 md/arch#arm64 lang/python#3.12.8 md/pyimpl#CPython m/M cfg/retry-mode#legacy Botocore/1.37.9
```

### DIFF
--- a/tests/functional/test_rpc_v2_cbor.py
+++ b/tests/functional/test_rpc_v2_cbor.py
@@ -1,0 +1,90 @@
+# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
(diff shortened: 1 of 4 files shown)


tests/functional/test_rpc_v2_cbor.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+    client = patched_session.create_client(
+        'mockservice', region_name='us-west-2'
+    )
+    with ClientHTTPStubber(client) as stub_client:
+        stub_client.add_response()
+        # The mock CBOR operation registers `'PROTOCOL_RPC_V2_CBOR': 'M'`
+        client.mock_operation(MockOpParam='mock-op-param-value')
+    ua_string = get_captured_ua_strings(stub_client)[0]
+    feature_list = parse_registered_feature_ids(ua_string)
```
SamRemis, line 89: This function isn't quite right, so this assertion also isn't.  This only ever returns the string after `m/` but before the first comma.  So take for instance the valid user agent string `m/D,M` - if this got returned here, it would incorrectly fail.  Conversely, if there was a multi character UA string like `m/Ma,D`, that would pass even though we aren't looking for `Ma`, we are looking for `M`.

SamRemis (approved):

nateprewitt (approved):

PR #3438, opened 2025-04-01
Add a condition to validate that the event stream payload size doesn't exceed the max payload size

ubaskota:
This change adds a condition to validate that the payload length in an event stream is not greater than 24Mb.

Test: Various tests were performed with different prelude values to verify their expected functionality.

### DIFF
--- a/.changes/next-release/enhancement-Eventstream-60668.json
+++ b/.changes/next-release/enhancement-Eventstream-60668.json
@@ -1,5 +1,5 @@
 {
   "type": "enhancement",
   "category": "Eventstream",
-  "description": "The limit for event streams maximum payload size was raised from 16Mb to 24Mb."
+  "description": "The event streams maximum payload size is now required to be 24Mb or less."
 }


jonathan343 (changes requested):
The changes look good to me, but can we get a changelog entry? We added one when removing the check so we should probably add one for reintroducing it. If you haven't done it before, you can generate one by running `./scripts/new-change`.

Adding this will generate an entry in [botocore/CHANGELOG.rst](https://github.com/boto/botocore/blob/develop/CHANGELOG.rst) and [boto3/CHANGELOG.rst](https://github.com/boto/boto3/blob/develop/CHANGELOG.rst) and inform customers about the change.

.changes/next-release/enhancement-Eventstream-60668.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Eventstream",
+  "description": "The limit for event streams maximum payload size was raised from 16Mb to 24Mb."
```
nateprewitt, line 4: nit; we weren't enforcing 16Mb prior to this release, we should try to frame this as the effective change.
```suggestion
  "description": "The event streams maximum payload size is now required to be 24Mb or less."
```

jonathan343, line 4: This is inaccurate since the 16Mb validation doesn't exist currently. I suggested new wording below:
```suggestion
  "description": "Reintroduce a client-side maximum payload size limit, now set to 24Mb."
```

jonathan343: Wording was addressed in b1960151a2335f5f87a6a9d44ecb2b7f65c66768

nateprewitt (approved):

PR #3440, opened 2025-04-06
feat: Add user configuration directories

neilramsay:
Add the session configuration variable config_dir (defaults to ~/.aws/config.d).

This allows a user to separate their configuration into separate files. This is useful for separate identities (work versus personal), and for sharing configuration with others.

The config_dir is searched recursively, and files are loaded and merged in alphanumeric order.
The config_file (defaults to ~/.aws/config) takes precedence over files in the config_dir.


neilramsay:
I've also noticed the documentation for configuration files is over in the [boto3 configuration documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#), so we'll need to update it after this is merged.

PR #3185, opened 2024-05-23
Add support for Python 3.13

nateprewitt:
This PR will start provisional testing on the recently released beta for Python 3.13. This will be used to catch any issues early before the official release in October.

PR #3450, opened 2025-04-14
Test on 3.14 beta

nateprewitt:
This is a draft PR to see where we're at with the existing 3.14 beta build. We'll use this to track any current issues and path to support in October with the official release.


tests/functional/leak/test_resource_leaks.py
```diff
@@ -23,7 +23,7 @@ class TestDoesNotLeakMemory(BaseClientDriverTest):
     INJECT_DUMMY_CREDS = True
     # We're making up numbers here, but let's say arbitrarily
     # that the memory can't increase by more than 10MB.
-    MAX_GROWTH_BYTES = 10 * 1024 * 1024
+    MAX_GROWTH_BYTES = 15 * 1024 * 1024
```
nateprewitt, line 26: We had [similar memory issues](https://github.com/boto/botocore/pull/3185#issuecomment-2127895144) with the 3.13 beta builds that were resolved before launch.

This is a change we can make temporarily (12MB may be more reasonable, I'm only seeing it peak slightly over 11MB). We'll just want to make sure we come back and try to reduce it later if the issues are resolved. We'll update the comment above too once we arrive at a final number.

nateprewitt:
It looks like something potentially changed with reference counting. Not sure yet if it's an optimization or if we're losing something with the lru cache.


[Affected Test](https://github.com/boto/botocore/blob/ba6ea09b2e3998458fc3a9556115b9851b3c5a45/tests/unit/test_utils.py#L3564) - Jonas added these a couple years ago but it looks like we're also testing around a quirk of reference counting. Taking a look to see if we can streamline the test more once we root cause.
```
     def test_lru_cache_weakref():
        class ClassWithCachedMethod:
            @lru_cache_weakref(maxsize=10)
            def cached_fn(self, a, b):
                return a + b
    
        cls1 = ClassWithCachedMethod()
        cls2 = ClassWithCachedMethod()
    
        assert cls1.cached_fn.cache_info().currsize == 0
>       assert getrefcount(cls1) == 2
E       assert 1 == 2
E        +  where 1 = getrefcount(<tests.unit.test_utils.test_lru_cache_weakref.<locals>.ClassWithCachedMethod object at 0x324669a90>)
```

jonathan343 (approved):

PR #3453, opened 2025-04-15
Add support to register features related to checksums with required functional tests

ubaskota:
Description

This PR adds feature IDs for checksum algorithms and checksum request behaviors, and registers them for feature tracking.

Changes:

    Adds feature IDs for all checksum algorithms.
    Adds checksum feature IDs for all requests and responses.
    Registers feature IDs when the corresponding features are used.
    Adds functional tests.

Tests:

    Manually verified that the feature ID is correctly registered in the user agent.
    Validated the changes with functional tests.


botocore/httpchecksum.py
```diff
@@ -355,6 +356,17 @@ def apply_request_checksum(request):
     checksum_context = request.get("context", {}).get("checksum", {})
     algorithm = checksum_context.get("request_algorithm")
 
+    if "method" in request and (
```
jonathan343, line 359: I have a few questions about the logic here:
- Why is this being guarded by a conditional checking for the `HEAD` and `GET` HTTP methods? Do we only want to track this metric in cases where response checksum validation would occur? My understanding is we want to always track this metric, despite what operations are being used.
- This method is responsible for applying a flexible checksum to a request. To me it doesn't make sense to put this logic here since it's related to response checksum config. I think this makes more sense to go where we determine the config values for `response_checksum_validation` in `args.py` if possible.

botocore/httpchecksum.py
```diff
@@ -410,6 +435,19 @@ def _apply_request_trailer_checksum(request):
         headers["Content-Encoding"] = "aws-chunked"
     headers["X-Amz-Trailer"] = location_name
 
+    checksum_algorithm_name = algorithm["algorithm"]
```
jonathan343, line 438: This looks identical to the logic above, I'd prefer we consolidate this to one location which makes it easier to maintain. For example, if we make a change in the future we only have to remember to make it in one place.

botocore/httpchecksum.py
```diff
@@ -387,6 +399,19 @@ def _apply_request_header_checksum(request):
     digest = checksum_cls().handle(request["body"])
     request["headers"][location_name] = digest
 
+    checksum_algorithm_name = algorithm["algorithm"]
+    request_checksum_calculation = request["context"][
+        "client_config"
+    ].request_checksum_calculation
+    request_checksum_calculation_feature_id = (
+        "FLEXIBLE_CHECKSUMS_REQ_" + request_checksum_calculation.upper()
+    )
```
jonathan343, line 408: As mentioned above, I think this check specifically should go in `args.py` if possible. That's where we determine what the config values are.

tests/functional/test_httpchecksum.py
```diff
@@ -549,3 +598,96 @@ def test_unsuccessful_streaming_response_checksum_validation(
...
+        ),
+        (
+            "SHA1",
+            "when_required",
+            "when_supported",
+            request_payload,
+            ["X", "a", "b"],
+        ),
+        ("SHA256", None, "when_supported", request_payload, ["Y", "Z", "b"]),
+        ("SHA1", "when_required", None, request_payload, ["X", "a", "b"]),
```
jonathan343, line 636: The first case already tests the default values so I don't think these are needed

tests/functional/test_httpchecksum.py
```diff
@@ -549,3 +598,96 @@ def test_unsuccessful_streaming_response_checksum_validation(
...
+            [
+                (
+                    "SHA256",
+                    None,
+                    "when_required",
+                    request_payload,
+                    ["Y", "Z", "c"],
+                )
+            ]
+        )
```
jonathan343, line 652: I don't think this case is providing any value. We should instead include a test for `CRC64NVME` which is only available when the CRT is available, similar to CRC32C.

Lmk if you have any questions about this.

botocore/httpchecksum.py
```diff
@@ -432,6 +435,14 @@ def _apply_request_trailer_checksum(request):
     )
 
 
+def _register_checksum_algorithm(algorithm):
+    checksum_algorithm_name = algorithm["algorithm"]
+    checksum_algorithm_name_feature_id = (
+        "FLEXIBLE_CHECKSUMS_REQ_" + checksum_algorithm_name.upper()
```
jonathan343, line 441: We'll need to handle the `CRC64NVME` case since the feature id associated with it used `CRC64`.

botocore/httpchecksum.py
```diff
@@ -432,6 +435,16 @@ def _apply_request_trailer_checksum(request):
     )
 
 
+def _register_checksum_algorithm_feature_id(algorithm):
+    checksum_algorithm_name = algorithm["algorithm"]
```
jonathan343, line 439: We should just make this `checksum_algorithm_name = algorithm["algorithm"].upper()` instead of having to do it each time we refer to it down below.

botocore/useragent.py
```diff
@@ -60,6 +60,15 @@
     'S3_TRANSFER': 'G',
     'ENDPOINT_OVERRIDE': 'N',
     'SIGV4A_SIGNING': 'S',
+    'FLEXIBLE_CHECKSUMS_REQ_CRC32': 'U',
+    'FLEXIBLE_CHECKSUMS_REQ_CRC32C': 'V',
+    'FLEXIBLE_CHECKSUMS_REQ_CRC64': 'W',
```
jonathan343, line 65: I'm thinking we should just refer to this as `FLEXIBLE_CHECKSUMS_REQ_CRC64NVME` instead of `FLEXIBLE_CHECKSUMS_REQ_CRC64`. At the end of the day, all that matters is that we register the `'W'` feature id. It's kinda weird to use`CRC64` and then special case for `CRC64NVME`.

What do you think?

botocore/httpchecksum.py
```diff
@@ -432,6 +435,16 @@ def _apply_request_trailer_checksum(request):
...
+    if checksum_algorithm_name.upper() == "CRC64NVME":
+        checksum_algorithm_name = "CRC64"
+    checksum_algorithm_name_feature_id = (
+        "FLEXIBLE_CHECKSUMS_REQ_" + checksum_algorithm_name.upper()
+    )
```
jonathan343, line 444: I think it's better practice to use a f-string for cases like this. Something like:
```python
f"FLEXIBLE_CHECKSUMS_REQ_{checksum_algorithm_name}"
```

Note: This assumes you accepted my change above related to `upper()`

botocore/args.py
```diff
@@ -900,8 +900,20 @@ def _handle_checksum_config(
                 config_value=value,
                 valid_options=valid_options,
             )
+        self._register_checksum_config_feature_ids(value, config_key)
         config_kwargs[config_key] = value
 
+    def _register_checksum_config_feature_ids(self, value, config_key):
+        if config_key == "request_checksum_calculation":
+            checksum_calculation_feature_id = (
```
jonathan343, line 908: The `checksum_calculation_feature_id` name only makes sense in the if statement and not in the elif part. We should use something like `checksum_config_feature_id`.

Also, we should handle the case where `config_key` is not `"request_checksum_calculation"` or `"response_checksum_validation"`. We shouldn't fail for a user-agent related issue. I think this would currently fail with:
```
NameError: name 'checksum_calculation_feature_id' is not defined
```

PR #3457, opened 2025-04-18
Alias command for socialmessaging

SamRemis:
During a previous manual customization to rename this method, there was a human error which caused the command to be named incorrectly.  This PR allows both the old name and the new name to be used in botocore and boto3.  This must be released on the same day as [the associated CLI v1 PR](https://github.com/aws/aws-cli/pull/9450).

Generating the docs shows that this now has the correctly documented name and does not show the old name:
image.


alexgromero (approved):
LGTM :shipit:

PR #2816, opened 2022-11-19
Move signature version selection to use service model

nateprewitt:
## Overview

This PR will fix a long standing bug in Botocore, moving signature resolution off using `endpoints.json` as the source of truth.

In previous releases, Botocore has ignored the service model's configured `signatureVersion` ([example](https://github.com/boto/botocore/blob/2df5c5baf51325b16b8d2891b74e4731cdfea06f/botocore/data/lambda/2015-03-31/service-2.json#L9)). Instead, it relied on either a service-specific configuration in `endpoints.json` or falling back to the partition default (sigv4). While the partition-level fallback wasn't intended, it was masked from the testing infrastructure due to every active service supporting sigv4.

As we introduce new signing schemes such as the recently added `bearer` auth, we are seeing these schemes ignored at the service level. To address this, this patch will now prefer the service configured `signatureVersion`. It will still be possible to supplement these through the endpoints.json file, but this method is on the deprecation path and may be removed in a future release.

We'd advise anyone relying on this method to ensure any user-owned service models have an appropriately configured `signatureVersion` in the their metadata. Alternatively, you may use the botocore Config option, `signature_version` as [detailed in the docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-the-config-object) to choose specific signature versions per client.

## Testing

To ensure we're not introducing regressions around signing, we validated all existing models `signatureVersion`'s match what was previously resolved in `endpoints.json`.

**Before**
<details>

```json
{
   "accessanalyzer":"v4",
   "account":"v4",
   "acm":"v4",
   "acm-pca":"v4",
   "alexaforbusiness":"v4",
   "amp":"v4",
   "amplify":"v4",
   "amplifybackend":"v4",
   "amplifyuibuilder":"v4",
   "apigateway":"v4",
   "apigatewaymanagementapi":"v4",
   "apigatewayv2":"v4",
   "appconfig":"v4",
   "appconfigdata":"v4",
   "appflow":"v4",
   "appintegrations":"v4",
   "application-autoscaling":"v4",
   "application-insights":"v4",
   "applicationcostprofiler":"v4",
   "appmesh":"v4",
   "apprunner":"v4",
   "appstream":"v4",
   "appsync":"v4",
   "athena":"v4",
   "auditmanager":"v4",
   "autoscaling":"v4",
   "autoscaling-plans":"v4",
   "backup":"v4",
   "backup-gateway":"v4",
   "backupstorage":"v4",
   "batch":"v4",
[...]

PR #3465, opened 2025-05-09
Prefer signature version from service models over endpoints.json

ashovlin:
### Background
An upcoming service is re-using an existing endpoint prefix that is still present in `endpoints.json` with an explicit `signatureVersions`. 

The new service is specifying a different `signatureVersion` in its service model. However botocore is still preferring the version from `endpoints.json`

https://github.com/boto/botocore/pull/2816 in 2022 last touched this area. However this logic: https://github.com/boto/botocore/blob/6f783fe339578cd7444bbd0296aade4ebdb0ce43/botocore/client.py#L836-L843

 only prefers the service model  over `endpoints.json` if the service model's signature version is _not_ in the this list:
https://github.com/boto/botocore/blob/6f783fe339578cd7444bbd0296aade4ebdb0ce43/botocore/client.py#L69-L78

My interpretation is that this allowed new services to model new signature versions such as `bearer`, but preserved our original behavior of using `endpoints.json` as the source of truth.

### Approach
Now, I removed that `_LEGACY_SIGNATURE_VERSIONS` set. We'll prefer the service model it its set, but still fall back to `endpoints.json` if not.

This required some custom handling for S3, S3Control, and Import/Export to preserve their current resolved values.

**For Reviewers**: Long term, we likely want to remove `endpoints.json` entirely. I didn't do so yet, as that `resolved` is plumbed around pretty heavily, and we do have unit tests that are guarding the current code paths. But let me know if you'd rather do so now.

### Testing
1. Added a unit test exercising the scenario that the upcoming service is hitting.
2. Ran the following both before and after this change, confirmed that there was not a diff in the output:
```
import json
import botocore.session

signature_versions = {}
for service_name in session.get_available_services():
    client = session.create_client(service_name)
    signature_versions[service_name] = client.meta.config.signature_version

print(json.dumps(signature_versions, indent=2))
```

### DIFF
--- a/.changes/next-release/bugfix-endpoints-38893.json
+++ b/.changes/next-release/bugfix-endpoints-38893.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "``endpoints``",
+  "description": "Prefer signature version from service models over endpoints.json."
+}
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -66,18 +66,6 @@ from botocore.utils import S3RegionRedirector  # noqa
 from botocore import UNSIGNED  # noqa
 
 
-_LEGACY_SIGNATURE_VERSIONS = frozenset(
-    (
-        'v2',
-        'v3',
-        'v3https',
-        'v4',
-        's3',
-        's3v4',
-    )
-)
-
-
 logger = logging.getLogger(__name__)
 history_recorder = get_global_history_recorder()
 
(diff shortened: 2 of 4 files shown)


.changes/next-release/bugfix-endpoints-38893.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "``endpoints``",
+  "description": "Prefer signature version from service models over endpoints.json."
+}
```
ashovlin, line 5: Unsure if we want this, as it should be transparent. #2816 didn't have a changelog entry.

nateprewitt: This is probably a nuance we don't need to highlight. These aren't public interfaces and as long as the change is a no-op for the end user, it's kind of moot.

botocore/client.py
```diff
@@ -832,27 +820,34 @@ def _resolve_signature_version(self, service_name, resolved):
...
-            if service_name == 's3':
-                return 's3v4'
+        if self.service_signature_version is not None:
+            # Prefer the service model
+            potential_versions = [self.service_signature_version]
+            # importexport is modeled with V2, but we previously allowed
+            # endpoints.json to move it to v4 before switching this method
+            # to prefer the service models
+            if service_name == 'importexport':
+                return 'v4'
```
nateprewitt, line 836: Do we have any other options for this? S3 has it's own special cases throughout the code but adding more random service conditionals is ideally avoided. I'd maybe suggest we add a handler to set this explicitly for the service instead.

botocore/client.py
```diff
@@ -832,27 +857,29 @@ def _resolve_signature_version(self, service_name, resolved):
...
-            and self.service_signature_version
-            not in _LEGACY_SIGNATURE_VERSIONS
-        ):
-            # Prefer the service model as most specific
-            # source of truth for new signature versions.
-            potential_versions = [self.service_signature_version]
+        # These have since added the "auth" key to the service model
+        # with "aws.auth#sigv4", but preserve existing behavior from
+        # when we preferred endpoints.json over the service models
+        if service_name in ['s3', 's3-control']:
```
nateprewitt, line 863: nit; Static lists we should do as tuples instead of lists ideally so we're not allocating unnecessary memory.
```suggestion
        if service_name in ('s3', 's3-control'):
```

.changes/next-release/bugfix-endpoints-38893.json
ashovlin: Removed via https://github.com/boto/botocore/pull/3465/commits/0da938819815bf35605d85519752de3c0a2732ef/

nateprewitt (approved):
Comment on a nit, but otherwise LGTM.

botocore/client.py
ashovlin: One note on the approach in https://github.com/boto/botocore/pull/3465/commits/b2da232e0219bfa449aa1639daf98f9c338db3f3:

`client.meta.config.signature_version` for `importexport` will now return `v2` (instead of `v4`) since this doesn't execute the `choose-signer` event.

I removed that assertion from the functional test, but left the assertion that the request ultimately has the V4 header.

Issue #3468, opened 2025-05-13
S3 CRC64NVME breaks with awscrt==0.27.1 due to version detection


alexrashed:
### Describe the bug

With `awscrt==0.27.1`, [which was released on the 8th of May](https://pypi.org/project/awscrt/0.27.1/), the `awscrt` version detection in `botocore` introduced with https://github.com/boto/botocore/pull/3370 fails:
- `awscrt==0.27.1`:
  ```
  print(f"{awscrt.__version__}")
  # v0.27.1
  ```
- `awscrt==0.26.1`:
  ```
  print(f"{awscrt.__version__}")
  # 0.26.1
  ```
This results in boto3 rejecting `CRC64NVME` as valid S3 `ChecksumAlgorithm`.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

- `awscrt` version is correctly detected.
- `CRC64NVME` is accepted as valid S3 checksum algorithm if awscrt >= 0.27.1 is installed.

### Current Behavior

Using `ChecksumAlgorithm="CRC64NVME"` fails with:
```
  File ".../.venv/lib/python3.11/site-packages/botocore/context.py", line 123, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.11/site-packages/botocore/client.py", line 995, in _make_api_call
    resolve_checksum_context(request_dict, operation_model, api_params)
  File ".../.venv/lib/python3.11/site-packages/botocore/httpchecksum.py", line 256, in resolve_checksum_context
    resolve_request_checksum_algorithm(request, operation_model, params)
  File ".../.venv/lib/python3.11/site-packages/botocore/httpchecksum.py", line 296, in resolve_request_checksum_algorithm
    raise FlexibleChecksumError(
botocore.exceptions.FlexibleChecksumError: Unsupported checksum algorithm: crc64nvme
```

### Reproduction Steps

```
import boto3
import uuid
import awscrt

s3 = boto3.resource('s3')
bucket = s3.Bucket(str(uuid.uuid4()))
print(f"AWS CRT version string: {awscrt.__version__}")
try:
    bucket.create(CreateBucketConfiguration={'LocationConstraint': s3.meta.client.meta.region_name})
    bucket.put_object(
        Key="aws-crt-checksum-test",
        Body="yeah, this won't work anyways",
        ChecksumAlgorithm="CRC64NVME"
    )
finally:
[...]

PR #3469, opened 2025-05-13
fix awscrt version detection

alexrashed:
This PR contains a small adjustment to how the `awscrt` version is detected, such that it is compatible with the version string of `awscrt>=0.27.1`.

Fixes https://github.com/boto/botocore/issues/3468.


nateprewitt:
Resolving in favor of the upstream fix in https://github.com/awslabs/aws-crt-python/pull/654. Thanks for the proactive PR, @alexrashed!

PR #3473, opened 2025-05-16
Protocol selection bugfix

SamRemis:
This resolves a bug that has yet to impact customers where a different protocol could have theoretically been used to initialize the parser and serializer due to decentralized logic of initializing the serializer and parser objects.  

This could not yet impact customers due to a coincidence of the internal logic that dictates what gets set as the singular "protocol" for the protocol trait having previously always matched the protocol which we resolved.  This PR fixes the bug and allows us to control the priority order of our protocols to match the best customer experience.  

This priority order was previously static across SDKs, but now that we are able to choose what suits our customers best, we are also dropping CBOR down the priority list for the reasons mentioned in #3471.


botocore/args.py
```diff
@@ -26,6 +26,13 @@
 import botocore.serialize
 from botocore.config import Config
 from botocore.endpoint import EndpointCreator
+
+# We have moved the PRIORITY_ORDERED_SUPPORTED_PROTOCOLS list out of this file
+# so we import it here to maintain backwards compatibility.  The redundant "as"
+# statement tells tools like Ruff not to remove it
+from botocore.model import (
```
SamRemis, line 33: Note: importing this from args.py in model.py instead creates a circular import, so we unfortunately need to do it this way.

nateprewitt: Why does this need to live in `model.py`?

tests/functional/models/appflow-with-more-protocols/2020-08-23/endpoint-rule-set-1.json
```diff
@@ -0,0 +1,314 @@
+{
```
nateprewitt, line 1: Do we need this or the paginator file?

tests/functional/models/appflow-with-more-protocols/2020-08-23/service-2.json
```diff
@@ -0,0 +1,5851 @@
+{
```
nateprewitt, line 1: This is a lot of overhead and makes it difficult to understand what's important in the file. Would it make more sense to do something slimmed down? Either a basic foo service with a single protocol or something like [this](https://github.com/boto/botocore/blob/develop/tests/functional/models/test-discovery-endpoint/2020-08-20/service-2.json)?

tests/functional/test_protocol_selection.py
```diff
@@ -0,0 +1,81 @@
...
+        aws_secret_access_key='bar',
+    )
+
+    # This test would not be effective if the protocols were ever changed or
+    # the order of json and xml were flipped.  We need to make sure that
+    # json comes before xml and the list contains both
+    service_model_metadata = client.meta.service_model.metadata
+    assert PRIORITY_ORDERED_SUPPORTED_PROTOCOLS.index(
+        'rest-xml'
+    ) > PRIORITY_ORDERED_SUPPORTED_PROTOCOLS.index('rest-json')
```
nateprewitt, line 52: Is this piece relevant for the protocol selection or is it an implementation detail? If we get back `service_model_metadata['protocols'] == ['rest-json', 'rest-xml']` as True, that would imply this is also true, right?

tests/functional/test_protocol_selection.py
```diff
@@ -0,0 +1,81 @@
...
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+from botocore.model import PRIORITY_ORDERED_SUPPORTED_PROTOCOLS
```
nateprewitt, line 13: This may live better in `compat` or `utils` if we need it in a leaf node.

botocore/args.py
SamRemis: It doesn't - moved it to utils.py based on your comment below.  There were more similar constants there, so it made more sense there.

tests/functional/models/appflow-with-more-protocols/2020-08-23/endpoint-rule-set-1.json
SamRemis: Nope- vastly simplified the models for the test

tests/functional/test_protocol_selection.py
SamRemis: Well I'm not quite sure what you mean by this one.  

The purpose of this is effectively to make sure that the `protocol` field has a lower priority than the one we select.  Now that you've said that, I can change the test to check that explicitly instead of this weird coincidental check we are doing here.  But this is basically here to make sure the following can't happen:

1. We have changed the priority list so that now XML comes before JSON
2. We regress the fixes in this PR and we are using the `protocol` field for the parser again, but because XML is now higher priority, the tests don't catch it.  

I have updated this check, but if you have a better way to make sure that this test continues to be effective, I'm open to suggestions.

botocore/args.py
```diff
@@ -26,6 +26,13 @@
 import botocore.serialize
 from botocore.config import Config
 from botocore.endpoint import EndpointCreator
+
+# We have moved the PRIORITY_ORDERED_SUPPORTED_PROTOCOLS list out of this file
+# so we import it here to maintain backwards compatibility.  The redundant "as"
+# statement tells tools like Ruff not to remove it
```
kdaily, line 32: Couldn't you put a `# noqa` comment on the line to achieve this, and then wouldn't need to document this?

SamRemis: `noqa` and the redundant import are effectively the same thing, right?

I wanted to keep the comment there to explain why the import is necessary and prevent people from removing it in the future.  Do you think it's better without it?

tests/functional/test_protocol_selection.py
```diff
@@ -0,0 +1,72 @@
...
+                    'content-length': '17',
+                    'content-type': 'application/x-amz-json-1.1',
+                    'date': 'Fri, 31 Oct 2024 01:46:30 GMT',
+                    'x-amzn-requestid': 'abcd',
+                },
+                'HTTPStatusCode': 200,
+                'RequestId': 'abcd',
+                'RetryAttempts': 0,
+            },
+        }
```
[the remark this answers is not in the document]
SamRemis: That would be better for sure and I considered it; could also pass by coincidence if the parser happened to be similar (for instance different versions of JSON parsing).  

It's currently resolved [here](https://github.com/boto/botocore/blob/bb52a338ed992f83d4ec2317aae953380ae564d2/botocore/endpoint.py#L304), but not emitted as a part of any sort of event and a reference does not exist to it on the client.  Since there's no active reference to the parser, the only way that I could think of at the time test this would was to use the `gc` module and see if an instance of the class is set to be garbage collected, but that seems like an awful idea.

This comment helped to reconsider, and now I've used mocking a method on the expected parser to accomplish this.  If you have a better solution in mind, I'm open to discussion.

nateprewitt (approved):

PR #3479, opened 2025-05-23
feat: add support for accountId in imds credentials

ubaskota:
This PR is not ready to be merged because the new “-extended” metadata endpoint is not available yet.

This PR does the following:
```
* Add support for the new *-extended/credentials IMDS endpoint, with automatic fallback to the legacy endpoint if unavailable.
* Support the AWS_EC2_INSTANCE_PROFILE_NAME environment variable and related settings.
* Retrieve the AWS account ID during IMDS calls, for the new endpoint.
* Adds required unit tests as defined in the SEP.
```

Tests:
`* Validated the changes with unit tests `


botocore/configprovider.py
```diff
@@ -192,6 +192,12 @@
         None,
         utils.ensure_boolean,
     ),
+    'ec2_instance_profile_name': (
```
jonathan343, line 201: If we're allowing this to be configured as an environment variable and in the AWS config file, I'd assume we also want this to be added to the `Config` object which I don't see in this PR.

botocore/utils.py
```diff
@@ -563,44 +573,32 @@ def _log_imds_response(self, response, reason_to_log, log_body=False):
...
-                # an empty dictionary to signal that it was not able to
-                # retrieve credentials. These error will contain both a
-                # Code and Message key.
-                if 'Code' in credentials and 'Message' in credentials:
-                    logger.debug(
-                        'Error response received when retrieving'
-                        'credentials: %s.',
-                        credentials,
-                    )
-                return {}
```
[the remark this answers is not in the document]
ubaskota: I moved it inside `_resolve_get_credentials` function because that’s where the logic to check and validate the imds credentials response is located.

botocore/utils.py
```diff
@@ -379,25 +383,31 @@ def __init__(
         env=None,
         user_agent=None,
         config=None,
+        use_extended_api=None,
```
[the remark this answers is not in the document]
ubaskota: The SEP says we need to set it to none but doesn’t say anything about allowing users to set it up. I added that. We can also set it up to True like you said, but after some clean up, I decided to remove it.

botocore/configprovider.py
```diff
@@ -100,6 +100,12 @@
         False,
         utils.ensure_boolean,
     ),
+    'ec2_metadata_disabled': (
```
jonathan343: We usually keep this consistent with the config file naming. Introducing a third version of the name seems like it may cause confusion. To be clear, I'm suggesting we change `ec2_metadata_disabled` to `disable_ec2_metadata`.

botocore/exceptions.py
```diff
@@ -832,3 +832,15 @@ class UnsupportedServiceProtocolsError(BotoCoreError):
         'Botocore supports {botocore_supported_protocols}, but service {service} only '
         'supports {service_supported_protocols}.'
     )
+
+
+# class UnableToGetProfileNameError(BotoCoreError):
+#     fmt = "Error: Unable to get profile name"
```
jonathan343: Looks like this was missed during cleanup

botocore/utils.py
```diff
@@ -62,11 +62,12 @@
     urlunsplit,
     zip_longest,
 )
-from botocore.exceptions import (
+from botocore.exceptions import (  # UnableToGetProfileNameError,
```
jonathan343: This also looks like something that needs to be cleaned up

botocore/utils.py
```diff
@@ -386,18 +388,21 @@ def __init__(
...
             env.get('AWS_EC2_METADATA_DISABLED', 'false').lower() == 'true'
+            or self._config.get('disable_ec2_metadata', 'false') == 'true'
         )
         self._imds_v1_disabled = config.get('ec2_metadata_v1_disabled')
         self._user_agent = user_agent
         self._session = botocore.httpsession.URLLib3Session(
             timeout=self._timeout,
             proxies=get_environ_proxies(self._base_url),
         )
+        self.resolvedProfile = None
```
jonathan343: nit - In Python, snake case is used for variable names. We should rename this to `resolved_profile`.

> Function names should be lowercase, with words separated by underscores as necessary to improve readability.
>
>Variable names follow the same convention as function names.

Reference: [PEP8 - Function and Variable Names](https://peps.python.org/pep-0008/#function-and-variable-names)

botocore/utils.py
```diff
@@ -645,7 +752,12 @@ def _needs_retry_for_credentials(self, response):
         )
 
     def _contains_all_credential_fields(self, credentials):
-        for field in self._REQUIRED_CREDENTIAL_FIELDS:
+        required_credential_fields = self._REQUIRED_CREDENTIAL_FIELDS
+        if self._use_extended_api and "AccountId" in credentials:
```
jonathan343: Shouldn't this conditional just be `if self._use_extended_api:`? If we are using the extended API, shouldn't we always receive `AccountId`? If so, the current logic allows for a case when this isn't true.

botocore/utils.py
```diff
@@ -611,20 +600,138 @@ def retrieve_iam_role_credentials(self):
...
+        if self._use_extended_api is None or self._use_extended_api:
+            url_path = self._URL_PATH_EXTENDED
+        else:
+            url_path = self._URL_PATH
         return self._get_request(
-            url_path=self._URL_PATH,
+            url_path=url_path,
             retry_func=self._needs_retry_for_role_name,
             token=token,
-        ).text
```
jonathan343, line 619: Why did the `.text` get removed here?

botocore/utils.py
```diff
@@ -563,44 +568,28 @@ def _log_imds_response(self, response, reason_to_log, log_body=False):
...
-                    'expiry_time': credentials['Expiration'],
-                }
-                self._evaluate_expiration(credentials)
+            profileName = self.resolve_if_pre_defined_profile_name()
+            if not profileName:
+                profileName = self._resolve_get_profile_name(token)
+            credentials = self._resolve_get_credentials(profileName, token)
+            if credentials:
+                # register feature-id for IMDS credentials
+                register_feature_id('CREDENTIALS_IMDS')
```
jonathan343, line 580: I don't see any tests related to this feature id. Lmk if I missed something. Otherwise, we should add one that verifies this feature id gets registered when using IMDS for credentials.

botocore/utils.py
```diff
@@ -611,20 +600,138 @@ def retrieve_iam_role_credentials(self):
             logger.debug("Bad IMDS request: %s", e.request)
         return {}
 
+    def resolve_if_pre_defined_profile_name(self):
+        role_name = self.get_user_defined_profile_name()
```
jonathan343: nit - Throughout this PR, you're using `profile_name` and `role_name` synonymously which reads a bit weird, especially if someone doesn't have much context on EC2 instance profiles. We should chose one and stick to it. I generally lean toward what is used currently, which seems to be "role_name".

botocore/utils.py
```diff
@@ -563,44 +568,28 @@ def _log_imds_response(self, response, reason_to_log, log_body=False):
...
-                    'role_name': role_name,
-                    'access_key': credentials['AccessKeyId'],
-                    'secret_key': credentials['SecretAccessKey'],
-                    'token': credentials['Token'],
-                    'expiry_time': credentials['Expiration'],
-                }
-                self._evaluate_expiration(credentials)
+            profileName = self.resolve_if_pre_defined_profile_name()
+            if not profileName:
+                profileName = self._resolve_get_profile_name(token)
```
jonathan343: I still think these three lines can/should be handled in the same function. We would just have one line here like:
```python
profile_name = self._get_profile_name(token)
```

Everything else should be handled in the `_get_profile_name` method.

botocore/utils.py
```diff
@@ -563,44 +568,28 @@ def _log_imds_response(self, response, reason_to_log, log_body=False):
...
+            credentials = self._resolve_get_credentials(profileName, token)
+            if credentials:
+                # register feature-id for IMDS credentials
```
jonathan343: nit - I don't see much value in this comment. It doesn't provide any more information/context than the line below

botocore/utils.py
```diff
@@ -611,20 +600,138 @@ def retrieve_iam_role_credentials(self):
...
+                        self._num_attempts,
+                    )
+                retry_count += 1
+        except BadIMDSRequestError as e:
+            logger.debug("Bad IMDS request: %s", e.request)
+
+    def has_valid_user_defined_profile_name(self, val):
+        if len(val.strip()) == 0:
+            return False
+        return True
```
jonathan343: I think something like below is more straight forward and simple:
```python
def is_valid_profile_name(self, val):
    return bool(val.strip())
```
Any non-empty string should return `True`, so if `False` is returned, we know it's an empty string.

botocore/utils.py
```diff
@@ -611,20 +600,138 @@ def retrieve_iam_role_credentials(self):
...
+            retry_count += 1
+        if role_name is None:
+            raise self._RETRIES_EXCEEDED_ERROR_CLS()
+        return role_name
+
+    def _perform_fallback(self):
+        if self._use_extended_api is None or self._use_extended_api is True:
+            self._use_extended_api = False
+        else:
+            # raise UnableToGetProfileNameError()
```
jonathan343: nit - This comment needs to be cleaned up

botocore/exceptions.py
```diff
@@ -832,3 +832,15 @@ class UnsupportedServiceProtocolsError(BotoCoreError):
...
+
+
+class Ec2ProfileNameMisconfigurationError(BotoCoreError):
+    fmt = "Error: Invalid profile name for ec2_instance_profile_name"
```
jonathan343: From this user POV, this error doesn't seem too helpful. We're telling them their profile name is invalid, but not telling them why. We should have something like:

>  "The EC2 instance profile name provided is invalid. Profile names must not be empty or contain only whitespace characters."

botocore/utils.py
```diff
@@ -386,18 +388,21 @@ def __init__(
             config = {}
         self._base_url = self._select_base_url(base_url, config)
         self._config = config
+        self._use_extended_api = None
 
         if env is None:
             env = os.environ.copy()
         self._disabled = (
             env.get('AWS_EC2_METADATA_DISABLED', 'false').lower() == 'true'
+            or self._config.get('disable_ec2_metadata', 'false') == 'true'
```
jonathan343: I don't think we actually need the `env.get(...)` line anymore right? This is now coming from the config provider which checks both locations for us. We should only need the `_config` line now.

Also, we're using the `utils.ensure_boolean` validation now. Shouldn't we be expecting `bool` instead of strings?

botocore/utils.py
```diff
@@ -611,20 +600,138 @@ def retrieve_iam_role_credentials(self):
...
+        ):
+            raise Ec2ProfileNameMisconfigurationError()
+        elif self.resolvedProfile:
+            role_name = self.resolvedProfile
+        return role_name
+
+    def _resolve_get_profile_name(self, token):
+        retry_count = 0
+        role_name = None
+        while role_name is None and retry_count < 2:
```
jonathan343: 1. I don't really see the value in the `role_name` variable introduced here. I don't think there will ever be a case when the condition `role_name is None` is `False`. 
2. I think a for loop makes more sense here since we know the number of iterations ahead of time. It cleans things up a bit by removing the need for the counter and increment lines.

botocore/utils.py
```diff
@@ -611,20 +600,138 @@ def retrieve_iam_role_credentials(self):
...
+                                    'credentials: %s.',
+                                    credentials,
+                                )
+                            return {}
+                        self._evaluate_expiration(return_credentials)
+                        return return_credentials
+                    else:
+                        self._perform_fallback()
+                except self._RETRIES_EXCEEDED_ERROR_CLS:
+                    self._perform_fallback()
```
jonathan343: Can you explain why we're calling `self._perform_fallback()` in the case of a retries exceeded error?

botocore/exceptions.py
ubaskota: It was intentionally left there incase its needed later. It will be removed before the final push if deemed unnecessary.

botocore/utils.py
ubaskota: same as above.

botocore/utils.py
ubaskota: same as above.

botocore/utils.py
ubaskota: Not always. We may receive a response without `AccountId`. There's a test case where we need to handle a scenario where a request to the new API returns a response without `AccountId`.

botocore/utils.py
ubaskota: I initially removed it to validate the response code before getting the response content.

botocore/utils.py
ubaskota: addressed.

botocore/utils.py
ubaskota: All responses that do not return a 200 status code are automatically retried. Currently, the default number of retry attempts is set to 1. This means that if a 404 error occurs, a 'retries exceeded' error is raised. We want to implement a fallback mechanism to handle this situation by falling back to legacy API when this happens.

botocore/exceptions.py
```diff
@@ -832,3 +832,11 @@ class UnsupportedServiceProtocolsError(BotoCoreError):
...
         'supports {service_supported_protocols}.'
     )
+
+
+class Ec2RoleNameMisconfigurationError(BotoCoreError):
+    fmt = "Error: Invalid role name for ec2_instance_profile_name. Role names must not be empty or contain only whitespace characters."
+
+
+class UnableToGetCredentialsError(BotoCoreError):
+    fmt = "Error: Unable to get credentials for both Legacy and Extended APIs"
```
jonathan343, line 842: We should remove this exception if it isn't being used anywhere

botocore/utils.py
jonathan343: Got it, thanks for the clarification.

In that case, it seems like the `_REQUIRED_CREDENTIAL_FIELDS_EXTENDED` is unnecessary and actually inaccurate. It currently includes `AccountId` which indicates it's required. If you agree, we can probably revert the changes you made in `_contains_all_credential_fields` and remove the `_REQUIRED_CREDENTIAL_FIELDS_EXTENDED` variable.

botocore/utils.py
```diff
@@ -611,21 +592,116 @@ def retrieve_iam_role_credentials(self):
...
+
+    def _perform_fallback(self):
+        if self._use_extended_api is None or self._use_extended_api is True:
+            self._use_extended_api = False
+        else:
+            raise self._RETRIES_EXCEEDED_ERROR_CLS()
+
+    def get_user_defined_role_name(self):
+        role_name = self._config.get('ec2_instance_profile_name')
+        return role_name
```
jonathan343: I don't think this function is necessary. We can just replace the following line:
```py
role_name = self.resolve_if_pre_defined_role_name()
```
with:
```py
role_name = self._config.get('ec2_instance_profile_name')
```

botocore/utils.py
```diff
@@ -611,21 +592,116 @@ def retrieve_iam_role_credentials(self):
...
+                        "Max number of attempts exceeded (%s) when "
+                        "attempting to retrieve data from metadata service.",
+                        self._num_attempts,
+                    )
+        except BadIMDSRequestError as e:
+            logger.debug("Bad IMDS request: %s", e.request)
+
+    @staticmethod
+    def _has_valid_user_defined_role_name(val):
+        return bool(val.strip())
```
jonathan343: Similar to my comment above, I don't think making this a function is necessary. This is only used once and can be inlined:

So instead of:
```py
if role_name and not self._has_valid_user_defined_role_name(role_name):
```

we have:
```py
if not (isinstance(role_name, str) and role_name.strip()):
```

The updated line also validates if the configured role name is a string. I don't think that validation is being done elsewhere

botocore/utils.py
```diff
@@ -611,21 +592,116 @@ def retrieve_iam_role_credentials(self):
...
+            try:
+                role_name = self._get_iam_role(token)
+                if self._use_extended_api is None:
+                    self._use_extended_api = True
+                self.resolved_role_name = role_name
+                break
+            except self._RETRIES_EXCEEDED_ERROR_CLS:
+                self._perform_fallback()
+        if role_name is None:
+            raise self._RETRIES_EXCEEDED_ERROR_CLS()
```
jonathan343: When do we hit this case?

botocore/utils.py
```diff
@@ -611,21 +592,116 @@ def retrieve_iam_role_credentials(self):
...
+        return role_name
+
+    def _perform_fallback(self):
+        if self._use_extended_api is None or self._use_extended_api is True:
+            self._use_extended_api = False
+        else:
+            raise self._RETRIES_EXCEEDED_ERROR_CLS()
```
jonathan343, line 628: I think it makes more sense to `raise` in the function where this gets called instead

botocore/utils.py
```diff
@@ -611,21 +592,116 @@ def retrieve_iam_role_credentials(self):
...
         r = self._get_request(
-            url_path=self._URL_PATH + role_name,
+            url_path=url_path + role_name,
             retry_func=self._needs_retry_for_credentials,
             token=token,
         )
         return json.loads(r.text)
 
+    def _get_iam_role_credentials(self, role_name, token):
+        try:
```
jonathan343: Why do we need this outer `try`? Is there an exception we're not catching properly inside the for loop?

botocore/configprovider.py
```diff
@@ -192,6 +198,12 @@
         None,
         utils.ensure_boolean,
     ),
+    'ec2_instance_profile_name': (
+        'ec2_instance_profile_name',
+        'AWS_EC2_INSTANCE_PROFILE_NAME',
+        None,
+        None,
```
jonathan343, line 205: We should consider adding a custom validation function that validates the profile name isn't whitespace or an empty string.

botocore/exceptions.py
```diff
@@ -832,3 +832,11 @@ class UnsupportedServiceProtocolsError(BotoCoreError):
         'Botocore supports {botocore_supported_protocols}, but service {service} only '
         'supports {service_supported_protocols}.'
     )
+
+
+class Ec2RoleNameMisconfigurationError(BotoCoreError):
+    fmt = "Error: Invalid role name for ec2_instance_profile_name. Role names must not be empty or contain only whitespace characters."
```
jonathan343, line 838: We should reuse an existing config error class instead of creating a new one. We can customize the message as needed.

jonathan343 (commented):
Two small comments from our review meeting, please include them in the refactored version of this PR

PR #3481, opened 2025-05-28
Track retry mode and gzip feature ids

arandito:
#### Description

This PR adds support for tracking the following feature IDs related to retry modes and gzip request compression in the user agent string.

| Feature ID | Metric Value |
| --- |  :---:  |
| RETRY_MODE_LEGACY | D |
| RETRY_MODE_STANDARD | E |
| RETRY_MODE_ADAPTIVE | F |
| GZIP_REQUEST_COMPRESSION | L |

These values indicate which retry mode was configured and if the request should use gzip compression.

#### Example user agent string:
```
Boto3/1.38.23 md/Botocore#1.38.23 ua/2.1 os/macos#24.4.0 md/arch#arm64 lang/python#3.9.22 md/pyimpl#CPython m/F,L,Z,b cfg/retry-mode#adaptive Botocore/1.38.23
```
- `F` indicates the retry mode was set to `adaptive`.
- `L` indicates the request uses gzip compression.


botocore/args.py
```diff
@@ -601,6 +602,11 @@ def _compute_retry_mode(self, config_kwargs):
             retry_mode = 'legacy'
         retries['mode'] = retry_mode
 
+    def _register_retry_mode_feature_id(self, config_kwargs):
+        retries = config_kwargs.get('retries')
+        retry_mode = retries.get('mode')
+        register_feature_id(f'RETRY_MODE_{retry_mode.upper()}')
```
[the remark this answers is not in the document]
arandito: This makes sense. I'll move the register call to `_register_retries()` in client.py to ensure the call is non-blocking. I'll also add an additional `else` statement to return for unsupported retry modes before registering the feature id.

tests/functional/test_retry.py
```diff
@@ -349,3 +367,35 @@ def test_can_exhaust_default_retry_quota(self):
...
+            # Add two failed responses followed by a success
+            http_stubber.add_response(status=502, body=b'{}')
+            http_stubber.add_response(status=502, body=b'{}')
+            http_stubber.add_response(status=200, body=b'{}')
+            client.list_tables()
+        ua_strings = get_captured_ua_strings(http_stubber)
+        # Confirm all requests register `'RETRY_MODE_ADAPTIVE': 'F'`
+        for ua_string in ua_strings:
+            feature_list = parse_registered_feature_ids(ua_string)
+            assert 'F' in feature_list
```
nateprewitt, line 401: nit; this seems like something we could/should be reusing rather than rewriting the same test many times if it can be parameterized. When you find yourself copy pasting scaffolding, unless there's something truly unique about it, try to find opportunities to reuse.

A simple way to do this is to either implement a standalone function, or a method on BaseRetryTest to help for this case. It could look something like:

```python
    def _get_ua_strings_from_retries(self, client):
        with ClientHTTPStubber(client) as http_stubber:
            # Add two failed responses followed by a success
            http_stubber.add_response(status=502, body=b'{}')
            http_stubber.add_response(status=502, body=b'{}')
            http_stubber.add_response(status=200, body=b'{}')
            client.list_tables()
        ua_strings = get_captured_ua_strings(http_stubber)
        return  [parse_registered_feature_ids(ua_string) for ua_string in ua_strings]
```

and your test(s) would look something like:

```python
def test_user_agent_has_standard_mode_feature_id(self):
    client = self.create_client_with_retry_mode(
        'dynamodb', retry_mode='standard'
    )
    feature_lists = self._get_ua_strings_from_retries(client)
    # Confirm all requests register `'RETRY_MODE_ADAPTIVE': 'E'`
    assert all('E' in feature_list for feature_list in feature_lists)
    
def test_user_agent_has_adaptive_mode_feature_id(self):
    client = self.create_client_with_retry_mode(
        'dynamodb', retry_mode='adaptive'
    )
    feature_lists = self._get_ua_strings_from_retries(client)
    # Confirm all requests register `'RETRY_MODE_ADAPTIVE': 'F'`
    assert all('F' in feature_list for feature_list in feature_lists)
```

You can also parametrize these as well with pytest if you're not using the base unittest parent class. That's a separate consideration though. I don't think this is blocking, but it's something to consider for development.

nateprewitt (approved):
A quick testing note to look at but this otherwise looks fine.

tests/functional/test_retry.py
arandito: Got it, I didn't catch that repeated logic but will keep that in mind for future development. I added a reusable helper to simplify testing if we add new retry modes in the future.

PR #3486, opened 2025-06-06
Use usedforsecurity=False for md5

nateprewitt:
Now that we've dropped Python 3.8, we should have this available on all supported versions. We've removed large portions of our MD5 out of the library and the remaining bits are all for non-security related functionality. We can start flagging these as such for FIPS compliant environments and other deployments that restrict use.

### DIFF
--- a/botocore/compat.py
+++ b/botocore/compat.py
@@ -148,7 +148,7 @@ def total_seconds(delta):
 # Checks to see if md5 is available on this system. A given system might not
 # have access to it for various reasons, such as FIPS mode being enabled.
 try:
-    hashlib.md5()
+    hashlib.md5(usedforsecurity=False)
     MD5_AVAILABLE = True
 except (AttributeError, ValueError):
     MD5_AVAILABLE = False
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -328,9 +328,8 @@ def _sse_md5(params, sse_member_prefix='SSECustomer'):
     key_as_bytes = params[sse_key_member]
     if isinstance(key_as_bytes, str):
         key_as_bytes = key_as_bytes.encode('utf-8')
-    key_md5_str = base64.b64encode(get_md5(key_as_bytes).digest()).decode(
-        'utf-8'
-    )
+    md5_val = get_md5(key_as_bytes, usedforsecurity=False).digest()
+    key_md5_str = base64.b64encode(md5_val).decode('utf-8')
     key_b64_encoded = base64.b64encode(key_as_bytes).decode('utf-8')
     params[sse_key_member] = key_b64_encoded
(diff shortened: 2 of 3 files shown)


botocore/utils.py
```diff
@@ -3247,14 +3247,14 @@ def calculate_md5(body, **kwargs):
 
 def _calculate_md5_from_bytes(body_bytes):
```
SamRemis, line 3248: This is the most ambiguous one because this is a generic usage.  

It's easy to imagine someone stumbling upon the `calculate_md5` function which calls these two functions and "un-deprecating" it to use it, potentially for security purposes.

The odds of anyone intentionally using md5 for security related purposes and that making it through code review are very low, so this is safe.  It's common knowledge at this point that MD5 is a broken algorithm and should not be used for security.

botocore/handlers.py
```diff
@@ -328,9 +328,8 @@ def _sse_md5(params, sse_member_prefix='SSECustomer'):
     key_as_bytes = params[sse_key_member]
     if isinstance(key_as_bytes, str):
         key_as_bytes = key_as_bytes.encode('utf-8')
-    key_md5_str = base64.b64encode(get_md5(key_as_bytes).digest()).decode(
-        'utf-8'
-    )
+    md5_val = get_md5(key_as_bytes, usedforsecurity=False).digest()
```
SamRemis, line 331: This md5 parameter [maps to](https://github.com/boto/botocore/blob/8cb37cd9d449ac1a3ea2c70fbb5ff3f94d9d7b45/botocore/data/s3/2006-03-01/service-2.json#L2366-L2371) the `x-amz-server-side​-encryption​-customer-key-MD5 ` header, which according to [the S3 docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html) is used for integrity checks, not security:
> Use this header to provide the base64-encoded 128-bit MD5 digest of the encryption key according to [RFC 1321](http://tools.ietf.org/html/rfc1321). Amazon S3 uses this header for a message integrity check to ensure that the encryption key was transmitted without error.

botocore/compat.py
```diff
@@ -148,7 +148,7 @@ def total_seconds(delta):
 # Checks to see if md5 is available on this system. A given system might not
 # have access to it for various reasons, such as FIPS mode being enabled.
 try:
-    hashlib.md5()
+    hashlib.md5(usedforsecurity=False)
```
SamRemis, line 151: Recording why each of these is used to justify that it's not security related, but I have no concerns with any of them.  First one: 

This one is used for checking if the MD5 algorithm is used in the customer's version of hashlib; this is sometimes removed in FIPS compliant environments, so we use this check just to set the `MD5_AVAILABLE` attribute in compat.py.

PR #3488, opened 2025-06-09
Update Requests to 2.32.4

nateprewitt:
Update Requests to the latest release to address recent security fixes.


nateprewitt:
Resolving in favor of #3489 since we didn't get around to reviews.

PR #3491, opened 2025-06-11
fix: convert utcnow() to now(UTC)

nicolapace:
# Fix deprecated `datetime.datetime.utcnow()` usage
Replace deprecated `datetime.datetime.utcnow()` calls with `datetime.datetime.now(datetime.UTC)` to resolve deprecation warnings and ensure future compatibility.

### What changed:

Updated all instances of `datetime.datetime.utcnow()` to use `datetime.datetime.now(datetime.UTC)`
This change maintains identical functionality while using the recommended timezone-aware approach

### Why:

`datetime.datetime.utcnow()` is deprecated in Python 3.12+ and scheduled for removal
The new approach explicitly uses timezone-aware datetime objects, which is the recommended best practice
Prevents deprecation warnings in logs and ensures forward compatibility

### Testing:

Verified all datetime operations continue to work as expected
No functional changes to application behavior

### DIFF
--- a/botocore/auth.py
+++ b/botocore/auth.py
@@ -422,7 +422,7 @@ class SigV4Auth(BaseSigner):
     def add_auth(self, request):
         if self.credentials is None:
             raise NoCredentialsError()
-        datetime_now = datetime.datetime.utcnow()
+        datetime_now = datetime.datetime.now(datetime.UTC)
         request.context['timestamp'] = datetime_now.strftime(SIGV4_TIMESTAMP)
         # This could be a retry.  Make sure the previous
         # authorization header is removed first.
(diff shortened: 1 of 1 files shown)


botocore/auth.py
```diff
@@ -422,7 +422,7 @@ def signature(self, string_to_sign, request):
     def add_auth(self, request):
         if self.credentials is None:
             raise NoCredentialsError()
-        datetime_now = datetime.datetime.utcnow()
+        datetime_now = datetime.datetime.now(datetime.UTC)
```
kurtmckee, line 425: This change makes the datetime instances timezone-aware, and changes how they're serialized.

In addition, `datetime.UTC` is not compatible with all supported Python versions. It's only available on Python 3.11+.

kurtmckee:
#3239 addresses this issue without introducing Python compatibility issues or serialization changes.

Unfortunately, I've been waiting for almost a year for a review from the devs.

botocore/auth.py
AndrewTsai0401: how about using `timezone.utc` as suggested in 
https://docs.python.org/3.9/library/datetime.html#timezone-objects

nateprewitt:

PR #3494, opened 2025-06-17
Avoid a race condition on directory creation

andyferris:
If a command like `aws s3 cp` is spawned simultaneously (and for the first time), different processes might try to create this directory simultaneously, resulting in an unnecessary exception.

Fixes aws/aws-cli#6489


nateprewitt (approved):
Thanks @andyferris!

PR #3498, opened 2025-06-19
feat: Add support for accountId in imds credentials(2)

ubaskota:
This PR is not ready to be merged because the new “-extended” metadata endpoint is not available yet.

This PR does the following:

* Add support for the new *-extended/credentials IMDS endpoint, with automatic fallback to the legacy endpoint if unavailable.
* Support the AWS_EC2_INSTANCE_PROFILE_NAME environment variable and related settings.
* Retrieve the AWS account ID during IMDS calls, for the new endpoint.
* Adds required unit tests as defined in the SEP.

Tests:
* Validated the changes with unit tests


botocore/config.py
```diff
@@ -419,6 +428,14 @@ def _validate_s3_configuration(self, s3):
                     s3_addressing_style=addressing_style
                 )
 
+    def _validate_ec2_instance_role_name(self, role_name):
+        if role_name is not None and (
+            not bool(role_name.strip()) or not isinstance(role_name, str)
+        ):
```
nateprewitt: We probably want to change the ordering of the checks here and maybe make them more descriptive. We haven't verified the `role_name` is a string before calling `strip()` which can result in unexpected errors like:

```python
>>> object().strip()
AttributeError: 'Foo' object has no attribute 'strip'
```

We should never have to bail out unexpectedly when doing validation. This is a way we can progressively refine the checks without calling unsupported methods.

```suggestion
    if role_name is None:
        return
    elif not isinstance(role_name, str) or not role_name.strip():
```

This skips over the None default and then progressively checks 1.) is it not a string and 2.) is it an empty string. For the case it's not a string, we stop evaluating at the first part of the conditional and won't call `strip()` on the unsupported object.

botocore/config.py
```diff
@@ -419,6 +428,14 @@ def _validate_s3_configuration(self, s3):
...
+            raise InvalidConfigError(
+                error_msg="Error: Invalid role name for ec2_instance_profile_name. Role names must not be empty or contain only whitespace characters."
```
nateprewitt: We're already raising an exception here, we don't need to worry about prefixing it with "Error: ".
```suggestion
                error_msg="Invalid role name for ec2_instance_profile_name. Role names must not be empty or contain only whitespace characters."
```

We should also [run the formatter](https://github.com/boto/botocore/blob/develop/CONTRIBUTING.rst#codestyle) over this code, lines we want to try to keep <80 characters.

botocore/configprovider.py
```diff
@@ -100,6 +100,18 @@
         False,
         utils.ensure_boolean,
     ),
+    'disable_ec2_metadata': (
+        'disable_ec2_metadata',
+        'AWS_EC2_METADATA_DISABLED',
+        False,
+        utils.ensure_boolean,
+    ),
```
nateprewitt: Did we reach a conclusion on this? I think we want to align with to `ec2_metadata_disabled` to match `ec2_metadata_v1_disabled` before we merge.

botocore/utils.py
```diff
@@ -572,28 +571,37 @@ def _log_imds_response(self, response, reason_to_log, log_body=False):
...
                     'token': credentials['Token'],
                     'expiry_time': credentials['Expiration'],
                 }
-                self._evaluate_expiration(credentials)
-                return credentials
+                if self._use_extended_api is True:
+                    if 'AccountId' in credentials:
+                        return_credentials['account_id'] = credentials[
+                            'AccountId'
+                        ]
```
nateprewitt: I think we talked about this one yesterday, is this needed? Regardless of the AccountId being present we should be able to add a single line following expiry_time like:

```python
                    [...]
                    'expiry_time': credentials['Expiration'],
                    'account_id': credentials.get('AccountId'),
                }
```

If it's None, it should be ignored when creating the credentials, otherwise it will be populated. Where you hitting issues while trying to implement that?

botocore/utils.py
```diff
@@ -620,16 +628,48 @@ def retrieve_iam_role_credentials(self):
...
+                if index == 0:
+                    self._use_extended_api = True
+                else:
+                    self._use_extended_api = False
+                self.resolved_role_name = role_name
+                return credentials, role_name
+            except self._RETRIES_EXCEEDED_ERROR_CLS:
+                logger.debug(
+                    "Max number of attempts exceeded (%s) when "
+                    "attempting to get iam role name. Fallback to legacy.",
```
nateprewitt: This isn't necessarily the role name, it could be either the role or credentials in this case now, right?
```suggestion
                    "attempting to get iam role name. Fallback to legacy.",
```

botocore/utils.py
```diff
@@ -572,28 +571,37 @@ def _log_imds_response(self, response, reason_to_log, log_body=False):
...
-    _REQUIRED_CREDENTIAL_FIELDS = [
+    _URL_PATH_EXTENDED = 'latest/meta-data/iam/security-credentials-extended/'
+    _REQUIRED_CREDENTIAL_FIELDS = (
         'AccessKeyId',
         'SecretAccessKey',
         'Token',
         'Expiration',
-    ]
+    )
+    _URL_PATHS = [_URL_PATH_EXTENDED, _URL_PATH]
```
nateprewitt: I think we discussed we want this to be constructed dynamically? With the current implementation, we will always hit both URLs if the extended API isn't available/functioning. The purpose of `_use_extended_api` is to explicitly short circuit those calls in the event we know they're going to fail. We may need to change where this is constructed as it's not necessarily constant for the lifetime of the `InstanceMetadataFetcher`.

botocore/utils.py
```diff
@@ -620,16 +628,48 @@ def retrieve_iam_role_credentials(self):
...
-    def _get_iam_role(self, token=None):
+    def _retrieve_resolve_role_credentials(self, token):
+        for index, url_path in enumerate(self._URL_PATHS):
+            try:
+                role_name = self._get_iam_role(url_path, token)
+                credentials = self._get_credentials(url_path, role_name, token)
+                if index == 0:
+                    self._use_extended_api = True
+                else:
+                    self._use_extended_api = False
```
nateprewitt: This won't always be true if we're not calling the extended API after setting this to False. With the current code, we will toggle this each time. A better option may be to set this if we hit the exception handling code rather than trying to intuit a value based off the _URL_PATHS.

botocore/utils.py
```diff
@@ -620,16 +628,48 @@ def retrieve_iam_role_credentials(self):
...
+        return {}, None
+
+    def _get_iam_role(
+        self,
+        url,
+        token=None,
+    ):
+        role_name = self.resolve_if_pre_defined_role_name()
+        if role_name:
+            return role_name
```
nateprewitt: We might consider simplifying this a little.

```suggestion
        role_name = self.resolved_role_name or self._config.get('ec2_instance_profile_name')
        if role_name:
            return role_name
```

and we can delete `resolve_if_pre_defined_role_name`.

botocore/utils.py
```diff
@@ -620,16 +628,48 @@ def retrieve_iam_role_credentials(self):
...
             retry_func=self._needs_retry_for_role_name,
             token=token,
         ).text
 
-    def _get_credentials(self, role_name, token=None):
+    def resolve_if_pre_defined_role_name(self):
+        role_name = self._config.get('ec2_instance_profile_name')
+        if self.resolved_role_name:
+            role_name = self.resolved_role_name
+        return role_name
```
nateprewitt: If we don't go with the suggestion above to remove this, we should consider rewriting the function. There's no need to check the config at the beginning, we can skip over that until we need it.

```suggestion
        if self.resolved_role_name:
            return self.resolved_role_name
        return self._config.get('ec2_instance_profile_name')
```

tests/functional/test_utils.py
```diff
@@ -57,7 +57,7 @@ def test_catch_retryable_http_errors(self):
             fetcher = InstanceMetadataFetcher()
             send_mock.side_effect = ConnectionClosedError(endpoint_url="foo")
             creds = fetcher.retrieve_iam_role_credentials()
-        self.assertEqual(send_mock.call_count, 2)
+        self.assertEqual(send_mock.call_count, 3)
```
nateprewitt, line 60: Where is the 3rd call coming from here? Because we're failing to hit the new API and falling back?

tests/unit/test_credentials.py
```diff
@@ -1748,6 +1748,29 @@ def test_load_from_instance_metadata(self):
...
+        timeobj = datetime.now(tzlocal())
+        timestamp = (timeobj + timedelta(hours=24)).isoformat()
+        fetcher = mock.Mock()
+        fetcher.retrieve_iam_role_credentials.return_value = {
+            'access_key': 'a',
+            'secret_key': 'b',
+            'token': 'c',
+            'expiry_time': timestamp,
+            'role_name': 'myrole',
+            'account_id': '1234567890',
```
nateprewitt, line 1761: Can we test this when `account_id` is None?

botocore/utils.py
```diff
@@ -620,16 +628,48 @@ def retrieve_iam_role_credentials(self):
...
+            try:
+                role_name = self._get_iam_role(url_path, token)
+                credentials = self._get_credentials(url_path, role_name, token)
+                if index == 0:
+                    self._use_extended_api = True
+                else:
+                    self._use_extended_api = False
+                self.resolved_role_name = role_name
+                return credentials, role_name
+            except self._RETRIES_EXCEEDED_ERROR_CLS:
```
nateprewitt, line 647: Have we verified that the API not existing/not returning what we expect produces an error within this class set?

botocore/utils.py
```diff
@@ -620,16 +628,48 @@ def retrieve_iam_role_credentials(self):
...
+                    "Max number of attempts exceeded (%s) when "
+                    "attempting to get iam role name. Fallback to legacy.",
+                    self._num_attempts,
+                )
+        return {}, None
+
+    def _get_iam_role(
+        self,
+        url,
+        token=None,
```
nateprewitt: Generally changing API signatures like this is considered breaking. Since they're private functions, this is less likely to cause impact, but it's something to be mindful of. Most of the time, you'll need to add parameters to the end of the signature to avoid breaking customers using positional arguments.

tests/unit/test_utils.py
```diff
@@ -2770,6 +2788,8 @@ def add_get_role_name_imds_response(self, role_name=None):
 
     def add_get_credentials_imds_response(self, creds=None):
         if creds is None:
+            creds = self._creds_extended
+        elif creds == 'legacy':
             creds = self._creds
```
nateprewitt: Do we need this? We should probably just supply `self._creds` for `creds` in the case we'd be passing "legacy". Having magic strings for config like this should generally be avoided if we can.

tests/unit/test_utils.py
```diff
@@ -2801,19 +2821,27 @@ def _test_imds_base_url(self, config, expected_url):
...
 
+    @mock.patch.dict(os.environ, {'AWS_EC2_METADATA_DISABLED': 'true'})
     def test_disabled_by_environment(self):
-        env = {'AWS_EC2_METADATA_DISABLED': 'true'}
-        fetcher = InstanceMetadataFetcher(env=env)
+        disable_ec2_metadata = (
+            os.environ.get('AWS_EC2_METADATA_DISABLED').lower() == 'true'
+        )
+        config = {'disable_ec2_metadata': disable_ec2_metadata}
+        fetcher = InstanceMetadataFetcher(config=config)
```
nateprewitt: This test is a little strange and may actually be pointing out a breaking change we've made.

What are we trying to validate by setting the environment variable in this case? We've changed the behavior of `InstanceMetadataFetcher` so that it no longer considers the `env` parameter and only operates off a config. This technically isn't public, but that's definitely breaking behavior. We may want to consider how we handle this.

Regardless of the above though, setting the variable in the env and then putting it in the config is redundant. We should probably just set it in the config and look at how to refactor this test more holistically.

tests/unit/test_utils.py
```diff
@@ -2801,19 +2821,27 @@ def _test_imds_base_url(self, config, expected_url):
...
-    def test_disabled_by_environment_mixed_case(self):
-        env = {'AWS_EC2_METADATA_DISABLED': 'tRuE'}
-        fetcher = InstanceMetadataFetcher(env=env)
+    @mock.patch.dict(os.environ, {'AWS_EC2_METADATA_DISABLED': 'tRuE'})
+    def test_disabled_by_environment_mixed(self):
+        disable_ec2_metadata = (
+            os.environ.get('AWS_EC2_METADATA_DISABLED').lower() == 'true'
+        )
+        config = {'disable_ec2_metadata': disable_ec2_metadata}
+        fetcher = InstanceMetadataFetcher(config=config)
```
nateprewitt: This test also isn't providing a lot of value now. We should consider adding something at a functional level that's actually using the Config/Configstore.

tests/unit/test_utils.py
```diff
@@ -3025,11 +3056,213 @@ def test_exhaust_retries_on_credentials_request(self):
...
+        self.assertEqual(result1, self._expected_creds_extended)
+
+        self.add_get_token_imds_response(token='token')
+        self.add_get_credentials_imds_response()
+        result2 = imdsFetcher.retrieve_iam_role_credentials()
+        self.assertEqual(result2, self._expected_creds_extended)
+
+    def test_returns_valid_credentials_with_account_id_when_unstable_profile(
+        self,
+    ):
```
nateprewitt, line 3109: What does "unstable profile" mean here?

tests/unit/test_utils.py
```diff
@@ -3025,11 +3056,213 @@ def test_exhaust_retries_on_credentials_request(self):
...
+
+    def test_invalid_profile_name_throws_an_error(self):
+        imdsFetcher = InstanceMetadataFetcher(num_attempts=1)
+
+        self.add_get_token_imds_response(token='token')
+        self.add_get_role_name_imds_response('invalid-role-name')
+        self.add_imds_response(b'', status_code=404)
+        self.add_imds_response(b'', status_code=404)
+        result = imdsFetcher.retrieve_iam_role_credentials()
+        self.assertEqual(result, {})
```
nateprewitt: We're not actually throwing an error for this test or assert one was thrown. We should potentially make the intent clearer and validate we're getting an empty result for the reason we expect.

tests/unit/test_utils.py
```diff
@@ -3025,11 +3056,213 @@ def test_exhaust_retries_on_credentials_request(self):
...
+
+    def test_returns_valid_credentials_when_accountId_unavailable(self):
```
nateprewitt:
```suggestion
    def test_returns_valid_credentials_when_account_id_unavailable(self):
```

nateprewitt: How are we verifying the account id isn't available with this test? Ideally we want a more explicit assertion.

tests/unit/test_utils.py
```diff
@@ -3025,11 +3056,213 @@ def test_exhaust_retries_on_credentials_request(self):
...
+        self.add_get_credentials_imds_response('legacy')
+        result1 = imdsFetcher.retrieve_iam_role_credentials()
+        self.assertEqual(result1, self._expected_creds)
+
+        self.add_get_token_imds_response(token='token')
+        self.add_get_credentials_imds_response('legacy')
+        result2 = imdsFetcher.retrieve_iam_role_credentials()
+        self.assertEqual(result2, self._expected_creds)
+
+    def test_returns_valid_credentials_when_role_name_provided_and_accountId_unavailable(
```
nateprewitt:
```suggestion
    def test_credentials_with_role_name_but_account_id_unavailable(
```

Same note here, I don't know what we're actually validating reading through the test code. We should add some specific assertions for the state we're trying to validate beyond that we got back credentials.

tests/unit/test_utils.py
```diff
@@ -3025,11 +3056,213 @@ def test_exhaust_retries_on_credentials_request(self):
...
+        self.assertEqual(result1, self._expected_creds)
+
+        self.add_get_token_imds_response(token='token')
+        self.add_get_credentials_imds_response('legacy')
+        result2 = imdsFetcher.retrieve_iam_role_credentials()
+        self.assertEqual(result2, self._expected_creds)
+
+    def test_returns_valid_credentials_when_unstable_profile_and_accountId_unavailable(
+        self,
+    ):
```
nateprewitt:
```suggestion
    def test_credentials_with_unstable_profile_and_account_id_unavailable(
        self,
    ):
```

tests/unit/test_utils.py
```diff
@@ -3025,11 +3056,213 @@ def test_exhaust_retries_on_credentials_request(self):
...
+            self.assertEqual(result2, {})
+
+            self.add_get_token_imds_response(token='token-1')
+            self.add_get_credentials_imds_response('legacy')
+            result3 = imdsFetcher.retrieve_iam_role_credentials()
+            self.assertEqual(result3, self._expected_creds)
+
+    def test_throws_error_when_account_id_unavailable_and_profile_name_invalid(
+        self,
+    ):
```
nateprewitt: Same note here, we're not actually verifying which error got thrown. We should try to be more concrete with the case we're catching here.

nateprewitt (changes requested):
Good start! I left some comments on things we should look into before we consider releasing. The main pieces are ensuring we're not making unnecessary calls that might impact performance, and ensuring we're properly testing behaviors.

I'd recommend taking a closer look at how we're approaching tests in `test_utils.py` and see if we have options to be more precise with the cases. If that proves to be difficult, we may consider doing more functional testing instead with higher level behaviors. Let me know if you have any questions!

botocore/utils.py
ubaskota: This is needed to ensure that `AccountId` is added to credentials only when available. Doing `'account_id': credentials.get('AccountId')` will add `AccountId` to credentials with a value of `None` when `AccountId` is unavailable. Some of our tests will fail because of this. We'd still need another conditional statement to ignore it when None.

botocore/utils.py
ubaskota: True.

tests/functional/test_utils.py
ubaskota: Yes, this is because of the fallback.

botocore/utils.py
nateprewitt: Can we list the tests that are failing? I would expect [Credentials](https://github.com/boto/botocore/blob/e37a7a63bbac485a01928de5eca134feadbc2d3e/botocore/credentials.py#L307-L331) to operate identically given the default for account_id is None. If we're checking this at a different layer, like the dictionary, we might talk about updating those.

botocore/utils.py
ubaskota: Updated and verified.

botocore/utils.py
ubaskota: Noted. Adding a non-default argument after a default argument raises a syntax error, so I added it like this.

tests/unit/test_utils.py
ubaskota: This is an existing test that had to be modified to match our change. I agree that this and the test below should be removed and added at a functional level that's actually using the config/configstore.

botocore/utils.py
nateprewitt: Yeah, you would normally do something like this in that case:
```suggestion
        self,
        token=None,
        url=None,
        
        [...]
        if url is None:
            url = some_default_value
```

tests/unit/test_utils.py
nateprewitt: What I'm highlighting is we've broken the behavior in this test. What's currently written doesn't test anything except the test itself.

We were previously validating that `env` worked and we were pulling data from the environment. We don't do that anymore though in the current rewrite.

tests/unit/test_utils.py
ubaskota: This is from the test suite. Think of a scenario where a user replaces or removes their IAM instance profile on EC2 at any time even while the instance is running. This change will be reflected immediately and a 404 will be returned when the user tries to get the credentials. 

Also, I am modifying this and a few other similar tests as I've found some potential issues with caching.

tests/unit/test_utils.py
ubaskota: We are using the same response that a legacy API would return. Which is without an account id.

tests/unit/test_utils.py
ubaskota: Account Id unavailable refers to a response similar to the response received with legacy API.

tests/unit/test_credentials.py
ubaskota: Added.

tests/unit/test_utils.py
ubaskota: I updated the test to get the value from config instead of env variable.

tests/unit/test_utils.py
ubaskota: I removed this test as it's no longer needed.

botocore/config.py
```diff
@@ -379,6 +386,8 @@ def __init__(self, *args, **kwargs):
 
         self._validate_retry_configuration(self.retries)
 
+        self._validate_ec2_instance_role_name(self.ec2_instance_profile_name)
```
jonathan343: nit: I brought this up in the previous PR, but it might have gotten lost in translation:

We should keep the naming of variables and functions as close as possible. Here we have a config value named `ec2_instance_profile_name`. I think it makes sense to use the same naming for the related validation function.
```suggestion
        self._validate_ec2_instance_profile_name(self.ec2_instance_profile_name)
```

botocore/config.py
```diff
@@ -419,6 +428,14 @@ def _validate_s3_configuration(self, s3):
                     s3_addressing_style=addressing_style
                 )
 
+    def _validate_ec2_instance_role_name(self, role_name):
```
jonathan343:
```suggestion
    def _validate_ec2_instance_profile_name(self, profile_name):
```

botocore/config.py
```diff
@@ -419,6 +428,14 @@ def _validate_s3_configuration(self, s3):
...
+        if role_name is None:
```
jonathan343:
```suggestion
        if profile_name is None:
```

botocore/config.py
```diff
@@ -419,6 +428,14 @@ def _validate_s3_configuration(self, s3):
...
+            return
+        if not isinstance(role_name, str) or not bool(role_name.strip()):
+            raise InvalidConfigError(
+                error_msg="Invalid role name for ec2_instance_profile_name. Role names must not be empty or contain only whitespace characters."
```
jonathan343: nit: Breaking this up makes it a bit easier to read. ruff doesn't support this yet (see https://github.com/astral-sh/ruff/issues/9634) so we need to do it ourselves.

I also replaced instances of the word "role" with "profile"
```suggestion
                error_msg=(
                    "Invalid profile name for ec2_instance_profile_name. Profile"
                    " names must not be empty or contain only whitespace"
                    " characters."
                )
```

botocore/config.py
```diff
@@ -419,6 +428,14 @@ def _validate_s3_configuration(self, s3):
                     s3_addressing_style=addressing_style
                 )
 
+    def _validate_ec2_instance_role_name(self, role_name):
+        if role_name is None:
+            return
+        if not isinstance(role_name, str) or not bool(role_name.strip()):
```
jonathan343:
```suggestion
        if not isinstance(profile_name, str) or not bool(profile_name.strip()):
```

botocore/utils.py
```diff
@@ -572,28 +570,32 @@ def _log_imds_response(self, response, reason_to_log, log_body=False):
...
         try:
             token = self._fetch_metadata_token()
-            role_name = self._get_iam_role(token)
-            credentials = self._get_credentials(role_name, token)
+            credentials, role_name = self._retrieve_resolve_role_credentials(
+                token,
+            )
             if self._contains_all_credential_fields(credentials):
-                credentials = {
+                return_credentials = {
```
jonathan343: Just curious for my own understanding. Why did this variable name need to change?

botocore/utils.py
```diff
@@ -620,16 +622,53 @@ def retrieve_iam_role_credentials(self):
...
+                self.resolved_role_name = role_name
+                if self._use_extended_api is None:
+                    self._use_extended_api = True
+                return credentials, role_name
+            except self._RETRIES_EXCEEDED_ERROR_CLS:
+                logger.debug(
+                    "Max number of attempts exceeded (%s) when "
+                    "attempting to get iam role name or credentials. Fallback to legacy if legacy API not used already; otherwise, raise an error.",
+                    self._num_attempts,
+                )
```
jonathan343: nit: Breaking down the long string to make this more readable
```suggestion
                logger.debug(
                    "Max number of attempts exceeded (%s) when attempting to"
                    " get iam role name or credentials. Fallback to legacy if"
                    " legacy API not used already; otherwise, raise an error.",
                    self._num_attempts,
                )
```

botocore/utils.py
```diff
@@ -620,16 +622,53 @@ def retrieve_iam_role_credentials(self):
...
+                if self._use_extended_api in (True, None):
+                    self._use_extended_api = False
+                else:
+                    logger.debug(
+                        "Clear the cache and raise an error if the requests fail (non-200 response), addressing potential issues caused by the cached role name or API usage flag."
+                    )
```
jonathan343: nit: Breaking down the long string to make this more readable
```suggestion
                    logger.debug(
                        "Clear the cache and raise an error if the requests"
                        " fail (non-200 response), addressing potential issues"
                        " caused by the cached role name or API usage flag."
                    )
```

botocore/utils.py
```diff
@@ -620,16 +622,53 @@ def retrieve_iam_role_credentials(self):
...
+                    self.resolved_role_name = None
+                    self._use_extended_api = None
```
jonathan343: What's the reason for setting these variables to `None` before raising the error below?

botocore/utils.py
```diff
@@ -620,16 +622,53 @@ def retrieve_iam_role_credentials(self):
...
+            return role_name
         return self._get_request(
-            url_path=self._URL_PATH,
+            url_path=url,
             retry_func=self._needs_retry_for_role_name,
             token=token,
         ).text
 
-    def _get_credentials(self, role_name, token=None):
+    def _get_credentials(self, url_path, role_name, token=None):
```
jonathan343: I think @nateprewitt commented on something similar for the `_get_iam_role` method above. We should add the new parameter to the end of the function signature.

Also, one function uses `url` and the other uses `url_path`. These are small things that we want to try and be as consistent as possible with because they can add up over time and make the code base even more confusing.

botocore/utils.py
```diff
@@ -385,7 +386,6 @@ def __init__(
         timeout=DEFAULT_METADATA_SERVICE_TIMEOUT,
         num_attempts=1,
         base_url=METADATA_BASE_URL,
-        env=None,
```
jonathan343: Removing this is raising a bit of a red flag. We don't use this, but it' possible someone is doing something like:

```python
from botocore.utils import IMDSFetcher
imds_fetcher = IMDSFetcher(env=...)
```

This removal breaks this customer.

I imagine this case is rare though because people most likely would want us to use their actual environment variables.

I'll think on this and let's come back to it tomorrow

botocore/config.py
ubaskota: This was addressed in one of the old commits, where all the names that had `..profile..` in them were changed to `..role..`, but I can see that it makes more sense to use `..profile..`.

botocore/utils.py
ubaskota: This was required when we had a conditional to check for account id. Reverting it back as we don't need it anymore.

PR #3502, opened 2025-07-01
Remove outdated/unnecessary compat shims

nateprewitt:
Removes extraneous comment and an unnecessary xml import now that we've dropped Python 3.8 support.


jonathan343 (approved):
:shipit:

PR #3504, opened 2025-07-01
Add official warnings for the removal of six from vendoring

nateprewitt:
With the release of Botocore 1.11.0 in 2018, the `botocore.vendored` module was deprecated and a majority of the contents were removed. The remaining pieces for Requests have been left along with a copy of `six`. The `six` module was deprecated in 2021 when [Python 2 support was dropped](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-python-2-7-in-aws-sdk-for-python-and-aws-cli-v1/) but left for backwards compatibility.

Now that botocore hasn't supported Python 2.7 for 4 years, we intend to remove this from the library in a future release. In the interim, warnings will be raised when direct imports are made against either the `compat` or `vendor` instances of `six`.

This includes both of the below syntaxes:

```python
from botocore.compat import six
```
```python
import botocore.vendored.six
```

Customers who still need [`six`](https://pypi.org/project/six) should add it to their direct dependencies to receive the latest updates.


tests/functional/test_six_threading.py
```diff
@@ -1,62 +0,0 @@
-"""
-Regression test for six issue #98 (https://github.com/benjaminp/six/issues/98)
```
SamRemis, line 2: Just for my understanding - why are we removing the tests now as opposed to when we actually remove the module?  

Is there a benefit, or more of a "this has to be done sooner or later and since this is deprecated, the tests are of little value"

botocore/compat.py
```diff
@@ -360,3 +360,20 @@ def has_minimum_crt_version(minimum_version):
...
+    if name == "six":
+        warnstr = (
+            "The botocore.compat.six module is deprecated and will be removed "
+            "in a future version. Please use six as a direct dependency."
+        )
+        warnings.warn(warnstr, DeprecationWarning, stacklevel=2)
+        from botocore.vendored import six
+        return six
+
+    raise AttributeError(f"Module {__name__} has no attribute {name}.")
```
SamRemis, line 379: Why isn't this breaking?  Probably something I'm missing, but wouldn't this fail to return existing attributes on the module and always raise the error on any attribute other than `six`?

SamRemis: Also since you removed the import statement above, wouldn't this just be breaking now?

nateprewitt: > Why isn't this breaking? Probably something I'm missing, but wouldn't this fail to return existing attributes on the module and always raise the error on any attribute other than six?

This is the fallback to the standard import mechanism.

> Also since you removed the import statement above, wouldn't this just be breaking now?

This is the import statement now. Imports are going through `__getattr__`. You can try it out with the PR and our existing test suite.

tests/functional/test_six_threading.py
nateprewitt: Six hasn't been supported for several years and this is an issue for a significantly older version of six. It's no longer used in the library itself anymore so this test isn't relevant, that was all removed with Python 2.7 and the follow up work to clear six imports.

botocore/compat.py
nateprewitt: If you're not familiar with module level __getattr__, [PEP 562](https://peps.python.org/pep-0562/#rationale) is a good read. That's what this implementation is based off.

SamRemis (approved):

a-detiste:
Thank you for doing this @nateprewitt 

I was planning to propose a more intrusive patch on Debian side something of the likes of

```
-from botocore.vendored import six
+try:
+     import six
+except ImportError:
+     pass
```

`+` setting a loose `Recommends: python3-six` dependency in the package metadata,
because we are very far in the `six` removal (we patch the biggest culprit `dateutil`);
but now I will first wait to see how this very nice PR resolves.

https://wiki.debian.org/Python3-six-removal

Greetings

PR #3506, opened 2025-07-01
Fix generic error parsing to handle 4xx responses

adev-code:
When 4XX error is returned by AWS Services, the actual 4XX error information is not returned. Instead, depending on where the error exception has happened (for example:`ResponseParserError`) `ResponseParserError` exceptions are returned to users instead of informative error messages.

This change updates `_is_generic_error_response()`is to detect generic HTML error responses for all 4xx status codes, not just 5xx. This allows botocore to handle errors gracefully and provide more informative error messages to customers.

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -300,7 +300,7 @@ class ResponseParser:
         # non sensical parsed data.
         # To prevent this case from happening we first need to check
         # whether or not this response looks like the generic response.
-        if response['status_code'] >= 500:
+        if response['status_code'] >= 400:
             if 'body' not in response or response['body'] is None:
                 return True
 


SamRemis:
Thanks for the PR, it's a good start.  Still a few things we are going to have to do here prior to merging:
1. Add a test to make sure this behavior is working as expected
2. Investigate the test failures in the CI to make sure we aren't adding any breaking changes
3. Add some more context to the PR description to make it clear what types of cases this will happen in.

I'm happy to help with or elaborate on any/all of the above

adev-code:
Closing this one in favor of: https://github.com/boto/botocore/pull/3602

PR #3509, opened 2025-07-07
Add changes to parse all output members from the response body even when the same shape is reused for input.

ubaskota:
This change adds support to parse all the response members including those with 'uri' or 'querystring' locations. This ensures complete and accurate response data when shapes are reused between input and output, as parameters sent in querystring or URI during request should still be parsed from response body.

Changes
```
- Implements a logic to only treat a member as "handled outside the body" if its location is one of  ("statusCode", "header", "headers").
- Ensures that the members with locations "uri" and "querystring" are parsed.
- Removes "IgnoreQueryParamsInResponse" from a list of protocol tests to be ignored.
```

Testing:
```
- Verified that all unit and functional tests pass.
```

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -214,7 +214,7 @@ class ResponseParser:
     # serialization dict. The location key tells us where in the response
     # to parse the value. Members with locations that aren't in this list
     # will be parsed from the body.
-    KNOWN_LOCATIONS = ('header', 'headers', 'status_code')
+    KNOWN_LOCATIONS = ('header', 'headers', 'statusCode')
 
     def __init__(self, timestamp_parser=None, blob_parser=None):
         if timestamp_parser is None:


botocore/parsers.py
```diff
@@ -460,10 +461,14 @@ def _handle_structure(self, shape, node):
             member_shape = members[member_name]
             if (
                 'location' in member_shape.serialization
-                or member_shape.serialization.get('eventheader')
-            ):
+                and member_shape.serialization['location']
+                not in ('uri', 'querystring')
```
jonathan343: I suggest we use our `KNOWN_LOCATIONS` list here instead of using an ignore list.

Instead of using an ignorelist approach with `not in ('uri', 'querystring')`, I recommend using the `KNOWN_LOCATIONS` allowlist. This would be more maintainable since we won't need to update the ignore list for every new request serialization trait that get's added.

```python
            location = member_shape.serialization.get('location')
            if (
                location in self.KNOWN_LOCATIONS
                or member_shape.serialization.get('eventheader')
            ):
```

botocore/parsers.py
```diff
@@ -460,10 +461,14 @@ def _handle_structure(self, shape, node):
...
+            ) or member_shape.serialization.get('eventheader'):
                 # All members with locations have already been handled,
-                # so we don't need to parse these members.
+                # so we don't need to parse these members except for uri
+                # and querystring, which we will need to parse again for
+                # deserialization even if they were already parsed during
+                # request construction
```
jonathan343: This comment isn't fully accurate. I think we're okay with keeping it as is and just adding "known locations"

```python
# All members with known locations have already been handled,
# so we don't need to parse these members.
```

botocore/parsers.py
```diff
@@ -142,6 +142,7 @@
 LOG = logging.getLogger(__name__)
 
 DEFAULT_TIMESTAMP_PARSER = parse_timestamp
+KNOWN_LOCATIONS = ("statusCode", "header", "headers")
```
jonathan343: To keep consistent with the pattern we use in `serialize.py`, I think it makes sense to add `KNOWN_LOCATIONS` as a class constant in the base `ResponseParser` class so it gets inherited in all sub-classes. We can also add a comment above it to provide clarity on what the constant is used for:

```python
    # This is a list of known values for the "location" key in the
    # serialization dict. The location key tells us where in the response
    # to parse the value. Locations not in this list will be ignored.
    KNOWN_LOCATIONS = ('statusCode', 'header', 'headers')
```

botocore/parsers.py
```diff
@@ -1173,7 +1178,7 @@ def _parse_non_payload_attrs(
         for name in member_shapes:
             member_shape = member_shapes[name]
             location = member_shape.serialization.get('location')
-            if location is None:
+            if location is None or location not in KNOWN_LOCATIONS:
```
jonathan343: @ubaskota and I discussed offline. I agree that we don't need this line since the conditionals implicitly handle each know location.

botocore/parsers.py
nateprewitt: We should keep it as a static tuple if we don't need to expand it later.

jonathan343: Yup, good call. I didn't notice when copying over from `serialize.py`

botocore/parsers.py
```diff
@@ -210,6 +210,7 @@ class ResponseParser:
 
     DEFAULT_ENCODING = 'utf-8'
     EVENT_STREAM_PARSER_CLS = None
+    KNOWN_LOCATIONS = ('header', 'headers', 'status_code')
```
jonathan343: I mentioned this in the last roud of reviews and I still think it's a good idea to leave a comment here to provide more context:

```suggestion
    # This is a list of known values for the "location" key in the serialization
    # dict. The location key tells us where in the response to parse the value.
    # Members with locations that aren't in this list will be parsed from the body.
    KNOWN_LOCATIONS = ('header', 'headers', 'status_code')
```

jonathan343 (changes requested):

botocore/parsers.py
ubaskota: Addressed.

botocore/parsers.py
ubaskota: Addressed.

botocore/parsers.py
ubaskota: Addressed.

botocore/parsers.py
ubaskota: There is a comment about it inside the if statement where `KNOWN_LOCATIONS` is used, so I didn't think commenting here was necessary.

```
Locations not in this list will be ignored.
```

Plus, aren't we ignoring the locations in this list?

jonathan343: > Plus, aren't we ignoring the locations in this list?

At a high-level, no, we aren't ignoring members with locations that are know. We will parse them from the location specified (e.g. We will parse from the header or status code). However, when checking for which members to parse from the body, we do ignore them since they were already parsed somewhere else.

I updated the suggested comment above with hopefully more clear wording.

ubaskota: I see. I added the comment.

botocore/parsers.py
```diff
@@ -210,6 +210,11 @@ class ResponseParser:
 
     DEFAULT_ENCODING = 'utf-8'
     EVENT_STREAM_PARSER_CLS = None
+    # This is a list of known values for the 'location' key  in the
+    # serialization dict. The location key tells us where in the response
+    # to parse the value. Members with locations that aren't in this list
+    # will be parsed from the body.
+    KNOWN_LOCATIONS = ('header', 'headers', 'status_code')
```
jonathan343: Not sure how I also messed this up in my original comment, but this should be `statusCode` not `status_code`.

```suggestion
    KNOWN_LOCATIONS = ('header', 'headers', 'statusCode')
```

jonathan343 (changes requested):
Looks good after this typo is fixed

jonathan343:
Posting this for historical context, but I've confirmed all output shapes for every service only define locations that exist in our `KNOWN_LOCATIONS` list. This PR is being merged to prevent a future case where a new output shape has a member with an unknown location. This might happen if the same structure is reused for both input and output.

Here is the script used to confirm all existing output member locations:
```python
import botocore.session

loader = session.get_component("data_loader")
services = loader.list_available_services("service-2")


locations = set()
for service in services:
    service_model = session.get_service_model(service)

    for operation in service_model.operation_names:
        operation_model = service_model.operation_model(operation)
        output_shape = operation_model.output_shape

        if output_shape:
            for member, member_shape in output_shape.members.items():
                location = member_shape.serialization.get("location")
                if location:
                    locations.add(location)
print(locations)
```

Output:
```
{'header', 'headers', 'statusCode'}
```

PR #3513, opened 2025-07-10
Change the default retry mode to standard from legacy and add required tests

ubaskota:
Description:
This PR changes the default retry mode to `standard` from `legacy` and modifies/adds the necessary unit and functional tests.

Tests:
- Verified that `standard` retry mode was used and its feature-id was registered when retry mode wasn't specified.
- Verified that all unit and functional tests pass.


tests/functional/test_config_provider.py
```diff
@@ -118,6 +118,6 @@ def test_defaults_mode_resolved_invalid_mode_exception(self, session):
             config = Config(defaults_mode='invalid_default_mode')
             session.create_client('sts', 'us-west-2', config=config)
 
-    def test_defaults_mode_resolved_legacy(self, session):
+    def test_defaults_mode_resolved_standard(self, session):
```
jonathan343: Why is this test name changing? That doesn't seem right. We aren't testing the `standard` defaults mode

tests/functional/test_retry.py
```diff
@@ -356,9 +366,7 @@ def test_adaptive_mode_retry_transient_error(self):
     def test_can_exhaust_default_retry_quota(self):
         # Quota of 500 / 5 retry costs == 100 retry attempts
         # 100 retry attempts / 2 retries per API call == 50 client calls
-        client = self.create_client_with_retry_mode(
-            'dynamodb', retry_mode='standard'
-        )
```
jonathan343, line 361: I think it might make sense for all tests to directly configure the `retry_mode`, even it it's the default. This prevents us from having to make these type of u[dates in the future. We just need to make sure there is a test that asserts what the default mode is.

tests/functional/test_retry.py
```diff
@@ -387,3 +393,20 @@ def test_user_agent_has_adaptive_mode_feature_id(self):
         feature_lists = self._get_feature_id_lists_from_retries(client)
         # Confirm all requests register `'RETRY_MODE_ADAPTIVE': 'F'`
         assert all('F' in feature_list for feature_list in feature_lists)
+
+    def test_service_specific_defaults_do_not_mutate_general_defaults(self):
```
jonathan343: This test isn't actually checking anything if both assertions are `2`. If we weren't falling back to the general defaults, we wouldn't know since both are asserting the same thing.

I also don't think this test is applicable to the standard retry mode. We only load in the `_retry.json` when using the `legacy` retry mode ([here](https://github.com/boto/botocore/blob/f9d81c2cdc3c236f08b6237f18a25acf78a35833/botocore/client.py#L271-L275)). This file defines service specific defaults which the `test_service_specific_defaults_do_not_mutate_general_defaults` checks aren't reused between clients for different services.

jonathan343 (commented):
Thanks Ujjwal!

The code changes look correct to me. I left a few comments on some of the tests.

The other thing we'll need to make sure get's released alongside this PR is an update to our [Boto3 Retries](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html) docs page. We'll need to update saying standard is the new default mode. It might also be worth updating the [Botocore Config reference](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html).

tests/functional/test_retry.py
ubaskota: Noted. I'll remove this test.

tests/functional/test_config_provider.py
ubaskota: It previously used 'legacy' as its default setting, but now 'standard' is the new default. The test `test_defaults_mode_resolved_standard,` uses `assert_client_uses_defaults `which verifies that clients are using 'standard' for retry when no specific mode is specified. Hence the name change.

tests/functional/test_retry.py
ubaskota: While I don't see any harm in doing that, I do think that it's semantically meaningful to not configure `retry_mode` when testing for default.

botocore/config.py
```diff
@@ -134,6 +134,8 @@ class Config:
 
     :type retries: dict
     :param retries: A dictionary for configuration related to retry behavior.
+        You can read more about retries in `boto3's retries page
+        <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html>`__.
```
nateprewitt: Putting direct links in code has generally been a bad idea as our domain/path structure may change over time. Also linking from Botocore to Boto3 isn't ideal. If we really want to include a link, we should look at adding it in the docs generator for the config page.

.changes/next-release/feature-retries-29845.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "retries",
+  "description": "Update the default retry strategy to standard"
```
jonathan343:
```suggestion
  "description": "Update the default retry mode from ``legacy`` to ``standard``."
```

botocore/config.py
```diff
@@ -149,16 +151,21 @@ class Config:
           example, setting this value to 2 will result in the request
           being retried at most two times after the initial request. Setting
           this value to 0 will result in no retries ever being attempted after
-          the initial request. If not provided, the number of retries will
-          default to the value specified in the service model, which is
-          typically four retries.
+          the initial request. If not provided, it will default to using the
+          number of retries in ```standard``` retry mode.
```
jonathan343: This isn't true if I set my mode to something else (`legacy` or `adaptive`). We should make this more general. It will default to the number of retries for their configured retry mode.

botocore/config.py
```diff
@@ -149,16 +151,21 @@ class Config:
...
+
         * ``mode`` -- A string representing the type of retry mode botocore
-          should use.  Valid values are:
+          should use. Valid values are:
 
-          * ``legacy`` - The pre-existing retry behavior.
+          * ``legacy`` - Legacy mode uses an older (v1) retry handler that
+            has limited functionality. The number of retries in this retry
+            mode for some services is specified in the service model. This
+            will default to 4 retries unless specified in the service model.
```
jonathan343: This is kinda confusing since here you give a number that represents retry count and below we provide a number that represents the max attempts. I think both should use the max attempts amount for consistency. I can see consumers getting confused by this. Should be 5 for legacy and 3 for standard.

botocore/configprovider.py
```diff
@@ -144,13 +144,13 @@
         'auto',
         None,
     ),
+    'retry_mode': ('retry_mode', 'AWS_RETRY_MODE', 'standard', None),
```
jonathan343: nit, is there a reason why we moved this above the `sts_regional_endpoints` entry away from it's original position? Or is this just because  a result of how you resolved the merge conflict?

tests/functional/test_config_provider.py
jonathan343: I don't think that's right. `defaults_mode` is a different concept and we aren't changing that default to `standard`. We're changing`retry_mode` to standard which is why this name change seems wrong to me.

The client in this test gets created with `defaults_mode` as `legacy` since that is the default

ubaskota: Noted. I'll fix it.

botocore/configprovider.py
ubaskota: This is the result of resolving the merge conflict.

botocore/config.py
ubaskota: I changed it to "4 retries (5 max attempts) for legacy and 2 retries (3 max attempts)" for standard.

botocore/config.py
ubaskota: When I updated this statement, I tried to match it with whatever we currently have in our docs site. After a few back and forth on this, I've changed it to something more generic.

.changes/next-release/feature-retries-29845.json
ubaskota: updated

botocore/config.py
ubaskota: I decided to remove both of these lines.

botocore/configprovider.py
jonathan343: nit - Can you move this back to its original spot below? It makes the history look a bit weird if you're moving an entire line for no reason

botocore/retries/standard.py
```diff
@@ -37,6 +37,10 @@
 from botocore.retries.base import BaseRetryableChecker, BaseRetryBackoff
 
 DEFAULT_MAX_ATTEMPTS = 3
+# This is a special case where we maintain the legacy retry behavior of
+# some services by setting max_attempts to 10 instead of the standard
+# retry mode's default of 3 attempts.
```
nateprewitt: We can probably be more generic with this comment. This is a general purpose override per service.
```suggestion
# Service-specific overrides for the default max_attempts.
```

.changes/next-release/feature-retries-29845.json
```diff
@@ -1,5 +1,5 @@
 {
   "type": "feature",
   "category": "retries",
-  "description": "Update the default retry mode from ``legacy`` to ``standard``."
+  "description": "Update the default retry mode from ``legacy`` to ``standard``. See [Boto3 Retries](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html) for more information about retry behavior and how to configure different retry options."
```
nateprewitt: Our changelog is RST instead of Markdown so we'll need to use [different syntax](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links) for the link.

botocore/configprovider.py
ubaskota: moved

.changes/next-release/feature-retries-29845.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "retries",
+  "description": "Update the default retry mode from ``legacy`` to ``standard``. See `Boto3 Retries <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html>`__ for more information about retry behavior and how to configure different retry options."
```
nateprewitt: I just realized this is in botocore and not boto3. The botocore commits get propagated to both Boto3 _and_ the AWSCLI. We should avoid pushing up Boto3 specific changelogs from this package.

If we're really concerned about pointing people to the guide, we can open a tracking issue and point to that for Botocore with all the information.

jonathan343:
Update from the [Updating AWS SDK defaults](https://aws.amazon.com/blogs/developer/updating-aws-sdk-defaults-aws-sts-service-endpoint-and-retry-strategy/) blog post:
> Update: After careful analysis, we found potential concerns with updating the default retry strategy at this time. We will re-evaluate the best course of action and schedule this change at a later time.

Closing this PR for now. We can re-open at a later date if needed.

PR #3514, opened 2025-07-10
Experimental plugin autoloading

SamRemis:
The contents of this PR are for internal use and subject to abrupt, breaking changes.  

This PR adds support for a new environment variable to autoload plugins on behalf of a botocore user.


botocore/plugin.py
```diff
@@ -0,0 +1,30 @@
+"""
+NOTE: This interface is considered private and is subject to abrupt breaking
```
nateprewitt:
```suggestion
NOTE: This module is considered private and is subject to abrupt breaking
```

botocore/session.py
```diff
@@ -1164,6 +1167,24 @@ def _get_ignored_credentials(self, aws_session_token, aws_account_id):
             credential_inputs.append('aws_account_id')
         return ', '.join(credential_inputs) if credential_inputs else None
 
+    def _register_client_plugins(self, client, config):
+        plugins_list = get_botocore_experimental_plugins()
```
nateprewitt: We should avoid naming the functions experimental if we can. We can add that in the docstring.
```suggestion
        plugins_list = get_botocore_plugins()
```

botocore/session.py
```diff
@@ -1164,6 +1167,24 @@ def _get_ignored_credentials(self, aws_session_token, aws_account_id):
...
+
+        client_plugins = {}
+        for plugin in plugins_list.split(','):
+            try:
+                name, module = [part.strip() for part in plugin.split('=')]
+                client_plugins[name] = module
+            except ValueError:
+                logger.warning(
+                    f"Invalid plugin format: {plugin}. Expected 'name=module'"
+                )
```
nateprewitt, line 1182: To make sure I understand the intent here, this is going to be an environment variable that's a comma delimited key-value pair string?

botocore/utils.py
```diff
@@ -364,6 +366,39 @@ def is_global_accesspoint(context):
     return is_global
 
 
+# All code related to experimental plugins is considered a private interface and
```
nateprewitt: Are we able to group this with the other plugin code? Do we think this is more generally applicable outside of the plugin context?

botocore/utils.py
```diff
@@ -364,6 +366,39 @@ def is_global_accesspoint(context):
...
+# subject to abrupt breaking changes or removal without a minor version bump
+BOTOCORE_EXPERIMENTAL__PLUGINS_CONTEXT = contextvars.ContextVar(
+    'BOTOCORE_EXPERIMENTAL__PLUGINS', default=None
```
nateprewitt: Do we want to have this use the value from the environment so we're checking a singular place?

tests/functional/test_plugins.py
```diff
@@ -0,0 +1,173 @@
...
+        client.list_buckets()
+
+
+@pytest.fixture
+def test_plugin():
+    """Fixture to create and register test plugin module in sys.modules."""
+    module = TestPluginModule()
+    sys.modules['test_plugin'] = module
+    yield module
+    sys.modules.pop('test_plugin', None)
```
nateprewitt: Do we have another way to register this without having to patch sys.modules? It seems like we could accidentally break other libraries.

tests/functional/test_plugins.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    Restore sys.modules after each test for isolation.
+
+    This fixture is autouse for this file, so any test that modifies sys.modules
+    will not leak changes to other tests.
+    """
+    old_modules = sys.modules.copy()
+    yield
+    sys.modules.clear()
+    sys.modules.update(old_modules)
```
nateprewitt: Is this safe? This seems like it's going to potentially break our tests, especially if they're parallelized.

tests/functional/test_plugins.py
```diff
@@ -0,0 +1,173 @@
...
+    def test_environment_variable(self, test_plugin):
+        """Tests that plugin is loaded and event registered via env variable."""
+        with (
+            mock.patch(
+                'os.environ',
+                {
+                    'AWS_ACCESS_KEY_ID': 'access_key',
+                    "AWS_SECRET_ACCESS_KEY": "secret_key",
+                    "BOTOCORE_EXPERIMENTAL__PLUGINS": "plugin_name=test_plugin",
+                },
```
nateprewitt: Do we want to just define this once for the test class and then modify a copy of it for each test? That will save us having to redefine it every test.

tests/functional/test_plugins.py
```diff
@@ -0,0 +1,173 @@
...
+            ),
+            mock.patch(
+                'botocore.httpsession.URLLib3Session.send'
+            ) as mock_send,
+        ):
+            session = get_session()
+            client = session.create_client('dynamodb', region_name='us-east-1')
+            mock_send.return_value = mock.Mock(
+                status_code=200, headers={}, content=b''
+            )
```
nateprewitt: Should we be using the ClientHTTPStubber here? Generally we want to avoid patching URLLib3Session directly in functional tests.

nateprewitt (commented):
Left some high level comments. It looks like the tests may be having issues if we can take a quick look at that.

botocore/session.py
SamRemis: Yes; if we want to update it, we don't really need a name and can just do the module. 

The reason for the key/value pair was only for readability of log messages.

tests/functional/test_plugins.py
SamRemis: If we are testing loading arbitrary modules, it needs to be exposed to the interpreter in one form or another.  I've modified this to be appending the sys path instead now, but am not sure if there's a better way.

botocore/utils.py
SamRemis: This is removed now, so resolving.  The reason that would have been difficult is that it would have been resolved at module import time, before we had the chance to mock out any environment variables.

tests/unit/test_plugin.py
```diff
@@ -0,0 +1,46 @@
...
+
+from botocore.plugin import (
+    PluginContext,
+    get_botocore_plugins,
+    reset_plugin_context,
+    set_plugin_context,
+)
+
+
+def test_get_botocore_plugins_env():
```
nateprewitt: This is covered in our functional testing currently, are we worried about something specifically for the unit tests?

tests/unit/test_plugin.py
```diff
@@ -0,0 +1,46 @@
...
+def test_get_botocore_plugins_ctx_takes_precedence():
+    ctx = PluginContext(plugins="a=b")
+    token = set_plugin_context(ctx)
+    try:
+        with mock.patch.dict(
+            os.environ, {'BOTOCORE_EXPERIMENTAL__PLUGINS': 'DISABLED'}
+        ):
+            assert get_botocore_plugins() == 'a=b'
+    finally:
+        reset_plugin_context(token)
```
nateprewitt: This test strikes me as a little odd. When/how is the context ever populated with something other than the environment variable? Are we testing that if we wrote a test mocking the environment that the variable is ignored after initialization?

botocore/plugin.py
```diff
@@ -0,0 +1,71 @@
...
+    _plugin_context.reset(token)
+
+
+def get_botocore_plugins():
+    context = get_plugin_context()
+    if context is not None:
+        plugins = context.plugins
+        if plugins is not None:
+            return plugins
+    return os.environ.get('BOTOCORE_EXPERIMENTAL__PLUGINS')
```
nateprewitt: Do we not want to initialize the plugins to the environment variable if the context is `None`? Having the definitions live in two separate places without a clear indicator if we're actually using the context seems strange.

botocore/session.py
```diff
@@ -1164,6 +1166,24 @@ def _get_ignored_credentials(self, aws_session_token, aws_account_id):
...
```
nateprewitt, line 1182: Do we want to do this every time we create a client or is it reasonable for the context to have this for us already? Are we expecting these to change over the lifecycle of the interpreter?

tests/functional/test_plugins.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def test_multiple_plugins_and_malformed():
+    plugin1_mod = importlib.import_module("first_of_two_plugins")
+    plugin2_mod = importlib.import_module("second_of_two_plugins")
+    first_plugin = plugin1_mod.plugin_instance
+    second_plugin = plugin2_mod.plugin_instance
+
+    client_test_with_plugins(
+        plugins="plugin1=first_of_two_plugins,plugin2=second_of_two_plugins,malformedplugin"
```
nateprewitt: Not for this PR, but an oddity with this is it seems like the key part of the pair is arbitrary and we've confined it to be a single class per module.

Was there a reason we couldn't have the key be the actual class and the value be the module so we're able to handle multiple plugins. That also avoids having the loose function requirement in each module. If we have this as a class method it makes it easier to formalize a plugin spec with something like a [Protocol](https://typing.python.org/en/latest/spec/protocol.html). Something to think about if we're ever going to expand this.

tests/functional/test_plugins.py
```diff
@@ -0,0 +1,87 @@
...
+    plugins_dir = os.path.join(os.path.dirname(__file__), "plugins")
+    sys.path.append(plugins_dir)
+    try:
+        yield
+    finally:
+        sys.path = _preserved_sys_path
+
+
+def client_test_with_plugins(plugins):
+    ctx = PluginContext(plugins=plugins)
```
nateprewitt: Is there a reason we're not doing the full lifecycle for the functional tests? Should we be using the environment variables instead of setting the context directly, or is this an intended use case for the design?

tests/unit/test_plugin.py
SamRemis: My intention here was to test that we are properly disabling the plugins.  It's written a bit strangely I guess, it should flip which one has DISABLED and which one has a=b.

botocore/session.py
SamRemis: It's a reasonable optimization.  It's one that could be a follow up if this ever becomes a bottleneck though considering that this is likely to be an extremely fast operation.

Do you feel strongly about this?

nateprewitt: The question is more stemming from this introducing the ability to hot swap plugins at runtime since it's reevaluated with each client. From offline discussions, it sounds like that is the long term intent for the design. We should probably make sure that's well documented/understood going forward because it seems like an area that may be surprising.

tests/functional/test_plugins.py
```diff
@@ -0,0 +1,98 @@
...
+        try:
+            session = get_session()
+            session.set_credentials('key', 'secret')
+            client = session.create_client('dynamodb', region_name='us-east-1')
+            with ClientHTTPStubber(client) as http_stubber:
+                http_stubber.add_response(status=200, body=b'')
+                client.list_tables()
+        finally:
+            reset_plugin_context(token)
+    assert disabled_plugin.invocations == 0
```
nateprewitt, line 92: I don't know if I understand this test. The `disabled_plugin` doesn't have any incrementing capability so it will always have `invocations == 0` regardless of if it gets called right? Isn't setting the environment variable to "DISABLED" and calling any of the other plugins, asserting they're zero closer to the correct end result?

What specific behavior are we trying to assert here?

tests/functional/test_plugins.py
```diff
@@ -0,0 +1,98 @@
+import importlib
```
nateprewitt, line 13:
```suggestion
# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
#
# Licensed under the Apache License, Version 2.0 (the "License"). You
# may not use this file except in compliance with the License. A copy of
# the License is located at
#
# http://aws.amazon.com/apache2.0/
#
# or in the "license" file accompanying this file. This file is
# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
# ANY KIND, either express or implied. See the License for the specific
# language governing permissions and limitations under the License.
import importlib
```

tests/functional/plugins/simple_plugin_test_module.py
```diff
@@ -0,0 +1,19 @@
+class TestPluginModule:
```
nateprewitt, line 13:
```suggestion
# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
#
# Licensed under the Apache License, Version 2.0 (the "License"). You
# may not use this file except in compliance with the License. A copy of
# the License is located at
#
# http://aws.amazon.com/apache2.0/
#
# or in the "license" file accompanying this file. This file is
# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
# ANY KIND, either express or implied. See the License for the specific
# language governing permissions and limitations under the License.
class TestPluginModule:
```

tests/functional/plugins/second_of_two_plugins.py
```diff
@@ -0,0 +1,17 @@
+class SecondPluginModule:
```
nateprewitt, line 13:
```suggestion
# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
#
# Licensed under the Apache License, Version 2.0 (the "License"). You
# may not use this file except in compliance with the License. A copy of
# the License is located at
#
# http://aws.amazon.com/apache2.0/
#
# or in the "license" file accompanying this file. This file is
# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
# ANY KIND, either express or implied. See the License for the specific
# language governing permissions and limitations under the License.
class SecondPluginModule:
```

tests/functional/plugins/recursive_plugin_test_module.py
```diff
@@ -0,0 +1,34 @@
+from botocore import utils
```
nateprewitt, line 13:
```suggestion
# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
#
# Licensed under the Apache License, Version 2.0 (the "License"). You
# may not use this file except in compliance with the License. A copy of
# the License is located at
#
# http://aws.amazon.com/apache2.0/
#
# or in the "license" file accompanying this file. This file is
# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
# ANY KIND, either express or implied. See the License for the specific
# language governing permissions and limitations under the License.
from botocore import utils
```

tests/functional/plugins/first_of_two_plugins.py
```diff
@@ -0,0 +1,17 @@
+class FirstPluginModule:
```
nateprewitt, line 13:
```suggestion
# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
#
# Licensed under the Apache License, Version 2.0 (the "License"). You
# may not use this file except in compliance with the License. A copy of
# the License is located at
#
# http://aws.amazon.com/apache2.0/
#
# or in the "license" file accompanying this file. This file is
# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
# ANY KIND, either express or implied. See the License for the specific
# language governing permissions and limitations under the License.
class FirstPluginModule:
```

tests/functional/plugins/disabled_test_plugin.py
```diff
@@ -0,0 +1,17 @@
+class DisabledPluginModule:
```
nateprewitt, line 13:
```suggestion
# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
#
# Licensed under the Apache License, Version 2.0 (the "License"). You
# may not use this file except in compliance with the License. A copy of
# the License is located at
#
# http://aws.amazon.com/apache2.0/
#
# or in the "license" file accompanying this file. This file is
# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
# ANY KIND, either express or implied. See the License for the specific
# language governing permissions and limitations under the License.
class DisabledPluginModule:
```

botocore/plugin.py
```diff
@@ -0,0 +1,73 @@
+"""
```
nateprewitt, line 13:
```suggestion
# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
#
# Licensed under the Apache License, Version 2.0 (the "License"). You
# may not use this file except in compliance with the License. A copy of
# the License is located at
#
# http://aws.amazon.com/apache2.0/
#
# or in the "license" file accompanying this file. This file is
# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
# ANY KIND, either express or implied. See the License for the specific
# language governing permissions and limitations under the License.
"""
```

tests/functional/plugins/disabled_test_plugin.py
```diff
@@ -0,0 +1,17 @@
...
+        self.invocations = 0
+
+    def register_event(self, client):
+        client.meta.events.register('before-call', self.increment_invocations)
+
+    def increment_invocations(self, **kwargs):
+        self.invocations += 1
+
+
+plugin_instance = DisabledPluginModule()
```
nateprewitt, line 24: Ideally we don't want to have this plugin at all, but due to limitations in the current design, we need an individual plugin per test (or a plugin per test in each plugin module). Moving to a setup where we can get a copy of the plugin per invocation will simplify testing/state issues.

nateprewitt (changes requested):

PR #3515, opened 2025-07-11
Query compatibility bugfix

SamRemis:
There's a fundamental problem with our query compatibility mode, rooted in how SQS implemented protocol migration and how error handling has evolved in AWS SDKs.

### Background

The Query protocol is unique because it can model error code information within errors themselves. Other protocols (JSON, CBOR) cannot include "error" fields in exception shapes.

Error codes help map exceptions to specific error classes (e.g., cloudwatch.exceptions.ResourceNotFound).  Each modeled error has a "code" associated with it which will map to the specific class.  This code comes straight from the "code" key from the "error" fields in the json model.  If the error does not have a code, the shape name is used as a fallback.

### Historical Context
 -  SQS migrated protocols before services were able to model support for multiple protocols
 -  During this migration, SQS models lost critical information from their "error" sections
 -  This information loss was never properly addressed, leading to some persistent breaking changes

### Current State

Multi-protocol support allows services to maintain their "error" section information when they support the `query` protocol as well as other protocols.  However, our Query protocol logic is broken because of the following:

When receiving a query-compatible error header, we store the original error code in a key called `QueryErrorCode` and the header-provided query error code in a key called `Code`.

During error mapping, we [prioritize QueryErrorCode (original code) over the code from the header](https://github.com/boto/botocore/blob/e566e05467f221f81bccd2fbd1b0fa20b92c6be2/botocore/client.py#L1019-L1021).  This means we will effectively never use the error code that gets sent back from a service.

### Impact
Currently, this only affects SQS as it's the only query-compatible service, but future query-compatible services would face incorrect error mapping.

### Proposed Solution
Change error code lookup from:
```
error_code = error_info.get("QueryErrorCode") or error_info.get("Code")
```
to:
```
error_code = error_info.get("Code")
```

This will prioritize the proper code for future services.  To preserve backwards compatibility for SQS specifically, we will create an SQS specific customization that prioritizes the original code over the one sent back in the header.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -237,7 +237,7 @@ def set_operation_specific_signer(context, signing_name, **kwargs):
         return signature_version
 
 
-def _handle_sqs_compatible_error(parsed, http_response, **kwargs):
+def _handle_sqs_compatible_error(parsed, context, **kwargs):
     """
     Ensures backward compatibility for SQS errors.
 
@@ -252,7 +252,7 @@ def _handle_sqs_compatible_error(parsed, http_response, **kwargs):
         return
 
     if query_code := parsed_error.get("QueryErrorCode"):
-        parsed["Error"]["ErrorCodeOverride"] = query_code
+        context['error_code_override'] = query_code
 
 
 def _resolve_sigv4a_region(context):
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -241,10 +241,10 @@ def _handle_sqs_compatible_error(parsed, context, **kwargs):
     """
     Ensures backward compatibility for SQS errors.
 
-    SQS' migration from the query protocol to JSON was done prior SDKs allowing a
-    service to support multiple models.  Because of this SQS is missing the "error"
+    SQS's migration from the Query protocol to JSON was done prior to SDKs allowing a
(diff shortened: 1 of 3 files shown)


botocore/handlers.py
```diff
@@ -237,6 +237,24 @@ def set_operation_specific_signer(context, signing_name, **kwargs):
...
 
+def _handle_sqs_compatible_error(parsed, context, **kwargs):
+    """
+    Ensures backward compatibility for SQS errors.
+
+    SQS' migration from the query protocol to JSON was done prior SDKs allowing a
+    service to support multiple models.  Because of this SQS is missing the "error"
+    key from its modeled exceptions, which is used by most query compatible services
+    to map exceptions to the proper error.  SQS instead uses the error's shape name,
+    which is preserved in the QueryErrorCode key.
```
jonathan343:
```suggestion
    SQS's migration from the Query protocol to JSON was done prior to SDKs allowing a
    service to support multiple protocols.  Because of this, SQS is missing the "error"
    key from its modeled exceptions, which is used by most query compatible services
    to map error codes to the proper exception.  Instead, SQS uses the error's shape name,
    which is preserved in the QueryErrorCode key.
```

jonathan343 (approved):

PR #3523, opened 2025-07-16
Correct DynamoDB's Count/ScannedCount when paginating over numeric result keys

ashovlin:
### Before

When paginating over DynamoDB Scan or Query and using a `--starting-token`, the AWS CLI incorrectly reports `Count` and `ScannedCount` as `0`. In the debug logs for the second call below, DynamoDB is still returning `5` for each.

This looks like a decision made in https://github.com/boto/botocore/pull/700/ and https://github.com/boto/botocore/pull/702  

```
~ aws dynamodb scan --table-name Superhero --max-items 1
{
    "Items": [
        {
            "name": {
                "S": "Hulk"
            }
        }
    ],
    "Count": 5,
    "ScannedCount": 5,
    "ConsumedCapacity": null,
    "NextToken": "eyJFeGNsdXNpdmVTdGFydEtleSI6IG51bGwsICJib3RvX3RydW5jYXRlX2Ftb3VudCI6IDF9"
}

~ aws dynamodb scan --table-name Superhero --max-items 1 --starting-token eyJFeGNsdXNpdmVTdGFydEtleSI6IG51bGwsICJib3RvX3RydW5jYXRlX2Ftb3VudCI6IDF9
{
    "Items": [
        {
            "name": {
                "S": "Batman"
            }
        }
    ],
    "Count": 0,
    "ScannedCount": 0,
    "ConsumedCapacity": null,
    "NextToken": "eyJFeGNsdXNpdmVTdGFydEtleSI6IG51bGwsICJib3RvX3RydW5jYXRlX2Ftb3VudCI6IDJ9"
}
```

### After
Now, in the logic where we initialize the pagination results while processing the first page, we'll use the first page's values for numeric result keys. I think this makes sense for DynamoDB, since the service is still processing those items, even if the client is truncating and resuming from an item within the first page.

Question for reviewers:
1. I confirmed manually that DynamoDB is the only case where we support numeric result keys, but did not add a test to guard that. Let me know if you think it's worth it.
2. I did not add a changelog entry, as this is intended for CLI V1. I don't believe users typically access the secondary result keys via the usual `for page in paginator.paginate()` pattern.


ashovlin:
> > I confirmed manually that DynamoDB is the only case where we support numeric result keys, but did not add a test to guard that. Let me know if you think it's worth it.
> 
> Non-blocking, but I think it's worth it

Added via https://github.com/boto/botocore/pull/3523/commits/dcfd79dba84b55d73242d7fb0d195f01824b4355

Issue #3528, opened 2025-07-22
Performance regression in S3 URL signing


akx:
### Describe the bug

While profiling our app, I noticed quite a bit of time being spent signing URLs with Boto3/Botocore, so I decided to dig a little deeper, and turns out that there's been pretty significant performance regressions (that aren't helped by awscrt) – namely, current versions of botocore are around 3x slower when calling [`S3.Client.get_presigned_url`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/generate_presigned_url.html). 

The below results were generated with [this benchmark code](https://github.com/akx/botocore-sign-perf) on my macOS machine.

| boto3 | botocore | awscrt | count | mean (ms) | median (ms) | std (ms) | min (ms) | max (ms) | Visual Chart |
|-------|----------|--------|-------|-----------|-------------|----------|----------|----------|--------------|
| 1.20.54 | 1.23.54 | - | 10 | 680.1 | 676.1 | 20.3 | 658.0 | 720.9 | `███████░░░░░░░░░░░░░` |
| 1.20.54 | 1.23.54 | 0.27.4 | 10 | 600.8 | 592.3 | 19.3 | 586.5 | 648.4 | `██████░░░░░░░░░░░░░░` |
| 1.21.46 | 1.24.46 | - | 10 | 671.7 | 666.3 | 21.6 | 654.9 | 729.1 | `███████░░░░░░░░░░░░░` |
| 1.21.46 | 1.24.46 | 0.27.4 | 10 | 590.9 | 589.0 | 8.3 | 582.1 | 611.0 | `██████░░░░░░░░░░░░░░` |
| 1.22.13 | 1.25.13 | - | 10 | 657.2 | 656.1 | 10.0 | 646.1 | 678.2 | `███████░░░░░░░░░░░░░` |
| 1.22.13 | 1.25.13 | 0.27.4 | 10 | 583.9 | 582.3 | 8.8 | 569.8 | 600.7 | `██████░░░░░░░░░░░░░░` |
| 1.23.10 | 1.26.10 | - | 10 | 649.6 | 650.0 | 4.0 | 640.8 | 654.0 | `███████░░░░░░░░░░░░░` |
| 1.23.10 | 1.26.10 | 0.27.4 | 10 | 588.5 | 587.0 | 11.5 | 571.8 | 614.7 | `██████░░░░░░░░░░░░░░` |
[...]

PR #3529, opened 2025-07-23
Remove unreferenced arguments from S3 endpoint ruleset

akx:
None of these arguments are referenced by the ruleset, but the set of arguments is used for caching endpoint resolution results via https://github.com/boto/botocore/blob/7ddc232fff4297fc19f89a70ce95014beb2d77bf/botocore/endpoint_provider.py#L707-L708.

This means that the entire endpoint ruleset is evaluated (only to end up with the same result) for every S3-related request with a different Key, IOW for every file stored in S3. Since evaluating the rule tree is pretty slow, this causes significant slowdowns in mass operations, not to mention memory bloat for the (useless) cache.

The three removed parameters are not in the [equivalent JS SDK code]( https://github.com/aws/aws-sdk-js-v3/blob/2e7d107003aa85474670fbd4e03212c473e23432/clients/client-s3/src/endpoint/endpointResolver.ts#L9-L27) either.

Refs #3528.

### DIFF
--- a/botocore/data/s3/2006-03-01/endpoint-rule-set-1.json
+++ b/botocore/data/s3/2006-03-01/endpoint-rule-set-1.json
@@ -58,21 +58,6 @@
             "documentation": "Internal parameter to use object lambda endpoint for an operation (eg: WriteGetObjectResponse)",
             "type": "Boolean"
         },
-        "Key": {
-            "required": false,
-            "documentation": "The S3 Key used to send the request. This is an optional parameter that will be set automatically for operations that are scoped to an S3 Key.",
-            "type": "String"
-        },
-        "Prefix": {
-            "required": false,
-            "documentation": "The S3 Prefix used to send the request. This is an optional parameter that will be set automatically for operations that are scoped to an S3 Prefix.",
-            "type": "String"
-        },
-        "CopySource": {
-            "required": false,
-            "documentation": "The Copy Source used for Copy Object request. This is an optional parameter that will be set automatically for operations that are scoped to Copy Source.",
-            "type": "String"
-        },
         "DisableAccessPoints": {
             "required": false,
             "documentation": "Internal parameter to disable Access Point Buckets",


SamRemis:
Hi @akx,

Thank you for the PR.  Unfortunately those models are owned upstream by the service team and can be updated at any time.  Even if we were to merge this today, if the service team makes any unrelated changes to their endpoints tomorrow, this file would just be overwritten and the keys would come back.  

While these aren't used in ruleset evaluation for botocore, my current understanding is that they are used by other AWS products and cannot be removed from the upstream ruleset.  I am still working on gathering more specifics.

Thank you for calling this out - it doesn't make sense for us to slow down evaluation for botocore users based on parameters we don't use.  We should either come up with an alternate solution here that causes unused keys to be ignored for caching purposes or we should find a way to remove these upstream.

akx:
@SamRemis Hi, thanks for looking into this!

Yeah, I figured this isn't the correct way to fix this given these lines had been added by an automated bot commit anyway, but I wanted to get the ball rolling.

> We should either come up with an alternate solution here that causes unused keys to be ignored for caching purposes [...]

As mentioned above, the JavaScript SDK has a better set of cache keys (`clients/client-s3/src/endpoint/endpointResolver.ts` added in https://github.com/aws/aws-sdk-js-v3/pull/6449/commits/4be028b96da4c70d0012159b17a7281cf6d4c3af, "chore: codegen sync").

It looks like the [`smithy-typescript-codegen` codegen code checks for this exact eventuality](https://github.com/smithy-lang/smithy-typescript/blob/b15137dba0091b26b3dd5d6efaac58545cf1c18a/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/endpointsV2/RuleSetParameterFinder.java#L70-L161) and elides parameters that aren't used by the ruleset for purposes of cache keying. (As it happens, [I wrote basically similar code earlier to prepare for this PR](https://github.com/akx/botocore-sign-perf/blob/master/find_unused_parameters.py) and based on it, it looks like that these 3 are apparently the only parameters unused across all of Botocore.)

One somewhat simple option could be to add some tribal knowledge to the resolver in botocore for this particular use: iff the ruleset is for S3, and the original set of parameters in the ruleset is this known one, then remove the three unused args before passing to caching/evaluation? Also probably a test that would fail if the S3 ruleset's parameters change, so maintainers would know to take action to adjust the code.

Another more labor-intensive option I can think of that would probably be for the best in the longer run is to also codegen ruleset evaluation into Python, instead of having some Python code evaluate the JSON at runtime.

akx:
I see the endpoint rule set [got updated recently](https://github.com/boto/botocore/commit/9a989600aa603bf27babd960f253b83434989e0a#diff-fdfd0c316c664afdf9aefea1ed3c1d30e66e000758eb93a61bbf74fe2ef47cad), but AISI, no changes to the unused arguments.

SamRemis:
We don't want to modify the endpoint rules directly, those are sourced internally from AWS service models and would be overwritten by S3's next release. Closing in favor of #3692, which preserves the original rules but drops the unused parameters later in the code path. 

Thanks again @akx for raising this and doing the research that informed the fix.

PR #3530, opened 2025-07-24
remove python 3.13 deprecation warning

maarten-k:
removes:
```
auth.py:422: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    datetime_now = datetime.datetime.utcnow()
```


nateprewitt:

PR #3536, opened 2025-08-04
Remove ASCII character restrictions from S3 metadata validation 

ubaskota:
This change does the following:
- Removes ASCII character restrictions from S3 metadata validation to allow Unicode characters in metadata values.
- Adds support to decode s3 encoded metadata values.
- Adds tests to validate that the encoded metadata are properly decoded.
- Adds tests to validate that the metadata with non-ascii values are allowed.

The following steps were performed to test the changes:
- Added and successfully ran the newly added tests along with the existing ones.
- Manually validated that s3 accepts non-ASCII unicode chars for metadata.
- Manually validated that the encoded metadata is decoded back to its original value.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -639,37 +639,6 @@ def document_base64_encoding(param):
     return append.append_documentation
 
 
-def validate_ascii_metadata(params, **kwargs):
-    """Verify S3 Metadata only contains ascii characters.
-
-    From: http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
-
-    "Amazon S3 stores user-defined metadata in lowercase. Each name, value pair
-    must conform to US-ASCII when using REST and UTF-8 when using SOAP or
-    browser-based uploads via POST."
-
-    """
-    metadata = params.get('Metadata')
-    if not metadata or not isinstance(metadata, dict):
-        # We have to at least type check the metadata as a dict type
-        # because this handler is called before param validation.
-        # We'll go ahead and return because the param validator will
-        # give a descriptive error message for us.
-        # We might need a post-param validation event.
-        return
-    for key, value in metadata.items():
-        try:
-            key.encode('ascii')
-            value.encode('ascii')
-        except UnicodeEncodeError:
-            error_msg = (
-                'Non ascii characters found in S3 metadata '
-                f'for key "{key}", value: "{value}".  \nS3 metadata can only '
-                'contain ASCII characters. '
-            )
-            raise ParamValidationError(report=error_msg)
-
-
 def fix_route53_ids(params, model, **kwargs):
     """
     Check for and split apart Route53 resource IDs, setting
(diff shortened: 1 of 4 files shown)


botocore/handlers.py
```diff
@@ -1561,6 +1574,7 @@ def get_bearer_auth_supported_services():
     ('after-call.s3.ListObjects', decode_list_object),
     ('after-call.s3.ListObjectsV2', decode_list_object_v2),
     ('after-call.s3.ListObjectVersions', decode_list_object_versions),
+    ('after-call.s3.HeadObject', decode_metadata),
```
SamRemis, line 1577: Please add this to GetObject as well.  

Do we have a way to programatically check for other methods that return custom metadata so we know to apply this customization in the future?

botocore/handlers.py
```diff
@@ -357,6 +357,56 @@ def _sse_md5(params, sse_member_prefix='SSECustomer'):
     params[sse_md5_member] = key_md5_str
 
 
+def decode_metadata(parsed, **kwargs):
```
SamRemis, line 360: Could this function be shortened to something like this?

```
if object_metadata := parsed.get('Metadata'):
    for metadata_key, metadata_value in object_metadata.items():
        if is_mime_encoded(metadata_value):
            object_metadata[metadata_key] = decode_mime_encoded_metadata_value(metadata_value)
```

botocore/handlers.py
```diff
@@ -357,6 +357,56 @@ def _sse_md5(params, sse_member_prefix='SSECustomer'):
...
+
+def is_mime_encoded(metadata_value):
+    import re
+
+    pattern = r'=\?[-\w]+\?[BQbq]\?[A-Za-z0-9+/=]+\?='
+    return bool(re.match(pattern, metadata_value))
+
+
+def decode_mime_encoded_metadata_value(encoded_value):
+    # Metadata is first utf-8 encoded before being sent in the s3 request.
```
SamRemis, line 384: Is it possible an object isn't double encoded because it was uploaded elsewhere?  How would this behave in that case?

Also, what is the current user experience for someone who uploads an object with non-ascii metadata using a different tool (for example the console), and then gets the metadata from it?  Would this break them?

SamRemis (commented):
First round of review leaving a few initial comments.  I'll need to do a deeper dive on how all of these encodings are handled across botocore, S3, and other tools to really be able to give this a full review.

botocore/handlers.py
ubaskota: All SDKs encode their requests before sending them to S3. Once received, S3 further mime encodes the data using both 'utf-8' and 'base64'. However, if the metadata was not double-encoded as intended, botocore may return the metadata in a form that's different from what the user uploaded.

S3 does not support non-ASCII metadata for objects uploaded through the console.

botocore/handlers.py
ubaskota: Upon running a short script that checks the output shapes and digging further, I found that there are a couple of other operations that deal with metadata configurations, but `GetObject` and `HeadObject` are the only ones that return metadata.

ubaskota:
During our review, we identified some oddities in how our encoding logic interacts with S3’s handling of object metadata, particularly for non-ASCII values. Supporting this reliably would likely result in a clunky implementation or a poor user experience. Additionally, we would want to ensure that this behavior is consistent across all our SDKs. For now, we are closing this out until there is a stronger demand for support.

Currently, S3 returns MIME-encoded responses for object metadata containing non-ASCII characters. For example, an object uploaded with boto3 requires two decoding steps to retrieve the same utf-8 metadata that was provided with an upload: first with "utf-8" and then with "base64".   While this works for the way boto3 uploads non-ASCII metadata, we can't guarantee that all tools will upload it with the same encoding.  This means that a user could upload an object with complex metadata with one tool, and we wouldn't be able to decode it.
 
Because of this, Botocore will maintain its current behavior and continue to not support non-ASCII metadata for S3 objects.  Closing this for now, but if it is necessary we can reopen this in the future.

PR #3539, opened 2025-08-13
Register feature ids for imds and http credentials

ubaskota:
This PR adds feature IDs for IMDS and HTTPS credentials, and registers them for tracking.

changes:
```
Adds feature IDs for "CREDENTIALS_HTTP" and "CREDENTIALS_IMDS".
Registers feature IDs when the corresponding features are used.
Adds functional tests that validate that the feature ids are registered each time they are used.
```

Tests:
`Validated the changes with functional tests.`


botocore/credentials.py
```diff
@@ -2180,6 +2182,15 @@ def load_credentials(self):
         # -js
         return None
 
+    def check_and_register_feature_id(self, method_name):
+        METHOD_FEATURE_MAP = {
```
SamRemis: Nit: the capital letters here suggests that this is intended to be a constant, which would usually be defined on a class constant or module level.  

This winds up creating this dictionary every time, which isn't going to be super time intensive or anything, but if we are going to use this as a constant, I'd rather see this defined higher up on the class level so it's only defined once and reused every time we call this method.

botocore/credentials.py
```diff
@@ -2170,6 +2171,7 @@ def load_credentials(self):
             logger.debug("Looking for credentials via: %s", provider.METHOD)
             creds = provider.load()
             if creds is not None:
+                self.check_and_register_feature_id(provider.METHOD)
```
SamRemis: You may have a better understanding of this than me, but I think the`load_credentials` method here is ultimately called from session.create_client, then those are cached and reused for multiple service calls with the same client.  

That means this feature id will only show up in the first call using a client unless I'm misunderstanding something here.  We may want to move adding this feature id to the signing method if we still have the info accessible at that point.  If not, I'm happy to discuss other alternatives.

SamRemis (changes requested):

botocore/signers.py
```diff
@@ -297,6 +303,8 @@ def get_auth_instance(
             return auth
 
         credentials = request_credentials or self._credentials
+        if credentials and (cred_method := credentials.method):
```
SamRemis: ~[very non-blocking, mega-nit]: is there a reason this is in parenthesis?  It makes it -slightly- harder to read for me~

Editing: this was me being dumb.  That would cause the interpreter to read this wrong and be invalid syntax.  A potential option would be nesting the if statements, but this is fine as well

SamRemis: Also as a note, this is potentially breaking if credentials doesn't have a method if someone is doing something hackey and using duck-typing to pass in a different credentials class.  This is a private interface though, so I think that's okay?

SamRemis (approved):

PR #3542, opened 2025-08-19
Fix use_ssl parameter ignored when custom HTTPS endpoint_url provided

adev-code:
Fixes a bug where setting `use_ssl=False` with a custom HTTPS endpoint incorrectly downgrades the connection to HTTP. According to the [documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html), when `endpoint_url` is provided, the `use_ssl` parameter should be ignored. The fix adds a check to ensure this downgrade only happens when no custom endpoint is specified, preserving the HTTPS protocol when customers explicitly provide an HTTPS URL.

**Links:**
- GitHub Issue: https://github.com/boto/boto3/issues/4252
- Code location: https://github.com/boto/botocore/blob/develop/botocore/regions.py#L524
- Documentation: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html

Fixes boto/boto3#4252

### DIFF
--- a/tests/unit/test_endpoint_provider.py
+++ b/tests/unit/test_endpoint_provider.py
@@ -14,7 +14,7 @@
 import json
 import logging
 import os
-from unittest.mock import Mock
+from unittest.mock import Mock, patch
 
 import pytest
 
@@ -24,6 +24,7 @@ from botocore.endpoint_provider import (
     ErrorRule,
     RuleCreator,
     RuleSet,
+    RuleSetEndpoint,
     RuleSetStandardLibrary,
     TreeRule,
 )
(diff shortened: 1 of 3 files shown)


tests/unit/test_endpoint_provider.py
```diff
@@ -542,3 +542,136 @@ def test_aws_is_virtual_hostable_s3_bucket_allow_subdomains(
...
+        with patch.object(
+            ssl_resolver_true,
+            '_get_provider_params',
+            return_value={'Endpoint': 'https://custom.com'},
+        ):
+            result = ssl_resolver_true.construct_endpoint(None, None, None)
+            assert result.url == 'https://custom.com'
+
+    # use_ssl=False, endpoint_url="http://..." → HTTP (http is not used anyway)
+    with patch.object(
```
SamRemis: Thanks for adding these tests!  

I'm hoping you don't mind me asking for a few updates:
1. it's hard to read what this test is doing without splitting this into six individual tests (`test_use_ssl_true_endpoint_http`, `test_use_ssl_false_endpoint_http`, etc.).  This could also be done with pymark.parametrize and make it one test that checks all scenarios.
3. The comments here are good for building an understanding, but they aren't self-explanatory, so they shouldn't be committed to develop.  For instance `# use_ssl=False, endpoint_url="http://..." → HTTP (http is not used anyway)`: reading this without context, I wouldn't understand what "http is not used anyway" means.

adev-code: Thanks Sam, I updated the test file.

SamRemis (approved):

Issue #3213, opened 2024-06-27
Race condition in botocore's use of JSONFileCache for credential loading


mikelambert:
### Describe the bug

I am getting the a rare crash when using botocore to load files.

It's a `JSONDecodeError: Expecting value: line 1 column 1 (char 0)` stemming from `JSONFileCache.__getitem__`, which looks to imply the json file is empty (didn't find any value at char 0). Someone helped point out it might be a race condition in the `JSONFileCache.__set__` , which appears to do:
```
            f.truncate()
            f.write(file_content)
```

We have multiple concurrent processes starting on a box that each are using botocore, so maybe this is just a race condition if one of them happens to look at the file post-truncate-pre-write? Not sure if a flock, or write-then-rename, or something else ends up a proper solution here?

### Expected Behavior

It shouldn't crash

### Current Behavior

```
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
  File "botocore/utils.py", line 3518, in __getitem__
    return json.load(f)
  File "__init__.py", line 293, in load
    return loads(fp.read(),
  File "__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None


  File "botocore/client.py", line 553, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "botocore/client.py", line 989, in _make_api_call
    http, parsed_response = self._make_request(
  File "botocore/client.py", line 1015, in _make_request
    return self._endpoint.make_request(operation_model, request_dict)
  File "botocore/endpoint.py", line 119, in make_request
    return self._send_request(request_dict, operation_model)
  File "botocore/endpoint.py", line 198, in _send_request
    request = self.create_request(request_dict, operation_model)
  File "botocore/endpoint.py", line 134, in create_request
    self._event_emitter.emit(
  File "botocore/hooks.py", line 412, in emit
    return self._emitter.emit(aliased_event_name, **kwargs)
  File "botocore/hooks.py", line 256, in emit
    return self._emit(event_name, kwargs)
  File "botocore/hooks.py", line 239, in _emit
    response = handler(**kwargs)
  File "botocore/signers.py", line 105, in handler
    return self.sign(operation_name, request)
  File "botocore/signers.py", line 186, in sign
[...]

Issue #3106, opened 2024-01-25
Credentials JSONFileCache explodes if cache file has malformed JSON


mildmojo:
### Describe the bug

The `JSONFileCache` class `__getitem__` accessor tries to load JSON from a file, but does not handle syntax errors in that file, raising an exception that bubbles up to the client application.

Given this is a cache, a failure to read from the cache should be treated as a cache miss, not an error.

### Expected Behavior

In `botocore/credentials.py`, in the `CachedCredentialFetcher` class, I expected `_load_from_cache()` to return `None` if the cache was invalid.

### Current Behavior

I found this while using `awscli`, which consumes `botocore`. When a credentials cache file was malformed, every command using that particular AWS account failed *super cryptically*, printing the bad file's cache key to STDERR:

```bash
$ aws ecr get-login-password --profile root

'c09eaf18178406053ca8802bc45deadbeefdc57a'

$ echo $?
255
```

Using a debugger, I located the failure at [`utils.py:3518`](https://github.com/boto/botocore/blob/e7c5b6ab22174797db551f44053a0b2245430649/botocore/utils.py#L3518), when `botocore` failed to parse a cache file. 

Cache file contents (prettified here):

```json
{
    "Credentials":
    {
        "AccessKeyId": "<redacted>",
        "SecretAccessKey": "<redacted>",
        "SessionToken": "<redacted>",
        "Expiration": "2024-01-24T02:49:26+00:00"
    },
    "AssumedRoleUser":
    {
        "AssumedRoleId": "<redacted>:botocore-session-1706060963",
        "Arn": "arn:aws:sts::<redacted>:assumed-role/Admin/botocore-session-1706060963"
    },
    "PackedPolicySize": 6,
    "ResponseMetadata":
    {
        "RequestId": "<redacted>",
        "HTTPStatusCode": 200,
        "HTTPHeaders":
        {
            "x-amzn-requestid": "<redacted>",
            "content-type": "text/xml",
            "content-length": "1569",
            "date": "Wed, 24 Jan 2024 01:49:26 GMT"
        },
        "RetryAttempts": 1
    }
}: true, "RetryAttempts": 2}}
```

(The bit at the bottom, `: true, "RetryAttempts": 2}}` invalidates the JSON.)

### Reproduction Steps

Here's a minimal reproduction:

```python
# json-cache-repro.py

from botocore.utils import JSONFileCache
from botocore.credentials import CachedCredentialFetcher
from os import remove

[...]

PR #3544, opened 2025-08-24
Add atomic write of creating json cache

ranlz77:
Add atomic file writing using `os.replace()` to prevent JSON file corruption that can occur during concurrent access to the file cache. 

**Fix:**
#3106 
#3213 

(aws-cli)
[#9632](https://github.com/aws/aws-cli/issues/9632)

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -22,6 +22,7 @@ import os
 import random
 import re
 import socket
+import tempfile
 import time
 import warnings
 import weakref
@@ -3578,11 +3579,34 @@ class JSONFileCache:
             )
         if not os.path.isdir(self._working_dir):
             os.makedirs(self._working_dir, exist_ok=True)
-        with os.fdopen(
-            os.open(full_key, os.O_WRONLY | os.O_CREAT, 0o600), 'w'
-        ) as f:
-            f.truncate()
-            f.write(file_content)
+        try:
+            temp_fd, temp_path = tempfile.mkstemp(
+                dir=self._working_dir,
+                suffix='.tmp'
+            )
+            
+            os.fchmod(temp_fd, 0o600)
+            with os.fdopen(temp_fd, 'w') as f:
+                temp_fd = None
+                f.write(file_content)
+                f.flush()
+                os.fsync(f.fileno())
+            
+            os.replace(temp_path, full_key)
+            temp_path = None
+            
+        except Exception:
+            if temp_fd is not None:
+                try:
+                    os.close(temp_fd)
+                except OSError:
+                    pass
+            if temp_path is not None and os.path.exists(temp_path):
+                try:
+                    os.unlink(temp_path)
+                except OSError:
+                    pass
+            raise
 
     def _convert_cache_key(self, cache_key):
         full_path = os.path.join(self._working_dir, cache_key + '.json')
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -14,6 +14,9 @@ import copy
 import datetime
 import io
 import operator
+import shutil
+import tempfile
+import os
 from contextlib import contextmanager
 from sys import getrefcount
 
@@ -54,6 +57,7 @@ from botocore.utils import (
     ArgumentGenerator,
     ArnParser,
     CachedProperty,
+    JSONFileCache, 
     ContainerMetadataFetcher,
     IMDSRegionProvider,
     InstanceMetadataFetcher,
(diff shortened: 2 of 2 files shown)


botocore/utils.py
```diff
@@ -3578,11 +3579,34 @@ def __setitem__(self, cache_key, value):
...
-        ) as f:
-            f.truncate()
-            f.write(file_content)
+        try:
+            temp_fd, temp_path = tempfile.mkstemp(
+                dir=self._working_dir,
+                suffix='.tmp'
+            )
+            
+            os.fchmod(temp_fd, 0o600)
```
ubaskota: This only works in unix based systems. You need to add a condition such that this line only runs when in unix based systems. For windows, the file permissions are set to read and write by default.

tests/unit/test_utils.py
```diff
@@ -3679,3 +3683,79 @@ def test_get_token_from_environment_returns_none(
...
+
+    @mock.patch('os.replace')
+    def test_uses_tempfile_and_replace_for_atomic_write(self, mock_replace):
+
+        self.cache['test_key'] = {'data': 'test_value'}
+        mock_replace.assert_called_once()
+
+        call_args = mock_replace.call_args[0]
+        temp_path = call_args[0]
+        final_path = call_args[1]
```
ubaskota: What is this line doing and why is it needed?

tests/unit/test_utils.py
```diff
@@ -3679,3 +3683,79 @@ def test_get_token_from_environment_returns_none(
...
+
+        assert '.tmp' in temp_path
+
+    def test_concurrent_writes_same_key(self):
```
ubaskota, line 3708: This test doesn't accurately validate your implementation. Your code creates a separate temporary file for each process attempting to write, but this test is only checking for data corruption during concurrent writes to the same key. The test should align with how your implementation actually handles file operations else it needs to be removed.

botocore/utils.py
ranlz77: Fixed it with adding a condition to check unix based function.

tests/unit/test_utils.py
ranlz77: Sorry for the mistake of unused var, fix it to try to check final_path of json file created by `os.replace`.

ranlz77:
Thank you for the detailed reviews, and apologies for the delayed feedback. 
I've made the suggested modifications in the latest commits.

tests/unit/test_utils.py
```diff
@@ -3679,3 +3683,92 @@ def test_get_token_from_environment_returns_none(
...
+
+        def track_temp_files(*args, **kwargs):
+            fd, path = original_mkstemp(*args, **kwargs)
+            temp_files_used.append(path)
+            return fd, path    
+        
+        def write_worker(thread_id):
+            try:
+                for i in range(3):
+                    self.cache[key] = {'thread': thread_id, 'iteration': i}
```
ubaskota, line 3725: I mentioned this in my previous review as well, could you clarify what this line is doing? It looks like you’re writing to a single key, but your implementation creates several temporary files. Can you explain the reasoning behind that?

ranlz77: Yes, I am trying to make sure the isolation of each write (even to the same key) via separated temp files and the last-write-wins behavior. The test validates that all 9 write operations target the same final file, but os.replace ensures the atomicity and no corruption.

ubaskota: I see. Just trying to understand this a bit better. What’s the difference between “writing to the same key via separate temp files” and simply “writing to separate temp files”? In this case, what’s the reason for tying all writes to the same key?

Also, if you look at the checks, you will see that some tests are failing on Windows because it doesn’t allow multiple processes to write to the same file concurrently. On Windows, file permissions are stricter than on Unix systems, so concurrent writes to the same target file can fail. Please, check and validate that all the tests are passing after you publish a revision.

ranlz77: Writing to different keys actually works well even without this implementation. However in real-world scenario, we are using automated scripts that trigger aws-cli to simultaneously refresh the same cached credential, and that often occurs corruption JSON file. The test of writing to the same key via separated temp files is to simulate this sort of race condition and to validate the last-write-wins. 

About the winsdow permission problem, I've made the modifications of adding some windows condition in the latest commit.

ranlz77:
Thank you for your reviews and progress in new PR. @ubaskota @ashovlin I think we can close this PR here, sorry for holding this issue for a long time.

ubaskota:
Implementation moved to: https://github.com/boto/botocore/pull/3597

PR #3549, opened 2025-08-27
Register feature ids for sso related credentials

ubaskota:
This PR adds feature IDs for SSO related credentials, and registers them for tracking.

**changes**:

- Adds feature IDs for `CREDENTIALS_PROFILE_SSO`, `CREDENTIALS_SSO`, `CREDENTIALS_PROFILE_SSO_LEGACY` and `CREDENTIALS_SSO_LEGACY`
- Registers feature IDs when the corresponding features are used.
- Adds functional tests that validate that the feature ids are registered each time they are used.

**Tests**:

- Validated the changes with manual and functional tests.
- Manual tests:
  1. With sso_session name:
     - Configured SSO credentials with `sso_session` name
     - Executed S3 operation
     - Verified feature IDs registration in user agent:
       - `CREDENTIALS_PROFILE_SSO` -> `r`
       - `CREDENTIALS_SSO` -> `s`

  2. Without sso_session name (legacy mode):
     - Configured SSO credentials without `sso_session` name
     - Executed S3 operation
     - Verified feature IDs registration in user agent:
       - `CREDENTIALS_PROFILE_SSO_LEGACY` -> `t`
       - `CREDENTIALS_SSO_LEGACY` -> `u`

### DIFF
--- a/tests/functional/test_credentials.py
+++ b/tests/functional/test_credentials.py
@@ -1483,11 +1483,6 @@ def test_user_agent_has_sso_legacy_credentials_feature_id(
     mock_load_sso_config.return_value = fake_fetcher_kwargs
     client_one = patched_session.create_client("s3", region_name="us-east-1")
     mock_load_sso_credentials.return_value = fake_response
-    with ClientHTTPStubber(client_one, strict=True) as http_stubber:
-        http_stubber.add_response()
-        http_stubber.add_response()
-        client_one.list_buckets()
-        client_one.list_buckets()
 
     _assert_feature_ids_in_ua(client_one, ['t', 'u'])
 


botocore/credentials.py
arandito, line 1: It looks like we are not registering the feature ids passed into the `SSOCredentialFetcher`.

We should call `register_feature_ids(self.feature_ids)`  before making the SSO client call to retrieve credentials: `response = client.get_role_credentials(**kwargs)` in line 2308.

This ensures we register either `CREDENTIALS_PROFILE_SSO` OR `CREDENTIALS_PROFILE_SSO_LEGACY` for the call to SSO.

tests/functional/test_credentials.py
```diff
@@ -1446,6 +1446,93 @@ def _assert_feature_ids_in_ua(client, expected_feature_ids):
             assert expected_id in feature_list
 
 
+@patch("botocore.credentials.CachedCredentialFetcher._load_from_cache")
```
arandito, line 1449: We may need to add a patch for `AssumeRoleProvider` to return `None` to both these tests.

tests/functional/test_credentials.py
```diff
@@ -1446,6 +1446,93 @@ def _assert_feature_ids_in_ua(client, expected_feature_ids):
...
+    mock_load_sso_config.return_value = fake_fetcher_kwargs
+    client_one = patched_session.create_client("s3", region_name="us-east-1")
+    mock_load_sso_credentials.return_value = fake_response
+    with ClientHTTPStubber(client_one, strict=True) as http_stubber:
+        http_stubber.add_response()
+        http_stubber.add_response()
+        client_one.list_buckets()
+        client_one.list_buckets()
+
+    _assert_feature_ids_in_ua(client_one, ['t', 'u'])
```
arandito, line 1487: I know we plan to edit these tests later but we should consider also testing for the initial SSO client call similar to the Assume Role feature id tests. The SSO client call should only register either `CREDENTIALS_PROFILE_SSO` OR `CREDENTIALS_PROFILE_SSO_LEGACY`.

botocore/credentials.py
ubaskota: Addressed.

tests/functional/test_credentials.py
ubaskota: While local testing worked without modifications, I implemented patches specifically for scenarios where AWS credentials are set for AssumeRoleProvider.

tests/functional/test_credentials.py
ubaskota: added

tests/functional/test_credentials.py
```diff
@@ -1446,6 +1446,103 @@ def _assert_feature_ids_in_ua(client, expected_feature_ids):
             assert expected_id in feature_list
 
 
+@patch("botocore.credentials.CachedCredentialFetcher._load_from_cache")
+@patch("botocore.credentials.SSOProvider._load_sso_config")
+@patch("botocore.credentials.ConfigProvider.load", return_value=None)
+@patch("botocore.credentials.SharedCredentialProvider.load", return_value=None)
```
jonathan343: Are these patches actually needed? These providers are lower precedence than SSO

tests/functional/test_credentials.py
```diff
@@ -1446,6 +1446,103 @@ def _assert_feature_ids_in_ua(client, expected_feature_ids):
...
+    }
+
+    mock_load_sso_config.return_value = fake_fetcher_kwargs
+    client_one = patched_session.create_client("s3", region_name="us-east-1")
+    mock_load_sso_credentials.return_value = fake_response
+    with ClientHTTPStubber(client_one, strict=True) as http_stubber:
+        http_stubber.add_response()
+        http_stubber.add_response()
+        client_one.list_buckets()
+        client_one.list_buckets()
```
jonathan343: This seems pretty redundant given the `_assert_feature_ids_in_ua` method already makes similar calls.

I do think the right way to do this is to make the calls in the test and simplify `_assert_feature_ids_in_ua` to take a subber and a list of feature ids and check if they were captured. Making the calls in `_assert_feature_ids_in_ua` seems wrong

tests/functional/test_credentials.py
```diff
@@ -1446,6 +1446,103 @@ def _assert_feature_ids_in_ua(client, expected_feature_ids):
...
+
+    _assert_feature_ids_in_ua(client_one, ['t', 'u'])
+
+
+@patch("botocore.credentials.CachedCredentialFetcher._load_from_cache")
+@patch("botocore.credentials.SSOProvider._load_sso_config")
+@patch("botocore.credentials.ConfigProvider.load", return_value=None)
+@patch("botocore.credentials.SharedCredentialProvider.load", return_value=None)
```
jonathan343: Same here. Why are the `ConfigProvider` and `SharedCredentialProvider` needed?

tests/functional/test_credentials.py
ubaskota: Removed. These remained after copying from previous test cases.

tests/functional/test_credentials.py
ubaskota: Removed. These remained after copying from previous test cases.

tests/functional/test_credentials.py
ubaskota: Yeah, those aren't required. This function (`_assert_feature_ids_in_ua`) is used by other tests too, so I'll refactor the tests in a separate PR.

PR #3551, opened 2025-08-28
Register feature ids for credentials set using environment variables and code

ubaskota:
This PR adds feature IDs for credentials set using environment variables and code, and registers them for tracking.

changes:

- Adds feature IDs for "CREDENTIALS_CODE" and "CREDENTIALS_ENV_VARS".
- Registers feature IDs when the corresponding features are used.
- Adds functional tests that validate that the feature ids are registered each time they are used.

Tests:

- Validated the changes with manual and functional tests.


tests/functional/test_credentials.py
```diff
@@ -1277,3 +1277,59 @@ def test_user_agent_has_http_credentials_feature_id(
...
+            client.list_buckets()
+            client.list_buckets()
+
+        ua_string = get_captured_ua_strings(http_stubber)
+        feature_list_one = parse_registered_feature_ids(ua_string[0])
+        feature_list_two = parse_registered_feature_ids(ua_string[1])
+        assert 'g' in feature_list_one and 'g' in feature_list_two
+
+    def test_user_agent_has_feature_id_set_for_credentials_set_via_code(
+        self,
```
SamRemis: These tests are starting to get really redundant.  Can we add two helper functions here?

1. Something to do the assertion:
```
    def _assert_feature_ids_in_ua(self, client, expected_feature_ids):
        """Helper to test feature IDs appear in user agent for multiple calls."""
        with ClientHTTPStubber(client, strict=True) as http_stubber:
            http_stubber.add_response()
            http_stubber.add_response()
            client.list_buckets()
            client.list_buckets()

        ua_strings = get_captured_ua_strings(http_stubber)
        for ua_string in ua_strings:
            feature_list = parse_registered_feature_ids(ua_string)
            for expected_id in expected_feature_ids:
                assert expected_id in feature_list
```
2. Consider using some sort of pymark.parametrize to duplicate less code- this would probably mean passing different stubbed clients into pymark.parametrize.  If it's unclear what I mean by that, feel free to ask for further clarification.

botocore/credentials.py
```diff
@@ -346,6 +346,8 @@ def _normalize(self):
         self.secret_key = botocore.compat.ensure_unicode(self.secret_key)
 
     def get_frozen_credentials(self):
+        if self.method == 'explicit':
+            register_feature_id('CREDENTIALS_CODE')
```
SamRemis: Genuine question here, not a suggestion.  I haven't done a deep dive on these code paths.  Could we just put this in the `__init__` method instead?  

My understanding is that this is a private interface, so it should only be called on client initialization by us, like [here](https://github.com/boto/botocore/blob/c99b6f200faec5936ad810a300d93bca57d1b180/botocore/session.py#L964-L969).  

There are *definitely* external users calling this directly though, so that may have a few issues if either:
- initializing this object on its own and attempting to register a feature id will raise an error, or
- creating an instance of this object then not using it and instead relying on the default chain will wind up still registering this feature id.

Putting this in the code I linked above would be another option.

ubaskota: The issue with placing the code inside `__init__` is that it fails to account for credentials provided at the time of session creation. If a user sets their AWS credentials when initializing a `boto3.Session` as shown in the example below, the feature does not get properly registered when `register_feature_id` is invoked within `__init__`

```
session = boto3.Session(
    aws_access_key_id="accesskeyid",
    aws_secret_access_key="secretaccesskey",
    aws_session_token="sessiontoken"
)
s3 = session.client('s3')
```

I wanted all calls for registering feature ids for credentials to be inside `credentials.py`. I'll look into registering it in `session.py`.

tests/functional/test_credentials.py
ubaskota: Noted. I will refactor the tests.

botocore/session.py
```diff
@@ -967,6 +971,9 @@ def create_client(
                 token=aws_session_token,
                 account_id=aws_account_id,
             )
+            # This enables feature id registration when credentials are set during client creation.
+            if getattr(credentials, 'method', None) == 'explicit':
+                register_feature_id('CREDENTIALS_CODE')
```
SamRemis: Maybe this wouldn't work, but since we call self.get_credentials just below here anyhow (line 986), could we potentially get both cases by just putting this code outside of these if statements (line 987)?  That way we only have to register this feature id in one place.

botocore/session.py
```diff
@@ -516,6 +516,10 @@ def get_credentials(self):
         credentials.
 
         """
+        # This enables feature id registration when credentials are set in a session
+        # and that session is used for client creation.
+        if getattr(self._credentials, 'method', None) == 'explicit':
```
SamRemis: I'm assuming the getattr is to support people using alternative credentials that don't have a `method` property - do we have a clear use case for this, or is this just out of caution?

botocore/session.py
ubaskota: Addressed.

botocore/session.py
ubaskota: This is just out of caution.

botocore/session.py
```diff
@@ -984,6 +987,7 @@ def create_client(
                     ignored_credentials,
                 )
             credentials = self.get_credentials()
+        register_feature_id('CREDENTIALS_CODE')
```
SamRemis: As the code is today, this will register it for all credentials, regardless of source.  I should have been clearer in my previous comment, the if statement is still required.  I have two potential suggestions here with no strong preference between them.  We can either:

1. Avoid the if statement by registering this in two places: 
    - once in this method on line 927 where we set credentials to a `botocore.credentials.Credentials` object
    - once in the [set_credentials method ](https://github.com/boto/botocore/blob/develop/botocore/session.py#L484-L508)that [boto3 uses](https://github.com/boto/boto3/blob/develop/boto3/session.py#L98) when credentials are set on a session.
2. Keep this here, but add back the if statement.  Since that's easy enough for me to add as a suggestion, I've done that below.

```suggestion
        if getattr(self._credentials, 'method', None) == 'explicit':
            register_feature_id('CREDENTIALS_CODE')
```


Please consider the pros and cons of each and pick the one you prefer (or a third option I'm not considering)

botocore/session.py
```diff
@@ -967,6 +969,7 @@ def create_client(
                 token=aws_session_token,
                 account_id=aws_account_id,
             )
+            # This enables feature id registration when credentials are set during client creation.
```
SamRemis:
```suggestion
```

I think you may have intended to delete this.

ubaskota: Yup. Thanks for pointing out out.

botocore/session.py
ubaskota: I decided to go with option 2. Also, I am assuming its a typo and you meant to write `if getattr(credentials, 'method', None) == 'explicit':`

SamRemis (approved):

PR #3554, opened 2025-09-02
Track assume-role and profile related credential feature ids

arandito:
#### Description

This PR adds support for tracking the following feature IDs related for assume role and profile credential providers in the user agent string.

| Feature ID | Metric Value |
| --- |  :---:  |
| CREDENTIALS_ENV_VARS_STS_WEB_ID_TOKEN | h |
| CREDENTIALS_STS_ASSUME_ROLE | i | 
| CREDENTIALS_STS_ASSUME_ROLE_WEB_ID | k |
| CREDENTIALS_PROFILE | n |
| CREDENTIALS_PROFILE_SOURCE_PROFILE | o |
| CREDENTIALS_PROFILE_NAMED_PROVIDER | p |
| CREDENTIALS_PROFILE_STS_WEB_ID_TOKEN | q |

These values indicate which credential providers were used to resolve credentials and their source.

#### Example user agent string:
```
Boto3/1.40.11 md/Botocore#1.40.18 ua/2.1 os/macos#24.5.0 md/arch#arm64 lang/python#3.12.9 md/pyimpl#CPython m/D,b,i,o,Z,n cfg/retry-mode#legacy Botocore/1.40.18
```
- `i` indicates the original request used credentials resolved from calling STS `assume-role`
- `o` indicates the original request used credentials resolved from a `source_profile` in a config profile
- `n` indicates the `assume-role` request used credentials resolved from a config profile with static credentials

> Note: Feature id tests for SharedCredentialProvider and ConfigProvider will be refactored to reduce repetition once #3551 is merged.

### DIFF
--- a/tests/functional/test_credentials.py
+++ b/tests/functional/test_credentials.py
@@ -1378,6 +1378,10 @@ def test_user_agent_feature_ids(
             '[default]\naws_access_key_id = FAKEACCESSKEY\naws_secret_access_key = FAKESECRET',
             [
                 patch(
+                    "botocore.credentials.AssumeRoleProvider.load",
+                    return_value=None,
+                ),
+                patch(
                     "botocore.credentials.EnvProvider.load", return_value=None
                 ),
             ],
@@ -1387,6 +1391,10 @@ def test_user_agent_feature_ids(
             'AWS_CONFIG_FILE',
             '[default]\naws_access_key_id = FAKEACCESSKEY\naws_secret_access_key = FAKESECRET',
             [
+                patch(
+                    "botocore.credentials.AssumeRoleProvider.load",
+                    return_value=None,
+                ),
                 patch(
                     "botocore.credentials.EnvProvider.load", return_value=None
                 ),


botocore/credentials.py
```diff
@@ -164,6 +164,12 @@ def create_credential_resolver(session, cache=None, region_name=None):
     return resolver
 
 
+def register_credential_feature_ids(feature_ids):
```
jonathan343: I think it makes more sense for us to add this in `useragent.py` instead. If we need this in a different module, importing from `credentials.py` is odd.

tests/functional/test_credentials.py
```diff
@@ -1277,3 +1277,214 @@ def test_user_agent_has_http_credentials_feature_id(
...
+                'AWS_ROLE_ARN': 'arn:aws:iam::123456789012:role/test-role',
+                'AWS_WEB_IDENTITY_TOKEN_FILE': '{token_file}',
+                'AWS_ROLE_SESSION_NAME': 'test-session',
+            },
+            ['h'],  # CREDENTIALS_ENV_VARS_STS_WEB_ID_TOKEN
+            'k',  # CREDENTIALS_STS_ASSUME_ROLE_WEB_ID
+        ),
+    ],
+)
+def test_user_agent_has_assume_role_feature_ids(
```
alexgromero: Would it make sense to split this into two tests, one covering Assume Role and another for Assume Role with Web Identity? That separation could make the scenarios and failures easier to follow.

alexgromero (commented):
Thanks, Antonio! Looks good overall. I left one minor readability suggestion.

tests/functional/test_credentials.py
arandito: I agree. The test became harder to follow when trying to consolidate. Will update!

botocore/useragent.py
```diff
@@ -78,6 +78,13 @@
     'FLEXIBLE_CHECKSUMS_REQ_WHEN_REQUIRED': 'a',
     'FLEXIBLE_CHECKSUMS_RES_WHEN_SUPPORTED': 'b',
     'FLEXIBLE_CHECKSUMS_RES_WHEN_REQUIRED': 'c',
+    'CREDENTIALS_ENV_VARS_STS_WEB_ID_TOKEN': 'h',
```
jonathan343, line 83: This list is missing an entry for `CREDENTIALS_ENV_VARS` which is resulting in `m/g` not getting registered when using assume role with `credential_source=Environment`.

```suggestion
    'CREDENTIALS_ENV_VARS': 'g',
    'CREDENTIALS_ENV_VARS_STS_WEB_ID_TOKEN': 'h',
```

tests/unit/test_credentials.py
```diff
@@ -746,6 +746,46 @@ def test_account_id_with_invalid_arn(self):
         self.assertEqual(response, expected_response)
         self.assertEqual(response['account_id'], None)
 
+    def test_feature_ids_property(self):
```
jonathan343: I don't see too much value in both of the `test_feature_ids_property` tests. It looks like they're just checking that `@property` works as expected?

jonathan343 (changes requested):

tests/unit/test_credentials.py
arandito: Yea I see that. I'll remove them.

botocore/useragent.py
arandito: `CREDENTIALS_ENV_VARS` was not part of this PR. It was merged in #3551. I did test this PR with environment variables and confirmed they are registered when `credential_source=Environment`

botocore/credentials.py
arandito: Agreed. Will update!

jonathan343 (approved):

PR #3555, opened 2025-09-03
Remove sms client following the service deprecation

ubaskota:
Amazon sms is no longer publicly active. This PR will remove the, now deprecated, sms service from the Botocore SDK.


kdaily:
It looks like we might be able to remove these as well? Doesn't seem to be using them in the test suite here, but would be good to clean up if they are unnecessary so we aren't confused later.

https://github.com/boto/botocore/blob/eaa97a36a863b6fcaa2c3e0553b71dadd73a1e3d/tests/functional/test_paginator_config.py#L119-L121

ubaskota:
Those unused lines removed with the last commit.

PR #3562, opened 2025-09-11
Register feature id for credentials set using boto2 config file

ubaskota:
This PR registers feature id for credentials set using boto2 config file.

**changes:**

Adds feature IDs for `CREDENTIALS_BOTO2_CONFIG_FILE`.
Registers feature ID when the corresponding feature is used.
Adds a functional test that validates that the feature id is registered each time they are used.
Applies patches to credentials obtained from the `BotoProvider` for functional testing, specifically targeting credentials processed after Boto2 config in the provider chain.

**Tests:**
Validated the changes manually and with a functional test.


tests/functional/test_credentials.py
```diff
@@ -1213,6 +1213,10 @@ def test_credential_context_override(self):
                     "botocore.credentials.ContainerProvider.load",
                     return_value=None,
                 ),
+                patch(
```
jonathan343: Why is this patch needed?

tests/functional/test_credentials.py
```diff
@@ -1245,6 +1249,10 @@ def test_credential_context_override(self):
                         "AccountId": "01234567890",
                     },
                 ),
+                patch(
```
jonathan343: Same here, I'm not sure why this is needed

tests/functional/test_credentials.py
```diff
@@ -1281,6 +1289,34 @@ def test_credential_context_override(self):
...
+                    },
+                ),
+                patch(
+                    "botocore.credentials.ConfigProvider.load",
+                    return_value=None,
+                ),
+                patch(
+                    "botocore.credentials.SharedCredentialProvider.load",
+                    return_value=None,
+                ),
```
jonathan343, line 1305: Is there some environmental factor requiring that these be needed? Removing them passes for me locally?

tests/functional/test_credentials.py
ubaskota: When BotoProvider (boto2) credentials are configured on your machine, the credential resolution logic gives priority to these credentials over those provided by the Instance Metadata Service (IMDS) or ContainerProvider. As a result, Boto2 credentials will be used and this test will fail. To avoid this, we patch the Boto2 credentials to None. This skips Boto2 credentials and continues searching for credentials from other providers in the resolution chain.

tests/functional/test_credentials.py
ubaskota: same as above

tests/functional/test_credentials.py
ubaskota: The test could fail when its set to receive credentials from `SharedCredentialsProvider` or `ConfigProvider` because they take precedence over Boto2 configuration settings. While the test currently passes without these providers, they were included as a precautionary measure for future.

tests/functional/test_credentials.py
jonathan343: We really need to consider rewriting these tests. These tests could break when we add a new credentials resolver in the future. We should make the tests more robust to prevent false failures down the line.

I think this is out of scope for the PR but I think you should follow up with a refactor after these set of PRs are merged.

For now, we will realistically never be using a `BotoProvider` in our CI or internal infrastructure. We should remove this patch here and down below.

tests/functional/test_credentials.py
jonathan343: see comment above

tests/functional/test_credentials.py
jonathan343: These are more common providers that devs might be using. I'm fine leaving them for now, but not ideal.

jonathan343 (changes requested):

tests/functional/test_credentials.py
ubaskota: Removed the patch for `BotoProvider`

tests/functional/test_credentials.py
ubaskota: Removed the patch for `BotoProvider`

jonathan343 (approved):

PR #3568, opened 2025-09-27
Support default millisecond timestamp precision for custom services

SamRemis:
### Overview:
Currently, only the ISO8601 timestamp format supports sub-second precision (microseconds). However, the [Smithy standard](https://github.com/smithy-lang/smithy/blob/main/designs/defining-timestamps.md#protocol-limitations-rationale) specifies millisecond precision as the maximum supported granularity.

This mismatch is causing issues for bedrock-agentcore, which requires millisecond-precise timestamps to function correctly. This PR adds service-level millisecond precision support that complies with Smithy standards while maintaining backward compatibility.

### Implementation:
Introduces a new `creating-serializer` event that allows services to override timestamp precision during serializer creation. We can customize services to allow them to opt into millisecond precision without requiring client-side configuration changes.

### Alternatives Considered:
- **Client-side configuration:** Environment variables, config files, or SDK parameters. Rejected because this service needs millisecond precision by default - requiring all customers to configure this would create a poor user experience.
- **Operation-level precision:** More granular control per API call. Would require significant architectural changes (similar to parsers.py request-level overrides) that exceed current requirements.
- **Global default change:** Change all services to millisecond precision. Quickly dismissed as a breaking change.

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -60,6 +60,7 @@ from botocore.exceptions import (
     UnsupportedTLSVersionWarning,
 )
 from botocore.regions import EndpointResolverBuiltins
+from botocore.serialize import TIMESTAMP_PRECISION_MILLISECOND
 from botocore.signers import (
     add_dsql_generate_db_auth_token_methods,
     add_generate_db_auth_token,
@@ -1069,9 +1070,9 @@ def remove_bedrock_runtime_invoke_model_with_bidirectional_stream(
         del class_attributes['invoke_model_with_bidirectional_stream']
 
 
-def customize_bedrock_agentcore_serializer(serializer_kwargs, **kwargs):
-    """Event handler to enable millisecond precision for bedrock-agentcore."""
-    serializer_kwargs['timestamp_precision'] = 'millisecond'
+def enable_millisecond_timestamp_precision(serializer_kwargs, **kwargs):
+    """Event handler to enable millisecond precision"""
+    serializer_kwargs['timestamp_precision'] = TIMESTAMP_PRECISION_MILLISECOND
 
 
 def add_retry_headers(request, **kwargs):
@@ -1478,7 +1479,7 @@ BUILTIN_HANDLERS = [
     ),
     (
         'creating-serializer.bedrock-agentcore',
-        customize_bedrock_agentcore_serializer,
+        enable_millisecond_timestamp_precision,
     ),
     ('after-call.iam', json_decode_policies),
     ('after-call.ec2.GetConsoleOutput', decode_console_output),
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -1069,6 +1069,11 @@ def remove_bedrock_runtime_invoke_model_with_bidirectional_stream(
         del class_attributes['invoke_model_with_bidirectional_stream']
 
 
+def customize_bedrock_agentcore_serializer(serializer_kwargs, **kwargs):
+    """Event handler to enable millisecond precision for bedrock-agentcore."""
+    serializer_kwargs['timestamp_precision'] = 'millisecond'
+
+
 def add_retry_headers(request, **kwargs):
     retries_context = request.context.get('retries')
     if not retries_context:
@@ -1470,6 +1475,10 @@ BUILTIN_HANDLERS = [
     (
         'creating-client-class.bedrock-runtime',
         remove_bedrock_runtime_invoke_model_with_bidirectional_stream,
+    ),
+    (
+        'creating-serializer.bedrock-agentcore',
+        customize_bedrock_agentcore_serializer,
     ),
     ('after-call.iam', json_decode_policies),
     ('after-call.ec2.GetConsoleOutput', decode_console_output),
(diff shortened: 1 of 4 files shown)


botocore/serialize.py
```diff
@@ -85,6 +98,11 @@ class Serializer:
     MAP_TYPE = dict
     DEFAULT_ENCODING = 'utf-8'
 
+    def __init__(self, timestamp_precision='second'):
```
hssyoo: Is it possible to pass `timestamp_precision` to the `serialize_to_request` method instead? I think it would be a bit cleaner if serializer objects don't need to be tied to a specific `timestamp_precision ` value. I haven't looked into if it's possible to grab the value at the method call sites and I don't feel super strongly about it so feel free to merge if we think this is the best place

SamRemis: I like that idea better in theory, but it would be a hackey shim.  Right now [our serialization logic](https://github.com/boto/botocore/blob/5d50ec01fc458b7592c090f21a565335ff818a59/botocore/serialize.py#L452-L455) just passes in the value and shape.  At this point in the code, we have access to the following (assuming JSON parser):
- self (the whole serializer object) - storing something here wouldn't be an option for setting it on a method level due to race conditions that could occur
- The "key" that it will be serialized to 
- The modeled shape
- The DateTime object that we are serializing

That leaves us with two options:
1. Add this to the operation model somehow, probably using `service-2.sdk-extras.json` or some sort of model pre-processing.  That would get pretty difficult to maintain as these services grow
2. Find some way to pass around an `operation_context` of sorts that contains it, and add the precision to the context in `serialize_to_request`.  
2a. We could pass around a dictionary to these methods, which gets messy since these are generally inside a structure (maybe nested a few layers), so we have to make sure that persists.  
2b. contextvars would be the other solution I could think of here, but I'd be concerned about behavior if for some reason the context var didn't exist.  Maybe that's a non-issue since we would have just created the context in the serialize_to_request method.


Based on all of this, do you have a preferred solution?  I'm open to chat here, but right now I'd say we should either go with the current solution or 2b.

botocore/handlers.py
```diff
@@ -1069,6 +1069,11 @@ def remove_bedrock_runtime_invoke_model_with_bidirectional_stream(
         del class_attributes['invoke_model_with_bidirectional_stream']
 
 
+def customize_bedrock_agentcore_serializer(serializer_kwargs, **kwargs):
```
jonathan343: There isn't anything bedrock specific happening here. I think we should make this a generic handler that would be reused for other services if we decide they also need millisecond precision. Maybe something like `enable_millisecond_timestamp_precision` or shorter.

botocore/serialize.py
```diff
@@ -67,9 +67,22 @@
 HOST_PREFIX_RE = re.compile(r"^[A-Za-z0-9\.\-]+$")
 
 
-def create_serializer(protocol_name, include_validation=True):
+def create_serializer(
+    protocol_name, include_validation=True, timestamp_precision='second'
+):
+    """Create a serializer for the given protocol.
+
+    :param protocol_name: The protocol name to create a serializer for.
```
jonathan343, line 83: nit - These docstring entries are usually paired with a `:type <parameter_name>: <type>` entry like `:type protocol_name: str`

tests/unit/test_serialize.py
```diff
@@ -616,3 +616,89 @@ def test_restxml_serializes_unicode(self):
             self.serialize_to_request(params)
         except UnicodeEncodeError:
             self.fail("RestXML serializer failed to serialize unicode text.")
+
+
+class TestTimestampPrecisionParameter(unittest.TestCase):
```
jonathan343, line 625: Should we add coverage to ensure rfc822 isn't affected?

botocore/serialize.py
```diff
@@ -139,18 +157,36 @@ def _create_default_request(self):
     # Some extra utility methods subclasses can use.
 
     def _timestamp_iso8601(self, value):
-        if value.microsecond > 0:
-            timestamp_format = ISO8601_MICRO
+        """Return ISO8601 timestamp with precision based on timestamp_precision."""
+        # Smithy's standard is milliseconds, so we truncate the timestamp if the millisecond flag is set to true
+        if self._timestamp_precision == 'millisecond':
```
jonathan343: We should turn this into a constant

botocore/serialize.py
```diff
@@ -85,6 +98,11 @@ class Serializer:
     MAP_TYPE = dict
     DEFAULT_ENCODING = 'utf-8'
 
+    def __init__(self, timestamp_precision='second'):
+        if timestamp_precision is None:
+            timestamp_precision = 'second'
+        self._timestamp_precision = timestamp_precision
```
jonathan343, line 117: If the value isn't `None`, we should probably have some sort of validation and raise an error if we receive something unsupported like `nanosecond`. I know the else would handle this, but might give users the impression we actually do support nanosecond precision

botocore/serialize.py
```diff
@@ -139,18 +157,36 @@ def _create_default_request(self):
...
+        if self._timestamp_precision == 'millisecond':
+            base_timestamp = calendar.timegm(value.timetuple())
+            milliseconds = (value.microsecond // 1000) / 1000.0
+            return base_timestamp + milliseconds
+        else:
+            return int(calendar.timegm(value.timetuple()))
 
     def _timestamp_rfc822(self, value):
+        """Return RFC822 timestamp (always second precision - RFC doesn't support sub-second)."""
+        # RFC 2822 doesn't support sub-second precision, so always use second precision format
```
jonathan343, line 200: We should call this out in the docstring for `create_serializer`

jonathan343 (approved):

PR #3570, opened 2025-09-29
Refactor the tests for credentials feature id registration

ubaskota:
### Overview
The PR refactors credential feature ID registration tests across some providers to improve test organization and maintainability. 

### Changes
The existing parameterized tests were removed, and `CredentialsResolver` was patched to include a `CredentialsProvider` instance directly, rather than patching each credential provider individually within each test. This approach reduced the number of individual patches and simplified the test setup. Tests for the following credential providers were refactored accordingly:
- Environment variables
- Code
- Instance Metadata Service (IMDS)
- Container
- Boto2 config
- Process provider
- Single Sign On (SSO)

### Testing
Tested and verified that all tests pass successfully.

### DIFF
--- a/tests/functional/test_credentials.py
+++ b/tests/functional/test_credentials.py
@@ -1458,17 +1458,11 @@ def _assert_deferred_credential_feature_ids(
 
 
 class TestFeatureIdRegistered:
-    @patch("botocore.credentials.get_credentials")
     def test_user_agent_has_env_vars_credentials_feature_id(
         self,
-        mock_config_provider,
         monkeypatch,
         patched_session,
     ):
-        env_provider = EnvProvider()
-        credential_resolver = CredentialResolver([env_provider])
-        mock_config_provider.return_value = credential_resolver
-
         env_vars = {
             'AWS_ACCESS_KEY_ID': 'FAKEACCESSKEY',
             'AWS_SECRET_ACCESS_KEY': 'FAKESECRET',
@@ -1499,7 +1493,7 @@ class TestFeatureIdRegistered:
         self,
         _unused_env_provider,
         mock_iam_fetcher,
-        mock_config_provider,
+        mock_get_credentials,
         patched_session,
     ):
         iam_role_fetcher = InstanceMetadataFetcher()
@@ -1514,7 +1508,7 @@ class TestFeatureIdRegistered:
             "expiry_time": "2099-01-01T00:00:00Z",
         }
         mock_iam_fetcher.return_value = fake_credentials
-        mock_config_provider.return_value = credential_resolver
+        mock_get_credentials.return_value = credential_resolver
 
         client = patched_session.create_client("s3", region_name="us-east-1")
         _assert_feature_ids_in_ua(client, ['0'])
@@ -1526,7 +1520,7 @@ class TestFeatureIdRegistered:
         self,
         _unused_env_provider,
         mock_container_metadata_fetcher,
-        mock_config_provider,
+        mock_get_credentials,
         monkeypatch,
         patched_session,
     ):
@@ -1550,7 +1544,7 @@ class TestFeatureIdRegistered:
         }
 
         mock_container_metadata_fetcher.return_value = fake_credentials
-        mock_config_provider.return_value = credential_resolver
+        mock_get_credentials.return_value = credential_resolver
 
         client = patched_session.create_client("s3", region_name="us-east-1")
         _assert_feature_ids_in_ua(client, ['z'])
@@ -1562,7 +1556,7 @@ class TestFeatureIdRegistered:
         self,
         _unused_env_provider,
         mock_boto2_config,
-        mock_config_provider,
+        mock_get_credentials,
         patched_session,
     ):
         boto_provider = BotoProvider()
@@ -1575,7 +1569,7 @@ class TestFeatureIdRegistered:
             }
         }
         mock_boto2_config.return_value = fake_credentials
-        mock_config_provider.return_value = credentials_resolver
+        mock_get_credentials.return_value = credentials_resolver
 
         client = patched_session.create_client("s3", region_name="us-east-1")
         _assert_feature_ids_in_ua(client, ['x'])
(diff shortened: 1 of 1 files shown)


tests/functional/test_credentials.py
arandito, line 1239: We should refactor these tests to be in the `TestFeatureIdRegistered` class as well.

arandito, line 1314: Same here. We should refactor these tests into `TestFeatureIdRegistered`.

arandito, line 1263: Once we refactor the remaining tests, we should move this inside the `TestFeatureIdRegistered` class.

arandito, line 1436: We should be able to reuse this function for the refactored Assume Role tests. If we do, we should also move it inside the `TestFeatureIdRegistered` class.

tests/functional/test_credentials.py
```diff
@@ -1709,3 +1455,235 @@ def _assert_deferred_credential_feature_ids(
...
+    @patch("botocore.credentials.get_credentials")
+    def test_user_agent_has_env_vars_credentials_feature_id(
+        self,
+        mock_config_provider,
+        monkeypatch,
+        patched_session,
+    ):
+        env_provider = EnvProvider()
+        credential_resolver = CredentialResolver([env_provider])
+        mock_config_provider.return_value = credential_resolver
```
arandito: Do we need this? According to [`credentials.py`](https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L125-L139), EnvProvider is the first in the resolver chain. I don't think we need to mock out the `get_credentials` call if we monkeypatch the env vars.

tests/functional/test_credentials.py
```diff
@@ -1709,3 +1455,235 @@ def _assert_deferred_credential_feature_ids(
...
+    @patch("botocore.credentials.get_credentials")
+    @patch(
+        "botocore.utils.InstanceMetadataFetcher.retrieve_iam_role_credentials"
+    )
+    @patch("botocore.credentials.EnvProvider.load", return_value=None)
+    def test_user_agent_has_imds_credentials_feature_id(
+        self,
+        _unused_env_provider,
+        mock_iam_fetcher,
+        mock_config_provider,
```
arandito: `mock_config_provider` is a confusing name since we are mocking the `get_credentials` method. We should rename it to something like `mock_get_credentials`. It looks like this is the same for all the new tests added.

tests/functional/test_credentials.py
```diff
@@ -1709,3 +1455,235 @@ def _assert_deferred_credential_feature_ids(
...
+
+        fake_credentials = {
+            "role_name": "FAKEROLE",
+            "access_key": "FAKEACCESSKEY",
+            "secret_key": "FAKESECRET",
+            "token": "FAKETOKEN",
+            "expiry_time": "2099-01-01T00:00:00Z",
+        }
+        mock_iam_fetcher.return_value = fake_credentials
+        mock_config_provider.return_value = credential_resolver
```
arandito: In [credentials.py](https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L246), `get_credentials` returns `credential_resolver.load_credentials` and not the `credential_resolver` object directly. Do you know why this still works?

tests/functional/test_credentials.py
```diff
@@ -1709,3 +1455,235 @@ def _assert_deferred_credential_feature_ids(
...
+            aws_secret_access_key='FAKESECRET',
+            aws_session_token='FAKETOKEN',
+        )
+        _assert_feature_ids_in_ua(client, ['e'])
+
+    @patch("botocore.credentials.get_credentials")
+    @patch(
+        "botocore.utils.InstanceMetadataFetcher.retrieve_iam_role_credentials"
+    )
+    @patch("botocore.credentials.EnvProvider.load", return_value=None)
```
arandito, line 1491: Why do we need to mock `EnvProvider` if we are bypassing the credential chain by creating a `CredentialResolver`?

tests/functional/test_credentials.py
arandito, line 1364: We need to refactor these as well.

ubaskota: The original scope of this PR was to replace the parameterized tests for the features ids I worked in with the current ones and to patch the credentials providers so that the `CredentialResolver` is patched instead of individual providers. That’s why I didn’t refactor the other tests outside that scope.

That said, I’m happy to include the tests you mentioned inside the class if those are the only changes needed.

ubaskota: same as above.

ubaskota: same as above.

ubaskota: same as above

tests/functional/test_credentials.py
ubaskota: `.load_credentials()` isn’t needed in this case because we’re mocking `get_credentials()` directly by patching it to return a `CredentialsResolver` instance with our desired providers. Since we’re bypassing the actual function logic entirely, the internal implementation, including any calls to `.load_credentials()` doesn’t impact the test. As long as our patch returns a valid object that the code under test expects (a `CredentialsResolver` which is a list of credential providers), the mock will behave as expected.

tests/functional/test_credentials.py
ubaskota: That’s because `patched_session` sets environment variable credentials in [conftest.py](https://github.com/boto/botocore/blob/develop/tests/functional/conftest.py). If we don’t explicitly mock the environment provider to return None, the credential resolution will fall back to those environment credentials, which will lead to unintended behavior in the test.

tests/functional/test_credentials.py
ubaskota: same as the first response above.

tests/functional/test_credentials.py
ubaskota: addressed

tests/functional/test_credentials.py
ubaskota: addressed

Issue #3571, opened 2025-09-29
apigateway - add paginator for GetDomainNameAccessAssociations


mrweir:
### Describe the bug

The API for GetDomainNameAccessAssociations is missing a paginator.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Paginator should work for the method `get_domain_name_access_associations`.

The missing entry in `botocore/data/apigateway/2015-07-09/paginators-1.json` should be:
```json
{
    "GetDomainNameAccessAssociations": {
        "input_token": "position",
        "output_token": "position",
        "limit_key": "limit",
        "result_key": "items"
    }
}
```

### Current Behavior

The paginator for this methods is not implemented (not present in `botocore/data/apigateway/2015-07-09/paginators-1.json`).

### Reproduction Steps

Run:

```python
import boto3
client = boto3.client('apigateway')
paginator = client.get_paginator("get_domain_name_access_associations")
```

Results in this error:
```text
Traceback (most recent call last):
  File "/path/to/test.py", line 5, in <module>
    paginator = client.get_paginator("get_domain_name_access_associations")
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/botocore/client.py", line 1151, in get_paginator
    if not self.can_paginate(operation_name):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/botocore/client.py", line 1222, in can_paginate
    actual_operation_name = self._PY_TO_OP_NAME[operation_name]
                            ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
KeyError: 'get_domain_name_access_associations'
```

### SDK version used

1.34.46

### Environment details (OS name and version, etc.)

Ubuntu 24.04.2 LTS

adev-code:
Hello @mrweir, thanks for reaching out. Looking at the [API Doc](https://docs.aws.amazon.com/apigateway/latest/api/API_GetDomainNameAccessAssociations.html), it can be seen that the API should support pagination from the description: *The current pagination position in the paged result set.*. I have looked at other API's that uses [Pagination](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigateway.html#paginators), they all have similar description.

adev-code:
[...]

PR #3572, opened 2025-09-30
Add paginator for GetDomainNameAccessAssociations API for API Gateway Service. 

adev-code:
As per the API Gateway Service API [documentation](https://docs.aws.amazon.com/apigateway/latest/api/API_GetDomainNameAccessAssociations.html), it mentions: "*The current pagination position in the paged result set.*".

Looking at [other API's](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigateway.html#paginators) that has pagination, for example: GetDeployments", "GetDomainNames" as examples, they have the same pagination description as mentioned in the GetDomainNameAccessAssociations API documentation. 

Fixes: https://github.com/boto/botocore/issues/3571

### DIFF
--- a/botocore/data/apigateway/2015-07-09/paginators-1.json
+++ b/botocore/data/apigateway/2015-07-09/paginators-1.json
@@ -30,6 +30,12 @@
       "limit_key": "limit",
       "result_key": "items"
     },
+    "GetDomainNameAccessAssociations": {
+      "input_token": "position",
+      "output_token": "position",
+      "limit_key": "limit",
+      "result_key": "items"
+    },
     "GetModels": {
       "input_token": "position",
       "output_token": "position",


SamRemis:
We cannot edit any files in `/data/*` directly.  This has to come from the service team upstream

SamRemis:
See [contributing.rst](https://github.com/boto/botocore/blob/develop/CONTRIBUTING.rst#contributing-code) for more details.  

The general rule of thumb (though there is a little bit of a special case with the paginators files) is that when a service team updates their models, changes we make will get overwritten.  To make changes similar to this, we generally want to make them in `paginators-1.sdk-extras.json` so that they don't get overwritten.  the `sdk-extras.json` files will get merged with the models provided by the service team at runtime.

In this case, we are assuming that these paginators will work as expected and conform to the behavior expected by our pagination code.  While this is likely the case, the service team should still add this upstream since it will enable pagination for all SDKs and not just boto3/CLI.

SamRemis:
Closing this, but please make a ticket to the service team and encourage them to add these paginators.

PR #3574, opened 2025-10-09
Support null values in parsers when services accepts and returns null values.

adev-code:
This PR addresses parsing where the `_handle_integer` method returns with a TypeError when a service (e.g. RDS Data API) supports and returns `null` or `None` values.

The root cause lies in the `_handle_integer` method in `RestJSONParser` class, which directly calls int(value) without validating whether the value is None or not This creates an inconsistency where successful cases like integers parse correctly, but any value containing null values triggers a `TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'.`

The fix implements proper null value validation in the integer parsing logic, ensuring that null values are handled gracefully while maintaining backward compatibility with existing functionality. This change allows AWS services to properly support null values.

Fixes: https://github.com/boto/boto3/issues/4508


SamRemis:
The issue you linked points to the shape being [boxed long](https://github.com/boto/botocore/blob/6e8d143b0a13ab67560e7f400bb47363e82287f0/botocore/data/rds-data/2018-08-01/service-2.json#L356-L359), which means it is a nullable int in python.  This suggests that we have a bug.

I'd hope that we can do a few things to move this forward, but it may be asking too much, so I'm open to some discussion to make sure we can move this forward at a reasonable rate:

1. We need to add tests.  Ideally this comes in the form of upstream additions to the protocol tests for all protocols that support the `box` trait.  That may be all of them, we will need to find a definition of what the `box` trait does to confirm.
2. We should make sure this works for _all_ protocols and primitive types, not just integers in `RestJson`.  This likely requires more code changes to ensure all of our `_handle*` methods can accept a null when it's marked as `box`.
3. We have encountered some issues in the past with our parsers being too lenient, we should keep this change impacting as few shapes as possible.  Instead of always supporting nulls for shapes modeled as integers/longs, what do you think of changing the logic so that it only accepts `null` as a valid value if the shape is modeled as `box`, and keeping the logic the same otherwise?

adev-code:
Closing this PR as the approach has changed based on internal investigation.

This PR attempted to fix null handling in a list by checking the service model's `box` trait. However, investigation revealed that regardless of how the service model defines the shape (boxed or not), if the service returns null in a list, the SDK must preserve it to avoid data loss. The new approach handles null values universally in list parsing without relying on service model traits, ensuring no data is dropped.

New PR: https://github.com/boto/botocore/pull/3618

PR #3579, opened 2025-10-14
 TCP Keep-Alive Environment Variable Support

adev-code:
Currently, TCP Keep-Alive can be configured using `Config` object 
and `~/.aws/config` file as per [Configuration Doc](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html).
This PR adds environment variable support via `BOTOCORE_TCP_KEEPALIVE`.

Example Usage:
export BOTOCORE_TCP_KEEPALIVE=True or true
or
os.environ['BOTOCORE_TCP_KEEPALIVE'] = 'True' or 'true'

Continuation of https://github.com/boto/botocore/pull/2268.  Thank you to @ryansb for the original PR that this is based off of and @miketheman for bringing this back to our attention.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -557,11 +557,9 @@ class ClientArgsCreator:
         # This disables Nagle's algorithm and is the default socket options
         # in urllib3.
 
-        # looks at Config object, config file and Environment variable
         socket_options = [(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]
         client_keepalive = client_config and client_config.tcp_keepalive
         if client_keepalive is None:
-            # config_store checks environment variable first, then config file.
             client_keepalive = self._config_store.get_config_variable(
                 'tcp_keepalive'
             )
--- a/tests/unit/test_args.py
+++ b/tests/unit/test_args.py
@@ -115,7 +115,6 @@ class TestCreateClientArgs(unittest.TestCase):
             'partition_data': {},
         }
         call_kwargs.update(**override_kwargs)
-
         return self.args_create.get_client_args(**call_kwargs)
 
     def call_compute_client_args(self, **override_kwargs):


adev-code:
Tests and [Configuration Documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html) edit will be added.

adev-code:
Created a PR for Doc change to add `BOTOCORE_TCP_KEEPALIVE`

https://github.com/boto/boto3/pull/4642

adev-code:
Changelog added

botocore/args.py
```diff
@@ -553,16 +551,16 @@ def _resolve_endpoint(
             service_name, region_name, endpoint_url, is_secure
         )
 
-    def _compute_socket_options(self, scoped_config, client_config=None):
```
arandito, line 556: I know this is a private API so it ***shouldn't*** be used by anyone, but are we confident this will not break any customers?

I did a quick search and it doesn't look like we use this elsewhere in our repo.

tests/unit/test_args.py
```diff
@@ -115,6 +115,17 @@ def call_get_client_args(self, **override_kwargs):
             'partition_data': {},
         }
         call_kwargs.update(**override_kwargs)
+
+        # Bridge test's scoped_config to config_store for new _compute_socket_options implementation
```
arandito: I understand this bridge avoids modifying existing tests to keep backwards compatibility, but since we are breaking `_compute_socket_options`, it may make sense to refactor all the tcp keepalive tests to use `config_store` instead of `scoped_config`.

This can cause confusion for future developers that see a mismatch in what the functions accepts (which is only `client_config`) and what is used in tests (`scoped_config`). They may lack the historical context to understand why this bridge is necessary.

Other tests in this file use`config_store.set_config_variable` to set the config value, we should be able to use this here too.

arandito (commented):
Thanks for this PR Adrian!

I left two comments regarding backwards compatibility with the changes to `_compute_socket_options` and the testing bridge. Curious to see what Jonathan thinks.

Let's also make sure to merge this PR at the same time as the docs change in boto3.

botocore/args.py
adev-code: As per checking our code, the `_compute_socket_options` is only used in `return` statement of the function `get_client_args()`. Before, the `scoped_keepalive` checks the config file. Now it is taking care of `tcp_keepalive` as it also checks the config file. I suppose this should not break as the functionality of `scoped_keepalive` is now covered by `tcp_keepalive`.

tests/unit/test_args.py
adev-code: Yes that make sense. I will refactor (next commit) all the TCP keepalive tests to use `config_store.set_config_variable()` directly and removed the bridge entirely. The tests are now consistent with the rest and match what `_compute_socket_options` actually expects. Thanks for pointing this out as it may cause confusion in the future for developers.

botocore/args.py
jonathan343: We should keep the `scoped_config` argument in the signature to prevent people who may be overwriting this method from breaking. With a quick Google search, I found two examples of people overwriting this method:
- https://github.com/boto/boto3/discussions/3603
- https://github.com/boto/botocore/issues/2916

botocore/args.py
```diff
@@ -553,16 +551,16 @@ def _resolve_endpoint(
...
         socket_options = [(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]
         client_keepalive = client_config and client_config.tcp_keepalive
-        scoped_keepalive = scoped_config and self._ensure_boolean(
-            scoped_config.get("tcp_keepalive", False)
-        )
-        # Enables TCP Keepalive if specified in client config object or shared config file.
-        if client_keepalive or scoped_keepalive:
+        tcp_keepalive = self._config_store.get_config_variable('tcp_keepalive')
+
+        if client_keepalive or tcp_keepalive:
```
jonathan343: This doesn't respect our config precedence (Config > Env var > Shared AWS Config file). For example, if I have `Config(tcp_keepalive=False)` set on my client and `tcp_keepalive = True` in the shared aws config file, we will evaluate to True even though `Config` should have the highest precedence.

The correct thing to do here would be something like:
```py
client_keepalive = client_config and client_config.tcp_keepalive
if client_keepalive is None:
    # config_store checks environment variable first, then config file.
    client_keepalive = self._config_store.get_config_variable(
        'tcp_keepalive'
    )
if client_keepalive:
    socket_options.append((socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1))
```

botocore/configprovider.py
```diff
@@ -204,6 +204,12 @@
         None,
         None,
     ),
+    'tcp_keepalive': (
+        'tcp_keepalive',
+        'BOTOCORE_TCP_KEEPALIVE',
+        None,
```
jonathan343, line 210: Why `None` instead of `False`?

adev-code: Looking at other config, they use `None`. I initially used `False` but that could mean *explicitly disabled*.  I wanted to stick with *default* value of `None` for code consistency with other configs rather than forcing a specific disabled state.

botocore/args.py
adev-code: This make sense. Thank you Jonathan.

botocore/args.py
adev-code: Placed the `scoped_config` back

botocore/args.py
```diff
@@ -556,13 +556,17 @@ def _resolve_endpoint(
     def _compute_socket_options(self, scoped_config, client_config=None):
         # This disables Nagle's algorithm and is the default socket options
         # in urllib3.
+
+        # looks at Config object, config file and Environment variable
```
SamRemis: This seems like it may have been left in by mistake.  It's a description of what's happening on 5 livnes below on 565.  We 
```suggestion
```

botocore/args.py
```diff
@@ -556,13 +556,17 @@ def _resolve_endpoint(
...
         socket_options = [(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]
         client_keepalive = client_config and client_config.tcp_keepalive
-        scoped_keepalive = scoped_config and self._ensure_boolean(
-            scoped_config.get("tcp_keepalive", False)
-        )
-        # Enables TCP Keepalive if specified in client config object or shared config file.
-        if client_keepalive or scoped_keepalive:
+        if client_keepalive is None:
+            # config_store checks environment variable first, then config file.
```
SamRemis:
```suggestion
```
This is documented in the config store itself and is used by many other argument resolvers here.  IMO it's out of place, so I'm adding a suggestion to remove it.

tests/unit/test_args.py
```diff
@@ -115,6 +115,7 @@ def call_get_client_args(self, **override_kwargs):
             'partition_data': {},
         }
         call_kwargs.update(**override_kwargs)
+
```
SamRemis: Nit: seems like another accidental addition.  Not harmful or helpful, but adding a suggestion to remove it.  
```suggestion
```

tests/unit/test_args.py
```diff
@@ -240,10 +241,10 @@ def test_region_does_not_resolve_if_not_s3_and_endpoint_url_provided(self):
         )
         self.assertEqual(client_args['client_config'].region_name, None)
 
-    def test_tcp_keepalive_enabled_scoped_config(self):
-        scoped_config = {'tcp_keepalive': 'true'}
+    def test_tcp_keepalive_enabled_config_store(self):
+        self.config_store.set_config_variable('tcp_keepalive', True)
```
SamRemis: Is this testing what we are intending it to?  

This [no longer mocks out the values from the config file](https://github.com/boto/botocore/blob/23581d12f23774b99e413cf3149cde0dffeecd30/botocore/configprovider.py#L528-L532), but stops the lookup entirely and returns this.

We should talk about what a different testing strategy could look like for this, but I'm not sure this unit test is necessary any longer.  We aren't using scoped_config directly, so what we were testing has been abstracted away.  To test TCP sources, we really only need to check three options:
- nothing set
- set in any of the config locations from the provider store
- set in the client config

tests/unit/test_args.py
```diff
@@ -252,30 +252,28 @@ def test_tcp_keepalive_enabled_scoped_config(self):
...
         with mock.patch('botocore.args.EndpointCreator') as m:
+            self.config_store.set_config_variable('tcp_keepalive', True)
             self.call_get_client_args(
-                scoped_config={'tcp_keepalive': 'true'},
                 client_config=Config(tcp_keepalive=False),
             )
             self.assert_create_endpoint_call(
                 m,
-                socket_options=self.default_socket_options
-                + [(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)],
```
SamRemis, line 275: Why did we change this assertion?  Shouldn't setting it to true enable keepalive?

tests/unit/test_args.py
```diff
@@ -284,18 +282,28 @@ def test_tcp_keepalive_enabled_if_set_anywhere(self):
...
         with mock.patch('botocore.args.EndpointCreator') as m:
-            self.call_get_client_args(scoped_config=scoped_config)
+            self.call_get_client_args()
             self.assert_create_endpoint_call(
                 m, socket_options=self.default_socket_options
             )
 
     def test_tcp_keepalive_enabled_case_insensitive(self):
-        scoped_config = {'tcp_keepalive': 'True'}
+        self.config_store.set_config_variable('tcp_keepalive', True)
```
SamRemis: Why did we change this from a string to a boolean?  Doesn't that defeat the point of the test?

SamRemis (changes requested):

tests/unit/test_args.py
adev-code: I mistakenly changed it to boolean during refactoring - will revert back to string 'True' to properly test case-insensitive.

tests/unit/test_args.py
adev-code: Yes and Good catch. Thank you.

tests/unit/test_args.py
adev-code: - **Nothing set**: `test_tcp_keepalive_not_specified`
- **Set in config_store**: `test_tcp_keepalive_enabled_environment_variable` and `test_tcp_keepalive_enabled_case_insensitive`
- **Set in client_config**: `test_tcp_keepalive_enabled_if_set_anywhere`

PR #3586, opened 2025-11-04
Prevent JSON file corruption with atomic writes

ubaskota:
*Description of changes:*
Use `os.replace()` for atomic file operations to eliminate race conditions during concurrent access to the file cache. This ensures that writes are either fully completed or not applied at all, preventing partial writes that could leave the JSON file in an invalid state when multiple processes access the cache simultaneously.

Note: This change is based on the approach suggested in [#3544](https://github.com/boto/botocore/pull/3544) by @ranlz77. Thanks for identifying this issue and proposing the initial solution.

*Tests:*
I verified that all tests including the the newly added unit tests pass.

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -22,6 +22,7 @@ import os
 import random
 import re
 import socket
+import tempfile
 import time
 import warnings
 import weakref
@@ -3577,11 +3578,33 @@ class JSONFileCache:
             )
         if not os.path.isdir(self._working_dir):
             os.makedirs(self._working_dir, exist_ok=True)
-        with os.fdopen(
-            os.open(full_key, os.O_WRONLY | os.O_CREAT, 0o600), 'w'
-        ) as f:
-            f.truncate()
-            f.write(file_content)
+        try:
+            temp_fd, temp_path = tempfile.mkstemp(
+                dir=self._working_dir, suffix='.tmp'
+            )
+            if hasattr(os, 'fchmod'):
+                os.fchmod(temp_fd, 0o600)
+            with os.fdopen(temp_fd, 'w') as f:
+                temp_fd = None
+                f.write(file_content)
+                f.flush()
+                os.fsync(f.fileno())
+
+            os.replace(temp_path, full_key)
+            temp_path = None
+
+        except Exception:
+            if temp_fd is not None:
+                try:
+                    os.close(temp_fd)
+                except OSError:
+                    pass
+            if temp_path is not None and os.path.exists(temp_path):
+                try:
+                    os.unlink(temp_path)
+                except OSError:
+                    pass
+            raise
 
     def _convert_cache_key(self, cache_key):
         full_path = os.path.join(self._working_dir, cache_key + '.json')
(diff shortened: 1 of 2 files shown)


botocore/utils.py
```diff
@@ -3577,11 +3578,33 @@ def __setitem__(self, cache_key, value):
...
-            os.open(full_key, os.O_WRONLY | os.O_CREAT, 0o600), 'w'
-        ) as f:
-            f.truncate()
-            f.write(file_content)
+        try:
+            temp_fd, temp_path = tempfile.mkstemp(
+                dir=self._working_dir, suffix='.tmp'
+            )
+            if hasattr(os, 'fchmod'):
+                os.fchmod(temp_fd, 0o600)
```
akx, line 3586: This shouldn't be necessary – `mkstemp`'s documentation says "The file is readable and writable only by the creating user ID.".

botocore/utils.py
```diff
@@ -3577,11 +3578,33 @@ def __setitem__(self, cache_key, value):
...
+            with os.fdopen(temp_fd, 'w') as f:
+                temp_fd = None
```
akx, line 3588: `os.fdopen` is an alias of `open`. Exiting the context manager (for example when unwinding from an exception) should close the underlying fd, so I'm not sure the `temp_fd = None` + cleanup dance in `except Exception` below is necessary?

botocore/utils.py
```diff
@@ -3577,11 +3578,33 @@ def __setitem__(self, cache_key, value):
...
+                f.write(file_content)
+                f.flush()
+                os.fsync(f.fileno())
```
akx, line 3591: Is there a particular reason to add an explicit flush + fsync?

botocore/utils.py
ubaskota: Yes, you are right. This isn't needed. Thanks for pointing it out.

botocore/utils.py
ubaskota: You're correct that `os.fdopen()` takes ownership of the file descriptor and the context manager will close it automatically. But the cleanup code handles exceptions that occur before `os.fdopen()` and prevents double-close issues after it. The exception handler also removes orphaned temporary files from disk in both scenarios. Think of a scenario where there’s an exception before `os.fdopen()` (`os.fchmod()` fails), `temp_fd` is still open and needs manual closing to prevent file descriptor leaks.

botocore/utils.py
ubaskota: Closing the file doesn’t guarantee the data has reached the disk. `f.write()` sends data to Python’s buffer, then to the OS buffer, and the OS writes it to disk later.

If you call `os.replace()` right after closing and the system crashes before the OS flushes its buffer, you can end up with a partial or corrupt file. `flush()` pushes Python’s buffer to the OS, and `fsync()` forces the OS to commit it to disk. Together they ensure the data is fully written before you proceed. More here:  https://docs.python.org/3/library/os.html#os.fsync

ubaskota:
This PR will be closed shortly. Per our open source contribution guidelines, we'll update the original PR #3544 instead.

ubaskota:
Implementation moved to: https://github.com/boto/botocore/pull/3597

PR #3590, opened 2025-11-12
[Fix] Remove weakref usage in RequestSigner

adev-code:
Remove weakref.proxy usage in RequestSigner. The weakref was added for Python 2.6 memory leak prevention and is no longer needed.

### DIFF
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -13,6 +13,7 @@
 import base64
 import datetime
 import json
+import weakref
 
 import botocore
 import botocore.auth
@@ -84,7 +85,7 @@ class RequestSigner:
         self._auth_token = auth_token
         self._service_id = service_id
 
-        self._event_emitter = event_emitter
+        self._event_emitter = weakref.proxy(event_emitter)
 
     @property
     def region_name(self):


tests/functional/leak/test_resource_leaks.py
```diff
@@ -22,10 +22,10 @@ class TestDoesNotLeakMemory(BaseClientDriverTest):
     # a substantial amount of time to the total test run time.
     INJECT_DUMMY_CREDS = True
     # We're making up numbers here, but let's say arbitrarily
-    # that the memory can't increase by more than 13MB.
+    # that the memory can't increase by more than 10MB.
 
     # TODO: Attempt to bring this back to 10MB once Python 3.14 releases
-    MAX_GROWTH_BYTES = 13 * 1024 * 1024
+    MAX_GROWTH_BYTES = 10 * 1024 * 1024
```
nateprewitt: Since these are two unrelated fixes, let's try to keep them in separate PRs. From the looks of it, the tests are still using elevated memory, so this change can't be done currently.

adev-code: Ok, thanks @nateprewitt 👍🏼

adev-code:
Based on the CI test failures, it looks like the weakref line needs to stay in place. Removing it causes test failures, which confirms that it's a required part of the implementation. The weakref helps free up memory more quickly. Without it, memory consumption grows substantially over time, making the removal not worthwhile.

Closing.

PR #3592, opened 2025-11-14
Add hostPrefix support to RPC v2 CBOR serializer

ubaskota:
*Description of changes:*
This change adds hostPrefix support to the RPC v2 CBOR serializer by including the host prefix in the serialized request.

*Testing*
Verified that all tests pass successfully.

### DIFF
--- a/tests/unit/test_serialize.py
+++ b/tests/unit/test_serialize.py
@@ -535,16 +535,6 @@ class TestInstanceCreation(unittest.TestCase):
         with self.assertRaises(ParamValidationError):
             self.assert_serialize_invalid_parameter(request_serializer)
 
-    def test_validation_raises_error_for_unknown_parameters(self):
-        operation_model = self.service_model.operation_model('TestOperation')
-        serializer = serialize.create_serializer(
-            self.service_model.metadata['protocol']
-        )
-
-        params = {'Timestamp-Non-existant': ''}
-        with self.assertRaises(ParamValidationError):
-            serializer.serialize_to_request(params, operation_model)
-
 
 class TestHeaderSerialization(BaseModelWithBlob):
     def setUp(self):


botocore/serialize.py
```diff
@@ -1057,6 +1057,10 @@ def serialize_to_request(self, parameters, operation_model):
         if input_shape is not None:
             self._serialize_payload(parameters, serialized, input_shape)
 
+        host_prefix = self._expand_host_prefix(parameters, operation_model)
+        if host_prefix is not None:
+            serialized['host_prefix'] = host_prefix
+
```
SamRemis, line 1063: I was hoping to point you to similar tests for other protocols, but if they exist, I'm not seeing them when I search the tests directory.

It's probably better to over-test than to under-test.  If it's not too difficult, can we add a test for this to `test_serialize.py`?

SamRemis (changes requested):
So I'm alright without a changelog since no services that are released today use CBOR in botocore, but I'd like to update the description a bit.  Can we clarify that this is bringing it inline with every other protocol, not just JSON, or remove that sentence?

tests/unit/test_serialize.py
```diff
@@ -737,3 +737,71 @@ def test_invalid_timestamp_precision_raises_error(self):
...
+    def test_no_host_prefix_when_not_configured(self):
+        operation_model = self.service_model.operation_model('TestOperation')
+        serializer = serialize.create_serializer('smithy-rpc-v2-cbor')
+
+        params = {'Foo': 'bound'}
+        serialized = serializer.serialize_to_request(params, operation_model)
+
+        self.assertNotIn('host_prefix', serialized)
+
+    def test_no_host_prefix_when_params_invalid(self):
```
SamRemis: This test isn't actually testing anything about host prefix, right?  We shouldn't include this unless it's testing related behavior.

tests/unit/test_serialize.py
```diff
@@ -737,3 +737,71 @@ def test_invalid_timestamp_precision_raises_error(self):
...
+                },
+                'StringType': {
+                    'type': 'string',
+                },
+            },
+        }
+        self.service_model = ServiceModel(self.model)
+
+    def test_host_prefix_added_to_serialized_request(self):
+        self.model['operations']['TestOperation']['endpoint'] = {
```
SamRemis: Instead of adding to the operation dynamically for this test, can we just make two operations like `TestNoHostPrefixOperation` and `TestHostPrefixOperation`?

If we do this, would also consolidate this into a single test with pymark.parametrize, but that's up to you.  Keeping them separate is fine too.

tests/unit/test_serialize.py
ubaskota: Yeah, the test doesn’t reach the change in `serialize.py`, but it validates that `ParamValidationError` is raised when a parameter with an invalid key name (that doesn’t exist in the shape : ’Foo-Non-existant’) is used.

tests/unit/test_serialize.py
```diff
@@ -535,6 +535,16 @@ def test_instantiate_with_validation(self):
...
+    def test_no_host_prefix_when_params_invalid(self):
+        operation_model = self.service_model.operation_model('TestOperation')
+        serializer = serialize.create_serializer(
+            self.service_model.metadata['protocol']
+        )
+
+        params = {'Timestamp-Non-existant': ''}
+        with self.assertRaises(ParamValidationError):
+            serializer.serialize_to_request(params, operation_model)
+
```
nateprewitt: How is this different than the test immediately above this?

tests/unit/test_serialize.py
```diff
@@ -535,6 +535,16 @@ def test_instantiate_with_validation(self):
         with self.assertRaises(ParamValidationError):
             self.assert_serialize_invalid_parameter(request_serializer)
 
+    def test_validation_raises_error_for_unknown_parameters(self):
```
nateprewitt: What is this testing specifically? I don't think this is related to the serializer. We already have tests for the validator [here](https://github.com/boto/botocore/blob/4e1ef91e7554d048a3cd709a4c5017bd4cbba965/tests/unit/test_validate.py#L75-L89).

ubaskota: Noted. I'll remove it.

tests/unit/test_serialize.py
ubaskota: Addressed.

PR #3597, opened 2025-11-21
Update json cache to eliminate race condition

ubaskota:
*Description of changes:*
Use `os.replace()` for atomic file operations to eliminate race conditions during concurrent access to the file cache. This ensures that writes are either fully completed or not applied at all, preventing partial writes that could leave the JSON file in an invalid state when multiple processes access the cache simultaneously.

Tests:
I verified that all tests including the the newly added unit tests pass.

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -17,7 +17,6 @@ import operator
 import os
 import shutil
 import tempfile
-import threading
 from contextlib import contextmanager
 from sys import getrefcount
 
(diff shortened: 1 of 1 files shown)


botocore/utils.py
SamRemis, line 3572: What would be the tradeoffs of making sure the cache key is always unique to avoid this race condition entirely?

botocore/utils.py
```diff
@@ -3577,11 +3579,31 @@ def __setitem__(self, cache_key, value):
...
-        with os.fdopen(
-            os.open(full_key, os.O_WRONLY | os.O_CREAT, 0o600), 'w'
-        ) as f:
-            f.truncate()
-            f.write(file_content)
+        try:
+            temp_fd, temp_path = tempfile.mkstemp(
+                dir=self._working_dir, suffix='.tmp'
+            )
+            with os.fdopen(temp_fd, 'w') as f:
```
SamRemis: So @ubaskota and I chatted about we are removing the nested `os.open(full_key, os.O_WRONLY | os.O_CREAT, 0o600)`, but I want to record it here for posterity:

The inner `os.open` that was formerly on  line 3582 was necessary because Python has less restrictive default permissions for a newly created file.  Now that we are using `tempfile.mkstemp`, the default permissions are already owner read/write for owner only, so we no longer need the nested open call.

botocore/utils.py
```diff
@@ -3577,11 +3579,31 @@ def __setitem__(self, cache_key, value):
...
+                temp_fd = None
```
SamRemis: What specific errors are we trying to protect against by setting `temp_fd` and `temp_path` to `None`?  

If we don't have a clear idea of the new failure modes we are accounting for by adding this logic, we should remove it.  If we do and these `None` checks are necessary/valuable, we should document that here so it's easier to maintain going forward.

ubaskota: We can remove the lines setting temp_fd and temp_path to None along with the exception handler, and the code will work fine in normal operation. However, it may introduce issues in error scenarios as described below:

1. File descriptor leak: If an exception occurs between mkstemp() and os.fdopen(), the file descriptor remains open
2. Orphaned temp files: If errors occur during write/flush/fsync or if os.replace() fails, temporary files will remain on disk forever 

These are edge cases that handle those rare scenarios. Since we haven't encountered these problems historically and they haven't caused issues in our environment, I'm okay with removing the cleanup code.

tests/unit/test_utils.py
```diff
@@ -3679,3 +3684,85 @@ def test_get_token_from_environment_returns_none(
...
+    def test_uses_tempfile_and_replace_for_atomic_write(self, mock_replace):
+        self.cache['test_key'] = {'data': 'test_value'}
+        mock_replace.assert_called_once()
+
+        call_args = mock_replace.call_args[0]
+        temp_path = call_args[0]
+
+        assert '.tmp' in temp_path
+
+    def test_concurrent_writes_to_multiple_temp_files(self):
```
SamRemis: This passes both before and after the above changes. It's not possible in pytest to simulate a partial write during `f.write(file_content)` without complex monkey-patching, and the added complexity may not be worth the value.

tests/unit/test_utils.py
```diff
@@ -3679,3 +3684,85 @@ def test_get_token_from_environment_returns_none(
...
+
+        for thread_id in range(3):
+            key = f'concurrent_test_{thread_id}'
+            final_data = self.cache[key]
+            self.assertIsInstance(final_data, dict)
+            self.assertEqual(final_data['thread'], thread_id)
+            self.assertIn('thread', final_data)
+            self.assertIn('iteration', final_data)
+
+    def test_atomic_write_preserves_data_on_failure(self):
```
SamRemis: Similar to the above, this passes both before and after the changes above.  It's simulating a random error during  at the start of `cache._dumps`, so I don't think it's really testing what it's intending to.  Would you be open to removing or rewriting this one as well?

tests/unit/test_utils.py
ubaskota: I decided to remove this test.

tests/unit/test_utils.py
ubaskota: I decided to remove this test.

PR #3602, opened 2025-11-25
Use generic error handling for 4XX empty responses

adev-code:
### Problem

When AWS services return 4XX errors with empty bodies, users receive XML parsing errors instead of meaningful HTTP status messages. For example, service returns HTTP 413 with an empty body when request size limits are exceeded, but users see:
```
botocore.parsers.ResponseParserError: Unable to parse response (no element found: line 1, column 0), invalid XML received.
```
### Solution

This PR extends generic error handling from 5XX to 4XX status codes, preventing parsing failures and surfacing the actual HTTP error, for example:
```
botocore.exceptions.ClientError: An error occurred (413) when calling the PutDashboard operation: Content Too Large
```
### Backwards Compatibility

- Preserves all existing 5XX behavior
- Only affects 4XX responses with empty/HTML bodies that would previously cause parsing failures


botocore/parsers.py
```diff
@@ -598,7 +598,7 @@ def _do_error_parse(self, response, shape):
             status_message = http.client.responses.get(
                 response['status_code'], ''
             )
-            if status_message:
+            if status_message and hasattr(e, 'add_note'):
```
nateprewitt: I think this requires Python 3.11+ so this is probably not the approach we want to take to make sure this works everywhere.

adev-code: Thanks Nate, yes I only found out from the CI tests and trying options. Will check for other ways.

robert-schmidtke:
Hi, I believe I am running into this issue and was wondering whether this fix was still being pursued in some way.

PR #3618, opened 2026-01-22
Add support for None values in list parsing

adev-code:
## Description

This PR adds support for parsing lists containing null (None) values, which previously caused a TypeError. This issue was discovered with the RDS Data API.

## Problem

When a service returns a list with null values e.g. `[123, null, 456]`, the parser fails with:
```
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```
## Solution

Added null value checking in `_handle_list()` before attempting to parse items. This fix applies to all parsers. 

## Testing

Unit tests covering:
- Simple lists with null values
- Nested lists with null values  

Fixes https://github.com/boto/boto3/issues/4508

### DIFF
--- a/botocore/parsers.py
+++ b/botocore/parsers.py
@@ -352,7 +352,7 @@ class ResponseParser:
         parsed = []
         member_shape = shape.member
         for item in node:
-            # Treat all lists as sparse during parsing to safely handle null 
+            # Treat all lists as sparse during parsing to safely handle null
             # elements that may be present in service responses.
             if item is None:
                 parsed.append(None)


botocore/parsers.py
```diff
@@ -352,7 +352,10 @@ def _handle_list(self, shape, node):
         parsed = []
         member_shape = shape.member
         for item in node:
-            parsed.append(self._parse_shape(member_shape, item))
+            if item is None:
```
SamRemis, line 357: Looks great, thanks Adrian!  One comment, we should add a code comment explaining why we are doing this.  Something like this:
`#When parsing lists, we treat all lists as sparse to ensure safe handling of null values returned from the service`

adev-code: Thanks Sam! Added this note.

botocore/parsers.py
```diff
@@ -352,7 +352,11 @@ def _handle_list(self, shape, node):
...
+                # When parsing lists, we treat all lists as sparse to ensure safe handling of null values returned from the service
```
SamRemis:
```suggestion
            # Treat all lists as sparse during parsing to safely handle null 
            # elements that may be present in service responses.
            if item is None:
```

adev-code: Thanks Sam, changed it.,

SamRemis (approved):

PR #3619, opened 2026-01-23
Add fast path for list validation

SamRemis:
This PR improves performance during the validation of long lists by skipping unnecessary method calls when no additional validations beyond type checking are required for a given shape.

This is accomplished by making the shape aware of whether it has constraints that we will validate. This resulted in a 1.7x improvement in serialization/validation performance (40% faster) when serializing ten batches of 500 vectors with 1,536 floats each (768,000 floats total) for S3 PutVectors calls. The test was run on python 3.12 on a c6a.8xlarge AL2023 instance in us-west-2 with a bucket in the same region.

serialization_improvement

### DIFF
--- a/botocore/validate.py
+++ b/botocore/validate.py
@@ -14,7 +14,6 @@ Validation Errors
 """
 
 import decimal
-import functools
 import json
 from datetime import datetime
 
@@ -202,7 +201,6 @@ class ParamValidator:
     # Metadata attributes that we validate beyond type checking
     VALIDATED_METADATA_ATTRS = {'required', 'min', 'document', 'union'}
 
-    @functools.lru_cache
     def _shape_has_constraints(self, shape):
         """Whether the shape has validated constraints beyond type checking."""
         return bool(self.VALIDATED_METADATA_ATTRS & set(shape.metadata.keys()))


botocore/validate.py
```diff
@@ -246,6 +263,7 @@ def _validate_document(self, params, shape, errors, name):
             )
 
     @type_check(valid_types=(dict,))
```
jonathan343: Looks like you forgot to remove this line. There are now two `@type_check` decorators on this method
```suggestion
```

botocore/model.py
```diff
@@ -196,6 +197,16 @@ def metadata(self):
                 metadata[attr] = model[attr]
         return metadata
 
+    @CachedProperty
+    def has_constraints(self):
+        """Whether the shape has validation constraints beyond type checking.
+
+        This checks if the shape has any metadata that requires extra validation
+        such as min/max ranges, patterns, enums, etc.
```
jonathan343: nit - I don't think this docstring was update to reflect the latest values in `VALIDATED_METADATA_ATTRS`. Looks like it wasn't updated with 0a283c5c6

botocore/validate.py
```diff
@@ -298,22 +316,41 @@ def _validate_string(self, param, shape, errors, name):
...
         range_check(name, len(param), shape, 'invalid length', errors)
+
+        # If a list member does not have validation constraints, we will only check the type
+        member_type = member_shape.type_name
+        if (
+            member_type in self.SCALAR_TYPES
+            and not member_shape.has_constraints
+        ):
+            valid_types = self.SCALAR_TYPES[member_type]
+            valid_type_names = [str(t) for t in valid_types]
```
jonathan343: nit - We can move this line into the conditional since it isn't needed otherwise

botocore/model.py
```diff
@@ -97,6 +97,7 @@ class Shape:
         'clientContextParams',
         'requiresLength',
     ]
+    VALIDATED_METADATA_ATTRS = {'required', 'min', 'document', 'union'}
```
jonathan343: I'm wondering if we should instead push `VALIDATED_METADATA_ATTRS` and `has_constraints` into. Shape is a general-purpose data model used throughout botocore and I don't think it shouldn't need to know which of its metadata keys the validator considers meaningful.

I don't feel strongly, but think it's worth considering

SamRemis: The logic behind putting it here was that it was next to where the other attributes are listed in the code.  That makes it less likely to be forgotten if additional validated attributes are added in the future.

It looks like your sentence got cut off here, but I'm assuming you mean "into `validate.py`".  That makes more logical sense.  I'll consider leaving this here, but right now I'm leaning towards moving it into `validate.py`.

botocore/validate.py
```diff
@@ -181,6 +182,31 @@ def report(self, name, reason, **kwargs):
...
+    CONTAINER_TYPES = {
+        'structure': (dict,),
+        'map': (dict,),
+        'list': (list, tuple),
+    }
+
+    # Metadata attributes that we validate beyond type checking
+    VALIDATED_METADATA_ATTRS = {'required', 'min', 'document', 'union'}
+
+    @functools.lru_cache
```
jonathan343: Is this worth caching? I'm wondering if the value this saves us is worth the memory it takes up when storing all service shapes. From my quick investigation, we reuse the `ParamValidator` for all serialization ([src](https://github.com/boto/botocore/blob/d30db46883f26201954402df918679b581c6db50/botocore/serialize.py#L98-L99)). The more operations someone uses, the more this cache will grow and grow. We may end up in a state where we're essentially storing a portion of the service model in memory to save a bit of CPU time

SamRemis: You're right.  It was a holdover from when I had this on the shape, it made more sense to cache it there.  It saves only negligible time, so it's not worth the extra space.  I've removed it.  

I'm not updating the graphs or numbers in the PR description because the difference is so small.  To preserve the final numbers:
- Last run with caching (described above in graphs/final output numbers): 743.32ms average
- Run with caching today: 738.36ms average
- Run without caching today (what made it into the below commit): 745.29ms average

PR #3624, opened 2026-01-29
Change default S3 presign signature from SigV2 to SigV4

adev-code:
Changes the default signature version for S3 presign from SigV2 to SigV4, enabling support for conditional headers. SigV2 silently ignores conditional headers like IfNoneMatch and IfMatch, creating security issues. SigV4 supports header signing and is the modern AWS standard. 

Breaking change: Users requiring SigV2 must explicitly set signature_version='s3' in client [config](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-a-configuration-file). 

Adresses: https://github.com/boto/boto3/issues/4367

### DIFF
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -457,7 +457,7 @@ class ClientCreator:
         # the customer hasn't set a signature version so we default the
         # signature version to sigv2.
         client_meta.events.register(
-            'choose-signer.s3', self._default_s3_presign_to_sigv2
+            'choose-signer.s3', self._default_s3_presign_to_sigv4
         )
 
     def _inject_s3_input_parameters(self, params, context, **kwargs):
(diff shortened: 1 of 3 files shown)


SamRemis:
While I'm in favor of us upgrading the default signature version, there's a few things we need to do before we can merge this:
1. Determine customer impact as best we can internally.  We may not be able to do this at all.  We may also require some smarter configurations where we still default to v2 in some cases.  
2. Make a pinned GitHub issue warning users of the change and give them a minimum of a few month's notice.  We should consider blog posts and potentially adding this to the README as well.  
3. Make a changelog entry that makes this a `feature` so it becomes a minor version bump.  We don't want to make a change that will break customers as a patch bump.  
4. Add docs that tell customers how to downgrade back to v2 if necessary for their use case.  

Given the longer timeline here, I don't think we should keep this PR open, so I'll close it for now.  Feel free to re-open when we have a better idea of when we can safely do this.

PR #3628, opened 2026-02-06
Start testing Python 3.15 Alpha

jonathan343:
> [!NOTE]
> This PR is modeled after https://github.com/psf/requests/pull/7195

## Overview
This PR adds Python 3.15 canary coverage in CI and fixes deterministic gzip fixture failures in `tests/unit/test_compress.py` on 3.15 pre-release runtimes.

## Background
1. We want early signal for Python 3.15 regressions.
2. CPython [gh-91349](https://docs.python.org/3.15/whatsnew/changelog.html#changelog) changed default gzip compression level from 9 to 6, which changed our byte-for-byte gzip fixture output.
3. In 3.15-dev jobs, some dependencies may be built from source (instead of wheels) and use Rust/PyO3. PyO3 can reject unknown/new CPython minors by default during pre-release windows.

## Summary
1. Add `3.15-dev` to CI matrices in:
    - `.github/workflows/run-tests.yml`
    - `.github/workflows/run-crt-test.yml`
2. Set `PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1` only for `3.15-dev` jobs.
    - This is a CI canary workaround for prerelease interpreter builds. It allows PyO3 abi3 extension builds to proceed on 3.15-dev and can be removed when https://github.com/crate-py/rpds/issues/210 is resolved.
3. Keep gzip fixtures byte-stable by pinning `mtime=1` and `compresslevel=9` in `tests/unit/test_compress.py`.
4. Updates the locked `typing_extensions` version to address `AttributeError: module 'typing' has no attribute 'no_type_check_decorator'` errors.

## Testing
```
# Sent PyO3 forward compatibility env var
export  PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1

# Create Python 3.15 virtual environment
uv venv -p 3.15
source .venv/bin/activate

# Install package and dev requirements
uv pip install -e .
uv pip install -r requirements-dev-lock.txt

# Run unit and functional tests
python -m pytest tests/unit tests/functional

# Run integration tests
python -m pytest tests/integration
```

### DIFF
--- a/.github/workflows/run-crt-test.yml
+++ b/.github/workflows/run-crt-test.yml
@@ -15,6 +15,7 @@ concurrency:
 jobs:
   build:
     runs-on: '${{ matrix.os }}'
+    continue-on-error: ${{ matrix.python-version == '3.15-dev' }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -15,6 +15,7 @@ concurrency:
 jobs:
   build:
     runs-on: '${{ matrix.os }}'
+    continue-on-error: ${{ matrix.python-version == '3.15-dev' }}
     strategy:
       fail-fast: false
       matrix:


.github/workflows/run-crt-test.yml
nateprewitt, line 16: You may want to make sure these have `continue-on-error: ${{ matrix.python-version == '3.15-dev' }}` too. Otherwise, you're going to block PR merges on failures you may not be able to address immediately. It'll still give you signal with a ⚠️ flagged build if something breaks later.

jonathan343: Good call. I got distracted with addressing the failures and forgot to port that over. I've added it in d6765c30de11b8301b8a842351603f62f056bbb6

PR #3631, opened 2026-02-11
Implement opting out of S3 Express session auth via config variables

aemous:
*Description of changes:*

* Added support for opting out of S3 Express session authentication via new ``disable_s3_express_session_auth`` client configuration setting, or the new ``AWS_S3_DISABLE_EXPRESS_SESSION_AUTH`` environment variable, or the ``disable_s3_express_session_auth`` S3 shared configuration setting.
* Added new tests against the new configuration and functionality.

*Description of tests:*

* Successfully ran all unit and functional tests.

### DIFF
--- a/botocore/args.py
+++ b/botocore/args.py
@@ -323,8 +323,8 @@ class ClientArgsCreator:
                 account_id_endpoint_mode=client_config.account_id_endpoint_mode,
                 auth_scheme_preference=client_config.auth_scheme_preference,
                 s3_disable_express_session_auth=(
-                    s3_config.get('disable_s3_express_session_auth')
-                    if s3_config is not None
+                    client_config.s3.get('disable_s3_express_session_auth')
+                    if client_config.s3 is not None
                     else None
                 ),
             )


SamRemis:
One high level comment (got a few more comments to come):

Your PR description says that there's a new client configuration setting, but this doesn't work:
```
config = Config(disable_s3_express_session_auth=True)
s3 = boto3.client('s3',config=config)
```

We never added it to [the Config class](https://github.com/ashovlin/botocore/blob/develop/botocore/config.py#L25); how are users expected to set this in code?

aemous:
> We never added it to [the Config class](https://github.com/ashovlin/botocore/blob/develop/botocore/config.py#L25); how are users expected to set this in code?

@SamRemis As we discussed offline, it is set in code-config via the S3 config.

```python
import boto3

my_config = Config(
    s3 = {
        'disable_s3_express_session_auth': True
    }
)

client = boto3.client('s3', config=my_config)
# client = boto3.client('s3')
```

SamRemis:
Thank you, you're right. I wasn't familiar with the code path for service specific configs and was expecting an update in config.py

tests/functional/test_disable_s3_express_auth.py
```diff
@@ -0,0 +1,76 @@
...
+import datetime
+from unittest.mock import patch
+
+import pytest
+from dateutil.tz import tzlocal
+
+from tests import ClientHTTPStubber
+
+
+@pytest.mark.parametrize(
```
SamRemis: I appreciate the use of pytest.mark.parametrize here, but I think it may make more sense to split this into two more readable test cases.  One where we assert that s3express auth is disabled, and one where we assert that it's not. 

We can use class constants for reusable variables like the method responses, bucket name, and date.  Then use a pytest fixture for the datetime.  

We should also be adding an assertion on `len(http_stubber.requests)` for the number of requests on the stubbed client (expect one for the non-session auth test and two for the one with session auth).  We could potentially even assert the URL of the first request using `stubber.requests[0].url.endswith('?session')`, but that's relying on the URL of the API not changing.  

I'm asking for a refactoring of the whole class, so happy to chat more about this if you disagree or find any of it confusing.

botocore/configprovider.py
```diff
@@ -255,6 +255,12 @@
         None,
         utils.ensure_boolean,
     ),
+    'disable_s3_express_session_auth': (
+        ('s3', 'disable_s3_express_session_auth'),
+        'AWS_S3_DISABLE_EXPRESS_SESSION_AUTH',
+        None,
+        utils.ensure_boolean,
```
SamRemis: This line isn't doing what we expect.  It's intended to allow inputs like "TRUE", but if a user sets the following:
```
config = Config(
    s3 = {
        'disable_s3_express_session_auth': "true"
    }
)
```

They get this error: `botocore.exceptions.EndpointResolutionError: Value (DisableS3ExpressSessionAuth) is the wrong type. Must be <class 'bool'>`.  

Can you look into why we are bypassing this type coercion?

SamRemis: This statement was incorrect, I had to do some re-learning of our config system.  The config provider does not apply type coercion to inputs supplied on the client directly.  You're doing the same thing as `s3_disable_multiregion_access_points`. 

It's not great that we allow a user to pass in these configuration values directly and don't do any sort of validation or type checking on what they give us.  We validate some configurations, but not others.  Take for instance this config:
```
config = Config(
    s3 = {
        'disable_s3_express_session_auth': "true",
        'use_accelerate_endpoint': "false",
    },
    retries = {
        'mode': 'fake-retry-mode'
    }
)
```

Each of these three configurations will behave differently.  
- `disable_s3_express_session_auth`: this one will be passed through unvalidated to the endpoint resolver and raise the exception above.  This exception is clear enough for a user to be able to fix on their own
- `use_accelerate_endpoint`: this one will wind up resulting in a truthy value [here](https://github.com/boto/botocore/blob/54dbf1b6481f4b8950409d85174d7064398b610e/botocore/args.py#L783).  This means that `False` and `"false"` behave in opposite ways without giving the user any indication.
- `mode`: this one will [validate the and raise an error](https://github.com/boto/botocore/blob/54dbf1b6481f4b8950409d85174d7064398b610e/botocore/config.py#L454-L458)

Given the varying behavior of different inputs here, I don't feel strongly about updating this to add a validation when the error is already clear enough for a user to action.

aemous: Good catch; I also don't feel strongly on this and I'm optimistic that our docs should prevent egregious misuse of this config.

I would like to keep the current state in that case.

tests/functional/test_disable_s3_express_auth.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @pytest.fixture
+    def fixed_datetime(self):
+        return datetime.datetime(2024, 11, 30, 23, 59, 59, tzinfo=tzlocal())
+
+    def test_s3_express_auth_disabled(
+        self, patched_session, monkeypatch, fixed_datetime
+    ):
+        auth_type = None
+
```
SamRemis, line 27: I think there's still some slimming down to be done here.  I'd rather we avoid the use of nonlocal if we can as well.  The intention of this test is to check if we use s3 express session auth, but the assertion that requires using the event system only is asserting on the context.  IMO we can remove that assertion entirely since we can instead just check if the client called `create_session`.

We also will run into concurrency issues with xdist if we use monkeypatch.setenv, so it makes more sense to put this on the client config instead of relying on the environment variable.  

Open to discussing further if you have questions or disagree

aemous: This makes sense. I've published a revision based on these suggetions.

tests/functional/test_disable_s3_express_auth.py
aemous: Published a revision.

tests/functional/test_disable_s3_express_auth.py
```diff
@@ -0,0 +1,67 @@
...
+class TestDisableS3ExpressAuth:
+    DATE = datetime.datetime(2024, 11, 30, 23, 59, 59, tzinfo=tzutc())
+    BUCKET_NAME = 'mybucket--usw2-az1--x-s3'
+
+    CREATE_SESSION_RESPONSE = b'<?xml version="1.0" encoding="UTF-8"?>\n<CreateSessionResult><Credentials><AccessKeyId>test-key</AccessKeyId><Expiration>2024-12-31T23:59:59Z</Expiration><SecretAccessKey>test-secret</SecretAccessKey><SessionToken>test-token</SessionToken></Credentials></CreateSessionResult>'
+    LIST_OBJECTS_RESPONSE = b'<?xml version="1.0" encoding="UTF-8"?>\n<ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Name>mybucket--usw2-az1--x-s3</Name><Prefix/><KeyCount>0</KeyCount><MaxKeys>1000</MaxKeys><EncodingType>url</EncodingType><IsTruncated>false</IsTruncated></ListBucketResult>'
+
+    @pytest.fixture
+    def mock_datetime(self):
```
SamRemis, line 32: Nit: we could have used [autouse](https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request) here, but it's not a significant change.  It would just have prevented you from needing to pass it in to the below tests explicitly.

ashovlin, line 32: Do we need to mock datetime? I don't see the fixture referenced in the tests.

botocore/config.py
```diff
@@ -133,6 +133,12 @@ class Config:
             configured to use the us-east-1 region. This is the default if
             the configuration option is not specified.
 
+        * ``disable_s3_express_session_auth`` -- Refers to whether to use S3
+          Express session authentication. The value must be a boolean. If True, the
+          client will NOT use S3 Express session authentication.
+
+          Defaults to Nene.
```
ashovlin: Typo in "None"

tests/functional/test_disable_s3_express_auth.py
```diff
@@ -0,0 +1,67 @@
...
+        with mock.patch('datetime.datetime', spec=True) as mock_dt:
+            mock_dt.now.return_value = self.DATE
+            yield mock_dt
+
+    def test_disable_s3_express_auth_enabled(
+        self, patched_session, mock_datetime
+    ):
+        config = Config(s3={'disable_s3_express_session_auth': True})
```
ashovlin, line 40: Please test the full hierarchy after https://github.com/aws/aws-cli/pull/10105#discussion_r2855907624, not just the client config.

botocore/args.py
```diff
@@ -412,6 +412,16 @@ def compute_s3_config(self, client_config):
...
+    def compute_s3_disable_express_session_auth(self, client_config):
+        disable_express = self._config_store.get_config_variable(
+            's3_disable_express_session_auth'
+        )
+        if client_config is not None and client_config.s3 is not None:
+            disable_express = client_config.s3.get(
+                'disable_s3_express_session_auth'
+            )
+        return disable_express if disable_express is not None else False
+
```
ashovlin: Should we be doing validation here? Looks like we often raise `InvalidConfigError` from args.py. The internal spec has:
> Flag is boolean with only acceptable values of true and false, with any other value resulting in an error.

For `config = Config(s3={'disable_s3_express_session_auth': 1})`, I am seeing `botocore.exceptions.EndpointResolutionError: Value (DisableS3ExpressSessionAuth) is the wrong type. Must be <class 'bool'>`, but I think we'd rather raise sooner than letting it get to endpoint resolution. 

For a profile like the following, I'm not seeing an error
```
[default]
region=us-east-1
s3_disable_express_session_auth=optional
```

tests/functional/test_disable_s3_express_auth.py
aemous: It gets used internally by the underlying boto code, this may be more of a "monkey patch."

It's used by the credentials fetcher to check if the returned token is expired. Without mocking it, eventually the expiration date in the mocked response will pass, causing boto to think the token is expired, which causes it to retry the request a few times until the test fails to get an mocked unexpired token response.

This monkey patch makes the test code timeless.

botocore/args.py
aemous: Good question.

I had a discussion with Sam offline, and we noted that we have existing config settings that do not check for valid values, so this would be more consistent with those. We also noted that since this is a new config setting, we could do validation if we prefer that.

This is just the added context, I will make the revision since you stated your preference for validation.

aemous: Latest revision meets your expectations:

```
botocore.exceptions.InvalidConfigError: Invalid value "xyz" for s3_disable_express_session_auth. Value must be a boolean
```

tests/functional/test_disable_s3_express_auth.py
aemous: Revised.

botocore/config.py
aemous: Resolved.

botocore/args.py
ashovlin: Thanks! We have to be careful adding validations to existing options, since that could break an application that is silently succeeding despite an invalid value. But for brand new settings, I'd rather we start with validation rather than let an application proceed without it.

ashovlin (approved):

PR #3637, opened 2026-02-19
Add support for sha512, xxhash algorithms, and MD5 for httpchecksums

SamRemis:
*Description of changes:*
This PR updates the flexible checksums feature to add support for SHA512, XXHASH64, XXHASH3, XXHASH128 as well as add support for passing MD5 through as a checksum to services where it is supported when pre-calculated by the user.

Included are some small refactorings such as updating some unit tests to use pytest.mark.parametrize to reduce code duplication as well as updating the private `has_checksum_header` function to use a new `get_checksum_header` and which returns a list of algorithms found in the checksum header.

### DIFF
--- a/.changes/next-release/enhancement-httpchecksums.json
+++ b/.changes/next-release/enhancement-httpchecksums.json
@@ -1,6 +1,6 @@
 {
   "type": "enhancement",
   "category": "``checksums``",
-  "description": "Add support for SHA512, XXHASH64, XXHASH3, XXHASH128, and self-computed MD5 for flexible checksums"
+  "description": "Added support for the SHA512 checksum algorithm. When the optional AWS CRT (``awscrt``) dependency is installed, support is also enabled for the XXHASH64, XXHASH3, and XXHASH128 checksum algorithms. Also added pass-through support for customer-provided MD5 checksum headers (without SDK-side MD5 calculation or validation)."
 }
 


.changes/next-release/enhancement-httpchecksums.json
```diff
@@ -0,0 +1,6 @@
+{
+  "type": "enhancement",
+  "category": "``checksums``",
+  "description": "Add support for SHA512, XXHASH64, XXHASH3, XXHASH128, and self-computed MD5 for flexible checksums"
```
jonathan343: You don't need to use this wording exactly, but I think this changelog entry needs to be more clear about how we're supporting each algorithm:

```suggestion
  "description": "Added support for the SHA512 checksum algorithm. When the optional AWS CRT (``awscrt``) dependency is installed, support is also enabled for the XXHASH64, XXHASH3, and XXHASH128 checksum algorithms. Also added pass-through support for customer-provided MD5 checksum headers (without SDK-side MD5 calculation or validation)."
```

SamRemis:
> Thanks Sam!
> 
> I think these changes look pretty good! I don't have much feedback other than improving the changelog entry description. Happy to approve when that's updated.
> 
> One question I do have is: Can you can remind me why we're not switching to the CRT implementation of SHA512 when installed in the customers environment?

https://github.com/awslabs/aws-crt-python/blob/d6ed76b4ed79ab694f852cdecec197690c861674/awscrt/checksums.py

As far as I know, it's not currently supported.  I'll double check with the CRT team to make sure I'm not missing something prior to merging this PR

SamRemis:
Confirming here that's it's not currently supported in python bindings

PR #3639, opened 2026-02-23
Add logic to skip invalid feature id

SamRemis:
*Description of changes:*
With a recent PR, we added new get_checksum_header_algorithms function to get a list of headers that start with `x-amz-checksum-`, so if we have `x-amz-checksum-sha512`, it would return `sha512`.  

This function is [used to set feature ids](https://github.com/boto/botocore/blob/e0dc0476df0ae37ed679c47ea51b699b37d3186a/botocore/httpchecksum.py#L499-L500), and has a false positive for the `x-amz-checksum-algorithm` header and currently attempts to register a non-existent feature id for a checksum called "algorithm".  

To fix this and simplify the logic, this PR separates out the responsibilities of the functions.  The old `get_checksum_header_algorithms` has been renamed to `get_checksum_algorithm_headers` and returns only the raw headers.  The function which registers the feature ids will now filter out the `x-amz-checksum-algorithm` header and extract the algorithm name from the header name for registration.

Considering these changes are internal only and not public facing, I've decided against including a changelog.  Happy to include one in a future revision if requested.  

### DIFF
--- a/botocore/httpchecksum.py
+++ b/botocore/httpchecksum.py
@@ -498,7 +498,11 @@ def _register_checksum_feature_ids(request):
     if algorithm_headers := get_checksum_algorithm_headers(request):
         for header in algorithm_headers:
             header = header.upper()
-            if header != "X-AMZ-CHECKSUM-":
+            if header not in (
+                "X-AMZ-CHECKSUM-ALGORITHM",
+                "X-AMZ-CHECKSUM-MODE",
+                "X-AMZ-CHECKSUM-TYPE",
+            ):
                 algorithm_name = header.removeprefix("X-AMZ-CHECKSUM-")
                 _register_checksum_algorithm_feature_id(algorithm_name)
         return


botocore/httpchecksum.py
```diff
@@ -495,9 +495,12 @@ def _apply_request_trailer_checksum(request):
 
 def _register_checksum_feature_ids(request):
     """Register feature IDs for checksum algorithms used in the request."""
-    if algorithm_list := get_checksum_header_algorithms(request):
-        for algorithm_name in algorithm_list:
-            _register_checksum_algorithm_feature_id(algorithm_name)
+    if algorithm_headers := get_checksum_algorithm_headers(request):
+        for header in algorithm_headers:
+            header = header.upper()
+            if header != "X-AMZ-CHECKSUM-":
```
ashovlin: Should this be `!= "X-AMZ-CHECKSUM-ALGORITHM"` if that's the case we're trying to skip?

Issue #2216, opened 2020-11-13
sso-oidc exceptions don't translate "error_description" into ClientError message field


benkehoe:
The sso-oidc method [CreateToken](https://docs.aws.amazon.com/singlesignon/latest/OIDCAPIReference/API_CreateToken.html) returns an error response that looks like:
```json
{"error": "invalid_grant", "error_description": "Invalid grant provided"}
```
but on the `botocore.errorfactory.InvalidGrantException` object, the `error.response["Error"]["Code"]` is correct but `error.response["Error"]["Message"]` is an empty string.

This is using botocore 1.19.13

swetashre:
@benkehoe - Thank you for your post. While parsing a service response we expect the service to send error message with message attribute 
https://github.com/boto/botocore/blob/develop/botocore/parsers.py#L638

But in this case as the service is sending ```error_description``` instead of message, botocore is not able to parse that. Marking it as feature request.

benkehoe:
Thanks. Seems like it should be part of the response shape. In this case, the `error_description` field is required by the OAuth spec: https://tools.ietf.org/html/rfc6749#section-5.2

adev-code:
Hello @benkehoe, thank you for bringing this up. I wanted to give an update that the team is aware of this feature request and we will keep it open for tracking and to monitor 👍 reactions.

benkehoe:
Thanks for the update. I would call it a bug report rather than a feature request, this is something I would expect to work as part of the basic functionality of the SDK.

SamRemis:
An update to map the errors to the expected field has been merged and will be released with the next version of botocore.

Thank you for bringing this to our attention

PR #3642, opened 2026-02-27
Handle Oauth2 errors for SSO OIDC service

SamRemis:
*Issue #, if available:*
#2216

*Description of changes:*
Botocore has an expectation that service teams will return any error message in the `Message` field.  While most services are able to follow this standard, SSO-OIDC follows the OAUTH2 standard which puts the human-readable error in the `error_description` in JSON.

This PR adds a `before-parse` event to map the error_description to the expected location to display to customers.  This handler only attempts to modify the response when the status code is >= 400 and is wrapped in a try block for fault tolerance.  It also will not overwrite any existing `Message` or `message` returned from the service in the case that the service returns both an `error` description and a `Message` or `message` in a location we expect.

Reproduction code of the error case:
```
import boto3
from botocore.exceptions import ClientError

sso_oidc = boto3.client('sso-oidc')

register_response = sso_oidc.register_client(clientName='test-client', clientType='public')

client_id = register_response['clientId']
client_secret = register_response['clientSecret']

response = sso_oidc.create_token(clientId=client_id, clientSecret=client_secret, grantType='refresh_token', refreshToken='invalid_refresh_token_here')
```

Prior to this PR, this would get raised:
```
botocore.errorfactory.InvalidGrantException: An error occurred (InvalidGrantException) when calling the CreateToken operation: 
```

After this PR, the error_description from the service response is now surfaced in the exception message:
```
botocore.errorfactory.InvalidGrantException: An error occurred (InvalidGrantException) when calling the CreateToken operation: Public client can't be used for PresignedUrl refresh grant
```

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -1311,6 +1311,8 @@ def _should_handle_200_error(operation_model, response_dict):
 
 
 def _map_oauth2_errors(response_dict, **kwargs):
+    # SSO OIDC follows the OAuth2 standard, which returns error messages in
+    # 'error_description' instead of the 'Message' field botocore expects.
     try:
         if response_dict.get('status_code') < 400:
             return
--- a/tests/unit/test_handlers.py
+++ b/tests/unit/test_handlers.py
@@ -2174,7 +2174,7 @@ def test_map_oauth2_errors_adds_message():
     [
         # Success response
         (200, {'access_token': 'foo'}),
-        # no error_description
+        # No error_description
         (400, {'error': 'invalid_grant'}),
         # Empty error description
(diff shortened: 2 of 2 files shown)


botocore/handlers.py
```diff
@@ -1310,6 +1310,19 @@ def _should_handle_200_error(operation_model, response_dict):
     return True
 
 
+def _map_oauth2_errors(response_dict, **kwargs):
```
jonathan343, line 1313: I think it's worth adding a comment or docstring here explaining why we're doing this. Something like:

```
# SSO OIDC follows the OAuth2 standard, which returns error messages in
# 'error_description' instead of the 'Message' field botocore expects.
```

PR #3643, opened 2026-03-03
Exclude additional hop-by-hop headers from SigV4 signing

jonathan343:
## Overview
This PR aligns SigV4 header signing behavior with IAM guidance by excluding additional hop-by-hop headers from signature calculation. This prevents signature mismatch failures when intermediaries (for example, proxies or load balancers) mutate transport headers in transit.

## Background
Botocore already excluded a subset of volatile headers (`expect`, `transfer-encoding`, `user-agent`, `x-amzn-trace-id`), but it did not exclude the full IAM-documented hop-by-hop set.

## Summary
- Added the following headers to `SIGNED_HEADERS_BLACKLIST` in `botocore/auth.py`:
  - `connection`
  - `keep-alive`
  - `proxy-authenticate`
  - `proxy-authorization`
  - `te`
  - `trailer`
  - `upgrade`
- Added unit tests in `tests/unit/auth/test_signers.py` under `TestS3SigV4Auth` to verify each new header is excluded from `SignedHeaders`.
- Updated `_test_blocklist_header` to assert against parsed `SignedHeaders` tokens instead of raw substring matching in the full `Authorization` header (avoids false positives for short header names like `te`).
- Added changelog entry: `.changes/next-release/enhancement-auth-88913.json`.

## Testing
- `python -m pytest tests/unit/auth/test_signers.py`
- Live verification against STS (`GetCallerIdentity`):
  - Signed requests with each new hop-by-hop header present
  - Mutated those headers after signing
  - Confirmed no signature mismatch errors (received successful responses)

## References
- [IAM SigV4 guidance](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html)

---

### DIFF
--- a/tests/unit/auth/test_signers.py
+++ b/tests/unit/auth/test_signers.py
@@ -386,7 +386,12 @@ class TestS3SigV4Auth(BaseTestWithFixedDate):
         )
         auth = self.AuthClass(credentials, 's3', 'us-east-1')
         auth.add_auth(request)
-        self.assertNotIn(header, request.headers['Authorization'])
+        signed_headers = (
+            request.headers['Authorization']
+            .split('SignedHeaders=', 1)[1]
+            .split(',', 1)[0]
+        )
+        self.assertNotIn(header.lower(), signed_headers.split(';'))
 
     def test_blocklist_expect_headers(self):
         self._test_blocklist_header('expect', '100-continue')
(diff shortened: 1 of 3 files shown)


tests/unit/auth/test_signers.py
```diff
@@ -402,6 +407,31 @@ def test_blocklist_user_agent_header(self):
...
+
+    def test_blocklist_te_header(self):
+        self._test_blocklist_header('te', 'trailers')
+
+    def test_blocklist_trailer_header(self):
+        self._test_blocklist_header('trailer', 'x-amz-checksum-sha256')
+
+    def test_blocklist_upgrade_header(self):
+        self._test_blocklist_header('upgrade', 'websocket')
+
```
ubaskota, line 434: Nit: Would parametrized tests work better here, or do you prefer it as is?

jonathan343: I kept individual test methods to match the existing `TestS3SigV4Auth` style and make failures easy to pinpoint by test name. Since this file is unittest-style (not pytest-parametrized), I preferred explicit tests here.

alexgromero (approved):
:shipit:

PR #3648, opened 2026-03-10
Emit after-endpoint-resolution event after endpoint reesolution

aemous:
*Description of changes:*

* Emit an ``after-endpoint-resolution`` event after endpoint resolution is completed. This can be used by downstream applications, such as the AWS CLI v2, to benchmark endpoint resolution time.

*Description of tests:*

* Ran a test script that measures endpoint resolution time, and verified the behavior appears correct:

```python
import time
import botocore.session

start_time = None
end_time = None

def before_handler(builtins, model, params, context, **kwargs):
    global start_time
    if start_time is None:
        start_time = time.time_ns()

def after_handler(**kwargs):
    global end_time
    if end_time is None:
        end_time = time.time_ns()
        print(f'url: {kwargs['provider_result'].url}')


client = session.create_client('s3')

client.meta.events.register_last(
    'before-endpoint-resolution', before_handler, unique_id='benchmark_eresolution'
)
client.meta.events.register_last(
    'after-endpoint-resolution', after_handler, unique_id='benchmark_eresolution_after'
)

response = client.list_objects_v2(
    Bucket='BUCKETNAME--usw2-az1--x-s3',
)

print(f'endpoint resolution time {(end_time - start_time) / 1000} ms')

```

Output:

```
url: https://BUCKETNAME--usw2-az1--x-s3.s3express-usw2-az1.us-west-2.amazonaws.com
endpoint resolution time 225.0 ms
```

### DIFF
--- a/.changes/next-release/enhancement-Events-52626.json
+++ b/.changes/next-release/enhancement-Events-52626.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Events",
+  "description": "Emit an ``after-endpoint-resolution`` event after endpoint resolution is completed."
+}
--- a/botocore/regions.py
+++ b/botocore/regions.py
@@ -494,6 +494,7 @@ class EndpointRulesetResolver:
         request_context,
     ):
         """Invokes the provider with params defined in the service's ruleset"""
+        service_id = self._service_model.service_id.hyphenize()
         if call_args is None:
             call_args = {}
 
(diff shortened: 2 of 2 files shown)


SamRemis:
I’d rather keep the events system focused on hooks that are intended to affect/extend runtime behavior, not primarily for performance measurement.

This PR introduces a new `after-endpoint-resolution` hook and exposes `provider_result`, `operation_model`, and `request_context`. That creates surface areas for users to read and modify those values. It’s not clear from the PR description why consumers of the event need access to those objects.

If the main goal is benchmarking, I’m not convinced an event is the right mechanism. Once we introduce a “benchmark boundary” event, code can be added just before/after emission that will either inflate the measured time (if added before emission) or be missed entirely (if added after emission), so the measurement may drift from “endpoint resolution work” over time.

For botocore’s own benchmarking, we can do this without introducing a new public event by timing the endpoint provider call directly (e.g., wrapping `_ruleset_resolver._provider.resolve_endpoint` in a test using `monkeypatch`/`wraps`). 

I'll leave this open for a bit to see if anyone on the team disagrees, but close it out if I don't hear any other opinions.

PR #3649, opened 2026-03-11
Raise TypeError on bare test decorator usage and add regression tests

Alan4506:
*Background:*
`@skip_if_windows`, `@requires_crt`, and `@skip_if_crt` are decorator factories used to conditionally skip tests. Using them without parentheses (e.g., bare `@skip_if_windows`) is unsafe because Python passes the test function as the first argument, causing the real test body to never run. Assertions are never evaluated, but the test may still appear to pass or skip silently. 

*Description of changes:*
This PR makes bare `@skip_if_windows`, `@requires_crt`, and `@skip_if_crt` fail immediately with a `TypeError`. All existing usages in botocore already use the correct parenthesized form.

*Testing:*
Regression tests are added in `tests/unit/test_test_decorators.py` to verify correct behavior for all three decorators. All tests including the new tests pass.

### DIFF
--- a/tests/unit/test_decorators.py
+++ b/tests/unit/test_decorators.py
@@ -1,3 +1,16 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


tests/unit/test_decorators.py
alexgromero, line 1: One small nit: we might consider renaming `test_test_decorators.py` to just `test_decorators.py` since the `test_test_*` pattern reads a bit odd. Thoughts? 👀

Alan4506: It makes sense to me!

alexgromero, line 1: Let's add the licensing header to this file as well

Issue #3651, opened 2026-03-17
S3 PutObject crashes when using non-seekable stream


jamesls:
### Describe the bug

Passing a non-seekable stream to `put_object` fails with a signature error:

```
$ uv run --with boto3 python -c '
import io
import boto3

class NonSeekableStream(io.RawIOBase):
    def __init__(self, data):
        self._data = data
        self._pos = 0

    def read(self, n=-1):
        if n == -1:
            n = len(self._data) - self._pos
        chunk = self._data[self._pos : self._pos + n]
        self._pos += len(chunk)
        return chunk

    def readable(self):
        return True

s3 = boto3.client("s3")
s3.put_object(
    Bucket="mybucket-test-sync",
    Key="test.txt",
    Body=NonSeekableStream(b"hello world"),
    ContentLength=11,
)

'
Traceback (most recent call last):
  File "<string>", line 21, in <module>
  File ".cache/uv/archive-v0/yqGESA53aCEP9s53_l_I4/lib/python3.12/site-packages/botocore/client.py", line 602, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".cache/uv/archive-v0/yqGESA53aCEP9s53_l_I4/lib/python3.12/site-packages/botocore/context.py", line 123, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File ".cache/uv/archive-v0/yqGESA53aCEP9s53_l_I4/lib/python3.12/site-packages/botocore/client.py", line 1078, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (SignatureDoesNotMatch) when calling the PutObject operation: The request signature we calculated does not match the signature you provided. Check your key and signing method.
```

Compared to a seekable stream which works without issue:

```
$ uv run --with boto3 python -c '
import io
import boto3

s3 = boto3.client("s3")
s3.put_object(
    Bucket="mybucket-test-sync",
    Key="test.txt",
    Body=io.BytesIO(b"hello world"),
    ContentLength=11,
)

'
```

We can see that the SDK is sending an incompatible set of headers in its DEBUG output:

```
Content-Length: 11
Transfer-Encoding: chunked
Content-Encoding: aws-chunked
```

This is explicitly disallowed in the HTTP RFC (7230), section 3.3:

[...]

PR #3652, opened 2026-03-17
Fix aws-chunked Content-Length handling for non-seekable S3 uploads

jonathan343:
## Overview
This PR fixes `s3.put_object()` with a non-seekable `Body` and explicit `ContentLength`.

Previously, the trailer-checksum path could send both `Content-Length` and `Transfer-Encoding: chunked` on the same request. That produced an invalid signed request and could fail with `SignatureDoesNotMatch`.

Addresses https://github.com/boto/botocore/issues/3651

## Summary
- Reuse an existing `Content-Length` header as `X-Amz-Decoded-Content-Length` when the body length cannot be determined directly.
- Always remove `Content-Length` after switching the request to `Transfer-Encoding: chunked`.
- Add unit coverage for non-seekable trailer-checksum requests with an explicit content length.
- Add S3 functional coverage for `put_object(..., Body=NonSeekableStream(...), ContentLength=11)`.

## Testing

```python
import io
import boto3


class NonSeekableStream(io.RawIOBase):
    def __init__(self, data):
        self._data = data
        self._pos = 0

    def read(self, n=-1):
        if n == -1:
            n = len(self._data) - self._pos
        chunk = self._data[self._pos : self._pos + n]
        self._pos += len(chunk)
        return chunk

    def readable(self):
        return True


s3 = boto3.client("s3")
s3.put_object(
    Bucket="my-test-bucket",
    Key="test.txt",
    Body=NonSeekableStream(b"hello world"),
    ContentLength=11,
)
```

Before these changes, the script above produced a signature error similar to the following:
```
botocore.exceptions.ClientError: An error occurred (SignatureDoesNotMatch) when calling the PutObject operation: The request signature we calculated does not match the signature you provided. Check your key and signing method.
```
After these changes, the script successfully uploads the content from the non-seekable stream. 

## Additional Notes
This change fixes header and signing correctness for the initial request. Retry and redirect behavior for unseekable aws-chunked bodies is unchanged. Since we can't rewind, the SDK will produce an exception similar to the one below:

```
    raise UnseekableStreamError(stream_object=self.body)
botocore.exceptions.UnseekableStreamError: Need to rewind the stream <botocore.httpchecksum.AwsChunkedWrapper object at 0x10a3933b0>, but stream is not seekable.
```

---

### DIFF
--- a/botocore/httpchecksum.py
+++ b/botocore/httpchecksum.py
@@ -473,6 +473,10 @@ def _apply_request_trailer_checksum(request):
 
     content_length = determine_content_length(body)
     if content_length is None and "Content-Length" in headers:
+        # determine_content_length() cannot resolve the length of non-seekable
+        # bodies, but the caller may have set Content-Length explicitly. Reuse
+        # that value for X-Amz-Decoded-Content-Length before the header is
+        # removed for chunked transfer encoding.
         content_length = int(headers["Content-Length"])
     if content_length is not None:
         # Send the decoded content length if we can determine it. Some


botocore/httpchecksum.py
```diff
@@ -472,16 +472,18 @@ def _apply_request_trailer_checksum(request):
     headers["X-Amz-Trailer"] = location_name
 
     content_length = determine_content_length(body)
+    if content_length is None and "Content-Length" in headers:
```
SamRemis, line 475: [Non-blocking, question]

Why bother with this if statement?  If we weren't using the header's value before, couldn't we just move the if statement from below (line 482) to where this if statement is?

Is this just so we can attempt to set the decoded content length if it's present as a best-effort?  This makes sense, just want to confirm that I understand the reasoning here.

jonathan343: Honestly, I'm not following the question. The new branch is there so non-seekable bodies can still get `X-Amz-Decoded-Content-Length` from an already-present `Content-Length` header before we remove it. Simply moving the delete block earlier would drop the header without preserving that value. 

Note:
> For all requests, you must include the `x-amz-decoded-content-length` header, specifying the size of the object in bytes.

Docs: https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html

SamRemis: That's probably because my question didn't make a ton of sense - I had forgotten that users can set the ContentLength as an input on an operation. Without that input, it's not likely that this header will be set for a non-seekable stream, so it wouldn't make as much sense to try looking in the header.

This code block is just making sure we respect that input - thank you for the explanation.

PR #3678, opened 2026-04-21
Implement updated retry behavior gated behind opt-in environment variable

ubaskota:
⚠️ This change is internal-only. External users are not affected and should not depend on the gated behavior. 

*Description of changes:*
Implements the updated retry behavior for standard retry mode. Changes include updated exponential backoff with service specific scale factors, revised retry quota costs (14 for non-throttling, 5 for throttling), DynamoDB-specific max attempts (4) and base backoff (25ms), long-polling operation backoff on quota exhaustion, and `x-amz-retry-after` header support. All changes are gated behind a `DEFAULT_NEW_RETRIES` flag imported from an internal customization, ensuring no behavior change for external users. The default retry mode is also updated to `standard` for internal users via the same flag.

The goal of this is to test new retries behavior internally before releasing it to external customers.  To test this out prior to release, please set the `AWS_NEW_RETRIES_2026` environment variable to `true`.  

*Tests:*
- Added tests to verify the updated retry behavior behind the gated flag works as expected. All new tests pass.               
- Verified that existing tests for the current retry behavior continue to pass. 

### DIFF
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -21,19 +21,22 @@ import os
 from botocore import utils
 from botocore.exceptions import InvalidConfigError
 
-# This is not a public interface and is subject to abrupt breaking changes.
-# Currently it's only available to internal users for testing and validation.
-# Any usage is not advised or supported in external code bases.
 try:
-    from botocore.customizations.retries import STANDARD_RETRY_MODE_VERSION
+    # This is not a public interface and is subject to abrupt breaking changes.
+    # Currently it's only available to internal users for testing and validation.
+    # Any usage is not advised or supported in external code bases.
+    from botocore.customizations.retries import DEFAULT_NEW_RETRIES
 except ImportError:
-    STANDARD_RETRY_MODE_VERSION = "2.0"
+    DEFAULT_NEW_RETRIES = False
+
+_env_new_retries = os.environ.get('AWS_NEW_RETRIES_2026')
+if _env_new_retries is not None:
+    NEW_RETRIES_ENABLED = _env_new_retries.lower() == 'true'
+else:
+    NEW_RETRIES_ENABLED = DEFAULT_NEW_RETRIES
+_DEFAULT_RETRY_MODE = 'standard' if NEW_RETRIES_ENABLED else 'legacy'
 
 logger = logging.getLogger(__name__)
-_DEFAULT_RETRY_MODE = (
-    'standard' if STANDARD_RETRY_MODE_VERSION == "2.1" else 'legacy'
-)
-
 
 #: A default dictionary that maps the logical names for session variables
 #: to the specific environment variables and configuration file names
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -21,7 +21,18 @@ import os
 from botocore import utils
 from botocore.exceptions import InvalidConfigError
 
+# This is not a public interface and is subject to abrupt breaking changes.
+# Currently it's only available to internal users for testing and validation.
+# Any usage is not advised or supported in external code bases.
+try:
+    from botocore.customizations.retries import STANDARD_RETRY_MODE_VERSION
+except ImportError:
+    STANDARD_RETRY_MODE_VERSION = "2.0"
+
 logger = logging.getLogger(__name__)
+_DEFAULT_RETRY_MODE = (
+    'standard' if STANDARD_RETRY_MODE_VERSION == "2.1" else 'legacy'
+)
 
 
 #: A default dictionary that maps the logical names for session variables
@@ -150,7 +161,7 @@ BOTOCORE_DEFAUT_SESSION_VARIABLES = {
         'regional',
         None,
     ),
-    'retry_mode': ('retry_mode', 'AWS_RETRY_MODE', 'legacy', None),
+    'retry_mode': ('retry_mode', 'AWS_RETRY_MODE', _DEFAULT_RETRY_MODE, None),
     'defaults_mode': ('defaults_mode', 'AWS_DEFAULTS_MODE', 'legacy', None),
     # We can't have a default here for v1 because we need to defer to
     # whatever the defaults are in _retry.json.
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -34,8 +34,7 @@ def _resolve_new_retries():
     _env_new_retries = os.environ.get('AWS_NEW_RETRIES_2026')
     if _env_new_retries is not None:
         return _env_new_retries.lower() == 'true'
-    else:
-        return DEFAULT_NEW_RETRIES
+    return DEFAULT_NEW_RETRIES
 
 
 NEW_RETRIES_ENABLED = _resolve_new_retries()
--- a/botocore/configprovider.py
+++ b/botocore/configprovider.py
@@ -29,11 +29,16 @@ try:
 except ImportError:
     DEFAULT_NEW_RETRIES = False
 
-_env_new_retries = os.environ.get('AWS_NEW_RETRIES_2026')
-if _env_new_retries is not None:
-    NEW_RETRIES_ENABLED = _env_new_retries.lower() == 'true'
-else:
-    NEW_RETRIES_ENABLED = DEFAULT_NEW_RETRIES
+
+def _resolve_new_retries():
+    _env_new_retries = os.environ.get('AWS_NEW_RETRIES_2026')
+    if _env_new_retries is not None:
+        return _env_new_retries.lower() == 'true'
+    else:
+        return DEFAULT_NEW_RETRIES
+
+
+NEW_RETRIES_ENABLED = _resolve_new_retries()
 _DEFAULT_RETRY_MODE = 'standard' if NEW_RETRIES_ENABLED else 'legacy'
 
 logger = logging.getLogger(__name__)
--- a/botocore/retries/standard.py
+++ b/botocore/retries/standard.py
@@ -53,17 +53,20 @@ _SERVICE_MAX_ATTEMPTS = {
 logger = logging.getLogger(__name__)
 
 
-def register_retry_handler(client, max_attempts=DEFAULT_MAX_ATTEMPTS):
+def register_retry_handler(client, max_attempts=None):
     service_id = client.meta.service_model.service_id
     service_event_name = service_id.hyphenize()
+    retry_event_adapter = RetryEventAdapter()
 
     if STANDARD_RETRY_MODE_VERSION == "2.1":
         if (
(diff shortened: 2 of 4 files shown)


botocore/retries/standard.py
```diff
@@ -36,28 +37,66 @@
...
 
 def register_retry_handler(client, max_attempts=DEFAULT_MAX_ATTEMPTS):
-    retry_quota = RetryQuotaChecker(quota.RetryQuota())
-
     service_id = client.meta.service_model.service_id
     service_event_name = service_id.hyphenize()
+
+    if STANDARD_RETRY_MODE_VERSION == "2.1":
+        if (
+            max_attempts == DEFAULT_MAX_ATTEMPTS
```
SamRemis: [Missing functionality, change requested]
If a customer wants to set max attempts for dynamodb to 3, how would they do that here?  

Since it's the same as default max attempts, we don't know if it was explicitly set by the customer or if it was the default.  We could probably change the function signature to default to None, and set max attempts at the top of the function.

That should work, but if it doesn't for some reason, we may need to do another [config wrapper](https://github.com/boto/botocore/blob/0db98ca4a6b8b783a0e1f17f767c288a53bd90c1/botocore/args.py#L1043-L1051).

botocore/retries/standard.py
```diff
@@ -74,10 +113,25 @@ class RetryHandler:
     as an event handler.
     """
 
-    def __init__(self, retry_policy, retry_event_adapter, retry_quota):
+    _LONG_POLLING_OPERATIONS = {
```
SamRemis, line 118: [Change requested]

Since this will ultimately be a modeled trait, can we add a code comment here saying that this is temporary and will be replaced by a modeled trait.

botocore/retries/standard.py
```diff
@@ -93,6 +147,12 @@ def needs_retry(self, **kwargs):
                     retry_delay,
                 )
             else:
+                if STANDARD_RETRY_MODE_VERSION == "2.1":
+                    if self._is_long_polling_operation(context):
+                        polling_delay = self._retry_policy.compute_retry_delay(
+                            context
+                        )
+                        self._sleep(polling_delay)
```
SamRemis, line 157: [Question, bug if I'm right]

This sleeps at the proper time, but how are we preventing a double sleep when we hit this code path?  Won't we sleep once here, then again [here](https://github.com/boto/botocore/blob/7ebb8b8f04a355e08f90e77637b61b5bf43b1770/botocore/endpoint.py#L379)?  

Even if we are returning `None`, we still could wind up getting a non-None response from a later handler on needs-retry which [would get used instead](https://github.com/boto/botocore/blob/7ebb8b8f04a355e08f90e77637b61b5bf43b1770/botocore/endpoint.py#L369), even though we have already done the sleep.

botocore/retries/standard.py
```diff
@@ -252,11 +323,25 @@ def compute_retry_delay(self, context):
 class ExponentialBackoff(BaseRetryBackoff):
     _BASE = 2
     _MAX_BACKOFF = 20
+    _THROTTLING_BASE_SCALE = 1
+    _NON_THROTTLING_BASE_SCALE = 0.05
+    _DYNAMODB_BASE_SCALE = 0.025
```
SamRemis: [Change requested]

Can we make this more generic with a variable overrides dictionary so it's extensible?

I'm thinking something like this, but this is just for demonstration, it needs to be improved upon: 

```
_DYNAMODB_OVERRIDES = {"non-throttling-base-scale": 0.025}
#maybe this should be not specific to base scales for extensibility?
_BASE_SCALES  = {
    'dynamodb': _DYNAMODB_OVERRIDES,
    'dynamodb-streams': _DYNAMODB_OVERRIDES
    'default': {'non-throttling-base-scale': 0.05, 'throttling-base-scale': 0.025 }
}
```

Basically the idea is that right now we are only overriding one thing for one service, but in my mind it makes sense to group these overrides by service so we can easily see how each service is behaving.

SamRemis (changes requested):

botocore/retries/standard.py
```diff
@@ -36,28 +37,66 @@
...
     service_id = client.meta.service_model.service_id
     service_event_name = service_id.hyphenize()
+
+    if STANDARD_RETRY_MODE_VERSION == "2.1":
+        if (
+            max_attempts == DEFAULT_MAX_ATTEMPTS
+            and service_event_name in _SERVICE_MAX_ATTEMPTS
+        ):
+            max_attempts = _SERVICE_MAX_ATTEMPTS[service_event_name]
+        throttling_detector = ThrottlingErrorDetector(RetryEventAdapter())
```
SamRemis: [Nit, non blocking]

This is fine, but we could reuse the RetryEventAdapter on line 80 and only initialize one.  It's stateless, so we can just reuse it as input to both the `ThrottlingErrorDetector` and the `RetryHandler`.

botocore/retries/standard.py
ubaskota: Updated to return `False` after sleeping. Since `False` is `non-None`, `first_non_none_response` picks it up and prevents any later handler from sleeping again. Updated `endpoint.py` to treat `False` the same as `None` (don't retry, don't sleep).

botocore/retries/standard.py
```diff
@@ -272,10 +380,60 @@ def delay_amount(self, context):
...
+    def _get_retry_after_delay(self, context):
+        if context.http_response is None:
+            return None
+        retry_after_ms = context.http_response.headers.get(
+            self._RETRY_AFTER_HEADER
         )
+        if retry_after_ms is None:
+            return None
+        try:
+            return int(retry_after_ms) / 1000.0
```
SamRemis: [Missing corner case]

It may make sense to safeguard against values like negatives here so we don't blindly accept what the service sends.  This has to be a non-negative integer.

ubaskota: The SEP says `x-amz-retry-after` will always be a positive integer, so I didn't add a safeguard. But I agree that we should check for valid values. Added.

botocore/configprovider.py
```diff
@@ -21,8 +21,22 @@
...
+    from botocore.customizations.retries import DEFAULT_NEW_RETRIES
+except ImportError:
+    DEFAULT_NEW_RETRIES = False
+
+_env_new_retries = os.environ.get('AWS_NEW_RETRIES_2026')
+if _env_new_retries is not None:
+    NEW_RETRIES_ENABLED = _env_new_retries.lower() == 'true'
+else:
+    NEW_RETRIES_ENABLED = DEFAULT_NEW_RETRIES
+_DEFAULT_RETRY_MODE = 'standard' if NEW_RETRIES_ENABLED else 'legacy'
```
SamRemis:
```suggestion
def _resolve_new_retries():
    if env_value := os.environ.get('AWS_NEW_RETRIES_2026'):
        return env_value.lower() == 'true'
    return DEFAULT_NEW_RETRIES
  
NEW_RETRIES_ENABLED = _resolve_new_retries()
```

tests/unit/retries/test_standard_retry_v2_1.py
```diff
@@ -0,0 +1,446 @@
...
+    @mock.patch.dict(os.environ, {'AWS_NEW_RETRIES_2026': 'false'})
+    def test_env_var_false_disables_new_retries(self):
+        importlib.reload(configprovider)
+        self.assertFalse(configprovider.NEW_RETRIES_ENABLED)
+
+    @mock.patch.dict(os.environ, {}, clear=False)
+    def test_no_env_var_uses_default(self):
+        os.environ.pop('AWS_NEW_RETRIES_2026', None)
+        importlib.reload(configprovider)
+        self.assertFalse(configprovider.NEW_RETRIES_ENABLED)
```
SamRemis:
```suggestion
 class TestNewRetriesEnvironmentVariable(unittest.TestCase):
     @mock.patch.dict(os.environ, {'AWS_NEW_RETRIES_2026': 'true'})
     def test_env_var_true_enables_new_retries(self):
         self.assertTrue(configprovider._resolve_new_retries())

     @mock.patch.dict(os.environ, {'AWS_NEW_RETRIES_2026': 'false'})
     def test_env_var_false_disables_new_retries(self):
         self.assertFalse(configprovider._resolve_new_retries())

     def test_no_env_var_uses_default(self):
         with mock.patch.dict(os.environ, {}, clear=False):
             os.environ.pop('AWS_NEW_RETRIES_2026', None)
             self.assertFalse(configprovider._resolve_new_retries())
```
Fact check me here, but I think this will fix the failing CI along with my other suggestion since this doesn't rely on the global module state.

SamRemis (changes requested):

tests/unit/retries/test_standard_retry_v2_1.py
```diff
@@ -0,0 +1,446 @@
...
+import os
+from collections import Counter
+
+import pytest
+
+from botocore import configprovider
+from botocore.awsrequest import AWSResponse
+from botocore.exceptions import ReadTimeoutError
+from botocore.retries import quota, standard
+from tests import mock, unittest
```
SamRemis:
```suggestion
from tests import BaseEnvVar, mock, unittest
```

tests/unit/retries/test_standard_retry_v2_1.py
SamRemis:
```suggestion
class TestNewRetriesEnvironmentVariable(BaseEnvVar):
    def test_env_var_true_enables_new_retries(self):
        self.environ['AWS_NEW_RETRIES_2026'] = 'true'
        self.assertTrue(configprovider._resolve_new_retries())

    def test_env_var_false_disables_new_retries(self):
        self.environ['AWS_NEW_RETRIES_2026'] = 'false'
        self.assertFalse(configprovider._resolve_new_retries())

    def test_no_env_var_uses_default(self):
        self.assertFalse(configprovider._resolve_new_retries())
```

tests/unit/retries/test_standard_retry_v2_1.py
```diff
@@ -0,0 +1,446 @@
...
+
+These tests validate the updated retry behavior for standard retries,
+which is currently gated behind an internal flag and not yet available
+to external users. Once the changes are validated internally and released
+publicly, these tests will replace the corresponding tests in
+test_standard.py and this file will be removed.
+"""
+
+import importlib
+import os
```
SamRemis:
```suggestion
```

SamRemis (changes requested):
My bad, it still wasn't working.  Found some similar tests that use BaseEnvVar, and confirmed [this worked on my fork](https://github.com/SamRemis/botocore/pull/4/checks).

SamRemis (approved):

PR #3679, opened 2026-04-22
Deprecate Python 3.9 support

Alan4506:
This PR contains the changes that will go out with our release on 2026-04-29 to finalize our Python 3.9 support deprecation. As originally [announced](https://aws.amazon.com/blogs/developer/python-support-policy-updates-for-aws-sdks-and-tools/) in February 2022, we have followed a 6 month extended support model after the official end of support of a minor Python version. Afterwards, existing SDKs will continue to function as they have previously.

Customers wishing to upgrade to a version with newer features will need to migrate to Python 3.10 or later. No action should be needed as part of this deprecation, `pip` will manage version selection for you.


alexgromero (approved):

jonathan343 (changes requested):
Thanks Yuxuan, this looks good overall, just one issue:

This PR was generated before we updated our ruff version to v0.15.11. Instead, it was generated using v0.12.0 and addresses occurrences of `UP038` in the following files:

- compress.py
- httpchecksum.py
- httpsession.py
- paginate.py
- parsers.py
- serialize.py
- utils.py
- validate.py

This rule was removed in [v0.13.0](https://docs.astral.sh/ruff/rules/non-pep604-isinstance/#removed) for the following reason:

> This rule was removed as using [PEP 604](https://peps.python.org/pep-0604/) syntax in isinstance and issubclass calls isn't recommended practice, and it incorrectly suggests that other typing syntaxes like [PEP 695](https://peps.python.org/pep-0695/) would be supported by isinstance and issubclass. Using the [PEP 604](https://peps.python.org/pep-0604/) syntax is also slightly slower.

Can you rebase and regenerate these changes using ruff v0.15.11. This should revert the changes to those files.

jonathan343 (approved):
Thanks @Alan4506, looks good to me!

Approving, but please do not merge until 4/29.

PR #3686, opened 2026-04-23
Add changelog validation unit test

SamRemis:
*Description of changes:*
Add a unit test to validate that changelog entries are in the correct format.  


.github/workflows/validate-changelog.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: |
+          errors=0
+          for file in .changes/next-release/*; do
+            [ ! -f "$file" ] && continue
+            if ! jq -e '
+              type == "object" and
+              (keys | length) == 3 and
+              (.type | type) == "string" and
+              (.category | type) == "string" and
+              (.description | type) == "string" and
```
ubaskota: What happens if `category` or `description` is an empty string? Should we add a check to match the [validation](https://github.com/boto/botocore/blob/b88d178dd9bd410769f72e45f72e4d8e0716b2c4/scripts/new-change#L100) in the `new-change` script?

SamRemis: Good call, I made some updates to catch this scenario

ubaskota (approved):

ubaskota (approved):

PR #3687, opened 2026-04-24
Remove list_available_services from client creation happy path

SamRemis:
*Description of changes:*
For all service creation, we have a cached `list_available_services` method that we call, but the output is only used if we encounter an error during client creation.  This iterates through multiple directories to get all available service names, including the botocore/data directory with 424 seconds. 

This PR removes the call to this method if the service is able to be found.  Removing this step removes an average of 45.5ms for successful s3 client creation when run locally on my machine.  The performance improvement of this will vary between machines.

```
import time
import statistics
import botocore.session

times = []
for _ in range(500):
    start = time.perf_counter()
    session.create_client('s3', region_name='us-east-1')
    times.append((time.perf_counter() - start) * 1000)

print(f"Average:     {statistics.mean(times):.1f}ms")
p95_index = int(len(times) * 0.95)
print(f"P95:      {sorted(times)[p95_index]:.1f}ms")

# Before:
# Average:     626.4ms
# P95:      838.0ms

# After:
# Average:     580.9ms
# P95:      778.4ms
```

### DIFF
--- a/.changes/next-release/enhancement-Session-37279.json
+++ b/.changes/next-release/enhancement-Session-37279.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "``Session``",
+  "description": "Removes list_available_services from the happy path for client creation to save time on client creation"
+}
--- a/botocore/client.py
+++ b/botocore/client.py
@@ -127,7 +127,7 @@ class ClientCreator:
                 service_name, api_version
             )
             partition_data = self._loader.load_data('partitions')
-        except UnknownServiceError:
+        except (UnknownServiceError, DataNotFoundError):
             endpoints_ruleset_data = None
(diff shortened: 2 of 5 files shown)


jonathan343 (commented):
One edge I think we're breaking with this:

We currently have logic that catches an `UnknownServiceError` exception thrown when we attempt to load service endpoints and partitions files. In this case, we will fallback to using legacy endpoints. 

Problem: If a customer has custom models (pre-EP2.0 migration) and supplies a `api_version` parameter, we'll start throwing a `DataNotFoundError` which isn't caught and we'll start breaking customers for this scenario. 

https://github.com/boto/botocore/blob/61b379e9b0672aeba118030752b68272108dd1d7/botocore/client.py#L125-L137

PR #3688, opened 2026-04-24
Update vendored CA bundle

BurtPer:
*Description of changes:*

The vendored cacert.pem bundle seems a bit out of date. This brings it in line with the current Mozilla root store, matching what [certifi 2026.4.22](https://pypi.org/project/certifi/2026.4.22/) ships today.

This also removes expired root certificates that are no longer functional.


BurtPer:
@jonathan343 @aemous, you've had recent commits. Apologies if you are not the correct parties for review. Please take a look when you can.

SamRemis:
Hi @BurtPer,

Thank you for the PR - we cannot always use certifi's bundle, since some AWS services still use one of the certificates that certifi has removed on some legacy endpoints.  We can make small, more targeted updates like I have done in #3690.  

There's more updates that can be safely made to this file, but unfortunately we cannot accept a full rewrite.  Because of this, I'm closing this PR.

BurtPer:
Hi @SamRemis,

> we cannot always use certifi's bundle, since some AWS services still use one of the certificates that certifi has removed on some legacy endpoints.

Can you please be more specific with what you mean here? What's prohibiting you from porting the required certificate to a newer bundle?

SamRemis:
Sure - [this comment](https://github.com/boto/boto3/issues/2686#issuecomment-742579376) is the closest I can find to public info on this.  

Checking with `echo | openssl s_client -connect data.iot.eu-west-2.amazonaws.com:443 2>/dev/null | grep -A 20 "Certificate chain"`, this endpoint still uses [this cert](https://github.com/certifi/python-certifi/commit/45a64658872a94a83c4b70fce02a96f0f29895e6#diff-e1f673cf40a49b35c77f0550d8d3f940471f098e8e102bccc0ad72d05cf13d87L841-L875) as the root signing certificate.

SamRemis:
As for this:
> What's prohibiting you from porting the required certificate to a newer bundle?

We have no way of detecting which of the other removed certificates can be removed safely and which cannot, so we remove them on a case-by-case basis as we're able to verify they won't break existing AWS endpoints.

BurtPer:
> We have no way of detecting which of the other removed certificates can be removed safely and which cannot, so we remove them on a case-by-case basis as we're able to verify they won't break existing AWS endpoints.

@SamRemis this and #3690 are both deeply concerning coming as comment from a company like AWS. Have you consulted with AWS' security team about this? What I'm reading here in that the boto team does not have enough understanding of the product to make critical security decisions.

PR #3692, opened 2026-04-28
Improving caching of S3 endpoints by omitting unused parameters

ashovlin:
*Issue #, if available:* #3528, variation of PR https://github.com/boto/botocore/pull/3529, backport from https://github.com/aws/aws-cli/pull/10165

*Description of changes:* When resolving the endpoint for S3 operations, currently `{'Key', 'Prefix', 'CopySource'}` are defined as parameters in S3's rules, but aren't currently referenced. This is an issue for endpoint caching, since objects in the same bucket result in a unique set of endpoint parameters. This slows down repeated operations against objects in the same bucket, we re-resolve the endpoint each time.

Now, we drop these three parameters specifically before passing them into the rules. This comes with a test that will begin failing when the rules start referencing these parameters (or add new, unused parameters).

We believe this is safe for botocore, these extra parameters are in the rules for a plugin in a different SDK.


tests/functional/test_endpoint_rulesets.py
```diff
@@ -305,3 +308,34 @@ def builtin_overwriter_handler(builtins, **kwargs):
...
+            refs.add(rules['ref'])
+        for value in rules.values():
+            refs.update(_collect_refs_from_rules(value))
+    elif isinstance(rules, list):
+        for item in rules:
+            refs.update(_collect_refs_from_rules(item))
+    return refs
+
+
+@pytest.mark.validates_model
```
jonathan343: This should be `validates_models` not `validates_model`

```suggestion
@pytest.mark.validates_models
```

jonathan343 (changes requested):
Thanks @ashovlin! Looks good to me outside of one typo.

jonathan343 (approved):

PR #3697, opened 2026-05-04
Add public chain manipulation methods to TokenProviderChain

ubaskota:
*Issue #, if available:*
https://github.com/boto/boto3/issues/4723
https://github.com/langchain-ai/langchain-aws/issues/989
https://github.com/strands-agents/sdk-python/issues/1238

*Description of changes:*
Add `insert_before`, `insert_after`, `remove`, and `get_provider` methods to `TokenProviderChain`, mirroring the existing `CredentialResolver` API. This enables runtime manipulation of the token provider chain, allowing users to inject custom token providers at specific positions without relying on environment variables. Also adds `UnknownTokenProviderError` for when a referenced provider name doesn't exist in the chain, and makes `providers` a public attribute.

*Testing*
- Verified that the newly added tests pass 
- Manually verified end-to-end that a custom token provider injected via `insert_before` correctly resolves and produces a Bearer authorization header on outgoing requests
- Verified that all existing tests pass 

### DIFF
--- a/botocore/tokens.py
+++ b/botocore/tokens.py
@@ -27,6 +27,7 @@ from botocore.exceptions import (
     ClientError,
     InvalidConfigError,
     TokenRetrievalError,
+    UnknownTokenProviderError,
 )
 from botocore.utils import (
     CachedProperty,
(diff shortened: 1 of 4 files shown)


Alan4506 (approved):
LGTM!

botocore/tokens.py
```diff
@@ -167,10 +168,74 @@ class TokenProviderChain:
     def __init__(self, providers=None):
         if providers is None:
             providers = []
-        self._providers = providers
+        self.providers = providers
```
jonathan343: I think we should stick with `self._providers`. Historically, we have not exposed a public way for customers to hook into the token provider chain. Their only option was to hook into internals. Although this goes against the implied API contract, I still think we should be considerate and avoid breaking them if we can.

For example, LangChain accesses `._providers` directly:

https://github.com/langchain-ai/langchain-aws/blob/ab6b6f4f4fa48f921848dd255234b97afd683d37/libs/aws/langchain_aws/utils.py#L213

jonathan343 (changes requested):
Thanks Ujjwal! I have one request to avoid breaking customers currently hooking into the token provider chain

jonathan343 (commented):
Awesome, looks good! Last thing before I approve:

Can you add a changelog entry similar to:

```json
  {
    "type": "enhancement",
    "category": "Identity",
    "description": "Add public methods to insert, remove, and retrieve providers in the token provider chain."
  }
```

jonathan343 (approved):
LGTM, thanks @ubaskota!

PR #3698, opened 2026-05-04
Use an existing function to calculate offset

ubaskota:
*Issue #, if available:*

*Description of changes:*
Use the existing `_get_provider_offset` function to calculate offset in `insert_before` operation.


arandito (approved):
:shipit:

PR #3701, opened 2026-05-08
Add Free Threading support to CI

nateprewitt:
*Description of changes:*
The primary blocker for supporting [free threading](https://peps.python.org/pep-0779/) in Boto3/Botocore/Smithy Python was support in the upstream dependency, awscrt. Starting in awscrt 0.32.0 (https://github.com/awslabs/aws-crt-python/pull/725), releases began distributing artifacts for Python 3.13t and 3.14t. While 3.13t was released as experimental, 3.14t is starting to gain adoption.

Given there are no remaining blockers, this PR adds beta support for free threaded builds and makes sure that contract is validated in CI.

### DIFF
--- a/tests/functional/leak/test_resource_leaks.py
+++ b/tests/functional/leak/test_resource_leaks.py
@@ -10,6 +10,8 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import sysconfig
+
 from tests import BaseClientDriverTest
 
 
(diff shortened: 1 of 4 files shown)


tests/functional/leak/test_resource_leaks.py
```diff
@@ -22,10 +22,9 @@ class TestDoesNotLeakMemory(BaseClientDriverTest):
     # a substantial amount of time to the total test run time.
     INJECT_DUMMY_CREDS = True
     # We're making up numbers here, but let's say arbitrarily
-    # that the memory can't increase by more than 13MB.
+    # that the memory can't increase by more than 16MB.
 
-    # TODO: Attempt to bring this back to 10MB once Python 3.14 releases
```
nateprewitt, line 27: One thing that never got revisited after 3.14 released, is this test. Memory usage changed in 3.14 which raised our threshold. Free-Threading represents objects differently which increases memory usage ~15-20% in this case. It also is slower to trigger GC.

There are two options for this change, either increase the threshold to 16 (tests currently reach ~15.1), or use a different growth constant for free-threaded builds. That can be accomplished with something like:

```python
if sysconfig.get_config_var("Py_GIL_DISABLED") == 1:
    SCALING_FACTOR = 16
else:
    SCALING_FACTOR = 13
MAX_GROWTH_BYTES = SCALING_FACTOR * 1024 * 1024
```

jonathan343: Thanks Nate! I prefer keeping separate `SCALING_FACTOR`'s for GIL and no-GIL. I think a shared `SCALING_FACTOR = 16` would be fine, but if for some reason there is ever an increase in no-GIL and not with GIL, we'd want to catch that.

nateprewitt: GitHub's struggling with macOS runners right now but this passed on my fork, so I think [e18317f](https://github.com/boto/botocore/pull/3701/commits/e18317f23fe7a25379d6bdcfe2e1563fbd8df01b) should address this.

jonathan343 (approved):

PR #3703, opened 2026-05-12
Add MD5 checksum algorithm support

shreyaskommuri:
## Summary

Fixes `FlexibleChecksumError: Unsupported checksum algorithm: md5` when callers pass `ChecksumAlgorithm=MD5` to S3 operations.

MD5 is a valid S3 checksum algorithm but was not registered in `_CHECKSUM_CLS`, so botocore raised an error at the checksum computation stage even when the algorithm was accepted by the service model.

**Changes:**
- Import `md5` from `hashlib`
- Add `Md5Checksum` class following the same pattern as `Sha1Checksum` and `Sha256Checksum`
- Register `"md5": Md5Checksum` in `_CHECKSUM_CLS`
- Add `Md5Checksum` to the `_CHECKSUM_DIGEST_CASES` parametrized test list

## Related

Companion PR in aws/aws-cli adds `MD5` to the `--checksum-algorithm` choices for high-level `s3` commands.

Relates to aws/aws-cli#10295

## Test plan

- [x] `test_checksum_digest[Md5Checksum-XrY7u+Ae7tCTyyK7j1rNww==]` passes
- [x] All existing `_CHECKSUM_DIGEST_CASES` tests continue to pass

Developed with AI assistance, reviewed and tested by shreyaskommuri

### DIFF
--- a/botocore/httpchecksum.py
+++ b/botocore/httpchecksum.py
@@ -23,7 +23,7 @@ import base64
 import io
 import logging
 from binascii import crc32
-from hashlib import sha1, sha256, sha512
+from hashlib import md5, sha1, sha256, sha512
 
 from botocore.compat import HAS_CRT, has_minimum_crt_version, urlparse
 from botocore.exceptions import (
@@ -100,6 +100,16 @@ class CrtCrc32Checksum(BaseChecksum):
 
     def digest(self):
         return self._int_crc32.to_bytes(4, byteorder="big")
+
+class Md5Checksum(BaseChecksum):
+    def __init__(self):
+        self._checksum = md5()
+
+    def update(self, chunk):
+        self._checksum.update(chunk)
+
+    def digest(self):
+        return self._checksum.digest()
 
 
 class CrtCrc32cChecksum(BaseChecksum):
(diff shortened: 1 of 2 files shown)


SamRemis:
Hi @shreyaskommuri,

The absence of automatic calculation for MD5 was an intentional choice by AWS.  Customers looking to use the MD5 checksum are required to pass it in the pre-calculated parameter on upload using the `--checksum-md5` input parameter.  Any APIs which do not support this parameter do not support MD5 using the SDKs or the CLI.  

AWS generally avoids choosing MD5 for newer automatic checksum calculation because MD5 is considered a legacy checksum. It is still supported in older flows where it already existed, mainly for backward compatibility, but it is not the preferred direction for newer checksum mechanisms.

For newer S3 checksum behavior, AWS prefers algorithms like CRC32, CRC32C, CRC64NVME, SHA1, or SHA256 because they fit better with modern integrity checking, performance, and multipart upload behavior. CRC-based algorithms are especially useful at S3 scale because they can be efficiently combined across parts and are faster for large object workflows.  

I will close this PR to keep further discussion on the issue that you linked.

PR #3705, opened 2026-05-14
Update resource leak test scaling factors and CI Python resolution

jonathan343:
## Summary
Python 3.14.5 [reverted the incremental garbage collector](https://www.python.org/downloads/release/python-3145/) that was introduced in 3.14.0, restoring the generational GC from 3.13 due to reports of significant memory pressure in production environments.

The `MAX_GROWTH_BYTES` scaling factor in our resource leak tests had been bumped from 10 MB to 12 MB and then 13 MB to absorb the additional memory growth caused by the incremental GC (see #3522, which added a TODO to lower it once 3.14 stabilized). With the revert in 3.14.5, that headroom is no longer needed.

This PR includes the following changes:

- Drops the standard (non-free-threaded) `SCALING_FACTOR` from 13 to 4, giving us tighter regression detection.
- Bumps the free-threaded `SCALING_FACTOR` from 16 to 17 after observing repeated CI failures at 16. Per-object locking in free-threaded builds still requires the additional headroom.
- Adds `check-latest: true` to the `setup-python` step in `run-tests.yml` and `run-crt-test.yml` so CI always resolves to the latest available patch release rather than a cached older version, ensuring we pick up fixes like the 3.14.5 GC revert promptly.

---

### DIFF
--- a/.github/workflows/run-crt-test.yml
+++ b/.github/workflows/run-crt-test.yml
@@ -34,6 +34,7 @@ jobs:
           python-version: '${{ matrix.python-version }}'
           cache: 'pip'
           allow-prereleases: true
+          check-latest: true
       - name: Install dependencies and CRT
         run: |
           python scripts/ci/install --extras crt
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -34,6 +34,7 @@ jobs:
           python-version: '${{ matrix.python-version }}'
           cache: 'pip'
           allow-prereleases: true
+          check-latest: true
       - name: Install dependencies
         run: |
           python scripts/ci/install
--- a/tests/functional/leak/test_resource_leaks.py
+++ b/tests/functional/leak/test_resource_leaks.py
@@ -31,7 +31,7 @@ class TestDoesNotLeakMemory(BaseClientDriverTest):
         # memory for per-object locks and synchronization.
         SCALING_FACTOR = 16
     else:
-        SCALING_FACTOR = 13
+        SCALING_FACTOR = 4
     MAX_GROWTH_BYTES = SCALING_FACTOR * 1024 * 1024
 
     def test_create_single_client_memory_constant(self):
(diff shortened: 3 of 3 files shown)


jonathan343:
Hmm, builds are failing for 3.14 since `setup-python` has cached 3.14.4 and not finding the newer 3.14.5. They do have a `check-latest` option which I can toggle to `true`, however, this hurts performance and we rarely NEED the latest.

I'll look into what the cache TTL is to see how long we need to wait or if there are other options.

jonathan343:
Update: turns out there's no TTL like I though. `actions/setup-python` resolves to whatever's pre-installed on the runner image, and that only changes when GitHub publishes a new runner image (cadence is roughly weekly). With the 3.14.5 GC revert being the whole point of this PR, waiting indefinitely for the runner image to roll forward isn't ideal, and it would also mean every other PR's leak test breaks against cached 3.14.4 once this merges.

I decided to flip `check-latest: true` after all. I don't think the added time for installing the latest hurts us much and this only applies in the window between when a new patch is released and the runner image isn't updated.

Happy to revert and wait instead if anyone thinks otherwise

SamRemis (approved):
Looking at the commit history, this seems necessary, but wow, the scaling factor being so different for gil vs nogil is wild.  

Thanks Jonathan

arandito (approved):

jonathan343:
Yea, it was confusing to me at first. After studying how the GC was updated for free-threaded builds, it makes more sense. 

Basically the non-free-threaded builds relied on the GIL for doing reference counting. When there were no references to an object, the GIL ensured no race conditions could result in a race condition and GC was safe.

After removing the GIL, they needed to update GC to periodically stop all threads to delete objects safely to prevent the race conditions. Because this has a big impact on performance, the collection phases happens much less frequently. 

So higher memory doesn't mean a leak, it is a result of the deferred GC. We should look into a more accurate way of testing this for no-GIL builds.

PR #3710, opened 2026-05-22
Fix CloudFront signer wildcard question mark handling

### LINKED ISSUE #3377: CloudFrontSigner doesn't account for wildcard question marks
CloudFrontSigners can be used to sign an url with '*' and '?' wildcards: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-setting-signed-cookie-custom-policy.html#private-content-custom-policy-statement-cookies-values. The check in _build_url() to figure out whether it is adding a query or appending to an existing query string does not account for this and can result in incorrect results. For example, consider the url `"https://example.com/example_202?.zip"`. Rather than being modified with a new query string `"https://example.com/example_202?.zip?SomeParam=foo"`, it'll be incorrectly turned into `"https://example.com/example_202?.zip&SomeParam=foo"`.

I'm not exactly sure how this should (if at all) be fixed because the wildcard is in direct contraction with url specs. At the moment, we are working around it by detecting that a wildcard was used and swapping out the first '&' for a '?', but this only works in our case because we know there aren't any query strings to start with with our particular URLs.

rohan-patnaik:
Fixes #3377

## Summary
This fixes CloudFront signed URL generation for resource URLs that use `?` as a CloudFront wildcard instead of a query-string separator.

## Why
Before this change, `_build_url()` checked only whether `?` existed anywhere in the URL. A resource like `http://test.com/example_202?.zip` was treated as if it already had query parameters, so signing parameters were appended with `&` and produced an invalid URL.

## Test plan
- `.venv/bin/python -m pytest tests/unit/test_signers.py::TestCloudfrontSigner`
- `.venv/bin/python -m black --check botocore/signers.py tests/unit/test_signers.py`

### DIFF
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -443,8 +443,12 @@ class CloudFrontSigner:
         return self._build_url(url, params)
 
     def _build_url(self, base_url, extra_params):
-        separator = '&' if '?' in base_url else '?'
+        separator = '&' if self._has_query_string(base_url) else '?'
         return base_url + separator + '&'.join(extra_params)
+
+    def _has_query_string(self, url):
+        _, marker, query = url.rpartition('?')
+        return bool(marker and ('=' in query or '&' in query))
 
     def build_policy(
         self, resource, date_less_than, date_greater_than=None, ip_address=None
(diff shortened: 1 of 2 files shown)


rohan-patnaik:
@adev-code, is this fix still useful? I’m happy to refresh the branch if so. Thanks!

rohan-patnaik:
@adev-code, could you help triage or review this CloudFront signing fix when you have a moment? Let me know if anything else is needed from me. Thanks!

PR #3711, opened 2026-05-22
feat(credentials): support web identity duration seconds

### LINKED ISSUE #3411: Support duration_seconds on the web identity token provider
### Describe the feature

Right now, there's no way to provide `DurationSeconds` as an extra kwarg on the `AssumeRoleWithWebIdentityCredentialFetcher`. We'd like to be able to populate this from the profile config.

### Use Case

We need our provider to have a longer lifetime so we can mint pre-signed URLs which won't expire after 1 hour, the default STS timeout.

### Proposed Solution

Add a populate for `duration_seconds` here, the same way that already exists for the assume role provider.
https://github.com/boto/botocore/blob/728045e5c6392e4409223c37afebeee35b9d76ea/botocore/credentials.py#L1888-L1899

### Other Information

If this is a desired change, we'd be happy to put up the PR ourselves!

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

latest

EKS

rohan-patnaik:
Fixes #3411

## Summary
This lets the assume-role-with-web-identity provider read `duration_seconds` from profile config and pass it as `DurationSeconds` to STS.

## Why
The regular assume role provider already supports this config, but the web identity provider ignored it. That made web identity sessions keep the default STS duration even when a profile requested a longer session.

## Test plan
- `.venv/bin/python -m pytest tests/unit/test_credentials.py::TestAssumeRoleWithWebIdentityCredentialProvider`

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -1957,6 +1957,13 @@ class AssumeRoleWithWebIdentityProvider(CredentialProvider):
         if role_session_name is not None:
             extra_args['RoleSessionName'] = role_session_name
 
+        duration_seconds = self._get_profile_config('duration_seconds')
+        if duration_seconds is not None:
+            try:
+                extra_args['DurationSeconds'] = int(duration_seconds)
+            except ValueError:
+                pass
+
         fetcher = AssumeRoleWithWebIdentityCredentialFetcher(
             client_creator=self._client_creator,
             web_identity_token_loader=token_loader,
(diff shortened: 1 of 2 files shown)


rohan-patnaik:
@adev-code, is this change still useful? I’m happy to refresh the branch if so. Thanks!

rohan-patnaik:
@adev-code, could you help triage or review the web-identity duration support when convenient? Happy to make any adjustments needed. Thanks!

PR #3712, opened 2026-05-22
feat(credentials): allow docker host metadata URI

### LINKED ISSUE #2515: Add host.docker.internal to ContainerMetadataFetcher list of approved hosts
[`ContainerMetadataFetcher` currently allows three hosts for container metadata endpoints](https://github.com/boto/botocore/blob/ee45929f6e1d7476a480f90a77ada87b970a1bf5/botocore/utils.py#L2263): 169.254.170.2, localhost, and 127.0.0.1. However, neither localhost or 127.0.0.1 represent the local docker host. Instead, the host's localhost is reachable as host.docker.internal ([docs](https://docs.docker.com/desktop/mac/networking/#use-cases-and-workarounds)). On Linux, the workaround is to use `--network host`, but on Mac and Windows this doesn't work (the docker network is always isolated from the host).

host.docker.internal should be added to `ContainerMetadataFetcher`'s list of approved hosts, so that a custom metadata endpoint for testing can be run on the host rather than needing to be run in another container.

rohan-patnaik:
Fixes #2515

## Summary
This adds `host.docker.internal` to the allowed hosts for container metadata full URI retrieval.

## Why
Docker Desktop exposes the host machine through `host.docker.internal`. Without this allowlist entry, containerized local/test setups cannot point `AWS_CONTAINER_CREDENTIALS_FULL_URI` at a metadata endpoint running on the host over HTTP.

## Test plan
- `.venv/bin/python -m pytest tests/unit/test_utils.py::TestContainerMetadataFetcher`

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -3075,6 +3075,7 @@ class ContainerMetadataFetcher:
         '169.254.170.23',
         'fd00:ec2::23',
         'localhost',
+        'host.docker.internal',
     ]
 
     def __init__(self, session=None, sleep=time.sleep):
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -2775,6 +2775,11 @@ class TestContainerMetadataFetcher(unittest.TestCase):
     def test_localhost_with_port_http_is_allowed(self):
         self.assert_can_retrieve_metadata_from('http://localhost:8000/foo')
 
+    def test_docker_host_http_is_allowed(self):
+        self.assert_can_retrieve_metadata_from(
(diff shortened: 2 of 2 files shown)


rohan-patnaik:
@adev-code, is this change still useful? I’m happy to refresh the branch if so. Thanks!

rohan-patnaik:
@nateprewitt, would this host.docker.internal change fit the cross-SDK credential-host policy, or should it wait for agreement in aws/aws-sdk#562? Happy to adjust the approach. Thanks!

PR #3713, opened 2026-05-22
fix(auth): warn on empty bearer token env var

### LINKED ISSUE #3603: AWS_BEARER_TOKEN_BEDROCK with empty string causes incomplete signature errors
### Describe the bug

When AWS_BEARER_TOKEN_BEDROCK is set to an empty string (via export AWS_BEARER_TOKEN_BEDROCK=), boto3 treats it as a valid bearer token and uses it for authentication,
causing "incomplete signature" errors. The empty string should be treated as if the variable is unset.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

- Empty string should be ignored
- Fall back to AWS credentials (IAM role, assume role, etc.)

### Current Behavior

- boto3 uses empty bearer token in HTTP request headers
- Requests fail with incomplete signature errors
- Only unset AWS_BEARER_TOKEN_BEDROCK works (not export AWS_BEARER_TOKEN_BEDROCK=)

### Reproduction Steps

```
export AWS_BEARER_TOKEN_BEDROCK=

python3 << EOF
import boto3
import logging
boto3.set_stream_logger('', logging.DEBUG)

client = boto3.client('bedrock-runtime', region_name='us-east-1')
# Fails with incomplete signature error
# Debug log shows BEARER_TOKEN in request headers
EOF
```

### Possible Solution

_No response_

_No response_

### SDK version used

boto3: 1.35.69; botocore: 1.35.69

macOS

rohan-patnaik:
Fixes #3603

## Summary
This logs a warning when a bearer token environment variable is set to an empty string.

## Why
An empty bearer token is still treated as configured today, which matches the current environment-variable behavior. The hard part for users is diagnosing it: requests can fail later with confusing service-side auth errors. This warning makes the bad configuration visible without changing token resolution semantics.

## Test plan
- `.venv/bin/python -m pytest tests/unit/test_utils.py::test_get_token_from_environment_warns_for_empty_token tests/unit/test_utils.py::test_get_token_from_environment_returns_token tests/unit/test_utils.py::test_get_token_from_environment_returns_none`
- `git diff --check`

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -3630,7 +3630,14 @@ def get_token_from_environment(signing_name, environ=None):
     if environ is None:
         environ = os.environ
     env_var = _get_bearer_env_var_name(signing_name)
-    return environ.get(env_var)
+    token = environ.get(env_var)
+    if token == '':
+        logger.warning(
+            "%s is set to an empty value; requests using bearer "
+            "authentication may fail",
+            env_var,
+        )
+    return token
 
 
 def _get_bearer_env_var_name(signing_name):
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -13,6 +13,7 @@
 import copy
 import datetime
 import io
+import logging
 import operator
 import os
 import shutil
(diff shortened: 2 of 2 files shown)


rohan-patnaik:
@nateprewitt, is this warning-based approach still useful? I’m happy to refresh the branch if so. Thanks!

rohan-patnaik:
@nateprewitt, does this warning-only approach fit the SDK specification discussion in #3603? Please let me know if anything else is needed to move it forward. Thanks!

PR #3063, opened 2023-11-15
Simplify and speed up timestamp parsing

akx:
As mentioned in https://github.com/boto/botocore/pull/2972#issuecomment-1811904061:

`_parse_timestamp_with_tzinfo()` already attempts to do `fromtimestamp` parsing; not much point in doing that work twice for timestamp-esque strings (and failing in the cases described in #2972).

In the subsequent commits, this PR also cleans up remaining seemingly Useless Use of `tzlocal`.

PR #3715, opened 2026-05-25
Remove the python-dateutil dependency

### LINKED ISSUE #3070: Use built-in fromisoformat instead of python-dateutil
### Describe the feature

Currently, `python-dateutil` is used to parse dates, but `fromisoformat` is now a built-in option that should be used instead. It was [significantly improved in 3.11 to allow this use](https://docs.python.org/3/whatsnew/3.11.html#datetime) and it's available through a back port package for older versions:
`backports-datetime-fromisoformat; python_version < '3.11'`)

### Use Case

Currently, have to install an extra dependency that [has caused problems in the past](https://github.com/boto/botocore/issues/1872) so this removes the dependency to minimize and simplify `botocore`

### Proposed Solution

Replace uses of `python-dateutil` with built-in parsing

### Other Information

_No response_

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

1.32.5

N/A

miketheman:
*Description of changes:*

Replaces every `dateutil` usage with stdlib, then drops the requirement.
Please review commit by commit. Each pair (test, then refactor) handles one piece on its own.

### Why

`dateutil.tz` and `dateutil.parser` gave us timezone and string parsing.
Python 3.10+ stdlib (`datetime`, `email.utils`) handles every input shape we actually see.
Removing dateutil also removes its transitive `six` dep - one fewer thing for packagers
and security scanners to vet on every release.

(Note: botocore still carries a vendored version of `six` - that is out of scope of this change, and may come later)

### Review order

Tests first, then refactor:

1. `test: replace dateutil.tz.tzlocal`
2. `refactor: replace dateutil.tz.tzlocal`
3. `test: replace usages of dateutil.tz.tzutc`
4. `refactor: replace dateutil.tz.tzutc with stdlib`
5. `test: use consistent datetime fixtures in s3express`
6. `test: replace dateutil.parser with stdlib`
7. `refactor: replace dateutil.parser with stdlib`
8. `chore(deps): remove unused python-dateutil`

### Behavior change

`parse_timestamp` of an RFC 822 GMT string returns a datetime with `tzinfo=datetime.timezone.utc` instead of `dateutil.tz.tzutc()`.
Same instant, different `tm_isdst` (-1 vs 0). One doc-renderer test was updated to match.

### Follow-ups

- `botocore.compat.get_tzinfo_options` is now a one-line shim.
  The TODO in the function describes the cleanup; it removes a public
  symbol, so it should be considered separately.
- `botocore.endpoint._calculate_ttl` uses `strptime("%Z")`, which is
  platform-dependent. `email.utils.parsedate_to_datetime` would be a
  better fit. Separate concern, may come at another time.

*Issue #, if available:* Resolves #3070


botocore/utils.py
```diff
@@ -948,31 +948,55 @@ def _epoch_seconds_to_datetime(value, tzinfo):
...
-        # In certain cases, a timestamp marked with GMT can be parsed into a
-        # different time zone, so here we provide a context which will
-        # enforce that GMT == UTC.
-        return dateutil.parser.parse(value, tzinfos={'GMT': tzutc()})
-    except (TypeError, ValueError) as e:
-        raise ValueError(f'Invalid timestamp "{value}": {e}')
+        return email.utils.parsedate_to_datetime(value)
+    except (TypeError, ValueError):
+        pass
+    for fmt in ("%m/%d/%Y", "%d %b %Y"):
```
SamRemis, line 994: [Question]

This catches a few non-Smithy shapes but rejects others that dateutil handled.  For example, it accepts 06/25/2026 but not 2026/06/25, and 15 Jan 2026 but not January 15, 2026. Why these two formats specifically?

botocore/compat.py
```diff
@@ -28,7 +28,6 @@
 
 from botocore.vendored import six
 from botocore.exceptions import MD5UnavailableError
```
SamRemis, line 30: [Discussion point]

Starting off with my biggest comment: 

We have [a specification](https://smithy.io/2.0/spec/protocol-traits.html#timestampformat-trait) for what accepted timestamp shapes are, so we should only ever need to accept those for parsing.  This PR properly handles all officially supported datetime formats, so theoretically this is a fully safe change to make despite handling fewer formats than dateutil.

There are still a few places that I'm still concerned about breaking:
1. AWS services that are incorrectly returning these values which botocore accepts
2. Non-AWS third party services relying on botocore's overly permissive behavior

It's a _bad_ thing that botocore is so permissive.  We have seen AWS service teams test their service with other datetime formats against botocore and seen it work, then be surprised when the same datetime format broke the Java SDK.  If we ever want to tighten up this behavior, we have to rip the bandaid off at some point or fully commit to supporting this forever.  

One way we could make this a little more cautiously is to start by using this new code path first and emitting a warning when we have to fall back to the old one, then ripping out the legacy path in three months with a minor version bump.  What do you think?

SamRemis (commented):
[Note]

This did pass our all of our test suites when ran locally, including integ tests against live services

akx:
This seems to touch upon the same things as my #3063. Might be worth it to see if there's something in there that could/should be incorporated here (since it hasn't seen reviews in 3 years... 🥹).

(Also, I guess this branch should be rebased on `develop` so there wouldn't be 2449 commits in here?)

PR #3721, opened 2026-06-04
Document policyDocumentShape for iam operations as a dict

aemous:
*Issue #, if available:*

* https://github.com/boto/boto3/issues/229
* Resolves D95731475 (internal)

*Notes:*

* botocore has a built-in handler that automatically converts all `policyDocumentType` shapes for the `iam` service from a string to a dict, using JSON decoding. However, the boto3 docs describes fields with this type as a string, because it is unaware of the runtime transformation. This bugfix updates the documented type to match the runtime type, performing similar transformation logic to the documented type. There is existing precedent for this in boto3 for DynamoDB.
* Must be sim-shipped with https://github.com/boto/boto3/pull/4795.

*Description of changes:*

* Port `DocumentModifiedShape` class from `boto3`, with the following changes:
  * Class-level documentation comments.
  * When `new_description` is `None`, the description is unmodified.
* Add built-in handler that updates the documented type of all `iam` service `policyDocumentType` shapes. These shapes are auto-converted to a dict; the new handler auto-converts their documented type to match the runtime type.

*Description of tests:*

* Successfully pass all tests (see GitHub Actions).
* Ran the following script and observed expected behavior:

```python
import botocore.session
from botocore.docs.bcdoc.restdoc import DocumentStructure
from botocore.docs.method import document_model_driven_method

client = session.create_client('iam', region_name='us-east-1')
op_model = client.meta.service_model.operation_model('GetRole')

doc = DocumentStructure('test', target='html')
section = doc.add_new_section('method')
document_model_driven_method(
    section, 'get_role', op_model,
    event_emitter=client.meta.events,
    example_prefix='response = client.get_role',
)

content = doc.flush_structure().decode('utf-8')
assert "'AssumeRolePolicyDocument': {}" in content
assert '*(dict) --*' in content or '(*dict*) --' in content
print("PASS: AssumeRolePolicyDocument documented as dict")
```

* Connected local boto3 to a modified botocore with the changes in this PR. Then, I rebuilt the docs, and verified the HTML looks correct:

image

### DIFF
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -1545,10 +1545,11 @@ BUILTIN_HANDLERS = [
             'policyDocumentType',
             new_type='dict',
             new_description=(
-                'The policy document as a URL-encoded JSON-decoded dict. '
-                'This is automatically decoded from the original JSON string.'
+                ' This value is automatically parsed from the'
+                ' JSON string returned by the service.'
             ),
             new_example_value='{}',
+            append_description=True,
         ).replace_documentation_for_matching_shape,
     ),
     ('after-call.ec2.GetConsoleOutput', decode_console_output),
--- a/botocore/handlers.py
+++ b/botocore/handlers.py
@@ -50,6 +50,7 @@ from botocore.compat import (
 from botocore.docs.utils import (
     AppendParamDocumentation,
     AutoPopulatedParam,
+    DocumentModifiedShape,
     HideParamFromOperations,
 )
 from botocore.endpoint_provider import VALID_HOST_LABEL_RE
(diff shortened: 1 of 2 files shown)


botocore/handlers.py
```diff
@@ -1538,6 +1539,19 @@ def get_bearer_auth_supported_services():
         enable_millisecond_timestamp_precision,
     ),
     ('after-call.iam', json_decode_policies),
+    (
+        'docs.*.iam.*.complete-section',
+        DocumentModifiedShape(
+            'policyDocumentType',
+            new_type='dict',
+            new_description=(
+                ' This value is automatically parsed from the'
```
SamRemis: [definitely non-blocking]

I wonder if we should keep this a bit simpler and only document the runtime type that the customer sees.  From their perspective, the str -> dict conversion is an implementation detail and won't change how they interact with the code

I'd lean towards just removing this updated description and keeping the service-supplied one without appending any additional info.  

Either way, thank you for tackling this *very* old issue.  I appreciate the initiative and ownership here.

aemous: I have published a revision so that there is no appended details to the description.

SamRemis (approved):

PR #3722, opened 2026-06-09
validate region from S3 redirect response in S3RegionRedirector

metsw24-max:
**Unvalidated bucket region in S3RegionRedirector**

The deprecated `S3RegionRedirector.get_bucket_region` returns the region taken from a redirect response (the `x-amz-bucket-region` header, the error body `Region`, or the HeadBucket fallback) as-is, whereas its replacement `S3RegionRedirectorv2` already runs `validate_region_name` on the same value. The two classes therefore disagree on whether a server-supplied region is checked. That region then feeds endpoint resolution, the rewritten request URL and the SigV4 signing region, so an untrusted or misbehaving endpoint can push an arbitrary string through this path. I have kept the change to the v2 behaviour and applied the same `validate_region_name` call; nothing else in the flow changes. Added unit cover for the three sources. The class is deprecated, but it is still importable and used by third parties, so bringing it in line seems worthwhile.

### DIFF
--- a/.changes/next-release/bugfix-s3-42137.json
+++ b/.changes/next-release/bugfix-s3-42137.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "s3",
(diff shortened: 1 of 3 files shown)


metsw24-max:
any update?

PR #3724, opened 2026-06-10
reject trailing newline in host label validators

metsw24-max:
botocore's host label validators all anchor with `$`, which in Python also matches just before a trailing newline, so a value such as `us-east-1\n`, `mybucket\n` or a `data\n` host prefix passes validation and the unchanged string is then used to build the request host. `validate_region_name` even documents that its argument must be a valid host label, yet it accepts the newline, and the same gap exists in `validate_bucket_name`, the `HeaderToHostHoister` hostname check, the `S3ControlEndpointSetter` host label check and the serialiser host prefix check. The region case is the most exposed because that value can be sourced from an S3 cross-region redirect response and then feeds endpoint resolution and the SigV4 signing scope. `is_valid_endpoint_url` already screens for `\t\r\n` through `UNSAFE_URL_CHARS`, so these label validators are the remaining place a control character can slip past. I moved the trailing anchor to `\Z` in each one so a match has to reach the actual end of the string, and left the ARN and resource regexes alone since those re-extract a clean label and never carry a newline through. Valid regions, bucket names and host labels behave exactly as before.

### DIFF
--- a/.changes/next-release/bugfix-validators-newline.json
+++ b/.changes/next-release/bugfix-validators-newline.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Validators",
+  "description": "Anchor the host label, region, bucket name and host prefix validators with ``\\Z`` so a value with a trailing newline is rejected instead of slipping through ``$`` into the request host."
+}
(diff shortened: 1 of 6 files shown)


metsw24-max:
gentle ping

PR #3731, opened 2026-06-21
Improve NoRegionError message with guidance on setting a region

viveka76:
**Description of changes:** Expands the `NoRegionError` message to tell users
how to set a region (region_name argument, AWS_DEFAULT_REGION, or
~/.aws/config), and calls out the common mistake of putting region in
~/.aws/credentials. 
Message text only — no change to resolution behavior, precedence, or the
exception type. Backwards compatible. Unit test and changelog entry included.
By submitting this pull request, I confirm that you can use, modify, copy,
and redistribute this contribution, under the terms of your choice.
This pr message is generated by AI tools, and reviewed by Vivek Shaw

### DIFF
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -244,10 +244,9 @@ class NoRegionError(BaseEndpointResolverError):
     """No region was specified."""
 
     fmt = (
-        'You must specify a region. Set it via the region_name argument '
-        'to the client/Session, the AWS_DEFAULT_REGION environment '
-        'variable, or "region" in ~/.aws/config (note: region goes in '
-        '~/.aws/config, not ~/.aws/credentials).'
+        'You must specify a region. This can be done by passing the region_name '
+        'argument to the client or session, using the AWS_DEFAULT_REGION '
+        'environment variable, or setting "region" in the aws config file.'
     )
 
 
--- a/botocore/exceptions.py
+++ b/botocore/exceptions.py
@@ -243,7 +243,12 @@ class BaseEndpointResolverError(BotoCoreError):
 class NoRegionError(BaseEndpointResolverError):
     """No region was specified."""
 
-    fmt = 'You must specify a region.'
+    fmt = (
+        'You must specify a region. Set it via the region_name argument '
+        'to the client/Session, the AWS_DEFAULT_REGION environment '
+        'variable, or "region" in ~/.aws/config (note: region goes in '
+        '~/.aws/config, not ~/.aws/credentials).'
+    )
 
 
 class EndpointVariantError(BaseEndpointResolverError):
--- a/tests/unit/test_regions.py
+++ b/tests/unit/test_regions.py
@@ -212,9 +212,6 @@ class TestEndpointResolver(unittest.TestCase):
         with self.assertRaises(NoRegionError):
             resolver = regions.EndpointResolver(self._template())
             resolver.construct_endpoint('foo', None)
-        message = str(ctx.exception)
-        self.assertIn('You must specify a region', message)
-        self.assertIn('AWS_DEFAULT_REGION', message)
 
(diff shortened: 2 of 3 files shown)


SamRemis:
@viveka76 , thank you for the PR!  I agree, this error message could be more helpful.

There are a few notes before we can merge:
1. The AWS CLI v1 [catches this exception and expands on it](https://github.com/aws/aws-cli/blob/0cfabe510d6e9b4dda8a66a69416d32ed221fcc6/awscli/clidriver.py#L242-L243), so this would wind up with a confusing error message for CLIv1 users where we mention the SDK specific error message.  The CLIv1 is [scheduled to go into maintenance mode](https://aws.amazon.com/blogs/developer/cli-v1-maintenance-mode-announcement/) on July 15th of this year, so they will no longer receive these updates at that point.  Waiting until this date to merge would probably be the easiest way to handle this.  
2. The error message is good, but there are a few edits I'd like to make.  I'll suggest a change with those shortly.  
3. The updated test doesn't define `ctx` but references it, so it will fail when it runs.  Since error messages like this one are intended to be helpful and user-facing as opposed to part of the API contract that can be coded against, we generally don't assert on the content of the message directly.  I'd recommend just reverting the changes to the test file.
4. The PR description appears to be generated using an LLM.  Please see our [CONTRIBUTING.md file](https://github.com/aws/aws-cli/blob/develop/CONTRIBUTING.md#automated-tools) for guidance on how to use tools to contribute to this repo.  Please update the PR description to include the info outlined there.

tests/unit/test_regions.py
```diff
@@ -212,6 +212,9 @@ def test_ensures_region_is_not_none(self):
         with self.assertRaises(NoRegionError):
             resolver = regions.EndpointResolver(self._template())
             resolver.construct_endpoint('foo', None)
+        message = str(ctx.exception)
+        self.assertIn('You must specify a region', message)
+        self.assertIn('AWS_DEFAULT_REGION', message)
```
SamRemis:
```suggestion
```

botocore/exceptions.py
```diff
@@ -243,7 +243,12 @@ class BaseEndpointResolverError(BotoCoreError):
 class NoRegionError(BaseEndpointResolverError):
     """No region was specified."""
 
-    fmt = 'You must specify a region.'
+    fmt = (
+        'You must specify a region. Set it via the region_name argument '
+        'to the client/Session, the AWS_DEFAULT_REGION environment '
+        'variable, or "region" in ~/.aws/config (note: region goes in '
+        '~/.aws/config, not ~/.aws/credentials).'
```
SamRemis:
```suggestion
        'You must specify a region. This can be done by passing the region_name '
        'argument to the client or session, using the AWS_DEFAULT_REGION '
        'environment variable, or setting "region" in the aws config file.'
```

The reasoning behind these suggested edits:
 - `region` can be set in the credentials file and it will be respected by botocore
 - the filepath for the config file is not hard-coded and can be specified using the `AWS_CONFIG_FILE` environment variable (same with `AWS_SHARED_CREDENTIALS_FILE` for the credentials file)
 - While we could still include a reference to the credentials file, that would be a lot of detail for an error message.  Most of the common cases would be covered with this update.

SamRemis (changes requested):

botocore/exceptions.py
viveka76: Thank you for such detailed explanation on this, will apply your suggestion.

tests/unit/test_regions.py
viveka76: agreed as per the discussion on conversations tab

viveka76:
@SamRemis Thank you for your valuable suggestions, I have applied your suggestions and also added the AI use in commit message

PR #3733, opened 2026-06-23
Add static stability to credential refresh

alexgromero:
*Issue #, if available:*
Python-8911

### Overview
This PR adds static stability to credential refresh in `RefreshableCredentials`. When a credential source fails during refresh (outage, exception, missing keys, or expired response), the SDK now falls back to the most recently cached credentials and enters a jittered backoff window (5-10 minutes) instead of raising or hammering the source. If no cached credentials exist yet (e.g. `DeferredRefreshableCredentials` on first fetch), the error is still surfaced.

The new behavior is gated behind `DEFAULT_NEW_CREDENTIAL_REFRESH`, a flag that defaults to `False` in the public distribution. External users are unaffected until GA, at which point the flag and legacy path will be removed.

### Implementation

The new behavior lives in `_refresh_with_backoff` and its helpers. The legacy path (`_protected_refresh`, `_set_from_data`) is unchanged and kept under `_refresh_legacy`. A flag check in `_refresh` routes between the two. Credentials are validated before being cached so we never store missing or expired data.

### Testing
Following the same pattern as [test_standard_retry_v2_1.py](https://github.com/boto/botocore/blob/develop/tests/unit/retries/test_standard_retry_v2_1.py) from the retries internal release.

The file contains full copies of `TestRefreshableCredentials` and `TestRefreshLogic` with the flag forced on. Tests that change behavior have updated assertions (refresh failures return cached creds instead of raising). Unchanged tests are included so the full class can be swapped in at GA without cherry-picking individual methods.

The standalone pytest functions at the bottom cover backoff-specific behavior that does not exist in the legacy path (retry timing, not re-hitting the source immediately after failure, first-fetch failure raising).

At GA this file gets folded into `test_credentials.py` and removed.

---

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -17,6 +17,7 @@ import getpass
 import json
 import logging
 import os
+import random
 import subprocess
 import threading
 import time
@@ -69,6 +70,17 @@ from botocore.utils import (
     parse_key_val_file,
     resolve_imds_endpoint_mode,
 )
+
+try:
+    # This is not a public interface and is subject to abrupt breaking changes.
+    # Currently, it's only available to internal users for testing and validation
+    # of the new credential refresh behavior. Any usage is not advised or
+    # supported in external code bases.
+    from botocore.customizations.credentials import (
+        DEFAULT_NEW_CREDENTIAL_REFRESH,
+    )
+except ImportError:
+    DEFAULT_NEW_CREDENTIAL_REFRESH = False
 
 logger = logging.getLogger(__name__)
 ReadOnlyCredentials = namedtuple(
@@ -428,6 +440,7 @@ class RefreshableCredentials(Credentials):
         self._expiry_time = expiry_time
         self._time_fetcher = time_fetcher
         self._refresh_lock = threading.Lock()
+        self._refresh_blocked_until = None
         self.method = method
         self._frozen_credentials = ReadOnlyCredentials(
             access_key, secret_key, token, account_id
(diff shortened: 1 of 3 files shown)


botocore/credentials.py
```diff
@@ -563,7 +576,125 @@ def _is_expired(self):
...
+        # Apply a jittered 5-10 minute backoff after a failed refresh to avoid
+        # overwhelming the credential source during an outage.
+        backoff = random.uniform(300, 600)
+        now = self._time_fetcher()
+        self._refresh_blocked_until = now + datetime.timedelta(seconds=backoff)
+        return backoff
+
     def _refresh(self):
+        if DEFAULT_NEW_CREDENTIAL_REFRESH:
+            return self._refresh_with_backoff()
```
SamRemis: [nit, non-blocking]

This is a bit confusing to me since we have two return statements that both will always return `None`.  The `return` keywords on this line and the next don't do anything, but so it's a stylistic choice.

Just removing them would cause a double refresh, but I'd find this more readable as an if/else statement:
```
    if DEFAULT_NEW_CREDENTIAL_REFRESH:
        self._refresh_with_backoff()
    else:
        self._refresh_legacy()
```

It's not a functional issue, so if you disagree and you like this pattern better, just resolve this comment.

botocore/credentials.py
```diff
@@ -563,7 +576,125 @@ def _is_expired(self):
...
+        return self._refresh_legacy()
+
+    def _refresh_with_backoff(self):
+        # In the common case where we don't need a refresh, we
+        # can immediately exit and not require acquiring the
+        # refresh lock.
+        if not self.refresh_needed(self._advisory_refresh_timeout):
```
SamRemis, line 606: [non-actionable, just here for posterity]

The PR description mentions that the entire legacy path will be removed.  If we would allow an opt-out instead and allow customers to control this behavior, I'd rather see us not violate DRY here and find a way to merge this new codepath with the legacy.

Since the entire legacy path will be deleted (or this new one will), IMO it's alright for us to keep these code paths a little bit repetitive.  It keeps one clean, centralized gate for the new behavior.

botocore/credentials.py
```diff
@@ -563,7 +576,125 @@ def _is_expired(self):
...
+
+    def _handle_refresh_failure(self, error):
+        if self._frozen_credentials is None:
+            # No credentials have been successfully obtained yet, so we
+            # have nothing to fall back to.
+            raise CredentialRetrievalError(
+                provider=self.method, error_msg=str(error)
+            )
+        backoff = self._enter_refresh_backoff()
+        logger.warning(
```
SamRemis: [Discussion]

This is good for a single call and I'm glad we are warning here.  Since we may still make repeated calls for up to ten minutes, it may make sense for us to add a per-call debug log as well.  This way we link each call to the failed refresh and someone won't have to scroll through up to ten minutes of logs to figure out why they got a 403.

What do you think?

botocore/credentials.py
```diff
@@ -563,7 +576,125 @@ def _is_expired(self):
...
+                "credential refresh failed, response did not contain: "
+                f"{', '.join(missing_keys)}"
+            )
+
+        if parse(data['expiry_time']) <= self._time_fetcher():
+            return "credential source returned expired credentials"
+        return None
+
+    def _handle_refresh_failure(self, error):
+        if self._frozen_credentials is None:
```
SamRemis: [question, maybe blocking?]

This code path should be dead since we never set it to `None` in the `__init__` method.  Shouldn't this be on the DeferredRefreshableCredentials class, or am I missing something?

botocore/credentials.py
alexgromero: Yeah this makes sense. Will add in next revision.

botocore/credentials.py
alexgromero: Yeah, good call. That `None` case is specific to `DeferredRefreshableCredentials`, so I moved it there to make the behavior clearer and keep the base handler focused on the shared fallback/backoff path.

botocore/credentials.py
alexgromero: You're right. Will update in next revision.

botocore/credentials.py
alexgromero: Right, this is intentional. The legacy path is temporary and gets deleted at GA along with the flag, so I kept the two paths separate rather than merging them for one clean gate on the new behavior. No opt-out planned.

jonathan343 (approved):

ubaskota (approved):

PR #3743, opened 2026-07-04
anchor host label validators with \Z to reject trailing newline

ubeddulla:
several host and host-label validators anchor with `$`, which in python also matches just before a trailing newline, so a value like `us-west-2\n` slips through unchanged. that matters because these values are interpolated into the request hostname. validate_region_name gates the region taken from the `x-amz-bucket-region` response header (or the error body) in S3RegionRedirectorv2.get_bucket_region before it is used to re-resolve the endpoint, and the same shape guards the sigv host prefix in serialize (HOST_PREFIX_RE), the hoisted host header in HeaderToHostHoister, the s3-control arn host labels in S3ControlEndpointSetter, and the ruleset host-label check in endpoint_provider. switching the end anchor to `\Z` rejects the trailing newline and leaves every valid label matching as before. added a regression test that feeds a trailing newline through the redirector's response-header path.


ubeddulla:
any update?

PR #3751, opened 2026-07-17
Raise error on empty CA bundle value

alexgromero:
*Issue #, if available:*

*Description of changes:*
When a CA bundle value (`ca_bundle`, `AWS_CA_BUNDLE`, `REQUESTS_CA_BUNDLE`, or the client `verify` parameter) resolved to an empty string, botocore will now raise an error.

An empty string is not a valid value for these settings, which are defined as paths to a CA bundle.


ubaskota (approved):
LGTM

PR #3752, opened 2026-07-18
mask proxy credentials by rebuilding netloc, not substring replace

ubeddulla:
mask_proxy_url masks the userinfo with proxy_url.replace(username, mask, 1) and the same for the password, but str.replace swaps the first occurrence of that substring anywhere in the url, so when the credential value also appears earlier (in the scheme or host) the mask lands there and the real secret survives in the string that gets logged through ProxyConnectionError. for instance https://user:https@proxy.example.com masks to https://***:https@proxy.example.com with the password still visible. rebuilding the netloc from the parsed userinfo and host masks only the actual credential fields.

### DIFF
--- a/botocore/httpsession.py
+++ b/botocore/httpsession.py
@@ -210,13 +210,17 @@ def mask_proxy_url(proxy_url):
 
     :return: Masked proxy url, i.e. https://***:***@proxy.com
     """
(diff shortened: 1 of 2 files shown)


ubeddulla:
any update?

PR #3753, opened 2026-07-20
create the credential cache directory with 0700 in JSONFileCache

ubeddulla:
JSONFileCache creates its working directory with os.makedirs and no mode, so it lands at 0o777 & ~umask. That directory backs `~/.aws/sso/cache` (SSO access and refresh tokens) and `~/.aws/boto/cache` (cached assume-role credentials); the entries themselves are already 0600 since they come from mkstemp and os.replace keeps the mode, but under umask 000, which is easy to hit in containers and in systemd units with no UMask=, the directory is world-writable and another local user can unlink an entry and drop in their own for botocore to load back. Passing mode=0o700 lines this up with what the CLI does for the directories it creates, and I added a regression test alongside the existing atomic-write ones.

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -3589,7 +3589,7 @@ class JSONFileCache:
                 f"Value cannot be cached, must be JSON serializable: {value}"
             )
         if not os.path.isdir(self._working_dir):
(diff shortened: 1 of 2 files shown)


ubeddulla:
gentle ping

PR #3760, opened 2026-07-29
Add new refresh window defaults

alexgromero:
*Issue #, if available:*
Python-8913

*Description of changes:*
### Overview
This PR updates the flag-gated the credential refresh path to use the new refresh window defaults. The legacy path is unchanged and all behavior changes here are scoped to  `DEFAULT_NEW_CREDENTIAL_REFRESH`.

#### Mandatory refresh window
The new refresh path now uses a 1-minute mandatory refresh window.

 #### Advisory refresh window
On the new refresh path, the effective advisory refresh window is selected dynamically from the credential lifetime (`Expiration - now`) when a new credential set is obtained:
- `<= 20 minutes` -> `5 minutes`
- `> 20 minutes and < 90 minutes` -> `15 minutes`
- `>= 90 minutes` -> `60 minutes`
That window stays fixed for the current credential set, is recomputed only after a successful refresh, and is not recomputed after a failed refresh.

#### Internal override behavior
- Preserves explicitly passed internal refresh windows on the new path instead of replacing them with computed defaults.
- This keeps existing internal callers such as S3 Express on their explicit windows.

#### New path refresh flow
- Advisory refresh checks on the new path now use the effective advisory window for the current credential set.
- Deferred credentials continue to fetch on first use and compute their advisory window once a credential set is successfully obtained.

### Tests
The flag-on coverage lives in `tests/unit/test_credential_refresh.py` for now. At GA, that coverage will be folded into `tests/unit/test_credentials.py` and the temporary file will be removed.

### Notes
- The legacy refresh path is unchanged, including the existing 15-minute advisory and 10-minute mandatory defaults.
- S3 Express behavior is unchanged since it is out of scope for this work.

---

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -17,6 +17,7 @@ import getpass
 import json
 import logging
 import os
+import random
 import subprocess
 import threading
 import time
@@ -69,6 +70,17 @@ from botocore.utils import (
     parse_key_val_file,
     resolve_imds_endpoint_mode,
 )
+
+try:
+    # This is not a public interface and is subject to abrupt breaking changes.
+    # Currently, it's only available to internal users for testing and validation
+    # of the new credential refresh behavior. Any usage is not advised or
+    # supported in external code bases.
+    from botocore.customizations.credentials import (
+        DEFAULT_NEW_CREDENTIAL_REFRESH,
+    )
+except ImportError:
+    DEFAULT_NEW_CREDENTIAL_REFRESH = False
 
 logger = logging.getLogger(__name__)
 ReadOnlyCredentials = namedtuple(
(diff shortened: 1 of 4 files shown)


botocore/credentials.py
```diff
@@ -563,7 +586,160 @@ def _is_expired(self):
...
+            return self._explicit_mandatory_refresh_timeout
+        return _DEFAULT_MANDATORY_REFRESH_TIMEOUT_V2
+
+    def _compute_effective_advisory_refresh_timeout(self):
+        # Compute the effective advisory refresh timeout from the credential
+        # lifetime at the time the credential set is created or refreshed:
+        # <= 20 min -> 5 min
+        # > 20 min and < 90 min -> 15 min
+        # >= 90 min -> 60 min
+        lifetime = total_seconds(self._expiry_time - self._time_fetcher())
```
ubaskota, line 628: Nit: In this calculation, what are the chances of one being tz-aware and the other being naive? If one side is tz-aware and the other is naive, it raises a `TypeError`. I see that this calculation isn't new and is also done in `_seconds_remaining()`. Given that it's been there this long and hasn't been an issue, the chances of it being an issue in the future are slim, but wanted to point it out.

alexgromero: Agree the odds are slim. This follows the existing helper so I'm keeping it consistent for now.

ubaskota (approved):
LGTM

PR #3765, opened 2026-08-03
Include max attempts in the amz-sdk-request header on the initial attempt

ubaskota:
*Issue #, if available:*

*Description of changes:*
``amz-sdk-request`` omitted ``max`` on the initial attempt because the value was only set once a response was received. Seed it at client registration so it's present on every attempt.

*Testing:*
- Added unit tests asserting ``max`` is present on the initial request , across retries and for service-specific overrides (DynamoDB)
- Full test suite passes


Alan4506 (approved):
LGTM. Thanks @ubaskota!

PR #3766, opened 2026-08-05
Remove paginator tests specific to AWS CLI

kdaily:
*Issue #, if available:*

Followup after https://github.com/aws/aws-cli/pull/10414

*Description of changes:*

The AWS CLI auto-pagination requires stricter validation on the known parameters, including those that would come from customized configuration through paginator extras files.

This change removes two of those assertions:

- A check for parameters that aren't accounted for in the paginator config, like `non_aggregate_keys`
- A check for summing over integer columns, like for DynamoDB Scan

These do not apply to boto3/Python SDK, only the AWS CLI, and the tests have been vended as they currently are to the AWS CLI v1 codebase.

### Testing

I manually removed a `paginators-1.sdk-extras.json` file before making any code changes, and observed that the `_validate_output_keys_match` raises an error.

Then, I modified the test to remove the assertion that would be made, added the extras file back, and observed the test passes.


tests/functional/test_paginator_config.py
```diff
@@ -161,7 +161,7 @@ def test_lint_pagination_configs(operation_name, page_config, service_model):
     _validate_operation_has_output(operation_name, service_model)
     _validate_input_keys_match(operation_name, page_config, service_model)
     _validate_output_keys_match(operation_name, page_config, service_model)
-    _validate_new_numeric_keys(operation_name, page_config, service_model)
+    # _validate_new_numeric_keys(operation_name, page_config, service_model)
```
kdaily: Missed removing this instead of commenting it out.

tests/functional/test_paginator_config.py
```diff
@@ -28,109 +28,6 @@
...
-    'storagegateway.ListVolumes.GatewayARN',
-    'workdocs.DescribeUsers.TotalNumberOfUsers',
-    'xray.BatchGetTraces.UnprocessedTraceIds',
-    'xray.GetServiceGraph.EndTime',
-    'xray.GetServiceGraph.ContainsOldGroupVersions',
-    'xray.GetServiceGraph.StartTime',
-    'xray.GetTraceSummaries.TracesProcessedCount',
-    'xray.GetTraceSummaries.ApproximateTime',
-]
 KNOWN_PAGINATORS_WITH_INTEGER_OUTPUTS = (
```
jonathan343: I don't think `KNOWN_PAGINATORS_WITH_INTEGER_OUTPUTS` is being used anywhere anymore either.

jonathan343 (approved):

PR #3767, opened 2026-08-05
Document build_full_result contract and log dropped members

jonathan343:
## Summary

`PageIterator.build_full_result()` silently drops any top-level output member that isn't a `result_key` or `non_aggregate_key`. This PR adds a docstring describing the aggregation contract and a DEBUG log naming dropped members, so users who hit a missing field get a breadcrumb instead of debugging a silent omission.

## Background

`build_full_result()` collapses a paginated response into a single dict, aggregating only `result_key` values (summed across pages) and members listed in `non_aggregate_keys` (supplied per-service via `paginators-1.sdk-extras.json` overlays). Everything else is dropped from the final response. We maintained those overlays strictly because AWS CLI v1 shared our release stack and enforced pagination validation due to it's auto-pagination requirements. CLI v1 now vendors its own botocore, so this improves the experience for the remaining direct `build_full_result()` callers without requiring per-service customizations.

## Customer experience

No behavior change. A user hitting a missing field in the future now gets a DEBUG breadcrumb (visible under `boto3.set_stream_logger('')`) naming the dropped members.

```
2026-08-05 17:29:21,194 botocore.paginate [DEBUG] The following top-level output members are not accounted for in the pagination config and were dropped from build_full_result: RequestId, Status. Iterate the pages directly if you need these values.
```

---

### DIFF
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -483,7 +483,16 @@ class PageIterator:
         ]
 
     def build_full_result(self):
+        """Aggregate all pages into a single combined result.
+
+        ``result_key`` values are concatenated/summed across pages and
+        ``non_aggregate_keys`` are taken from the first page (assumed static).
+        Other top-level members are not guaranteed and are generally dropped;
+        iterate the pages directly if you need them.
+        """
         complete_result = {}
+        # Last page's top-level members, for the dropped-member debug log.
+        last_page_keys = set()
         for response in self:
             page = response
             # We want to try to catch operation object pagination
@@ -493,6 +502,8 @@ class PageIterator:
             # uses. We can remove it though once operation objects are removed.
             if isinstance(response, tuple) and len(response) == 2:
                 page = response[1]
+            if isinstance(page, dict):
+                last_page_keys = set(page)
             # We're incrementally building the full response page
             # by page.  For each page in the response we need to
             # inject the necessary components from the page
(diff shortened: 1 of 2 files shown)


botocore/paginate.py
```diff
@@ -529,6 +540,19 @@ def build_full_result(self):
...
             complete_result['NextToken'] = self.resume_token
+        # Log top-level members that weren't aggregated (not a result_key or
+        # non_aggregate_key) and so were dropped from the combined result.
+        if log.isEnabledFor(logging.DEBUG):
+            dropped = last_page_keys - set(complete_result)
+            dropped.discard('ResponseMetadata')  # never part of paginated data
+            if dropped:
+                log.debug(
+                    "The following top-level output members are not "
+                    "accounted for in the pagination config and were dropped "
```
Alan4506, line 551: Nit: I'm wondering if the wording "not accounted for in the pagination config" is accurate or not. For example, IAM `ListUsers` declares [IsTruncated as more_results](https://github.com/boto/botocore/blob/4df81b85d393d56c0d4fe8f1686f55d5214ad0db/botocore/data/iam/2010-05-08/paginators-1.json#L42) in its pagination config, but since it never lands in the merged result, the log shows:
```
The following top-level output members are not accounted for in the
pagination config and were dropped from build_full_result: IsTruncated.
Iterate the pages directly if you need these values.
```
I agree that `IsTruncated` is dropped from build_full_result, but it is indeed accounted for in the paginator config, right?

Correct me if my understanding is wrong.

Alan4506 (commented):
Thanks @jonathan343! LGTM overall but I have a question. Please see my comment below.

PR #3769, opened 2026-08-06
Skip legacy IMDS expiry extension on new refresh path

alexgromero:
*Issue #, if available:*
Python-9032

### Overview
Updates the IMDS credential-fetching path to use the new shared credential refresh behavior when `DEFAULT_NEW_CREDENTIAL_REFRESH` is enabled.

On the flag-off path, IMDS keeps the existing legacy expiry-extension behavior. On the flag-on path, IMDS no longer calls `_evaluate_expiration()`, returns the real IMDS expiry, and relies on the shared `RefreshableCredentials` refresh-window behavior instead.

### Testing
Also updates the flag-on test suite in tests/unit/test_credential_refresh.py to reflect the intended GA migration shape:

- keep mirrored classes as pass when unchanged
- use additive subclass coverage where existing tests still apply
- fully copy the IMDS fetcher test class where legacy expiry-extension tests need to be removed/replaced

---

### DIFF
--- a/botocore/utils.py
+++ b/botocore/utils.py
@@ -39,6 +39,17 @@ import botocore
 import botocore.awsrequest
 import botocore.httpsession
 
+try:
+    # This is not a public interface and is subject to abrupt breaking changes.
+    # Currently, it's only available to internal users for testing and validation
+    # of the new credential refresh behavior. Any usage is not advised or
+    # supported in external code bases.
+    from botocore.customizations.credentials import (
+        DEFAULT_NEW_CREDENTIAL_REFRESH,
+    )
+except ImportError:
+    DEFAULT_NEW_CREDENTIAL_REFRESH = False
+
 # IP Regexes retained for backwards compatibility
 from botocore.compat import (
     HAS_CRT,
@@ -609,7 +620,8 @@ class InstanceMetadataFetcher(IMDSFetcher):
                     'token': credentials['Token'],
                     'expiry_time': credentials['Expiration'],
                 }
-                self._evaluate_expiration(credentials)
+                if not DEFAULT_NEW_CREDENTIAL_REFRESH:
+                    self._evaluate_expiration(credentials)
                 return credentials
             else:
                 # IMDS can return a 200 response that has a JSON formatted
--- a/tests/unit/test_credential_refresh.py
+++ b/tests/unit/test_credential_refresh.py
@@ -20,27 +20,50 @@ yet available to external users. Once the changes are validated
 internally and released publicly, the classes below will replace the
 corresponding classes in test_credentials.py and the standalone tests
 will be added to test_credentials.py. This file will then be removed.
+
+Mirror policy for GA migration:
+- use ``pass`` when the original class is unchanged on the new path
+- subclass the original class when all original tests still apply and we
+  only need additive flag-on coverage
+- fully copy the class when the new path needs removals or replacements,
+  so this file is the GA-ready replacement
 """
 
-from datetime import datetime, timedelta
+import operator
+from datetime import datetime, timedelta, timezone
 
 import pytest
 from dateutil.tz import tzlocal
 
-from botocore import credentials
-from botocore.exceptions import CredentialRetrievalError
-from tests import BaseEnvVar, IntegerRefresher, mock, unittest
-from tests.unit.test_credentials import (
-    TestEnvVar as _TestEnvVar,
+from botocore import credentials, utils
+from botocore.awsrequest import AWSResponse
+from botocore.compat import json
+from botocore.exceptions import (
+    ConnectionClosedError,
+    ConnectTimeoutError,
+    CredentialRetrievalError,
+    InvalidIMDSEndpointError,
+    MetadataRetrievalError,
+    ReadTimeoutError,
 )
-from tests.unit.test_credentials import (
-    TestProcessProvider as _TestProcessProvider,
+from tests import (
+    BaseEnvVar,
+    FreezeTime,
+    IntegerRefresher,
+    RawResponse,
+    mock,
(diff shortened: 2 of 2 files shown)


botocore/utils.py
```diff
@@ -39,6 +39,17 @@
...
+    # This is not a public interface and is subject to abrupt breaking changes.
+    # Currently, it's only available to internal users for testing and validation
+    # of the new credential refresh behavior. Any usage is not advised or
+    # supported in external code bases.
+    from botocore.customizations.credentials import (
+        DEFAULT_NEW_CREDENTIAL_REFRESH,
+    )
+except ImportError:
+    DEFAULT_NEW_CREDENTIAL_REFRESH = False
+
```
ubaskota, line 52: In [PR#1](https://github.com/boto/botocore/pull/3733/changes), the `try/except` block lives only in `credentials.py`, and `client.py` imports `DEFAULT_NEW_CREDENTIAL_REFRESH` from it, which is the right way to do it. But here try/except block is duplicated instead. Is there anything stopping us from importing directly from credentials.py?

tests/unit/test_credential_refresh.py
ubaskota: Something I missed in PR-1: why are we using real time with `datetime.now(tzlocal())`? Tests relying on this could become flaky. Can we use  fixed hard coded values instead so the tests remain deterministic?

tests/unit/test_credential_refresh.py
```diff
@@ -245,14 +268,512 @@ def test_refresh_giving_expired_credentials_returns_cached(self):
...
 
+class TestInstanceMetadataProviderFlagOn(
+    test_credentials.TestInstanceMetadataProvider
+):
+    # GA: keep all tests from test_credentials.TestInstanceMetadataProvider and
+    # add the shared static-stability coverage below.
+    def test_imds_refresh_failure_returns_cached_credentials(self):
+        # IMDS should pick up shared static stability from
+        # RefreshableCredentials on the new path.
+        timeobj = datetime.now(tzlocal())
```
ubaskota: Is there a reason you're using real `datetime` here? We should mock it with hardcoded values to keep the tests deterministic. Same applies to other tests where real values are used for date and time.

tests/unit/test_credential_refresh.py
```diff
@@ -245,14 +268,512 @@ def test_refresh_giving_expired_credentials_returns_cached(self):
...
+        return {
+            'access_key': creds['AccessKeyId'],
+            'secret_key': creds['SecretAccessKey'],
+            'token': creds['Token'],
+            'expiry_time': creds['Expiration'],
+            'role_name': self._role_name,
+        }
+
+    @FreezeTime(module=utils.datetime, date=DATE)
+    def test_returned_expiry_is_not_extended(self):
```
ubaskota: This test passes for both old and new behavior. I set `monkeypatch.setattr(utils, 'DEFAULT_NEW_CREDENTIAL_REFRESH', False)` and it passed. Shouldn't it fail when using the old behavior?

tests/unit/test_credential_refresh.py
```diff
@@ -245,14 +268,512 @@ def test_refresh_giving_expired_credentials_returns_cached(self):
...
+            iam_role_fetcher=fetcher
+        )
+
+        creds = provider.load()
+
+        self.assertEqual(creds.access_key, 'a')
+        self.assertEqual(fetcher.retrieve_iam_role_credentials.call_count, 2)
+
+
+class TestInstanceMetadataFetcherFlagOn(unittest.TestCase):
```
ubaskota, line 307: Could we add a test for when IMDS returns already-expired credentials? Something similar to `test_expired_expiry_extension` in the existing legacy tests.

botocore/utils.py
alexgromero: Yes. `utils.py` can't import the flag from `credentials.py` the same way `client.py` does because `credentials.py` already imports `InstanceMetadataFetcher` from `utils.py` during module initialization. Importing `credentials.py` back from `utils.py` would create a circular import, so `utils.py` could see a partially initialized `credentials.py` before `DEFAULT_NEW_CREDENTIAL_REFRESH` has been assigned.

tests/unit/test_credential_refresh.py
alexgromero: Yep, good call. Added in [2ea54ce](https://github.com/boto/botocore/pull/3769/commits/2ea54ce09bdb3984f873c32ada00dcb9b2fb66dd).

tests/unit/test_credential_refresh.py
alexgromero: Yep, good catch. That test was meant to prove that the fetcher stops rewriting near-expiry IMDS expirations on the new path, but the copied helper in that class was still using `datetime.now(...)` while the code under test was using the frozen `DATE` clock. I updated it to use `DATE` as well and patched the legacy jitter, so it now cleanly distinguishes the legacy expiry-extension behavior from the new behavior.

tests/unit/test_credential_refresh.py
alexgromero: I updated the new helpers and the IMDS replacement tests to use fixed timestamps, since those were the cases where wall-clock time was affecting determinism. The remaining `datetime.now(...)` uses are in the copied `TestRefreshableCredentials` coverage from the legacy file, where they're only used to build relative expiry times. I left those unchanged to avoid unrelated churn in the mirrored legacy tests.

tests/unit/test_credential_refresh.py
alexgromero: Good catch. I updated the new helpers to use fixed timestamps so they're deterministic now.

ubaskota (approved):

PR #3770, opened 2026-08-07
Add support for per-API invocation read timeouts

jamesls:
Botocore allows you to specify a read and connect timeout in the client config that's provided during client instantiation.  This sets the timeouts for the entire client.  There is no mechanism to override the timeout for specific operations or on a per-request (per-invocation basis).

Wanting to override timeouts on a more granular level comes up in several cases:

* Some service operations are blocking API calls that wait until some task is complete before returning, e.g. Lambda's `Invoke`, or Bedrock Runtime's `Converse`.
* Some specific invocations have a known minimal timeout based on the values of the parameters provided.  For example, SQS's `ReceiveMessage` has a `WaitTimeSeconds` which is the maximum number of seconds the request will wait for new messages to arrive.  Ideally we don't want a read timeout that's less than the value of `WaitTimeSeconds` if provided.

Per-request timeouts are implemented through an optional override in the request context dictionary as discussed with @jonathan343 . I updated the `URLLib3Session` class to check for a `read_timeout` in the request context and pass that value along to the `urlopen(...)` call if present.  The `urlopen(...)` call happens in the `send` method which only takes `send(self, request: AWSPreparedRequest)`.

In order to preserve that method signature, I opted for the less risky route of adding an optional `context` param to the `AWSPreparedRequest`. It's already in the `AWSRequest` so it didn't seem unreasonable to pass it through to the `AWSPreparedRequest`.  Given this object hasn't add new public attributes in many years, and that this type is publicly documented and accessible in various events emitted, I also opted for a safer `getattr(request, 'context', None)` when checking for a read timeout override.

## Example script

```python
import time

import botocore.session
from botocore.exceptions import ReadTimeoutError


QUEUE_URL = 'https://sqs.us-west-2.amazonaws.com/12345/YOUR-QUEUE'


def set_read_timeout(context, **kwargs):
    context['read_timeout'] = 1
    print("[handler] set read_timeout=1s for this invocation")


def main():
    # session.set_debug_logger()
    #
    # Turn off retries so we can see the total wall clock time should be
    # a little more than 1 second.
    client = session.create_client(
        'sqs', config=Config(retries={'total_max_attempts': 1})
    )

    client.meta.events.register(
        'before-call.sqs.ReceiveMessage', set_read_timeout
    )

    print(
        f"Calling SQS ReceiveMessage (WaitTimeSeconds=20 long poll) against\n"
        f"  {QUEUE_URL}\n"
    )
    start = time.monotonic()
    try:
        client.receive_message(QueueUrl=QUEUE_URL, WaitTimeSeconds=20)
    except ReadTimeoutError as e:
        total = time.monotonic() - start
        print(f"\n[OK] Got ReadTimeoutError after {total:.2f}s (expected ~1s)")
        print(f"     {e}")
        print(
            "\nThe client default read_timeout was 60s and the long poll "
            "would hold the connection for 20s, but the per-invocation handler"
            " overrode the\nread timeout to 1s so the call failed fast."
        )
    else:
        print(
            "\n[NOTE] Call returned before 1s (queue had a message ready), "
            "so the\ntimeout did not fire this run.  Re-run against an "
            "empty queue."
        )


if __name__ == '__main__':
    main()
```

**Output**

```
Calling SQS ReceiveMessage (WaitTimeSeconds=20 long poll) against
  https://sqs.us-west-2.amazonaws.com/1234/testq

[handler] set read_timeout=1s for this invocation

[OK] Got ReadTimeoutError after 1.30s (expected ~1s)
     Read timeout on endpoint URL: "https://sqs.us-west-2.amazonaws.com/"

The client default read_timeout was 60s and the long poll would hold the connection for 20s,
but the per-invocation handler overrode the
read timeout to 1s so the call failed fast.
```

### DIFF
--- a/botocore/awsrequest.py
+++ b/botocore/awsrequest.py
@@ -369,8 +369,11 @@ class AWSRequestPreparer:
         body = self._prepare_body(original)
         headers = self._prepare_headers(original, body)
         stream_output = original.stream_output
+        context = original.context
 
-        return AWSPreparedRequest(method, url, headers, body, stream_output)
+        return AWSPreparedRequest(
+            method, url, headers, body, stream_output, context
+        )
 
     def _prepare_url(self, original):
         url = original.url
@@ -499,14 +502,18 @@ class AWSPreparedRequest:
     :ivar headers: The HTTP headers to send.
     :ivar body: The HTTP body.
     :ivar stream_output: If the response for this request should be streamed.
+    :ivar context: The request context from the originating ``AWSRequest``.
     """
 
-    def __init__(self, method, url, headers, body, stream_output):
+    def __init__(
+        self, method, url, headers, body, stream_output, context=None
+    ):
         self.method = method
         self.url = url
         self.headers = headers
         self.body = body
         self.stream_output = stream_output
+        self.context = context
 
     def __repr__(self):
         fmt = (
--- a/botocore/endpoint.py
+++ b/botocore/endpoint.py
@@ -186,7 +186,9 @@ class Endpoint:
             retries_context['invocation-id'] = str(uuid.uuid4())
 
         if success_response:
-            read_timeout = context['client_config'].read_timeout
+            read_timeout = context.get('read_timeout')
+            if read_timeout is None:
+                read_timeout = context['client_config'].read_timeout
             self._set_ttl(retries_context, read_timeout, success_response)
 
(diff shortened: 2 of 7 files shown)


jonathan343 (approved):
LGTM! Thanks @jamesls!

PR #3773, opened 2026-08-12
Add non-recoverable credential refresh errors

alexgromero:
*Issue #, if available:*
Python-9033

### Overview
This updates the new credential refresh path to recognize non-recoverable refresh failures and surface them immediately instead of sending them through the normal refresh backoff flow.

For recoverable refresh failures, the SDK should continue using cached credentials and back off before retrying. For non-recoverable failures, the error requires customer action or is otherwise not expected to succeed without external action, so the SDK should raise it immediately. To avoid repeatedly hitting the credential source when callers retry in a loop, these non-recoverable errors are cached for a short jittered TTL and re-raised during that window.

This change adds that behavior for the new refresh path and wires in the current non-recoverable cases for STS, SSO, and Login providers.

#### What changed
- add short-lived caching for non-recoverable refresh errors and re-raise the cached error while its TTL is still in effect
- scope STS non-recoverable classification to STS-backed providers only
- treat SSO unauthorized service responses as non-recoverable
- treat SSO cached-token load failures and expired cached SSO tokens as non-recoverable
- introduce `LoginInvalidCachedTokenError` to distinguish invalid cached login tokens during refresh from other `LoginTokenLoadError` failures
- limit `LoginInvalidCachedTokenError` to missing token or missing required field cases during refresh and keep private-key parsing and invalid service response failures on the existing `LoginTokenLoadError` path

---

### DIFF
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -104,7 +104,7 @@ _NONRECOVERABLE_STS_ERROR_CODES = frozenset(
         'InvalidIdentityToken',
         'MalformedPolicyDocument',
         'PackedPolicyTooLarge',
-        'RegionDisabled',
+        'RegionDisabledException',
     )
 )
 _NONRECOVERABLE_STS_METHODS = frozenset(
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -803,6 +803,8 @@ class RefreshableCredentials(Credentials):
         return None
 
     def _handle_refresh_exception(self, error):
+        # Surface non-recoverable failures immediately because
+        # they require customer action rather than retry or backoff.
         if _is_nonrecoverable_refresh_error(self.method, error):
             self._cache_nonrecoverable_error(error)
             raise error
--- a/botocore/credentials.py
+++ b/botocore/credentials.py
@@ -627,14 +627,15 @@ class RefreshableCredentials(Credentials):
             return False
         return self._time_fetcher() < refresh_blocked_until
 
-    def _has_cached_nonrecoverable_error(self):
-        # Only suppress refresh attempts while the cached error's TTL
-        # is in effect.
-        if self._cached_nonrecoverable_error is None:
-            return False
-        return (
-            self._time_fetcher() < self._cached_nonrecoverable_error_expires_at
-        )
+    def _get_cached_nonrecoverable_error(self):
+        # Return the cached non-recoverable error while its TTL is still active.
+        cached_error = self._cached_nonrecoverable_error
+        expires_at = self._cached_nonrecoverable_error_expires_at
+        if cached_error is None or expires_at is None:
+            return None
+        if self._time_fetcher() >= expires_at:
+            return None
+        return cached_error
 
     def _cache_nonrecoverable_error(self, error):
         cache_ttl = random.uniform(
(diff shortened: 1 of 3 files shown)


botocore/credentials.py
```diff
@@ -92,6 +95,36 @@
...
+_NONRECOVERABLE_ERROR_CACHE_MIN_TTL = 1
+_NONRECOVERABLE_ERROR_CACHE_MAX_TTL = 5
+_NONRECOVERABLE_STS_ERROR_CODES = frozenset(
+    (
+        'AccessDenied',
+        'IDPRejectedClaim',
+        'InvalidIdentityToken',
+        'MalformedPolicyDocument',
+        'PackedPolicyTooLarge',
+        'RegionDisabled',
```
alexgromero: I just realized the error code here should be `RegionDisabledException`. Will update in next revision. 

See models: https://github.com/boto/botocore/blob/develop/botocore/data/sts/2011-06-15/service-2.json#L906-L907

botocore/exceptions.py
```diff
@@ -215,6 +215,12 @@ class CredentialRetrievalError(BotoCoreError):
     fmt = 'Error when retrieving credentials from {provider}: {error_msg}'
 
 
+class RefreshNonRecoverableError:
```
ubaskota, line 218: In `exceptions.py`, all other errors inherit from an exception base class except for this. It works now because it's always mixed with classes that already inherit from Exception, but if it were ever used by itself, it couldn't be caught with except or raised properly. Is there a reason it doesn't subclass from Exception?

botocore/credentials.py
```diff
@@ -741,6 +803,9 @@ def _validate_data(self, data):
         return None
 
     def _handle_refresh_exception(self, error):
```
ubaskota, line 807: `_handle_refresh_exception` raises non-recoverable errors immediately, even when the cached credentials are still valid. I confirmed that this is the expected behavior. Can we add a comment here noting that its intentional and why?

botocore/credentials.py
```diff
@@ -592,6 +627,31 @@ def _in_refresh_backoff(self):
             return False
         return self._time_fetcher() < refresh_blocked_until
 
+    def _has_cached_nonrecoverable_error(self):
+        # Only suppress refresh attempts while the cached error's TTL
```
ubaskota: This has the same double-read pattern that was fixed in PR #3775. Today it's safe because both `_has_cached_nonrecoverable_error` and `_clear_refresh_failure_state` are called in `_protected_refresh` which is under the lock, so thread A and thread B can't be in here simultaneously. But for defensive purposes, could we read into a local variable just like you did in [_in_refresh_backoff](https://github.com/boto/botocore/blob/34909c69bfb72c3ca273cc3cc59d52d8773fe312/botocore/credentials.py#L589) ? This way it keeps the method safe regardless of where it's called from.

botocore/exceptions.py
alexgromero: `RefreshNonRecoverableError` isn't intended to be raised on its own. It's just a marker used with `isinstance()` so the refresh path can recognize provider errors that are non-recoverable. We still want to raise the original provider-specific error to the caller rather than replace it with a generic non-recoverable error.

botocore/credentials.py
alexgromero: Sounds good, added in https://github.com/boto/botocore/pull/3773/commits/941823c04287fdf4df6b75123134c6eb94e8814b.

botocore/credentials.py
alexgromero: Added in https://github.com/boto/botocore/pull/3773/commits/7950eef98f32a6f4cfa33f89d876c9b3e48b304f.

botocore/credentials.py
alexgromero: Added in https://github.com/boto/botocore/pull/3773/commits/3afe7ebd6fabab4d78f8b46fea1439e6d4bd75a8.

ubaskota (approved):

PR #3777, opened 2026-08-15
Invalidate cached credentials on auth failures

alexgromero:
*Issue #, if available:*
Python-9269

### Overview
This updates credential refresh to invalidate cached refreshable credentials after service authentication failures.

When `DEFAULT_NEW_CREDENTIAL_REFRESH` is enabled, a service can reject otherwise unexpired credentials with `ExpiredToken` or `InvalidToken`, but botocore may still keep that credential set cached until its normal expiration time. This change marks the rejected credentials stale so the next request goes back through refresh instead of continuing to reuse the rejected credentials.

The legacy refresh path is unchanged.

### What changed
On the signing side, `RequestSigner` now records the access key that signed the request in `request.context`. This is only done for the standard credential-chain case; request-scoped credentials and identity-cache flows are intentionally excluded.

On the response side, a new `after-call` handler watches for `ExpiredToken` and `InvalidToken`. When one of those errors is returned, it resolves the client's current credentials and invalidates them only if the cached access key still matches the access key that signed the rejected request. That avoids mutating fresh credentials when another thread has already completed a refresh.

The invalidation itself sets the cached credentials' expiry to `now`, so the next credential lookup routes through refresh. It also uses a non-blocking acquire on the refresh lock: if a refresh or another invalidation is already in progress, it returns immediately instead of blocking.

For providers with an intermediate cache layer, invalidating only the outer `RefreshableCredentials` object is not enough. AssumeRole, AssumeRoleWithWebIdentity, and SSO can otherwise refresh back into the same cached provider result. This change wires those providers to also expire the matching cached provider entry so the next refresh goes back to the underlying source.

### Testing
Added coverage for:
- signer behavior that records `signing_access_key` and skips request-scoped / identity-cache cases
- auth-error invalidation event registration and handler behavior, including `error_code_override`
- refreshable credential invalidation when the access key matches or does not match
- provider-cache invalidation for AssumeRole, AssumeRoleWithWebIdentity, and SSO
- invalidation preserving the existing refresh backoff flow rather than bypassing it
- strict refreshable credentials still invalidating while preserving fail-fast refresh behavior when refresh itself fails

---


tests/unit/test_credential_refresh.py
```diff
@@ -1431,3 +1431,315 @@ def test_login_refresh_invalid_cached_token_raises_nonrecoverable_error(
             match=error_msg,
         ):
             fetcher.refresh_credentials()
+
+
+class TestInvalidate:
```
ubaskota, line 1436: Can we add a test for the case where the refresh lock is already held when `_invalidate` is called?

ubaskota (commented):
It looks good to me as is. I've added a comment to point out a test gap and will approve it after its addressed. Thanks

tests/unit/test_credential_refresh.py
alexgromero: Yep, added in https://github.com/boto/botocore/pull/3777/commits/ae17afe2e2b5c493e6c254accb3800aa56a11c3e.

ubaskota (approved):

PR #3779, opened 2026-08-15
fix(paginate): detect pagination tokens that cycle, not just repeat

### LINKED ISSUE #3778: Paginator: cycling next-token (not just a repeated one) causes infinite pagination instead of PaginationError
### Describe the bug

`PageIterator.__iter__` (`botocore/paginate.py`) has a guard meant to stop pagination from looping forever when a service stops returning progress: if the next token is the same as the token from the previous page, it raises `PaginationError`.

```python
previous_next_token = None
...
if (
    previous_next_token is not None
    and previous_next_token == next_token
):
    message = (
        f"The same next token was received twice: {next_token}"
    )
    raise PaginationError(message=message)
self._inject_token_into_kwargs(current_kwargs, next_token)
previous_next_token = next_token
```

This only remembers the *single most recent* token. It correctly catches a token repeating on two consecutive pages (`A, A`), but it does not catch a longer cycle, e.g. a token sequence that alternates `A, B, A, B, ...`. At every step, `next_token` differs from `previous_next_token` (the token right before it), so the check never fires, and pagination continues indefinitely, re-fetching the same pages and never terminating.

This is plausible with any backend that hands back cursors from a small rotating/partitioned pool (e.g. a shard-based or load-balanced pagination cursor) rather than a token that always strictly advances — a service-side anomaly, but one the existing check exists specifically to guard against, and currently doesn't for anything beyond period-1 cycles.

### Expected Behavior

Any next token that repeats a *previously seen* token (not just the immediately preceding one) should raise `PaginationError`, the same as it does today for immediate repeats.

### Current Behavior

Pagination with a cycling token of period ≥ 2 never raises and never terminates.

### Reproduction Steps

Constructing a `PageIterator` directly against a fake service method that alternates `NextToken` between two values:

```python
import jmespath
from botocore.paginate import PageIterator

token_sequence = ['B', 'A', 'B', 'A', 'B', 'A']
call_count = {'n': 0}

def fake_method(**kwargs):
    i = call_count['n']
    call_count['n'] += 1
    return {'Items': [f'item-{i}'], 'NextToken': token_sequence[i % len(token_sequence)]}

pi = PageIterator(
    method=fake_method,
    input_token=['NextToken'],
    output_token=[jmespath.compile('NextToken')],
    more_results=None,
    result_keys=[jmespath.compile('Items')],
    non_aggregate_keys=[],
    limit_key=None,
    max_items=None,
    starting_token=None,
    page_size=None,
    op_kwargs={},
)

for page in pi:
    pass  # never terminates today
```

For comparison, the equivalent consecutive-duplicate case (`NextToken` = `A, A`) correctly raises `PaginationError: The same next token was received twice: {'NextToken': 'A'}` today — confirming the guard's intent, and that it just doesn't generalize past a 1-token lookback.

### Possible Solution

Track every next token seen so far (not just the immediately preceding one) and raise `PaginationError` as soon as a repeat of any previously-seen token is encountered. I have a fix and a regression test ready and will open a PR shortly.

### Environment

Reproduced on current `main` (commit `481aaad`, botocore 1.43.72).

Adityaj0:
### Issue # (if applicable)

fixes #3778

### Reason for this change

`PageIterator.__iter__` guards against non-terminating pagination by comparing each new `next_token` against `previous_next_token`, a single-slot memory of the immediately preceding token:

```python
previous_next_token = None
...
if (
    previous_next_token is not None
    and previous_next_token == next_token
):
    raise PaginationError(...)
self._inject_token_into_kwargs(current_kwargs, next_token)
previous_next_token = next_token
```

This catches a token repeating on two consecutive pages, but misses a longer cycle (`A, B, A, B, ...`): at every step `next_token` differs from the token immediately before it, so the check never fires and pagination runs forever, re-fetching the same pages. This is exactly the failure mode the check exists to prevent — a service handing back a token that makes no forward progress — it just doesn't generalize past a one-token lookback.

### Description of changes

Replaced the single `previous_next_token` slot with a list of every `next_token` seen so far (`seen_next_tokens`), and check membership in that list instead of equality with only the last one:

```python
seen_next_tokens = []
...
if next_token in seen_next_tokens:
    raise PaginationError(...)
self._inject_token_into_kwargs(current_kwargs, next_token)
seen_next_tokens.append(next_token)
```

`next_token` is a `dict`, so a `set` isn't usable directly (unhashable); a list with `in` keeps the same semantics as the original equality check while covering cycles of any length. Pagination depth is bounded in practice, so the linear scan is not a concern.

### Describe any new or updated permissions being added

None.

### Description of how you validated changes

Added `test_exception_raised_if_next_token_cycles` to `tests/unit/test_paginate.py`, using the same `self.paginator`/`self.method` fixture as the existing `test_exception_raised_if_same_next_token`, but with a token sequence `token1, token2, token1, token2` that never repeats consecutively. It fails (hangs/never raises before this fix would need a manual timeout) on `main` and passes with this change.

```
$ python -m pytest tests/unit/test_paginate.py -v -k "same_next_token or cycles"
test_exception_raised_if_same_next_token PASSED
test_exception_raised_if_next_token_cycles PASSED

$ python -m pytest tests/unit/test_paginate.py -q
75 passed

$ python -m pytest tests/functional/test_paginate.py -q
9 passed
```

Also manually verified with a standalone repro driving `PageIterator` directly against a fake `method` cycling `NextToken` between two values — hangs indefinitely on `main`, raises `PaginationError` with this change. And confirmed the existing consecutive-duplicate case still raises correctly (no regression to the original behavior).

### Backwards compatibility

The only externally-visible change is that pagination which previously ran forever (a hang, effectively unusable) now raises `PaginationError` instead. No currently-terminating pagination changes behavior — a cycling token is never legitimate forward progress, so this only converts an infinite hang into a clear error.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

### DIFF
--- a/.changes/next-release/bugfix-Paginator-12345.json
+++ b/.changes/next-release/bugfix-Paginator-12345.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Paginator",
+  "description": "Fixed a bug where ``PageIterator`` only compared each next token against the immediately preceding one, so a service returning tokens from a cycling/rotating pool (e.g. alternating between two cursors) would paginate forever instead of raising ``PaginationError``. All previously-seen tokens are now tracked."
+}
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -260,7 +260,10 @@ class PageIterator:
         # being repeated on consecutive pages; it misses longer cycles
         # (e.g. a token sequence that alternates A, B, A, B, ...), which
         # would otherwise paginate forever.
-        seen_next_tokens = []
+        # next_token is a dict, so it isn't hashable as-is; store a
+        # frozenset of its items instead so membership checks stay O(1)
+        # even across paginations with very large numbers of pages.
+        seen_next_tokens = set()
         next_token = {key: None for key in self._input_token}
         if self._starting_token is not None:
             # If the starting token exists, populate the next_token with the
@@ -325,13 +328,24 @@ class PageIterator:
                     # next token to be the resume token.
                     self.resume_token = next_token
                     break
-                if next_token in seen_next_tokens:
+                next_token_key = self._hashable_token_key(next_token)
+                if next_token_key in seen_next_tokens:
                     message = (
                         f"The same next token was received twice: {next_token}"
                     )
                     raise PaginationError(message=message)
                 self._inject_token_into_kwargs(current_kwargs, next_token)
-                seen_next_tokens.append(next_token)
+                seen_next_tokens.add(next_token_key)
+
+    def _hashable_token_key(self, next_token):
+        # next_token's values are normally simple scalars (str/int/etc.),
+        # but fall back to a stable string representation if a service
+        # ever returns a nested/unhashable value so we still get O(1)
+        # membership checks instead of raising here.
+        try:
+            return frozenset(next_token.items())
+        except TypeError:
+            return repr(sorted(next_token.items(), key=lambda item: item[0]))
 
     def search(self, expression):
         """Applies a JMESPath expression to a paginator
--- a/botocore/paginate.py
+++ b/botocore/paginate.py
@@ -254,7 +254,13 @@ class PageIterator:
 
     def __iter__(self):
         current_kwargs = self._op_kwargs
-        previous_next_token = None
+        # Every next token we've injected into a request so far, used to
+        # detect a service handing back a token that doesn't make forward
+        # progress. A single "previous token" slot only catches the token
+        # being repeated on consecutive pages; it misses longer cycles
+        # (e.g. a token sequence that alternates A, B, A, B, ...), which
+        # would otherwise paginate forever.
+        seen_next_tokens = []
         next_token = {key: None for key in self._input_token}
         if self._starting_token is not None:
             # If the starting token exists, populate the next_token with the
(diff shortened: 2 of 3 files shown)


Adityaj0:
Updated this to fix a regression I found in my own original approach: `seen_next_tokens` was a `list` (since `next_token` is a dict and unhashable), and `next_token in seen_next_tokens` is an O(k) linear scan run once per page — O(n²) total across a full pagination.

Measured with a synthetic page loop:
```
list-based check:  16,000 pages -> 1.83s   |  200,000 pages -> didn't finish in 120s
set-based check:   16,000 pages -> 0.004s  |  200,000 pages -> 0.16s
```
That's a real problem for the workloads this touches — S3 listings, DynamoDB scans, CloudWatch Logs, etc. can legitimately run into tens/hundreds of thousands of pages, so the original fix would have traded an unobserved hang for a guaranteed slowdown on every large, perfectly healthy pagination.

Fixed by hashing `next_token` into a `frozenset(next_token.items())` and storing those in a `set` instead, restoring O(1) membership checks (with a `repr`-based fallback if a token value is ever itself unhashable). All existing and new unit/functional tests still pass; re-verified the original cycling repro still raises `PaginationError` correctly with this change.

PR #3781, opened 2026-08-18
Add credential refresh lifecycle tests

alexgromero:
*Issue #, if available:*
Python-9034

### Overview
This PR adds lifecycle coverage for the new credential refresh path behind `DEFAULT_NEW_CREDENTIAL_REFRESH`.

The earlier PRs in this stack add the behavior changes themselves. This PR focuses on the test coverage, including the modeled lifecycle scenarios and concurrency behavior.

> [!NOTE]                                                                                                                                                                                                 
> This branch currently includes commits #3777 so it can target feature/credential-refresh directly.                                                                               
> The primary review focus here is the lifecycle test coverage so **only the latest commit is the change under review here**.

##### 1. Modeled lifecycle coverage

* Copied the modeled credential refresh cases into `tests/unit/credential_refresh/credential-refresh-tests.json`
* Added a data-driven runner in `tests/unit/test_credential_refresh.py` to execute those cases
* Added support for `BOTOCORE_TEST_ID=<n>` to run an individual case while debugging
* Added a controllable clock and fixed backoff values so time-based assertions are deterministic

##### 2. Concurrency coverage

Added tests for concurrent credential access during refresh:

* during advisory refresh, one caller performs the refresh while other callers continue using cached credentials
* during mandatory or expired refresh, concurrent callers wait for the in-progress refresh and reuse its result

---

### DIFF
--- a/tests/unit/test_credential_refresh.py
+++ b/tests/unit/test_credential_refresh.py
@@ -2100,6 +2100,8 @@ class TestCredentialRefreshConcurrency:
             clock,
             refresher=refresher,
             expires_in=expires_in,
+            advisory_timeout=120,
+            mandatory_timeout=60,
         )
 
         refresh_owner_thread = CredentialFetchThread(creds)
--- a/tests/unit/test_credential_refresh.py
+++ b/tests/unit/test_credential_refresh.py
@@ -30,6 +30,8 @@ Mirror policy for GA migration:
 """
 
 import operator
+import os
+import threading
 from datetime import datetime, timedelta, timezone
 
 import pytest
@@ -62,6 +64,11 @@ from tests.unit import test_credentials
 
 DATE = datetime(2021, 12, 10, 0, 0, 0, tzinfo=timezone.utc)
 DT_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
+CREDENTIAL_REFRESH_TEST_CASES_FILE = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)),
+    'credential_refresh',
+    'credential-refresh-tests.json',
+)
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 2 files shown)


tests/unit/test_credential_refresh.py
```diff
@@ -1431,3 +1438,666 @@ def test_login_refresh_invalid_cached_token_raises_nonrecoverable_error(
...
+        assert refreshing_thread.error is None
+        assert refreshing_thread.frozen_credentials.access_key == 'NEW-ACCESS'
+        assert refresher.call_count == 1
+
+    @pytest.mark.parametrize(
+        'expires_in',
+        [timedelta(seconds=30), timedelta(seconds=-30)],
+        ids=['mandatory', 'expired'],
+    )
+    def test_mandatory_refresh_concurrent_callers_wait_for_single_refresh(
```
ubaskota, line 2091: The advisory concurrency test right above this explicitly passes `advisory_timeout=120`, `mandatory_timeout=60`, but the mandatory test relies on the defaults. Could we pass explicit timeouts here too so the test is self-contained? Or was there a reason you decided not to add it?

ubaskota:
Do we need to add functional/integration tests that cover the full invalidation chain?

tests/unit/test_credential_refresh.py
alexgromero: Yep, added in https://github.com/boto/botocore/pull/3781/commits/3635b418b6a7550aad43165b99d14376e3df6be9.

alexgromero (commented):
> Do we need to add functional/integration tests that cover the full invalidation chain?

@ubaskota I don't think we need additional functional/integration tests for this PR. The invalidation chain is already covered at the unit level across the signer, client, and credential refresh layers, and the lifecycle tests here cover the resulting credential behavior. I can take another look at the feature branch as a whole for any additional end-to-end coverage we may want before GA.

ubaskota (approved):

Issue #3755, opened 2026-07-21
`Expect: 100-continue` + zero-length PUT poisons the pooled connection: next response parsed against a stale status tuple (silent 60s stall + hidden retry + double-write)


ThomasBurgess2000:
### Describe the bug

When a `PutObject` with an **empty body** is sent over a pooled keep-alive connection to an S3-compatible server built on Go's `net/http` (SeaweedFS, MinIO, …), botocore's `Expect: 100-continue` handling permanently corrupts that connection object, and the **next** request on the connection has its response parsed against a stale, stashed status tuple instead of the socket.

Root cause walk-through (botocore 1.42.70):

1. `handlers.add_expect_header` adds `Expect: 100-continue` to every S3 PUT (registered on `before-call.s3`) — including zero-length bodies.
2. Go's `net/http` is RFC 9110-compliant: for a request with `Content-Length: 0` it never emits an interim `100 Continue` (`server.go`: `if req.ProtoAtLeast(1, 1) && req.ContentLength != 0`). RFC 9110 §10.1.1 explicitly permits omitting 100 "if the framing indicates that there is no content". So the **final 200 arrives in the interim-response slot**, within botocore's 1-second `wait_for_read` window (`awsrequest.py`, `AWSConnection._send_output`).
3. `AWSConnection._handle_expect_response` reads that early status line and — in its non-100 branch — permanently swaps `self.response_class` to `functools.partial(AWSHTTPResponse, status_tuple=(...))` so the already-consumed status line can be replayed for *this* response.
4. **The bug:** `AWSConnection.request()` resets `response_class` back to the default **only in the code path taken by requests that do *not* carry `Expect`**. In a burst of PUTs, every request carries `Expect`, so the poisoned `response_class` persists on the pooled connection.
5. On the next (non-empty) PUT, the server *does* send `100 Continue` + a final response — but `AWSHTTPResponse._read_status` **replays the stale stashed tuple without reading the socket**. Header parsing then starts at the real (unread) status line and the client is left holding a status-only response with no headers.

The wire is clean throughout — we verified with a byte-logging TCP tap that the server sends exactly one well-formed response per request. The "corrupt" bytes exist only in the client's parse.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

- A zero-length PUT carrying `Expect: 100-continue` against a server that (per RFC 9110 §10.1.1) skips the interim `100 Continue` should complete cleanly and leave the pooled connection in a clean state.
[...]

PR #3784, opened 2026-08-21
fix: Restore response_class on every request to prevent status leak

ubaskota:
*Issue #, if available:* #3755

*Description of changes:*
`AWSConnection.request()` only restored the `response_class` override in the non-Expect branch, so a cached status from an early final response could leak onto the next request on a pooled connection, dropping its headers (e.g. a missing S3 ETag). Move the restore so it runs on every request.

*Testing*
- Verified that the newly added test passes with the fix and fails without it.
- Verified that the existing passes

### DIFF
--- a/.changes/next-release/bugfix-s3-10143.json
+++ b/.changes/next-release/bugfix-s3-10143.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "s3",
+  "description": "Fixed an issue where reused connections could return a cached response\n  status, dropping response headers."
+}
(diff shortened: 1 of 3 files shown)


ashovlin (approved):

.changes/next-release/bugfix-s3-10143.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "s3",
```
alexgromero: The issue was the connection layer but s3 was just the trigger. so maybe:
```suggestion
  "category": "HTTP",
```

.changes/next-release/bugfix-s3-10143.json
```diff
@@ -0,0 +1,5 @@
...
+  "description": "Fixed an issue where reused connections could return a cached response\n  status, dropping response headers."
```
alexgromero: Is the `\n` here intentional?

alexgromero (approved):

PR #3791, opened 2026-09-02
Note SigV2 default for presigned URLs

Alan4506:
*Description of changes:*

Adds a warning to the `generate_presigned_url` docstring noting that S3 presigned URLs use SigV2 by default, and recommending using SigV4.

*Testing:*

- `make clean html` succeeds. 
- `pytest tests/unit/docs tests/functional/docs` passes. 
- Built with `generate_docs` scoped to `['s3', 'dynamodb']` to confirm the warning reaches only the S3 page. Same check repeated against boto3 with this branch installed locally.

The updated documentations for botocore and boto3 will look like the following:

image
image

### DIFF
--- a/botocore/signers.py
+++ b/botocore/signers.py
@@ -766,7 +766,7 @@ def generate_presigned_url(
         conditionals and ``Range``, are also silently dropped.
 
         It is recommended to configure the S3 client with Signature
-        Version 4, setting ``signature_version='s3v4'`` in the client's
+        Version 4 by setting ``signature_version='s3v4'`` in the client's
         ``Config``.
 
     :type ClientMethod: string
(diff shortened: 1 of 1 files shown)


botocore/signers.py
```diff
@@ -758,6 +758,17 @@ def generate_presigned_url(
...
 
+    .. warning::
+
+        For backwards compatibility, S3 presigned URLs use Signature
+        Version 2 by default in regions where S3 supports it. As a result,
+        some ``Params`` entries are not signed. Some headers, including the
+        conditionals and ``Range``, are also silently dropped.
+
+        It is recommended to configure the S3 client with Signature
+        Version 4, setting ``signature_version='s3v4'`` in the client's
```
arandito: minor nit when re-reading this, lets drop the comma and add "by" so it reads smoother.

"It is recommended to configure the S3 client with Signature Version 4 by setting..."

arandito (approved):
:shipit:

PR #3799, opened 2026-09-14
Expose calculated checksum on StreamingChecksumBody

arandito:
### Description

This PR makes two small additions to `StreamingChecksumBody` so that s3transfer can validate full object checksums on multipart downloads (boto/s3transfer#XXXX, a port of aws/aws-cli#10180):

- Adds a public `checksum` property that returns the running checksum object. When botocore has already wrapped a response body, s3transfer reads the part-level digest from it instead of computing a second checksum over the same bytes.
- Allows `expected=None`. `_validate_checksum` now returns early when there is no expected value, so the wrapper can be used to only calculate a checksum as the stream is read. S3 does not return a checksum header on ranged GETs of an object stored with a `FULL_OBJECT` checksum, so s3transfer wraps those bodies itself and combines the per-part results after all parts complete.

### Testing

- Added unit tests for the `checksum` property and for `read()`/`readinto()` without an expected checksum.
- Ran `tests/unit/test_httpchecksum.py` successfully
- Confirmed end to end with the s3transfer branch, which constructs `StreamingChecksumBody(body, content_length, checksum, expected=None)` on ranged GETs and reads `body.checksum` when botocore wrapped the body (see the Testing section of boto/s3transfer#XXXX).

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute
this contribution, under the terms of your choice.


jonathan343 (approved):
🚢

